PR #74, opened 2021-09-04
Simplify types consumption and use @temporalio/proto

feedmeapples:
## What was changed

Switched to `@temporalio/proto` NPM package for types consumption
Created custom `types/index.ts` that exports common types from various paths of `@temporalio/proto`
Refactored the app to consume `types/index.ts`

## Why?

Simplifies types consumption
Allows to remove custom type generation from the project 

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/lib/components/namespace-select.svelte
+++ b/src/lib/components/namespace-select.svelte
@@ -3,7 +3,7 @@
   import { getContext } from 'svelte';
   import { goto } from '$app/navigation';
 
-  import type { DescribeNamespaceResponse } from '$types/temporal/api/workflowservice/v1/request_response';
+  import type { DescribeNamespaceResponse } from '$types';
 
   $: namespaces = getContext<DescribeNamespaceResponse[]>('namespaces').map(
     (namespace) => namespace.namespaceInfo.name,
--- a/src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/events/events.d.ts
+++ b/src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/events/events.d.ts
@@ -1,85 +1,85 @@
 type HistoryEvent =
-  import('$types/temporal/api/history/v1/message').HistoryEvent;
+  import('$types').HistoryEvent;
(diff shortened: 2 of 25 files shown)


src/lib/components/namespace-select.svelte
```diff
@@ -3,7 +3,7 @@
   import { getContext } from 'svelte';
   import { goto } from '$app/navigation';
 
-  import type { DescribeNamespaceResponse } from '$types/temporal/api/workflowservice/v1/request_response';
+  import type { DescribeNamespaceResponse } from '$types/definitions';
```
feedmeapples, line 6: there is probably a better to do `$types` instead of `$types/definitions`. I didn't focus on this much and wanted to have a working PoC first

src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/events/events.d.ts
```diff
@@ -1,85 +1,85 @@
 type HistoryEvent =
-  import('$types/temporal/api/history/v1/message').HistoryEvent;
+  import('$types').HistoryEvent;
```
Comment (bergundy) [inline src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/events/events.d.ts:2]: Is this generated code?
Why do we need all of these separate import lines?

bergundy (approved):
Nice to see the protos library is being put into good use :)

src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/events/events.d.ts
Comment (feedmeapples) [reply src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/events/events.d.ts:2]: afaik these event types may end up having custom properties or the same properties but custom mapping handlers

cc @stevekinney to correct me or provide details

Comment (bergundy) [reply src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/events/events.d.ts:2]: You might want to take the interfaces instead of the classes, I find them easier to work with.

```suggestion
  import('$types').IHistoryEvent;
```

PR #76, opened 2021-09-07
Fix SSR issue with localStorage

stevekinney:
Makes sure we're running in a browser environment before trying to access `localStorage`.


feedmeapples (approved):

feedmeapples:
what benefits does svelte kit's SSR bring given we plan to serve the UI from a custom backend?

stevekinney:
I suspect we will end up disabling it, but [this warning in the documentation](https://kit.svelte.dev/docs#ssr-and-javascript-ssr) made me hesitate… this time.

PR #80, opened 2021-09-09
filter event history & swap key/value of table to json view

josephriosIO:
## What was changed
Added json view when clicking on table component and filter by event type 

## Why?
Apart of sprint tasks

## Video 

https://user-images.githubusercontent.com/43492172/132772925-70cd8894-95ec-4f94-a83f-066aa30c6801.mov

### DIFF
--- a/src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/events/_events.svelte
+++ b/src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/events/_events.svelte
@@ -5,6 +5,13 @@
 
   export let history: History;
   export let eventFormat: EventFormat;
+  export let eventType;
+
+  $: events = history.events.filter((event) => {
+    if (eventType && !String(event.eventType).startsWith(eventType))
+      return false;
+    return true;
+  });
 </script>
 
 <section>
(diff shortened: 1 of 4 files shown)


src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/events/_events.svelte
```diff
@@ -5,6 +5,13 @@
 
   export let history: History;
   export let eventFormat: EventFormat;
+  export let eventType;
+
+  $: events = history.events.filter((event) => {
+    if (eventType && !String(event.eventType).startsWith(eventType))
```
Comment (feedmeapples) [inline src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/events/_events.svelte:11]: is this for performance or why `startsWith` and not `includes` ?

feedmeapples (approved):

src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/events/_events.svelte
Comment (josephriosIO) [reply src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/events/_events.svelte:11]: I used the same logic as the workflow filters, but would you say ‘includes’ will be better in this case since users will be able to type in any part of the phrase they are searching for?

Comment (feedmeapples) [reply src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/events/_events.svelte:11]: @josephriosIO for UX - yes. For perf - no

Usually i would prioritize UX, unless there is a noticeable perf decrease (in which case this affects UX too)

PR #85, opened 2021-09-21
Add basic date filtering

stevekinney:
Initial ground work for selecting workflows in a given time range.

Also, this pull request adds `createStoreWithCallback`, which allows you to create a `writable` store that will trigger a function whenever `set` or `update` are called on the store.


feedmeapples (approved):

feedmeapples:
lg

PR #87, opened 2021-09-22
Fix time display; break out pagination and empty state

stevekinney:
Fixes bug with displaying the correct time format. Breaks out empty state and pagination controls into their own components. Gracefully hands no matching workflows better (e.g. displaying "No Workflows" instead of "Page 1 of 0").


src/routes/namespaces/[namespace]/workflows/__layout.svelte
```diff
@@ -30,35 +32,15 @@
...
-            Page {currentPage + 1} of {maximumPage}
-          </p>
-          <button
-            on:click={() => currentPage++}
-            disabled={currentPage >= maximumPage - 1}
-          >
-            Next
-          </button>
-        </section>
+        <WorkflowFilters bind:timeFormat />
```
Comment (feedmeapples) [inline src/routes/namespaces/[namespace]/workflows/__layout.svelte:35]: very cool

feedmeapples (approved):

PR #88, opened 2021-09-23
Fix bug with overwriting data; refactor query store

stevekinney:
Fixes a bug where new data was properly overwriting already-loaded data. This change also begins to pull out common utility functions for both workflow and event history fetching.


josephriosIO (approved):
I did a lot of googling but it looks good

PR #89, opened 2021-09-24
Fix bugs with query store and activity view

stevekinney:
Further refinements to how we update data. Reactive updating to current activity view.

### DIFF
--- a/src/lib/stores/workflows.ts
+++ b/src/lib/stores/workflows.ts
@@ -1,4 +1,4 @@
-import { derived, writable, Writable, get } from 'svelte/store';
+import { derived, writable, Writable, get as getFromStore } from 'svelte/store';
 import type { ListWorkflowExecutionsResponse } from '$types';
 import { unique } from '$lib/utilities/unique';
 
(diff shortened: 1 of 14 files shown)


src/lib/stores/workflows.ts
```diff
@@ -57,6 +58,15 @@ export const createStore = (namespace: string) => {
   const ids = derived(store, ($store) => Object.keys($store.ids));
   const range = createStoreWithCallback<Duration>({ hours: 24 }, update);
 
+  range.subscribe(() =>
+    store.set({
+      loading: false,
+      updating: true,
+      workflows: {},
```
feedmeapples, line 65: will this make the workflows screen render empty before fetching the data again? 
If so, what happens if we filter the existing store workflows based on range and fetch additional workflows in parallel?

also not a concern for this PR, just a side note

feedmeapples (approved):

src/lib/stores/workflows.ts
stevekinney: It will, for now. But, that's only a stop-gap measure. I had a bug where it wasn't removing stale executions. A forthcoming refactor will get this right.

PR #94, opened 2021-09-30
Add type checking to build process

stevekinney:
- Separates out linting and unit tests.
- Adds type checking to the CI/CD process

### DIFF
--- a/.github/workflows/check-types.yml
+++ b/.github/workflows/check-types.yml
@@ -1,0 +1,17 @@
+name: Check Types
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/check-types.yml
```diff
@@ -0,0 +1,17 @@
+name: Check Types
```
feedmeapples, line 1: if we add a build status badge in the readme, this seems like we would have separate badges for Types, Linting and Unit tests?

Do we add only the unit tests badge in the readme then, or all of them?

stevekinney: The trade-off is instantly knowing which step failed versus having to go and look at the logs. As far as `main` is concerned, Github keeps track as to whether or not _all_ of the steps passed or failed, so we can use that.

feedmeapples (approved):

PR #95, opened 2021-09-30
Standardize the API for individual and multiple workflows

stevekinney:
Modify `toWorkflowExecutions` so that it takes the API response directly, just like `toWorkflowExecution`.


feedmeapples (approved):

feedmeapples:
good call

PR #97, opened 2021-10-05
Refactor query stores

stevekinney:
Removes lots of duplicate code and makes the QueryStore abstraction a little less leaky. This version also supports dependencies and does not empty out the story when the dependencies are updated like the previous version.

Things that still need to be done:

- Implement `loading` and `updating` states.
- Needs more testing

### DIFF
--- a/src/lib/utilities/create-query-store.ts
+++ b/src/lib/utilities/create-query-store.ts
@@ -86,7 +86,6 @@ export const updateStore =
     formatter: Formatter<ResponseType, FormattedType>,
   ) =>
   (response: ResponseType) => {
-    console.log(response);
     const formatted = formatter(response);
     const ids = {};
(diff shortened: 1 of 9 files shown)


src/lib/utilities/create-query-store.ts
```diff
@@ -1,42 +1,136 @@
...
+  dependencies: Readable<Partial<Options>>[] = [],
 ) => {
-  const store = writable<S>(
+  let parameters = options;
+
+  const request = () => {
+    fetch({
+      ...parameters,
+      onUpdate: updateStore(store, format),
+      onComplete: removeStaleData(store, format),
```
feedmeapples, line 37: :100:

feedmeapples (approved):

PR #98, opened 2021-10-05
Dismissible menus

josephriosIO:
## What was changed
added dismissible for menus and made styling of all select menus the same

## Why?
apart of sprint goals for week of October 4th


src/lib/components/filter-select.svelte
```diff
@@ -2,6 +2,8 @@
   export let id: string;
   export let name: string;
   export let value: string | boolean;
+
+  let showDropdown = false;
```
stevekinney, line 6: `showDropdown` is declared but its value is never read.

src/lib/components/filter-select.svelte
```diff
@@ -13,5 +15,22 @@
     on:change
   >
     <slot />
+    <span
```
stevekinney, line 18: I don't think this does anything.

src/lib/utilities/click-outside.ts
```diff
@@ -0,0 +1,15 @@
+export function clickOutside(node: any) {
+  const handleClick = (event: MouseEvent) => {
+    if (node && !node.contains(event.target) && !event.defaultPrevented) {
+      node.dispatchEvent(new CustomEvent('click_outside', node));
```
stevekinney, line 4: I wonder if we need a custom event? A click event should open it. A click event on the `document` and if the target is *not* the dropdown, then we should close it.

stevekinney (commented):
The sign-in/sign-out dropdown is currently not working.

src/lib/components/filter-select.svelte
josephriosIO: will remove

src/lib/components/filter-select.svelte
josephriosIO: this was most likely left over code i forgot to remove this is like iteration 4 of this menu

src/lib/utilities/click-outside.ts
josephriosIO: will remake it to use without the custom event

josephriosIO (commented):
will add this to the sign in/sign out button as well

PR #108, opened 2021-10-12
Add tests and UI for notifications

stevekinney:
Adds notifications to the UI. Design is subject to change.

Screen Shot 2021-10-12 at 5 01 20 PM

### DIFF
--- a/src/lib/components/notifications.svelte
+++ b/src/lib/components/notifications.svelte
@@ -1,0 +1,58 @@
+<script lang="ts">
+  import { notifications } from '$lib/stores/notifications';
+  import { crossfade } from 'svelte/transition';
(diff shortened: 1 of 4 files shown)


src/lib/components/notifications.svelte
```diff
@@ -0,0 +1,58 @@
...
+
+    fallback(node) {
+      const style = getComputedStyle(node);
+      const transform = style.transform === 'none' ? '' : style.transform;
+
+      return {
+        duration: 600,
+        easing: quintOut,
+        css: (t) => `
+					transform: ${transform} scale(${t});
```
feedmeapples, line 18: nit: curious what the arguments for tabs vs spaces would be for cases when the code/markup is plain text

PR #123, opened 2021-10-21
Create a generalized button component

softwarecurator:
## What was changed
Changed all existing buttons to be used as a more generalized button with same stylings across the application 

## Why?
apart of sprint

### DIFF
--- a/src/lib/components/main-button.svelte
+++ b/src/lib/components/main-button.svelte
@@ -6,9 +6,7 @@
   export let active: boolean = false;
 </script>
 
-<button on:click class={styles} {disabled} class:active
-  ><slot /></button
->
+<button on:click class={styles} {disabled} class:active><slot /></button>
 
 <style lang="postcss">
   button {


src/lib/components/main-button.svelte
```diff
@@ -0,0 +1,31 @@
...
+  //Styles passed along to custom Button Component
+  export let styles: string = '';
+  export let disabled: boolean = false;
+  //Function Passed along to component that is used as a on:click
+  export let onClick;
+  //Boolean to state if button is active or not
+  export let active: boolean = false;
+</script>
+
+<button on:click={onClick} class={styles} {disabled} class:active
```
stevekinney, line 11: You can probably use [event forwarding](https://svelte.dev/docs#Element_directives) here instead of passing the prop through. For example:

```js
             👇
<button on:click class={styles} {disabled} class:active>
  <slot />
</button>
```

stevekinney (commented):
It might be interesting to look at the API and all of the use cases from a button in another framework. (E.g. https://paste.twilio.design/components/button)

PR #133, opened 2021-11-02
Button Component

josephriosIO:
## What was changed
A basic outline to a button component that has a few features

`<Button variant size disabled >`

Variants
- Primary
- Secondary
- Destroy

Size
- Small
- Large

Disabled and onclick works


src/lib/components/button/main-button.svelte
```diff
@@ -0,0 +1,91 @@
+<script lang="ts">
+  const getButtonState = (disabled?: boolean) => {
+    if (disabled) {
+      return 'disabled';
+    }
+    return 'default';
+  };
+
+  const getButtonStyle = (variant) => {
```
stevekinney, line 9: Is this any different than just returning the string directly?

We could probably just enforce this with TypeScript:

```ts
const getButtonStyle = (variant: 'primary' | 'secondary | 'destroy') => variant;
```

But, even then, there is probably a simpler way express this that I'll stumble across later on in the code.

src/lib/components/button/main-button.svelte
```diff
@@ -0,0 +1,91 @@
...
+        return 'secondary';
+      case 'destroy':
+        return 'destroy';
+      case 'primary':
+      default:
+        return 'primary';
+    }
+  };
+
+  const getButtonSize = (size) => {
```
stevekinney, line 21: The same goes here as well.

src/lib/components/button/main-button.svelte
```diff
@@ -0,0 +1,91 @@
...
+      case 'small':
+        return 'small';
+      case 'default':
+      default:
+        return 'default';
+    }
+  };
+
+  export let disabled: boolean = false;
+  export let variant: string = '';
```
stevekinney, line 32: Do we actually _need_ `getButtonStyle` or can we have TypeScript just handle this here?

```ts
export let variant: 'primary' | 'secondary | 'destroy' = 'secondary';
```

This will enforce the correct property and likely made it so that you can completely get rid of `getButtonStyle`.

src/lib/components/button/main-button.svelte
```diff
@@ -0,0 +1,91 @@
...
+  export let active: boolean = false;
+  export let size: string = 'default';
```
stevekinney, line 34: The same goes here as well.

src/lib/components/button/main-button.svelte
```diff
@@ -0,0 +1,91 @@
...
+
+  const buttonState = getButtonState(disabled);
+  const showDisabled = buttonState !== 'default';
+</script>
+
+<button
+  on:click
+  disabled={showDisabled}
+  class={`${getButtonSize(size)} ${getButtonStyle(variant)}`}
```
stevekinney, line 43: Now we can probably just use the props directly.

src/lib/components/button/main-button.svelte
josephriosIO: since all of these don't need to be set on the button "default" settings is there another way around this?

stevekinney (approved):

PR #138, opened 2021-11-22
Remove unneeded base-64 dependency

GiantRobots:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


stevekinney (approved):
Love it. Thanks!

PR #142, opened 2021-11-30
Render all the colors in our theme in a nice way so we can iterate on our color scheme

GiantRobots:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?
localhost_3000_styleguide_colors (1)


stevekinney:
I'm not totally sure that we want to put this page in the application itself, do we?

GiantRobots:
Fixed!

src/routes/styleguide/__layout.reset.svelte
```diff
@@ -0,0 +1,25 @@
+<script context="module" lang="ts">
+  import { mode } from '$app/env';
+
+  /** @type {import('@sveltejs/kit').Load} */
```
stevekinney, line 4: You don't need to to use the JSDoc syntax, you can use Typescript.

```js
<script context="module" lang="ts">
  import type { LoadInput } from '@sveltejs/kit';

  export async function load({ fetch }: LoadInput) {
    // …
  }
</script>
```

src/routes/styleguide/__layout.reset.svelte
```diff
@@ -0,0 +1,25 @@
...
+  export async function load({}) {
+    let responseObj: any = {
```
stevekinney, line 6: A tiny nit that you can ignore: Why even store this in a variable? You _could_ just return the error if they're not in development and then just go about your business.

```js
import type { LoadInput } from '@sveltejs/kit';

export async function load({ fetch }: LoadInput) {
   if (mode !== 'development') return { status: 404, error: 'Not Found' };

  return { status: 200 };
}
```

src/routes/styleguide/__layout.reset.svelte
```diff
@@ -0,0 +1,25 @@
...
+      status: 200,
+    };
+
+    if (mode !== 'development') {
+      responseObj = {
+        status: 404,
+        error: new Error('Could not find page'),
```
stevekinney, line 13: I'm pretty sure that this will actually throw an exception in this function and blow up (resulting in a 500 with a stack trace and not a 404). You probably just want to set the error property on the object as a string.

src/routes/styleguide/__layout.reset.svelte
```diff
@@ -0,0 +1,25 @@
...
+      };
+    }
+
+    return responseObj;
+  }
+</script>
+
+<script lang="ts">
+  import '../../app.postcss';
```
stevekinney, line 22: Do we have a path alias for this?

src/routes/styleguide/components/ColorBox.svelte
```diff
@@ -0,0 +1,12 @@
+<script lang="ts">
```
stevekinney, line 1: This kind of relates to what we were talking about yesterday: We now have an extra component that we're not _actually_ using. Over time this is just going to add a bit of friction and confusion. I can see something like this in something like a storybook (I'm using lowercase here because I don't care if it's _actually_ Storybook, but just the general idea), but it doesn't belong in the actual codebase.

src/routes/styleguide/components/ColorRenderer.svelte
```diff
@@ -0,0 +1,23 @@
+<script lang="ts">
```
stevekinney, line 1: Secondly, the `src/routes` founder is meaningful in SvelteKit as it adds it to the router. This means that a user _could_ navigate to `/styleguide/components/ColorRenderer` as well as `/styleguide/components`, which will error.

Shared components used throughout the codebase can go into `$lib/components` (this is a SvelteKit convention).

If you think you're only going to use a component with a given route, you can prefix it with an underscore (e.g. `_ColorRenderer.svelte`.

Also: We're currently going with kebab case (e.g. `color-renderer.svelte`. We can argue about whether or not that is the right choice for Svelte components, but for now we might as well stay consistent.

src/routes/styleguide/components/ColorRenderer.svelte
```diff
@@ -0,0 +1,23 @@
...
+  // This should be a circular reference but am Lazy and couldn't get it working
```
stevekinney, line 2: This should work:

```ts
type ColorValues = { [key: string]: string | ColorValues } 
```

theme.ts
```diff
@@ -0,0 +1,21 @@
+// Ignore this rule because tailwind is all cjs and we need to use requires here
+// Maybe come back to this but this seems to be the easy way out
+/* ts-ignore:@typescript-eslint/no-var-requires */
```
stevekinney, line 3: The other option is just to use JavaScript here. Alternatively, I'm not against just keeping this logic in `tailwind.config.cjs` for now. Other than the `omit` function that we're not actively using right now, we're not doing anything in this file.

stevekinney: Particularly because this is the file that is [failing the build](https://github.com/temporalio/ui/runs/4361170444?check_suite_focus=true#step:4:17).

src/routes/styleguide/__layout.reset.svelte
GiantRobots: I was copying what we were doing in our main __layout page but it seems that we get this injected another way. I deleted this import and all the tailwind css worked fine.

src/routes/styleguide/__layout.reset.svelte
GiantRobots: https://kit.svelte.dev/docs#loading

That's how they do the error message in the example and it seems to work. 

Throwing an error here results in an error page bubble up to the root error page (which we don't have) which eventually will throw a 500. I tried hitting some random routes and they hold onto the connection until they 500 out so we probably need to look into that eventually.

theme.ts
GiantRobots: I was waiting on choosing the colors later with Candace and brought the omit function because I figured we'd end up using it. I'll probably pick instead of omit in our more final version.

PR #148, opened 2021-12-01
Merge summary and events page for an individual workflow execution

stevekinney:
To match better with the new designs the existing summary and events page have been combined. Also: client-side sorting and filtering of events has been removed for now. The activities page is temporarily gone as we rework the state management.


josephriosIO (approved):
**Thano's snap**

GiantRobots (approved):

PR #149, opened 2021-12-01
empty workflow state

josephriosIO:
## What was changed
Changed the empty workflow state to match figma

## Image
Screen Shot 2021-12-01 at 12 46 39 PM

# NOTE
Images are emojis for now (placeholders)


src/routes/namespaces/[namespace]/workflows/index.svelte
```diff
@@ -43,18 +43,20 @@
...
-          {:else}
-            <WorkflowsEmptyState />
-          {/each}
-        {/await}
-      </tbody>
-    </WorkflowsSummaryTable>
+    {#await data}
+      <WorkflowsLoadingState />
+    {:then { workflows }}
+      {#if workflows.length > 0}
```
Comment (stevekinney) [inline src/routes/namespaces/[namespace]/workflows/index.svelte:49]: Tiny nit… you can just do `workflows.length` since zero is falsey and everything else is truthy.

PR #151, opened 2021-12-02
Filter by execution status

stevekinney:
Add the ability to filter by execution status.

### DIFF
--- a/src/lib/services/workflow-service.ts
+++ b/src/lib/services/workflow-service.ts
@@ -46,9 +46,10 @@ const emptyWorkflowRequest = (): Promise<ListWorkflowExecutionsResponse> => {
 
 export const fetchOpenWorkflows = (
   namespace: string,
-  { timeRange }: TimeRangeParameter,
+  { timeRange, status }: TimeRangeParameter & StatusParameters,
   request = fetch,
 ): Promise<ListWorkflowExecutionsResponse> => {
+  if (status && status !== 'Running') return emptyWorkflowRequest();
   return requestFromAPI<ListWorkflowExecutionsResponse>(
     `/namespaces/${namespace}/workflows/open`,
     {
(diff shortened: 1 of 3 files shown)


src/lib/services/workflow-service.ts
```diff
@@ -54,16 +63,39 @@ export const fetchWorkflowsByType = (
...
 
+export const fetchClosedWorkflows = (
+  namespace: string,
+  { timeRange, status }: TimeRangeParameter & StatusParameters,
+  request = fetch,
+): Promise<ListWorkflowExecutionsResponse> => {
+  if (status === 'Running') return emptyWorkflowRequest();
+
+  const params: Record<string, string> = {
+    'start_time_filter.earliest_time': createDate(timeRange || { hours: 24 }),
```
GiantRobots, line 74: Why || instead of ??

src/lib/utilities/get-workflow-status-filter-code.ts
```diff
@@ -0,0 +1,14 @@
...
+  status: WorkflowStatus,
+): ExecutionStatusCodes => {
+  if (status === 'Running') return '1';
+  if (status === 'Completed') return '2';
+  if (status === 'Failed') return '3';
+  if (status === 'Canceled') return '4';
+  if (status === 'Terminated') return '5';
+  if (status === 'ContinuedAsNew') return '6';
+  if (status === 'TimedOut') return '7';
+  return null;
```
GiantRobots, line 13: Note: this looks like an object lookup and you could probably do something like 
```typescript
let statusLookup = {
"running": 1
...
}

return statusLookup[status] ?? null;
```

But I lack context as to what is going on here. (I think a switch statement would also do exhaustiveness checking but ¯\_(ツ)_/¯ )

src/routes/namespaces/[namespace]/workflows/index.svelte
```diff
@@ -6,10 +6,11 @@
 
     const namespace = page.params.namespace;
     const timeRange = page.query.get('time-range');
+    const status = page.query.get('status') as WorkflowStatus;
```
Comment (GiantRobots) [inline src/routes/namespaces/[namespace]/workflows/index.svelte:9]: Should this as include null/undefined? in the case it doesn't exist in the query param? Or is it cool to pass that into the endpoints?

GiantRobots (approved):

src/routes/namespaces/[namespace]/workflows/index.svelte
Comment (stevekinney) [reply src/routes/namespaces/[namespace]/workflows/index.svelte:9]: `fetchClosedWorkflows` only adds it to the request if it's present since it's actively missing if we want everything.

PR #152, opened 2021-12-02
Get summary attributes for events

josephriosIO:
## What was changed
Gets summary events for for summary view


src/lib/utilities/find-value-if-it-includes-word.ts
```diff
@@ -0,0 +1,8 @@
+export const findValueIfItIncludesWord = (object, word) => {
+    for (var property in object) {
+      if (object.hasOwnProperty(property) && 
+         property.toString().includes(word)) {
+         return object[property];
+      }
+    }
+  }
```
GiantRobots: No newline on EOF

https://thoughtbot.com/blog/no-newline-at-end-of-file

Talks about why it matters, and also solutions for various editors.

src/lib/utilities/history-summary-plucker.ts
```diff
@@ -0,0 +1,93 @@
...
+};
+
+const getHistorySummary = (event) => {
+  const eventDetails = findValueIfItIncludesWord(event, 'EventAttributes');
+  return isKnownEventType(event.eventType)
+    ? summaryExtractors[event.eventType](eventDetails)
+    : eventDetails;
+};
+
+export { getHistorySummary };
```
GiantRobots, line 93: I think this file needs prettier ran on it.

src/lib/utilities/history-summary-plucker.ts
```diff
@@ -0,0 +1,93 @@
...
+
+const isKnownEventType = (eventType) => {
+  return eventType in summaryExtractors;
+};
+
+const getHistorySummary = (event) => {
+  const eventDetails = findValueIfItIncludesWord(event, 'EventAttributes');
+  return isKnownEventType(event.eventType)
+    ? summaryExtractors[event.eventType](eventDetails)
+    : eventDetails;
```
GiantRobots, line 90: You could simplify this line to be 

```typescript
return summaryExtractors[event.eventType](eventDetails) ?? eventDetails
```

The ?? is basically a `if (summaryExtractors[event.eventType](eventDetails)) == null/undefined` then use eventDetails. useful in statements with ?.

src/lib/utilities/history-summary-plucker.ts
```diff
@@ -0,0 +1,93 @@
...
+    input: d.input,
+    Name: d.activityType.name,
+  }),
+  ActivityTaskStarted: (d) => ({
+    attempt: d.attempt,
+    identity: d.identity,
+    requestId: d.requestId,
+  }),
+  ActivityTaskTimedOut: (d) => ({
+    'Timeout Type': d.failure?.timeoutFailureInfo?.timeoutType,
```
GiantRobots, line 22: What would the default value for this be? If the `?.` fails it will have undefined as a value so you'll return `{"Timeout Type": undefined}` not sure what the consuming code is like. Will that be okay?

src/lib/utilities/history-summary-plucker.ts
```diff
@@ -0,0 +1,93 @@
...
+    'Fire Timeout': d.startToFireTimeout?.duration,
+    'Timer ID': d.timerId,
+  }),
+  WorkflowExecutionStarted: (d) => {
+    const summary = {
+      'Close Timeout': d.workflowRunTimeout,
+      identity: d.identity,
+      input: d.input,
+      Parent: undefined,
+      Workflow: d.workflowType ? d.workflowType.name : '',
```
GiantRobots, line 61: Prefer `d.workflowType?.name ?? ''`

src/lib/utilities/history-summary-plucker.ts
```diff
@@ -0,0 +1,93 @@
...
+    };
+
+    return summary;
```
GiantRobots, line 64: Why use a return here instead of the `()=>({})` everywhere else?

src/lib/utilities/history-summary-plucker.ts
```diff
@@ -0,0 +1,93 @@
+import { findValueIfItIncludesWord } from './find-value-if-it-includes-word';
+
+const summaryExtractors = {
+  ActivityTaskCancelRequested: (d) => ({ Id: d.activityId }),
```
GiantRobots, line 4: I'm not really sure what `d` is in all of these. Can we use a more descriptive name?

GiantRobots: This code is also super unsafe from a type perspective. All of this would benefit greatly from typing the incoming value 

```typescript
  ActivityTaskCancelRequested: (d:WhateverNetworkValueThisRepresents) => ({ Id: d.activityId }),
```

src/lib/utilities/find-value-if-it-includes-word.ts
```diff
@@ -0,0 +1,7 @@
+export const findValueIfItIncludesWord = (object, word) => {
+  for (var property in object) {
```
GiantRobots, line 2: Typically we avoid var these days. 

https://levelup.gitconnected.com/stop-using-var-to-declare-variables-in-javascript-6c0caec16f43

Outlines some basics as to the why.

This is basically just checking if an object has a key with a value in it. You can generally check that like..

```typescript
const myRadObject= {
mojo: "sodope"
};
console.log(myRadObject["mojo"] ?? "Not Dope")
> "sodope"
console.log(myRadObject["Magical"] ?? "Not Dope")
> "Not Dope"
```

src/lib/utilities/history-summary-plucker.ts
```diff
@@ -0,0 +1,93 @@
...
+  ActivityTaskCompleted: (d) => ({ result: d.result }),
```
stevekinney, line 5: For a bunch of these, we can probably have a utility method like `pick`.

```js
const pick = (source, ...keys) => keys.reduce((result, key) => {
  result[key] = source[key];
  return result;
}, {});
```

Or—if you prefer:

```js
const pick = (source, ...keys) => {
  const result = {};
  for (const key of keys) {
    result[key] = source[key];
  };
  return result;
}
```

Then, you could just do something like:

```js
const summaryExtractors = {
  // …
  ActivityTaskCompleted: (d) => pick(d, 'result'),
  ActivityTaskFailed: (d) => pick('details', 'reason'),
  // …
}
```

Basically, we want to avoid having bugs do to typos.

You could add some additional type safety even…

```ts
function pick<T, K extends keyof T>(source: T, ...keys: K[]): Pick<T, K> {
  const result: any = {};
  for (const key of keys) {
    result[key] = source[key];
  }
  return result;
}
```

This will stop us is from accidentally typing a key that does not exist on the source object.

There is probably even fancier stuff we could do with currying but the gymnastics we'd need to do with types probably isn't worth it.

src/lib/utilities/find-value-if-it-includes-word.ts
GiantRobots: This can also probably be avoided in your code if you have types in your code.

src/lib/utilities/history-summary-plucker.ts
```diff
@@ -0,0 +1,93 @@
...
+  ChildWorkflowExecutionStarted: (d) => ({
+    Workflow: d,
+  }),
+  WorkflowTaskCompleted: (d) => ({ identity: d.identity }),
+  WorkflowTaskScheduled: (d) => ({
+    Taskqueue: d.taskQueue.name,
+    Timeout: d.startToCloseTimeout,
+  }),
+  WorkflowTaskStarted: (d) => ({ requestId: d.requestId }),
+  WorkflowTaskTimedOut: (d) => ({ 'Timeout Type': d.timeoutType }),
```
stevekinney, line 37: I don't think we need to worry about the inconsistent capitalization from the current UI. We can just pick properties as they show up in the JSON.

src/lib/utilities/history-summary-plucker.ts
stevekinney, line 64: We can probably just return the object like the rest.

src/lib/utilities/history-summary-plucker.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    return summary;
+  },
+  WorkflowExecutionCompleted: (d) => ({
+    result: d.result,
+    newExecutionRunId: d.newExecutionRunId || undefined,
+  }),
+  WorkflowExecutionFailed: (d) => ({
+    message: d.failure?.message,
+    newExecutionRunId: d.newExecutionRunId || undefined,
```
stevekinney, line 72: Wouldn't this be `undefined` anyway?

src/lib/utilities/history-summary-plucker.ts
```diff
@@ -0,0 +1,93 @@
...
+  }),
+  WorkflowExecutionTimedOut: (d) => ({
+    retryState: d.retryState,
+    newExecutionRunId: d.newExecutionRunId || undefined,
+  }),
+  WorkflowTaskFailed: (d) => ({ message: d.failure?.message }),
+  ChildWorkflowExecutionFailed: (d) => ({ message: d.failure?.message }),
+};
+
+const isKnownEventType = (eventType) => {
```
stevekinney: TypeScript should be able to handle this for us.

src/lib/utilities/history-summary-plucker.ts
stevekinney, line 87: You can probably just do `export const` on Line 86.

src/lib/utilities/find-value-if-it-includes-word.ts
GiantRobots: Also generally avoid `for( .. in ... )`. Prefer `for .. of` because it doesn't enumerate over object fields

https://stackoverflow.com/questions/500504/why-is-using-for-in-for-array-iteration-a-bad-idea

For of docs

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

Also also looking up objects this way is `O(N)` ie. it will iterate over every item in an object to find that instead of `O(1)` which will access the key directly. 

To put it another way. If you have an object with 10,000 keys you can do 10,000 checks or you can do 1 check depending on how you access the object.

src/lib/utilities/find-value-if-it-includes-word.ts
```diff
@@ -0,0 +1,7 @@
+export const findValueIfItIncludesWord = (object, word) => {
```
stevekinney, line 1: We already have a more TypeScript-friend (if also more verbose) version of this in `src/lib/utilities/get-component-for-event-type.ts`.

src/lib/utilities/history-summary-plucker.ts
stevekinney: @GiantRobots taught me about this today. I literally didn't know it existed.

src/lib/utilities/history-summary-plucker.ts
```diff
@@ -0,0 +1,87 @@
...
+  WorkflowExecutionTimedOut: (d) => ({
+    retryState: d.retryState,
+    newExecutionRunId: d.newExecutionRunId || undefined,
+  }),
+  WorkflowTaskFailed: (d) => ({ message: d.failure?.message }),
+  ChildWorkflowExecutionFailed: (d) => ({ message: d.failure?.message }),
+};
+
+const getHistorySummary = (event) => {
+  const eventDetails = findValueIfItIncludesWord(event, 'EventAttributes');
```
GiantRobots, line 83: eventDetails can be undefined here which may cause errors when using it with other functions

GiantRobots (changes requested):
history-summary-plucker is really difficult to understand without types and we should probably add a bunch

src/lib/utilities/history-summary-plucker.ts
GiantRobots: You can also do picks with destructuring like...

```typescript
  ActivityTaskFailed: ({details, reason}) => ({details, reason})
```

Although pick may be nicer.

src/lib/utilities/history-summary-plucker.test.ts
```diff
@@ -0,0 +1,27 @@
...
+    };
+
+    expect(getHistorySummary(event)).toStrictEqual({
+      'Close Timeout': '100s',
+      identity: 'some identity',
+      input: 'true',
+      Parent: undefined,
+      Workflow: 'Rainbow Workflow',
+    });
+  });
```
GiantRobots, line 26: When you add types to everything you can check if there are type errors with ts expect error comments

https://devblogs.microsoft.com/typescript/announcing-typescript-3-9-beta/#ts-expect-error-comments

GiantRobots (changes requested):

stevekinney (approved):

PR #153, opened 2021-12-02
Events tab bar changes

josephriosIO:
## What was changed
Made the tab bar look more like new UI without the pieces we don't have yet

# Images
Screen Shot 2021-12-02 at 3 18 06 PM


tailwind.config.cjs
```diff
@@ -5,6 +5,8 @@ const TemporalColors = {
   secondary: '#64748b',
   // Custom Colors
   primaryText: '#141414',
+  historybg: '#DCE4FB',
```
stevekinney, line 8: Do we want to capitalize the `b`?

stevekinney:
I love it… I just have the tiniest nit.

tailwind.config.cjs
josephriosIO: done and done

tailwind.config.cjs
```diff
@@ -5,6 +5,8 @@ const TemporalColors = {
   secondary: '#64748b',
   // Custom Colors
   primaryText: '#141414',
+  historyBg: '#DCE4FB',
```
stevekinney, line 8: Do we want to have a token dedicated to something so small? /cc @GiantRobots

canvanooo: It's technically the same as the Running status badge. We do use that pattern elsewhere..

josephriosIO: I can rename it or we can move it to just in a style sheet

src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/_tabs.svelte
```diff
@@ -1,29 +1,38 @@
...
+    @apply block;
   }
 
   a.active {
-    @apply border-b-8 border-purple-400 font-bold;
+    @apply border-b-2 rounded-b-sm border-primary font-medium;
+  }
+
+  span {
+    background-color: #dce4fb;
```
Comment (stevekinney) [inline src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/_tabs.svelte:35]: @canvanooo Should we just use `blue-700` and `blue-100` here?

PR #154, opened 2021-12-02
Data converter

GiantRobots:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


src/lib/stores/decrypt.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export const decryptPort = persistStore('port', null);
+
+// An enum or string union may be more appropriate here so we dont' get a
+// if thing and the falsiness evaluates to the null case, but start with this
+export const lastDecryptStatus = writable<null | boolean>(null);
+
+export const portNumber = writable<null | number>(null);
+
+export function lastDecryptFailure() {
```
stevekinney, line 12: Do you feel like the name of this function sounds like a property? But then it's really a setter?

src/lib/stores/decrypt.ts
```diff
@@ -0,0 +1,45 @@
+import { browser } from '$app/env';
+import { writable } from 'svelte/store';
+
+export const decryptPort = persistStore('port', null);
+
+// An enum or string union may be more appropriate here so we dont' get a
+// if thing and the falsiness evaluates to the null case, but start with this
+export const lastDecryptStatus = writable<null | boolean>(null);
```
stevekinney, line 8: You don't need `null` in the union type. It'll still work.

src/lib/stores/decrypt.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  const { subscribe, set } = writable(initialValue);
+
+  return {
+    subscribe,
+
+    set: (x) => {
+      if (browser) {
+        if (window?.localStorage) {
+          window.localStorage.setItem(name, x);
```
stevekinney, line 39: Would `window?.localStorage.setItem(name, x)` work here? And then we could skip the conditional?

src/lib/stores/decrypt.ts
```diff
@@ -0,0 +1,45 @@
...
+      }
+    }
+  }
+
+  const { subscribe, set } = writable(initialValue);
+
+  return {
+    subscribe,
+
+    set: (x) => {
```
stevekinney, line 36: Does `x` need a type?

src/routes/_header.svelte
```diff
@@ -15,6 +15,10 @@
 
 <script lang="ts">
   import { namespace } from '$lib/stores/namespace';
+  import { StatusOnline, StatusOffline } from 'svelte-hero-icons';
```
stevekinney, line 18: This is totally cool for now, but let's make a mental note that Candace has new icons coming.

src/lib/stores/decrypt.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export function lastDecryptFailure() {
+  lastDecryptStatus.set(false);
+}
+
+export function lastDecryptSuccess() {
+  lastDecryptStatus.set(true);
+}
+
+export function persistStore(name, initialValue) {
```
stevekinney, line 20: Do you like this name?

src/routes/_header.svelte
```diff
@@ -25,6 +29,23 @@
 >
   <NamespaceSelect />
   <div class="flex flex-row justify-center items-center">
+    {#if $decryptPort}
+      {#if $lastDecryptStatus === null}
```
stevekinney, line 33: Versus checking if something is null, should our logic be the opposite?

- `true`
- `false`
- `null`

I also kind of feel like having both `null` and `false` will eventually cause a bug like 9 months from now.

What if it _only_ could return `true` or `false` and we had another method for `$lastDecryptStatusExists` or something like that?

src/lib/stores/decrypt.ts
GiantRobots: Yeah, good call!

It needs another ?. in there after trying it out

`window?.localStorage?.setItem(name, x)`

src/lib/stores/decrypt.ts
GiantRobots: Setting it to string sounds like a reasonable call.

src/routes/_header.svelte
GiantRobots: Yup! I've talked to Candace about the how this works and the icons I'm using. Also about the future state of how this will be decrypted potentially on a per field basis, instead of the every field at once thing we have going on now.

src/lib/stores/decrypt.ts
GiantRobots: No, but I couldn't figure a better one. Open to suggestions

src/routes/_header.svelte
GiantRobots:
```// An enum or string union may be more appropriate here so we dont' get a
// if thing and the falsiness evaluates to the null case, but start with this
export const lastDecryptStatus = writable<null | boolean>(null);```

I totally have this as a comment for this store XD. I agree probably an enum/string union would be best.

`'unrequested' | 'success' | 'error'`

Was kind of what I was thinking.

src/lib/stores/decrypt.ts
GiantRobots: `persistentStore` `persistStoreToLocalStorage` `localStoragStore` are all I've got

PR #156, opened 2021-12-06
Data Converter 

GiantRobots:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


src/lib/services/workflow-service.ts
```diff
@@ -121,9 +124,24 @@ export async function fetchWorkflowWithEventHistory(
...
 ): Promise<{ workflow: WorkflowExecution; events: HistoryEventWithId[] }> {
+  let port = get(webDecoderPort);
+  // if port rawPayloads = true
+
   const [workflow, events] = await Promise.all([
     fetchWorkflow(parameters, request),
-    fetchEvents(parameters, request).then(toEventHistory),
+    fetchEvents(parameters, request)
+      .then(toEventHistory)
+      .then((events) => {
```
stevekinney, line 134: We seem to return `events` no matter what?

src/lib/services/workflow-service.ts
```diff
@@ -121,9 +124,24 @@ export async function fetchWorkflowWithEventHistory(
...
+        if (port === null) {
+          return events;
+        }
+
+        webDecodeEventPayloads(events, port).catch((error) => {
```
stevekinney, line 139: Should we be returning this? That promise's `return` will never do anything (it will just `return` into the abyss) and the `catch` will swallow the error.

src/lib/utilities/is-promise.ts
```diff
@@ -0,0 +1,11 @@
+export function isPromise(promiseTentially: any): Boolean {
```
stevekinney, line 1: The type should probably be `unknown` as opposed to any. And we probably want to use the `is` operator the the response.

```js
function isPromise(obj: unknown) obj is Promise<unknown> {
  return !!obj && (typeof obj === 'object' || typeof obj === 'function') && typeof obj.then === 'function';
}
```

src/routes/data-converter/[port].svelte
```diff
@@ -0,0 +1,17 @@
...
+<script lang="ts">
+  import { goto } from '$app/navigation';
+  import { page } from '$app/stores';
+  import { notifications } from '$lib/stores/notifications';
+  import { onMount } from 'svelte';
+  import { webDecoderPort } from '$lib/stores/web-decoder-config';
+  onMount(() => {
+    let { port } = $page.params;
+    webDecoderPort.set(port);
+    goto('/');
```
Comment (stevekinney) [inline src/routes/data-converter/[port].svelte:14]: Do you want to use `{ replaceState: true }`? I'm not sure.

src/lib/services/web-decoder.ts
```diff
@@ -0,0 +1,92 @@
...
+                case 'workflowExecutionStartedEventAttributes':
+                case 'activityTaskScheduledEventAttributes':
+                  event.attributes.input.payloads = currentPayload;
+                  break;
+              }
+            }),
+        );
+      });
+    });
+    await Promise.all(requests);
```
stevekinney, line 79: Do you want to return this? Otherwise any errors will be swallowed.

src/lib/services/web-decoder.ts
```diff
@@ -0,0 +1,92 @@
...
+          payloads = event.attributes.result.payloads;
+          break;
+        case 'signalExternalWorkflowExecutionInitiatedEventAttributes':
+        case 'workflowExecutionStartedEventAttributes':
+        case 'activityTaskScheduledEventAttributes':
+          payloads = event.attributes.input.payloads;
+          break;
+      }
+
+      console.log({ payloads });
```
stevekinney, line 49: 🪵

src/lib/services/web-decoder.ts
```diff
@@ -0,0 +1,92 @@
...
+      extractRequestId: (data) => data && data.requestId,
+    });
+  } catch (err) {
+    setLastWebDecoderFailure();
+    return Promise.reject(err);
+  }
+
+  try {
+    await sock.open();
+    console.log('opened', sock);
```
stevekinney, line 31: 🪵

src/lib/services/workflow-service.ts
GiantRobots: Good call this is definitely erroneous code, I was slinging this snippet around and it is definitely incorrect here.

src/lib/utilities/is-promise.ts
```diff
@@ -0,0 +1,12 @@
+export function isPromise(obj: any): obj is Promise<unknown> {
```
stevekinney, line 1: I know we talked about this a bit, but this function and its test file is what's failing the build. I wonder if we _need_ this functionality or if it's a nice to have that's adding complexity?

Do we _need_ it?

src/lib/utilities/is-promise.test.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  it('should return false if an object is partially promise like with a .catch', () => {
+    expect(isPromise({ catch: () => {} })).toBe(false);
+  });
+
+  it('should return false if an object is partially promise like with a .finally', () => {
+    expect(isPromise({ finally: () => {} })).toBe(false);
+  });
+
+  it('should return false if an object is promise like (contains .then, .catch. finally)', () => {
```
stevekinney, line 34: Why? If it adheres to the contract, do we care?

GiantRobots: I'm happy to remove the `obj.toString() === [object Promise]` because that's what the last check is. I'm happy to have it be a promise if it has those methods.

src/lib/services/workflow-service.ts
stevekinney: If you make this an `async`/`await`, then I think you can spare yourself all of the `Promise.resolve`s. Not a blocker. This code could probably use some refactoring, but it's not our biggest priority at this moment.

src/lib/utilities/is-promise.test.ts
```diff
@@ -0,0 +1,81 @@
+import { isPromise } from './is-promise';
```
stevekinney, line 1: It doesn't look like we're using this code at all? Let's cut it for now.

stevekinney (commented):
Let's remove `isPromise` for now if we're not using it.

src/lib/services/workflow-service.ts
GiantRobots: Yeah the convertEventPayloadFromDataConverter mutates the events object (not ideal) but the way the code was structured made this less than ideal. We can update this with a better access pattern using some more of Svelte's primitives in the future. I'm 100% in agreement this is not great code.

We've talked about what we want the future state to be and it doesn't really look anything like this.

stevekinney (approved):

PR #159, opened 2021-12-07
Workflow table

josephriosIO:
## What was changed
Matches figma (not pixel perfect) with screen scrolling


src/routes/namespaces/[namespace]/workflows/_workflows-summary-row.svelte
```diff
@@ -33,18 +33,18 @@
...
   </td>
   <td>
-    <a sveltekit:noscroll {href} class="font-mono text-right">
+    <a sveltekit:noscroll {href} class="text-left">
       <Time time={workflow.startTime} {timeFormat} />
     </a>
   </td>
   <td>
-    <a sveltekit:noscroll {href} class="font-mono text-right">
+    <a sveltekit:noscroll {href} class="text-let">
```
Comment (stevekinney) [inline src/routes/namespaces/[namespace]/workflows/_workflows-summary-row.svelte:47]: Should this be `text-left`?

src/routes/namespaces/[namespace]/workflows/_workflows-summary-row.svelte
```diff
@@ -56,15 +56,15 @@
   }
 
   tr {
-    @apply bg-gray-50;
+    @apply border-2;
   }
 
   tr:hover {
-    @apply bg-gray-100;
+    background-color: #e6fffa;
```
Comment (stevekinney) [inline src/routes/namespaces/[namespace]/workflows/_workflows-summary-row.svelte:63]: I'd just pick the closest Tailwind color.

stevekinney (commented):
Just a few minor tweaks.

src/routes/namespaces/[namespace]/workflows/_workflows-summary-row.svelte
Comment (josephriosIO) [reply src/routes/namespaces/[namespace]/workflows/_workflows-summary-row.svelte:47]: Oops yes let me push up the new changes in a few this reverts back the changes in the last commit back to a window type view

src/routes/namespaces/[namespace]/workflows/_workflows-summary-row.svelte
Comment (josephriosIO) [reply src/routes/namespaces/[namespace]/workflows/_workflows-summary-row.svelte:63]: Noted

src/routes/namespaces/[namespace]/workflows/index.svelte
```diff
@@ -50,9 +51,9 @@
       {#if workflows.length}
         <WorkflowsSummaryTable>
           <tbody slot="rows">
-            {#each workflows as workflow}
-              <WorkflowsSummaryRow {workflow} {timeFormat} />
-            {/each}
+            <VirtualList items={workflows} let:item>
```
Comment (stevekinney) [inline src/routes/namespaces/[namespace]/workflows/index.svelte:54]: This is kind of what we talked about the other day. I'm not convinced we can use a `<table>` for this view anymore. It the `VirtualList` might add something between the table heading and the table body that breaks the natural layout. I'm pretty sure you're going to have to fall back to an another element.

stevekinney (commented):
You'll definitely want to get a design review. There seems to be some issues with the borders around the tables as well as the alignment of the headings. I think you're not going to be able to use a table for this view.

src/routes/namespaces/[namespace]/workflows/_workflows-summary-table.svelte
```diff
@@ -1,18 +1,17 @@
...
+    <div class="w-2/12 text-left">End</div>
+  </div>
+  <slot />
+</section>
 
 <style lang="postcss">
-  th {
-    @apply bg-gray-200 text-gray-500 text-xs h-6 m-0 p-3 uppercase table-fixed;
+  .workflow-table {
+    display: block;
```
Comment (stevekinney) [inline src/routes/namespaces/[namespace]/workflows/_workflows-summary-table.svelte:14]: I'm not saying we should _never_ write custom CSS properties that aren't from Tailwind, but it's definitely something we want to avoid when possible.

There is a Tailwind class called `block` that you can just toss into the list of classes above.

src/routes/namespaces/[namespace]/workflows/_workflows-summary-table.svelte
```diff
@@ -1,18 +1,17 @@
...
+    height: calc(100vh - 25%);
```
Comment (stevekinney) [inline src/routes/namespaces/[namespace]/workflows/_workflows-summary-table.svelte:15]: This definitely has some weird implications on the layout. When we scroll down, there is a lot of extra space. We probably are going to want to get rid of this. I'm not going to block this PR over it though.

Now that we have no sidebar and we're about to remove the navigation on the other side, we can probably simplify the layout and remove this later.

src/routes/namespaces/[namespace]/workflows/_workflows-summary-table.svelte
Comment (stevekinney) [reply src/routes/namespaces/[namespace]/workflows/_workflows-summary-table.svelte:14]: But even then–it's a `section`, which should be a block by default.

src/routes/namespaces/[namespace]/workflows/_workflows-summary-table.svelte
```diff
@@ -1,18 +1,17 @@
...
-      <th class="w-3/12 text-left ">Workflow Name</th>
-      <th class="w-3/12 text-left">Workflow ID</th>
-      <th class="w-1/12">Status</th>
-      <th class="w-2/12 text-right">Started</th>
-      <th class="w-2/12 text-right">Ended</th>
-    </tr>
-  </thead>
-  <slot name="rows" />
-</table>
+<section class="workflow-table border-collapse w-full rounded-lg">
```
Comment (stevekinney) [inline src/routes/namespaces/[namespace]/workflows/_workflows-summary-table.svelte:1]: The table itself should probably have a border along the sides. Right now, it's missing a border along the bottom entirely. I suggest adding the following classes: `border-2 border-gray-300 overflow-hidden`

src/routes/namespaces/[namespace]/workflows/_workflows-summary-row.svelte
```diff
@@ -2,83 +2,50 @@
...
 
   $: href = getWorkflowExecutionUrl($namespace, workflow);
-  $: isActive = pathMatches(href, $page.path);
 </script>
 
-<tr class:active={isActive}>
-  <td class="">
-    <a sveltekit:noscroll {href}>
+<a sveltekit:noscroll {href}>
+  <article class="row flex flex-row border-2">
```
Comment (stevekinney) [inline src/routes/namespaces/[namespace]/workflows/_workflows-summary-row.svelte:16]: Do you see how this kind of doubles up the borders between rows and along the sides?

I'd recommend just doing `border-b-2`.

src/routes/namespaces/[namespace]/workflows/_workflows-summary-row.svelte
```diff
@@ -2,83 +2,50 @@
...
   export let timeFormat: string;
 
   $: href = getWorkflowExecutionUrl($namespace, workflow);
-  $: isActive = pathMatches(href, $page.path);
 </script>
 
-<tr class:active={isActive}>
-  <td class="">
-    <a sveltekit:noscroll {href}>
+<a sveltekit:noscroll {href}>
```
Comment (stevekinney) [inline src/routes/namespaces/[namespace]/workflows/_workflows-summary-row.svelte:15]: This is so much nice and behaves better than the table rows. The scroll is _way_ smoother now. Nice work.

src/routes/namespaces/[namespace]/workflows/_workflows-summary-row.svelte
```diff
@@ -2,83 +2,50 @@
...
+  <article class="row flex flex-row border-2">
+    <div class="links w-3/12 text-left">
+      {workflow.id}
+    </div>
+    <div class="links w-3/12 text-left">
       <h3>
```
Comment (stevekinney) [inline src/routes/namespaces/[namespace]/workflows/_workflows-summary-row.svelte:21]: It's not a blocker, but I wonder if we really need the `<h3>` tags? 🤷🏻

src/routes/namespaces/[namespace]/workflows/_workflows-summary-row.svelte
```diff
@@ -2,83 +2,50 @@
...
 <style lang="postcss">
-  p {
-    @apply m-0 text-gray-500 text-sm;
+  .row {
+    @apply w-full h-full flex no-underline p-2;
   }
 
-  tr {
-    @apply bg-gray-50;
+  .row:hover {
```
Comment (stevekinney) [inline src/routes/namespaces/[namespace]/workflows/_workflows-summary-row.svelte:44]: Much better—awesome.

src/routes/namespaces/[namespace]/workflows/index.svelte
```diff
@@ -49,15 +50,19 @@
...
       {:else}
         <WorkflowsEmptyState />
       {/if}
     {/await}
   </div>
 </section>
+
+<style lang="postcss">
+  .workflow-container {
+    margin: 0 2rem;
```
Comment (stevekinney) [inline src/routes/namespaces/[namespace]/workflows/index.svelte:66]: We want to have it fit onto the Tailwind grid and _not_ use bespoke margins.

stevekinney (commented):
Nice! It's getting there. Make sure you schedule a design review for this morning and we'll finish this up today.

src/routes/namespaces/[namespace]/workflows/index.svelte
Comment (stevekinney) [reply src/routes/namespaces/[namespace]/workflows/index.svelte:64]: Is this being used?

stevekinney (approved):

PR #162, opened 2021-12-08
Expand events and add support for pending activities

stevekinney:
Adds expanded detail view as well as updating the URL hash. Adds Font Awesome.

### DIFF
--- a/src/lib/models/workflow-execution.ts
+++ b/src/lib/models/workflow-execution.ts
@@ -1,6 +1,5 @@
 import type {
   WorkflowExecutionStatus,
-  PendingActivityInfo,
   DescribeWorkflowExecutionResponse,
   ListWorkflowExecutionsResponse,
 } from '$types';
@@ -17,7 +16,7 @@ export type WorkflowExecution = {
   status: WorkflowExecutionStatus;
   taskQueue?: string;
   historyEvents: Long;
-  pendingActivities: PendingActivityInfo[];
+  pendingActivities: PendingActivity[];
   url: string;
 };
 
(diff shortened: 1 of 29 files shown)


src/lib/models/workflow-execution.ts
```diff
@@ -38,7 +37,8 @@ export const toWorkflowExecution = (
   const historyEvents = response.workflowExecutionInfo.historyLength;
   const url = `/workflows/${id}/${runId}`;
   const taskQueue = response?.executionConfig?.taskQueue?.name;
-  const pendingActivities = response.pendingActivities || [];
+  const pendingActivities: PendingActivity[] =
+    (response.pendingActivities as unknown as PendingActivity[]) || [];
```
stevekinney, line 41: We need to do this because the type that was generated from the protobuf is _not_ what the API actually produces. This aligns it with the _actual_ values returned from the API.

src/lib/utilities/format-camel-case.ts
```diff
@@ -1,9 +1,9 @@
-const isUpperCase = (label: string, index: number) => {
+const isUpperCase = (label: string, index: number): boolean => {
```
GiantRobots, line 1: 👌🏽 That module boundary typing

src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/index.svelte
```diff
@@ -21,22 +21,20 @@
...
   import Event from './_event.svelte';
 
   export let workflow: WorkflowExecution;
-  export let events: HistoryEvent[];
+  export let events: HistoryEventWithId[];
 
-  let format: EventFormat = 'grid';
-
-  $: pendingActivities = workflow?.pendingActivities;
+  let pendingActivities = workflow?.pendingActivities;
```
Comment (GiantRobots) [inline src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/index.svelte:29]: This looks nullish here and we just don't have ts to yell at us. Is it okay for this to be null/undefined?

GiantRobots (approved):

src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/index.svelte
Comment (stevekinney) [reply src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/index.svelte:29]: It is. The component checks for its existence and its length.

PR #163, opened 2021-12-09
Remove side navigation; break out DataConvertorStatus

stevekinney:
Removes the side navigation, removes the sticky header, breaks out the Data Converter status into its own component.


src/lib/components/data-convertor-status.svelte
```diff
@@ -0,0 +1,25 @@
...
+  {:else if $lastDataConverterStatus === 'error'}
+    <Tooltip text={`Data converter couldn't connect to the remote converter`}>
+      <Icon src={StatusOffline} class="mx-auto text-red-400 w-8 h-8" />
+    </Tooltip>
+  {:else if $lastDataConverterStatus === 'success'}
+    <Tooltip text={'Data converter succesfully converted content'}>
+      <Icon src={StatusOnline} class="mx-auto text-green-500 w-8 h-8" />
+    </Tooltip>
+  {/if}
+{/if}
```
GiantRobots, line 25: Much better to leave this as a component.

src/routes/namespaces/[namespace]/workflows/index.svelte
```diff
@@ -39,25 +39,23 @@
...
+      <WorkflowsSummaryTable>
+        <tbody slot="rows">
+          {#each workflows as workflow}
+            <WorkflowsSummaryRow {workflow} {timeFormat} />
+          {/each}
+        </tbody>
+      </WorkflowsSummaryTable>
+    {:else}
+      <WorkflowsEmptyState />
+    {/if}
```
Comment (GiantRobots) [inline src/routes/namespaces/[namespace]/workflows/index.svelte:59]: Looks like we're missing an error state here? Or are we expecting code further up to error on it?

GiantRobots (approved):

src/routes/namespaces/[namespace]/workflows/index.svelte
Comment (stevekinney) [reply src/routes/namespaces/[namespace]/workflows/index.svelte:59]: @josephriosIO is currently refactoring this in #159.

PR #164, opened 2021-12-09
Make the table play nicer with skinnier widths

stevekinney:
Add better responsive design to event history.

### DIFF
--- a/src/lib/components/code-block.svelte
+++ b/src/lib/components/code-block.svelte
@@ -27,7 +27,7 @@
 </script>
 
 {#if inline}
-  <code class="language-json text-xs">
+  <code class="language-json" style="white-space: nowrap;">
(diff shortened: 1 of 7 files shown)


src/lib/components/code-block.svelte
```diff
@@ -27,7 +27,7 @@
 </script>
 
 {#if inline}
-  <code class="language-json text-xs">
+  <code class="language-json" style="white-space: nowrap;">
```
GiantRobots, line 30: Break word may be better

GiantRobots (approved):

src/lib/components/code-block.svelte
stevekinney: I literally never want it to wrap. I need the height to stay consistent.

PR #168, opened 2021-12-10
Add event filtering

stevekinney:
Adds event filtering using the logic—or something close to it—as found in the existing UI.

### DIFF
--- a/src/lib/utilities/get-event-categorization.test.ts
+++ b/src/lib/utilities/get-event-categorization.test.ts
@@ -1,0 +1,320 @@
+import {
+  eventTypeInCategory,
+  isCategoryType,
+} from './get-event-categorization';
+
+describe(eventTypeInCategory, () => {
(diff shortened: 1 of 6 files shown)


src/lib/utilities/get-event-categorization.test.ts
```diff
@@ -0,0 +1,320 @@
...
+  });
+
+  it('should categorize a "UpsertWorkflowSearchAttributes" as a "command"', () => {
+    expect(
+      eventTypeInCategory('command')({
+        eventType: 'UpsertWorkflowSearchAttributes',
+      }),
+    ).toBe(true);
+  });
+});
```
GiantRobots, line 294: All of these are true it would probably be good to test the opposite of these functions

```
    expect(
      eventTypeInCategory('workflow')({
        eventType: 'UpsertWorkflowSearchAttributes',
      }),
    ).toBe(false);
    ```

GiantRobots (approved):

PR #169, opened 2021-12-11
Move payload decoding to eventHistory. Use payloadPromise attribute f…

GiantRobots:
…or the payload.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


src/lib/models/event-history.ts
```diff
@@ -1,24 +1,31 @@
...
-): HistoryEventWithId[] => {
-  return response.history.events.map(
-    (event): HistoryEventWithId => ({
-      ...event,
-      id: String(event.eventId),
-      eventType: event.eventType as unknown as EventType,
-      attributes: getEventAttributes(event),
-    }),
+): Promise<HistoryEventWithId[]> => {
+  return await Promise.all(
```
stevekinney, line 21: My only fear with `Promise.all()` is that if _one_ of them fails, the entire thing fails. But, I think we can deal with that later. I know you don’t love observables, but we’re teetering on the edge eventually, but that’s an undertaking outside of the scope of this PR, so I think this  works for now and we can refine our thinking down the road.

stevekinney: You probably don’t need the await since the consumer is going to need to resolve the promise anyway.

src/lib/models/event-history.ts
```diff
@@ -1,24 +1,31 @@
...
+    response.history.events.map(
+      async (event): Promise<HistoryEventWithId> => ({
+        ...event,
+        id: String(event.eventId),
+        eventType: event.eventType as unknown as EventType,
```
stevekinney, line 26: This is not your problem, but maybe we just need to maintain our own types.

src/lib/services/data-converter.ts
```diff
@@ -10,6 +10,35 @@ interface WebSocketResponse {
   content: string;
   requestId: string;
 }
+
+export async function convertSinglePayloadDataConverter(
```
stevekinney, line 14: You can probably just name this `convertPayload`. TypeScript ensures it’s a single payload and it’s coming from `data-converter.ts`.

src/lib/services/data-converter.ts
```diff
@@ -10,6 +10,35 @@ interface WebSocketResponse {
...
+  payload: Payload,
+  websocket: any,
```
stevekinney, line 16: Do we have a better type for this or are we using a library that doesn’t have TypeScript types?

src/lib/services/events-service.ts
```diff
@@ -49,18 +47,7 @@ export const fetchEvents = async (
...
-          await convertEventPayloadFromDataConverter(events, port);
-        } catch (error) {
-          console.error(error);
-        }
-      }
-      return Promise.resolve(events);
-    });
+  ).then(toEventHistory);
 
   return events;
```
stevekinney, line 52: We can probably just return the promise chain at this point.

src/lib/stores/data-converter-websocket.ts
```diff
@@ -0,0 +1,43 @@
+import { get } from 'svelte/store';
+import WebSocketAsPromised from 'websocket-as-promised';
+import {
+  dataConverterPort,
+  setLastDataConverterFailure,
+} from './data-converter-config';
+
+export const dataConverterWebsocket = (function () {
```
stevekinney, line 8: Why does this need to be an IIFE? Modules are inherently IIFEs.

src/lib/stores/data-converter-websocket.ts
```diff
@@ -0,0 +1,43 @@
...
+        Object.assign({ requestId: requestId }, data),
+      extractRequestId: (data) => data && data.requestId,
+    });
+  } catch (err) {
+    setLastDataConverterFailure();
+  }
+
+  sock.open();
+
+  return {
```
stevekinney, line 37: So, to my previous comment, You can just export this object and it’s the same as the IIFE.

src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/_event.svelte
```diff
@@ -0,0 +1,70 @@
+<script lang="ts">
```
Comment (stevekinney) [inline src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/_event.svelte:1]: Is this an old file?

src/lib/utilities/decode-payload.ts
```diff
@@ -1,31 +1,57 @@
...
+  const port = get(dataConverterPort);
 
-  if (event.workflowExecutionStartedEventAttributes) {
-    const input = event.workflowExecutionStartedEventAttributes.input;
-    if (input === null) return 'null';
-    if (input) {
-      const [data] = input.payloads.map((payload) => payload.data);
-      return window.atob(String(data));
-    }
+  const potentialPayload = (anyAttributes?.input?.payloads ??
```
stevekinney, line 21: This feels like it needs to be broken out into a utility function with a test.

stevekinney (commented):
The one thing that comes to mind is that it might be easier to break this into two separate bodies of work.

1. A piece just focused around Data Converter.
  - The whole lifecycle of getting Data Converter up and running.
  - The UI pieces around the Data Converter status.
  - A function that uses Data Converter is active if it’s configured and is a no-op if it’s not.
  - A change from “Convertor” to “Converter” globally.
2. All of the ceremony around decoding payloads.

src/lib/models/event-history.ts
GiantRobots: I'm down with observables in the future.

Probably right on the await, I had done it because of difficulty with types and it just stuck and hadn't reviewed it. I had pushed this before checking first.

src/lib/models/event-history.ts
GiantRobots: Honestly the types were a huge issue for getting this right and it was easier to any several things because they were wrong.

src/lib/services/data-converter.ts
GiantRobots: The latter

src/lib/services/data-converter.ts
GiantRobots: Yeah, the other function should be deleted but I was keeping it around for reference.

src/lib/stores/data-converter-websocket.ts
GiantRobots: Stack overflow copy pasta from before vacation didn't fix it after you made this comment last time, will fix.

src/lib/utilities/decode-payload.ts
GiantRobots: Yeah this entire function needs to become a well tested function if it's working to satisfaction I'll focus on that tomorrow

src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/_event.svelte
Comment (GiantRobots) [reply src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/_event.svelte:1]: This looks like an old file, I'll delete.

src/lib/services/data-converter.ts
stevekinney: It'll join `svelte-virtual-list` on my technical debt naughty pile.

src/lib/utilities/decode-payload.ts
```diff
@@ -1,31 +1,56 @@
...
-    if (result === null) return 'null';
-    if (result) {
-      const [data] = result.payloads.map((payload) => payload.data);
-      return window.atob(String(data));
-    }
+  // This is a very sneaky array[0].data It's a holdover from a previous version of this code we could probably actually iterate
+  // over this and get each entry decoded from base64
+  // This was set to any because we just want whatever comes out of here to be base64 decoded and reset to the other value
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  let [JSONPayload]: any[] = potentialPayload.map((payload) => payload.data);
```
stevekinney, line 33: Isn't this always going to be a string?

stevekinney (approved):

src/lib/utilities/decode-payload.ts
GiantRobots: The types have it as a Uint8Array, which is stringish. But we re-use the JSONPayload variable. When it comes back decoded from the websocket it could be a string/array/object so setting it to any was convenient to get this working. I'd like to refine this in a later ticket when we do tests for this.

PR #171, opened 2021-12-14
Add pipeline to update ui-server on changes

feedmeapples:
## What was changed

Adds pipeline that notifies ui-server about new commits

## Why?

Automating UI updates

## Checklist

1. Closes 

2. How was this tested:

Tested with test repositories https://github.com/feedmeapples/test2
tbu. testing in current repo by manually running pipelines

3. Any docs updates needed?

### DIFF
--- a/.github/workflows/trigger-ui-server-update.yml
+++ b/.github/workflows/trigger-ui-server-update.yml
@@ -1,0 +1,26 @@
+name: 'Trigger ui-server Update'
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  notify:
+    name: 'trigger ui-server update'
+    runs-on: ubuntu-latest
+
+    defaults:
+      run:
+        shell: bash
+
+    steps:
+      - name: Dispatch ui-server Github Action
+        env:
+          PAT: ${{ secrets.PAT }}
+          PARENT_REPO: temporalio/ui-server
+          PARENT_BRANCH: main
+          WORKFLOW_ID: update-ui.yml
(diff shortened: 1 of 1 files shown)


.github/workflows/trigger-ui-server-update.yml
```diff
@@ -0,0 +1,26 @@
...
+
+    steps:
+      - name: Dispatch ui-server Github Action
+        env:
+          PAT: ${{ secrets.PAT }}
+          PARENT_REPO: temporalio/ui-server
+          PARENT_BRANCH: main
+          WORKFLOW_ID: update-ui.yml
+        run: |
+          curl -fL -X POST -H "Accept: application/vnd.github.v3+json" -H "Authorization: token ${{ env.PAT }}" https://api.github.com/repos/${{ env.PARENT_REPO }}/actions/workflows/${{ env.WORKFLOW_ID }}/dispatches -d '{"ref":"${{ env.PARENT_BRANCH }}"}'
```
stevekinney, line 26: I imagine that there is probably away to abstract this into a re-usable action—but I think this totally works for now.

stevekinney (approved):

.github/workflows/trigger-ui-server-update.yml
feedmeapples: that's true. I've looked at few github actions available in the marketplace that do this instead of manually calling `curl`, though i was a bit hesitant to use them because of the added security implication and having to rely on a third-party lib. Figured to have a direct REST call instead

Still open to the idea though

PR #178, opened 2021-12-17
table scroll fix

josephriosIO:
## What was changed
 Scroll on table wasn't working if a lot of workflows were shown, this should fix it
 
 ## Tested
  on my 13in macbook and on my 27in monitor


src/routes/namespaces/[namespace]/workflows/_workflows-summary-table.svelte
```diff
@@ -10,3 +10,9 @@
   </div>
   <slot />
 </section>
+
+<style lang="postcss">
+  section {
+    height: calc(100vh - 20%);
```
Comment (stevekinney) [inline src/routes/namespaces/[namespace]/workflows/_workflows-summary-table.svelte:16]: Is this different than just saying `80vh`? Like, this should work, but only because it's guessing at the other elements on the page, right?

PR #179, opened 2021-12-22
Add filtering for workflow ID and type

stevekinney:
Wires up logic for filtering the workflow execution list by ID or type.

- Syncs filters with query params.
- Adds debouncing to filter inputs.
- Adds tests for utility functions.


jest.config.js
```diff
@@ -1,3 +1,5 @@
+process.env.TZ = 'UTC';
```
feedmeapples, line 1: why is this one needed? for executing `jest` directly in terminal?

feedmeapples (approved):

jest.config.js
stevekinney: Or for editor plugins that might use Jest as well. Basically, if you're reading the configuration, you're probably running Jest—so, set this appropriately.

I'm thinking that Cypress is a lot better at mocking time and maybe it makes more sense to test this functionality in Cypress.

PR #184, opened 2021-12-27
Implement page for viewing workers

josephriosIO:
## What was changed
Added workers view

## Images
Screen Shot 2021-12-27 at 1 50 18 PM


src/lib/components/workers-row.svelte
```diff
@@ -0,0 +1,36 @@
...
+    </h3>
+  </div>
+  <div class="w-3/12 text-left">
+    {#if poller.taskQueueTypes.includes('WORKFLOW')}
+      <Icon icon={faCheck} color="blue" />
+    {:else}
+      <Icon icon={faTimes} color="black" />
+    {/if}
+  </div>
+  <div class="w-2/12 text-left">
```
stevekinney, line 23: Why is this column smaller?

src/lib/components/workers-row.svelte
```diff
@@ -0,0 +1,36 @@
...
+    {#if poller.taskQueueTypes.includes('ACTIVITY')}
+      <Icon icon={faCheck} color="blue" />
+    {:else}
+      <Icon icon={faTimes} color="black" />
+    {/if}
+  </div>
+</article>
+
+<style lang="postcss">
+  .row {
```
stevekinney, line 33: We might as well just add these classes to the element, right?

src/lib/components/workers-table.svelte
```diff
@@ -0,0 +1,17 @@
+<section
```
stevekinney, line 1: Do you think we benefit from breaking this out to its own component instead of just combining this and `workers-row.svelte` into one component?

I guess my other question is should these be in the `$lib/components` considering they're dedicated to one particularly page—like the workflows summary table.

src/lib/components/workers-table.svelte
```diff
@@ -0,0 +1,17 @@
...
+    <div class="w-3/12 text-left">Last Accessed</div>
+    <div class="w-3/12 text-left">Workflow Table Handler</div>
+    <div class="w-2/12 text-left">Activity Handler</div>
+  </div>
+  <slot />
+</section>
+
+<style lang="postcss">
+  .workers-table {
+    height: calc(100vh - 280px);
```
stevekinney, line 15: Are we using a Virtual list here? If not, I'm not sure we need this. Is the 280px coming from anywhere in particular?

src/lib/utilities/request-from-api.ts
```diff
@@ -47,6 +47,9 @@ export const requestFromAPI = async <T>(
     retryInterval = 5000,
   } = init;
 
+  const endpointParams = endpoint.split('?')[1];
```
stevekinney, line 50: You could probably collapse this into one line:

```js
const [path, params] = endpoint.split('?');
```

src/lib/utilities/request-from-api.ts
```diff
@@ -47,6 +47,9 @@ export const requestFromAPI = async <T>(
...
+  endpoint = endpoint.split('?')[0];
+
   if (!endpoint.startsWith('/')) endpoint = '/' + endpoint;
   const nextPageToken = token ? { next_page_token: token } : {};
   const query = new URLSearchParams({
```
stevekinney, line 55: This is _also_ the query—so, do we already have the query parameters stored? If so, we don't need to do it twice.

src/lib/utilities/request-from-api.ts
```diff
@@ -55,9 +58,8 @@ export const requestFromAPI = async <T>(
   });
 
   const url = toURL(base + '/api/v1' + encode(endpoint), query);
```
stevekinney, line 60: Since we're attaching the query on here, I'm not sure if you _also_ need to put them back on in Line 62.

src/lib/utilities/request-from-api.ts
```diff
@@ -55,9 +58,8 @@ export const requestFromAPI = async <T>(
...
-
   try {
-    const response = await request(url, options);
+    const response = await request(url + endpointParams, options);
```
stevekinney, line 62: Did we already add the query params on in Line 60?

stevekinney (commented):
Let me know when you've gotten a design review also.

src/lib/components/empty-state.svelte
```diff
@@ -0,0 +1,19 @@
...
+  export let content: string = '';
+</script>
+
+<div class="mt-12 flex flex-col justify-start items-center gap-2">
+  <span
+    class="flex rounded-full items-center justify-center w-16 h-16 bg-gray-200"
+  >
+    <Icon icon={faMeteor} scale={1.2} class="block w-full h-full" /></span
+  >
+  <h2 class="text-xl font-medium">No {title} Found</h2>
```
stevekinney, line 15: You probably want to _just_ include `{title}` instead of hard-coding "No" and "Found." Let whoever is using the component pass it in.

stevekinney (approved):

stevekinney (approved):

PR #187, opened 2021-12-30
Stack trace

josephriosIO:
## What was changed
Added view to see stack traces for each workflow

### Stack Trace
Screen Shot 2022-01-04 at 1 59 52 PM

### Workers
Screen Shot 2022-01-04 at 1 50 08 PM


src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/stack-trace.svelte
```diff
@@ -0,0 +1,60 @@
...
+        Refresh</Button
+      >
+      <p class="ml-9">Stack Trace at {datetime}</p>
+    </div>
+    {#await data}
+      <div>loading</div>
+    {:then { queryResult }}
+      <div class="flex items-start h-full">
+        <CodeBlock
+          content={window.atob(String(queryResult.payloads[0].data))}
```
Comment (josephriosIO) [inline src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/stack-trace.svelte:53]: typscript issue here

Comment (stevekinney) [reply src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/stack-trace.svelte:53]: You're telling it that this is a Promise that resolves to null [here](https://github.com/temporalio/ui/pull/187/files#diff-ee9488af68819b334c1e89c9977f90ab4e9f7ce9a3ac2c9498ba02c4ac77b7edR12). If you swap out `null` for the data structure's type, you should be good.

Comment (stevekinney) [reply src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/stack-trace.svelte:53]: For example: https://github.com/temporalio/ui/blob/main/src/lib/services/workflow-service.ts#L106

src/lib/components/button.svelte
```diff
@@ -12,15 +12,16 @@
     on:click
     class:selected={active}
     {disabled}
-    class={`${size} ${variant}`}><slot /></button
+    class={`flex items-center justify-center ${size} ${variant}`}
```
stevekinney, line 15: Do these work with Tailwind's tree-shaking/purging?

stevekinney: Also: Svelte allows you to interpolate with double-quoted strings. (e.g. `class="flex items-center justify-center {size} {variant}"`, but I'm still hesitant about the tree shacking. I'd probably use Svelte's [class toggles](https://svelte.dev/repl/e3ea17e8e09044999bf7cb4bc882adea?version=3.19.2).

src/lib/services/query-service.ts
```diff
@@ -0,0 +1,47 @@
+import type { WorkflowExecution } from '$lib/models/workflow-execution';
+import { requestFromAPI } from '$lib/utilities/request-from-api';
+
+type StackTraceOptions = {
+  workflow: WorkflowExecution;
+  namespace: string;
+};
+
+type QueryPayload = {
```
stevekinney, line 9: Are these included in the Temporal types?

src/lib/services/query-service.ts
```diff
@@ -0,0 +1,47 @@
...
+    encoding?: string;
+  };
+};
+
+type QueryType = {
+  payloads: QueryPayload[];
+};
+
+type StackTraceExecution = {
+  queryRejected?: null;
```
stevekinney, line 21: So, this is optional, but when it's present, it's always null?

josephriosIO: fixed this one it wasn't optional it's either a string or null

src/lib/components/button.svelte
josephriosIO: how would it work since its different styles like for size it could be small or large?

stevekinney: So, this is a bit of an exercise in API design.

Above, you have:

```ts
export let size: string = 'small';
```

This is a little problematic because the consumer can pass in _any_ string.

You _could_ tighten the types up a bit.

```ts
export let size: 'large' | 'small' = 'small';
```

No were down to only two options. You'd do the same with `variant` and `as` as well.

You could now do something like this:

```svelte
<button
  on:click
  class:selected={active}
  {disabled}
  class="flex items-center justify-center"
  class:small={variant === 'small'}
  class:large={variant === 'large'}
>
  <slot />
</button>
```

The other option, of course since "small" is the default, just allowing a `large` boolean prop that overrides the default.

```html
<Button large>Hello</Button>
```

And now, you can do something like this:

```svelte
<button
  on:click
  class:selected={active}
  {disabled}
  class="flex items-center justify-center"
  class:large
>
  <slot />
</button>
```

The pattern only works for two settings where one is the default. So, if you suspect there will be more, than you're going to need to stick with the `variant` pattern.

src/lib/components/button.svelte
```diff
@@ -12,15 +12,16 @@
...
+    ><slot /></button
   >
 {:else}
   <a
     {href}
     on:click
     class:selected={active}
     {disabled}
-    class={`${size} ${variant}`}><slot /></a
+    class={`flex items-center justify-center ${size} ${variant}`}><slot /></a
```
stevekinney, line 24: This is duplicated code, you might want to just store all of these attributes in an object and [then spread them across](https://svelte.dev/tutorial/spread-props) both the `a` and the `button`, that way we're not implementing everything twice.

src/lib/services/query-service.ts
```diff
@@ -0,0 +1,47 @@
...
+    encoding: string;
+  };
+};
+
+type QueryType = {
+  payloads: QueryPayload[];
+};
+
+type StackTraceExecution = {
+  queryRejected: string | null;
```
stevekinney, line 21: Do you ever intend for it to be `null` or just `undefined`? If not, you can just say `string` here and keep it optional.

src/lib/services/query-service.ts
```diff
@@ -0,0 +1,47 @@
...
+  metadata: {
+    encoding: string;
+  };
+};
+
+type QueryType = {
+  payloads: QueryPayload[];
+};
+
+type StackTraceExecution = {
```
stevekinney, line 20: Is this something we can pull from the Temporal types?

src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/_tabs.svelte
```diff
@@ -18,7 +18,7 @@
     href={routeFor('workflow.events.full', { namespace, workflowId, runId })}
     amount={historyEvents}
   />
-  <Tab label="Workers" href="#not-implemented" />
+  <Tab label="Workers" href="workers" />
```
Comment (stevekinney) [inline src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/_tabs.svelte:21]: We need to use a `routeFor` here.

stevekinney:
Based on that screenshot, it also looks like the spacing is a bit off between the button and the time stamp. You might want to get a design review from Candace.

PR #191, opened 2022-01-03
Show authenticated user's avatar

feedmeapples:
## What was changed

- fixes log in button to force hitting ui-server's auth URL
- shows authenticated user's avatar 
image

## Why?

let's user know when auth succeeds

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/lib/stores/user.ts
+++ b/src/lib/stores/user.ts
@@ -1,0 +1,12 @@
+import { writable } from 'svelte/store';
+
+import { fetchUser } from '$lib/services/user-service';
+
+export const user = writable<User>(null);
+
+const loadUser = async (): Promise<void> => {
+  const userRes = await fetchUser();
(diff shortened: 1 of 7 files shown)


src/lib/stores/user.ts
```diff
@@ -0,0 +1,12 @@
+import { writable } from 'svelte/store';
+
+import { fetchUser } from '$lib/services/user-service';
+
+export const user = writable<User>(null);
+
+const loadUser = async (): Promise<void> => {
```
stevekinney, line 7: Would this be better off just as a single API call here? https://github.com/temporalio/ui/blob/main/src/routes/__layout.svelte#L11-L21

feedmeapples: done

PR #197, opened 2022-01-06
Add guard clause to verify that the document is available

stevekinney:
Makes sure that checking for CSRF tokens is running the browser.


src/lib/utilities/request-from-api.ts
```diff
@@ -85,6 +86,7 @@ const withSecurityOptions = (options: RequestInit): RequestInit => {
 };
 
 const withCsrf = (headers: HeadersInit): HeadersInit => {
+  if (!browser) return;
```
feedmeapples, line 89:
```suggestion
  if (!browser) return headers;
```

feedmeapples (approved):

PR #212, opened 2022-01-11
Address TypeScript hints

stevekinney:
Address some of the easier suggestions from our TypeScript checker.

### DIFF
--- a/src/lib/services/query-service.ts
+++ b/src/lib/services/query-service.ts
@@ -1,18 +1,19 @@
-import sanitize from 'sanitize-html';
-
-import { requestFromAPI } from '$lib/utilities/request-from-api';
+import {
+  isTemporalAPIError,
+  requestFromAPI,
+} from '$lib/utilities/request-from-api';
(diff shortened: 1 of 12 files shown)


stevekinney:
Closed in favor of #216.

src/lib/services/query-service.ts
```diff
@@ -79,7 +79,7 @@ export async function getQueryTypes(
   options: WorkflowParameters,
   request = fetch,
 ): Promise<string[]> {
-  return new Promise<string[]>((resolve, reject) => {
+  return new Promise<string[]>((resolve) => {
     fetchQuery(
```
GiantRobots, line 83: This isn't for this code change it's further up but something to consider.

```type QueryRequestParameters = {
  workflow: Eventual<{ id: string; runId: string }>;
  namespace: string;
  queryType: string;
};
type WorkflowParameters = Omit<QueryRequestParameters, 'queryType'>;```

Could be extended instead of omitted which is usually easier for me to understand

PR #219, opened 2022-01-11
Add event IDs to list

stevekinney:
image

### DIFF
--- a/src/lib/components/event.svelte
+++ b/src/lib/components/event.svelte
@@ -12,7 +12,7 @@
 
   export let event: HistoryEventWithId | PendingActivity | Activity;
 
-  let { routeFor, pending, timeStamp, name, tag, classification } =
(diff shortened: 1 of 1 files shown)


src/lib/components/event.svelte
```diff
@@ -27,20 +27,23 @@
...
-        {name}
-      </EventLabel>
-    </h2>
-    <p class="text-sm">
-      <Icon icon={faCalendar} class="inline" />
-      {formatDate(timeStamp)}
-    </p>
+  <article class="flex gap-4 items-center p-4">
+    <p class="w-5 text-center text-gray-500">{id}</p>
+    <div class="w-full">
```
GiantRobots, line 32: This could be `grow` and if it's not a high enough "growth" you could use `grow-[x]`

GiantRobots (approved):

PR #226, opened 2022-01-12
Add Signal and Marker events to compact view

feedmeapples:
## What was changed

Added Signals and Markers to history compact view:

image

## Why?

Important events for compact view

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/lib/models/events-group.ts
+++ b/src/lib/models/events-group.ts
@@ -77,12 +77,12 @@ export class EventsGroup {
     }
 
     if (isSignalEvent(event)) {
-      this.id = event?.id;
+      this.id = event.id;
       this.name = `Signal ${event?.workflowExecutionSignaledEventAttributes?.signalName}`;
     }
 
     if (isMarkerEvent(event)) {
-      this.id = event?.id;
+      this.id = event.id;
       this.name = `Marker ${event?.markerRecordedEventAttributes?.markerName}`;
     }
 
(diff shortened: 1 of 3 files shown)


src/lib/models/events-group.ts
```diff
@@ -128,7 +156,12 @@ export class EventGroups {
   }
 
   add(event: CompactEvent): void {
-    if (isActivityScheduledEvent(event)) {
+    if (
+      isActivityScheduledEvent(event) ||
```
stevekinney, line 160: Eventually we will probably want to break this out into something like `isGroupableEvent` or something.

src/lib/models/events-group.ts
```diff
@@ -53,11 +71,21 @@ export class EventsGroup {
...
     }
 
-    if (isTimerStartedEvent(event)) {
+    if (isTimerEvent(event)) {
       this.id = event?.timerStartedEventAttributes?.timerId;
       this.name = `Timer ${event?.timerStartedEventAttributes?.timerId}`;
     }
 
+    if (isSignalEvent(event)) {
+      this.id = event?.id;
```
stevekinney, line 80: You probably don't need `.?` on a single property. Either way, it will turn out to be `undefined`. It probably adds some extra weight after compilation since it _might_ be compiled down to checking if the property exists—unless the compiler is really smart.

stevekinney (approved):
Looks great.

src/lib/models/events-group.ts
feedmeapples: right

PR #228, opened 2022-01-13
Add child workflows to history compact view

feedmeapples:
## What was changed

Added child workflows to history compact view

image

## Why?

helpful at understanding history of large workflows

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/lib/models/events-group.ts
+++ b/src/lib/models/events-group.ts
@@ -14,11 +14,22 @@ const activityTypes = [
 
 const timerTypes = ['TimerStarted', 'TimerCanceled', 'TimerFired'];
 
-const signalType = ['WorkflowExecutionSignaled'];
+const signalTypes = ['WorkflowExecutionSignaled'];
 
-const markerType = ['MarkerRecorded'];
+const markerTypes = ['MarkerRecorded'];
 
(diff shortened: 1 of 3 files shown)


src/lib/models/events-group.ts
```diff
@@ -41,6 +53,12 @@ const isActivityScheduledEvent = (
   return event.eventType === 'ActivityTaskScheduled';
 };
 
+const isChildWorkflowInitializedEvent = (
```
stevekinney, line 56: This can happen later, but you can see a world where we break these out into their own file because we might want to use them elsewhere.

src/lib/models/events-group.ts
```diff
@@ -196,6 +220,18 @@ export class EventGroups {
       const { scheduledEventId } = event.activityTaskCompletedEventAttributes;
       this.get(scheduledEventId).set(event.eventType, event);
     }
+
+    if (event.eventType === 'ChildWorkflowExecutionStarted') {
```
stevekinney, line 224: Eventually we're going to need to go back and address the test coverage. I think that we should ship the functionality for now, but we'll need to pay down the technical debt over the next few weeks.

stevekinney (approved):

PR #229, opened 2022-01-13
Transform payloads

GiantRobots:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


src/lib/utilities/decode-payload-lazy.ts
```diff
@@ -0,0 +1,23 @@
+import type { Payload } from '$types';
```
GiantRobots, line 1: This was named lazy because I was being lazy fighting with jest about imports not working correctly with the other parts of the decode-payload file.  Definitely needs to be remediated differently.

stevekinney (approved):

PR #230, opened 2022-01-14
Make history export work for replay

feedmeapples:
## What was changed

- Fixes Export History button 
- Changes to export raw history instead of dataconverted

image

## Why?

- helpful at history replay

## Checklist

1. Closes 

2. How was this tested:

 
3. Any docs updates needed?


src/lib/components/export-history.svelte
```diff
@@ -0,0 +1,30 @@
...
+    const events = await fetchRawEvents({
+      namespace,
+      executionId: workflowId,
+      runId: runId,
+    });
+
+    const content = JSON.stringify(events, null, 2);
+    download(content, `${runId}/events.json`, 'text/plain');
+
+    function download(content: string, fileName: string, contentType: string) {
```
stevekinney, line 16: Does the data URL method we were using not work?

src/lib/services/events-service.ts
```diff
@@ -51,3 +51,23 @@ export const fetchEvents = async (
 
   return events;
 };
+
+export const fetchRawEvents = async (
```
stevekinney, line 55: We are purposely not running this through data converter, right?

stevekinney (approved):

src/lib/services/events-service.ts
feedmeapples: yes, for replay functionality we want payloads to stay in the original form

src/lib/components/export-history.svelte
feedmeapples: as we are working with potentially 10s of MBs, Blob + URL.createObjectURL seemed more appropriate instead of bloating the html with the data url

plus that's less custom logic and just using existing api

feedmeapples: switching to a dynamic <a> also allows to delay loading of 10s of MBs until that's needed

stevekinney: Sold.

src/lib/services/events-service.ts
stevekinney: Cool. That's what I thought.

PR #234, opened 2022-01-17
Fix UI rendering on API call failures

feedmeapples:
## What was changed

Fixed UI showing up as blank page on API errors

## Why?

Improves visibility into what's going on

## Checklist

1. Closes 

2. How was this tested:

- turn of temporal service so the API calls will fail
- open UI

image

3. Any docs updates needed?

### DIFF
--- a/src/lib/services/settings-service.ts
+++ b/src/lib/services/settings-service.ts
@@ -9,7 +9,7 @@ export const fetchSettings = async (request = fetch): Promise<Settings> => {
 
   return {
     auth: {
-      enabled: settings.Auth.Enabled,
+      enabled: settings?.Auth?.Enabled,
     },
   };
 };
--- a/src/routes/__layout.svelte
+++ b/src/routes/__layout.svelte
@@ -15,10 +15,10 @@
   import '../app.postcss';
 
   export async function load({ fetch }: LoadInput) {
-    const { namespaces }: ListNamespacesResponse = await requestFromAPI(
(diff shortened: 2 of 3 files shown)


src/lib/services/settings-service.ts
```diff
@@ -9,7 +9,7 @@ export const fetchSettings = async (request = fetch): Promise<Settings> => {
 
   return {
     auth: {
-      enabled: settings.Auth.Enabled,
+      enabled: settings?.Auth?.Enabled,
```
GiantRobots, line 12: This will be null instead of false. We probably want a true/false value here (even though it'll probably work in js for most if statements if someone does if auth.enabled == false we'll have a bad time) I would recommend adding

`settings?.Auth?.Enabled ?? false`

stevekinney: Or you can use a `!!` to coerce it into a boolean.

src/routes/__layout.svelte
```diff
@@ -15,10 +15,10 @@
   import '../app.postcss';
 
   export async function load({ fetch }: LoadInput) {
-    const { namespaces }: ListNamespacesResponse = await requestFromAPI(
+    const { namespaces }: ListNamespacesResponse = (await requestFromAPI(
       routeForApi('namespaces'),
       { request: fetch },
-    );
+    )) || { namespaces: [] };
```
GiantRobots, line 21: `|| -> ??`

GiantRobots (approved):

PR #238, opened 2022-01-18
Test data converter

GiantRobots:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


src/lib/services/data-converter.ts
```diff
@@ -11,9 +12,9 @@ interface WebSocketResponse {
 
 export async function convertPayload(
   payload: Payload,
-  websocket: any,
+  websocket: WebSocketAsPromised,
 ): Promise<string | Payload> {
-  if (!websocket.isOpened && !websocket.isOpening) {
+  if (!websocket.isOpened) {
```
GiantRobots, line 17: If we're opening but not opened we would send before the websocket was opened and send would be broken.

src/lib/services/data-converter.ts
```diff
@@ -24,20 +25,25 @@ export async function convertPayload(
   if (!websocket.isOpened) {
     return Promise.resolve(payload);
   }
+
   const socketResponse: Promise<string> = websocket
     .sendRequest({
       payload: JSON.stringify(payload),
     })
     .then((response: WebSocketResponse) => {
+      setLastDataConverterSuccess();
```
GiantRobots, line 34: This was in the wrong place. Testing discovered this 🎉

src/lib/services/events-service.ts
```diff
@@ -16,33 +13,16 @@ type FetchEvents = NamespaceScopedRequest &
...
   request = fetch,
 ): Promise<HistoryEventWithId[]> => {
-  let params = {};
-
-  const port = get(dataConverterPort);
-
-  if (rawPayloads) {
-    params = { params: { rawPayloads: JSON.stringify(!!port) } };
-  }
-
```
GiantRobots, line 37: RawParams was an old UiServer artifact, we don't need it here.

src/lib/utilities/__mocks__/$app/env.ts
```diff
@@ -0,0 +1,2 @@
+// jsDom works for some dom implementation stuffs but $app/env browser fails whenever it's imported so lets mock it instead
+export const browser = jest.fn(() => false);
```
GiantRobots, line 2: This may be useful in more places, but this allowed me to run store logic and the like in jest.

src/lib/services/data-converter.ts
```diff
@@ -3,6 +3,7 @@ import {
   setLastDataConverterSuccess,
 } from '$lib/stores/data-converter-config';
 import type { Payload } from '$types';
+import type WebSocketAsPromised from 'websocket-as-promised';
```
stevekinney, line 6: 💪🏻

stevekinney (approved):

stevekinney (approved):

PR #241, opened 2022-01-21
Added routes, first run at rendering the page.

GiantRobots:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


src/lib/services/workflow-service.ts
```diff
@@ -51,3 +51,67 @@ export async function fetchWorkflow(
     toWorkflowExecution,
   );
 }
+
+export const fetchAllArchivedWorkflows = async (
```
stevekinney, line 55: Nit: Maybe we just call it `fetchArchivedWorkflows`?

src/lib/services/workflow-service.ts
```diff
@@ -51,3 +51,67 @@ export async function fetchWorkflow(
...
+  const query = toListWorkflowQuery(parameters);
+
+  const { executions, nextPageToken } =
+    (await requestFromAPI<ListWorkflowExecutionsResponse>(
+      routeForApi('archive', { namespace }),
+      {
+        params: { query },
+        request,
+      },
+    )) ?? { executions: [], nextPageToken: '' };
```
stevekinney, line 69: Should we come up with something better here? Like should `requestFromAPI` take an empty value? What does this request do if Archival isn't enabled?

src/lib/utilities/route-for-api.ts
```diff
@@ -23,6 +25,7 @@ type APIRouteParameters = {
   queue: string;
 };
 
+export type ArchiveRouteParameters = Pick<APIRouteParameters, 'namespace'>;
```
stevekinney, line 28: Do we need this type or can we just use `WorkflowListRouteParameters` since they're the same.

src/lib/utilities/route-for-api.ts
```diff
@@ -9,12 +9,14 @@ type WorkflowAPIRoutePath =
...
 type ParameterlessAPIRoutePath = 'cluster' | 'settings' | 'user' | 'namespaces';
+type ArchiveAPIRoutePath = 'archive';
 
 type APIRoutePath =
   | WorkflowsAPIRoutePath
   | WorkflowAPIRoutePath
   | ParameterlessAPIRoutePath
-  | TaskQueueAPIRoutePath;
+  | TaskQueueAPIRoutePath
+  | ArchiveAPIRoutePath;
```
stevekinney, line 19: I wonder if this should just go in with the `WorkflowsAPIRoutePath`.

src/lib/utilities/route-for-api.ts
```diff
@@ -60,6 +63,10 @@ export function routeForApi(
   route: TaskQueueAPIRoutePath,
   parameters: TaskQueueRouteParameters,
 ): string;
+export function routeForApi(
```
stevekinney, line 66: Again, if we fold this into the workflows list APIs, then we can basically just omit this.

PR #257, opened 2022-02-09
Add remote data encoder support.

robholland:
## What was changed

Added support for remote data encoders. This is a new API currently under review at https://github.com/temporalio/sdk-go/pull/717. This PR should not be merged until the SDK PR is merged.

For this first iteration only basic auth is supported via https://user:password@remote-encoder-url.com/, oauth for the remote encoder is not supported.

## Why?

This allows the web UI to make use of a shared remote encoder rather than needing to run tctl dc web command locally. In this deployment model developers don't need to use the tctl plugin system or have access to the encryption keys on their development machine.

## Checklist
1. How was this tested:

Tested on my integration branch of backgroud-checks which has a full encryption data converter setup for a sample Temporal application.

2. Any docs updates needed?

Yes, documentation will be needed to explain how to deploy such a system.


src/lib/utilities/decode-payload.test.ts
```diff
@@ -74,71 +70,17 @@ describe(decodePayload, () => {
...
-
-    expect(convertedPayload).toEqual(dataConvertedFailureWorkflowStartedEvent);
-
-    const dataConverterStatus = get(lastDataConverterStatus);
-    expect(dataConverterStatus).toEqual('error');
-  });
-
-  it('Should skip converting a payload and set the status to notRequested when the websocket and port is not set', async () => {
+  it('Should skip converting a payload and set the status to notRequested when the encoder endpoint is not set', async () => {
     const convertedPayload = await convertPayloadToJson(
```
GiantRobots, line 77: Hmm we definitely want to keep these tests around to validate the various feedback states for the data encoder.

robholland: I'll try and refactor to fit the flow without websockets.

stevekinney:
@GiantRobots Can you work with @robholland to get this merged in?

PR #258, opened 2022-02-10
Initial commit with functional errors from client/server side.

GiantRobots:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

https://user-images.githubusercontent.com/1903839/153312860-33481f31-25ca-4a41-b4d2-f7be45d19f6b.mov


src/lib/components/error-boundary.svelte
```diff
@@ -0,0 +1,46 @@
...
+  import { isNetworkError } from '$lib/utilities/handle-error';
+  import { networkError } from '$lib/stores/error-store';
+  import Error from './error.svelte';
+
+  export let error = null;
+  export let onError = null;
+
+  let theError;
+
+  // We know something broke and was un-recoverable so lets default to 500 and let the throwing code
```
stevekinney, line 11: We can probably remove this comment.

src/lib/components/error.svelte
```diff
@@ -0,0 +1,40 @@
...
+  // tell us what kind of stuff was broken
+  let status = 500;
+  if (error?.status) {
+    status = error.status;
+  }
+  if (error?.statusCode) {
+    status = error.statusCode;
+  }
+  $: currentLocation = window.location.toString();
+  console.log({ ThisIsTheWay: error });
```
stevekinney, line 20: 🤨

src/lib/utilities/handle-error.ts
```diff
@@ -1,8 +1,26 @@
+import { networkError } from '$lib/stores/error-store';
 import { notifications } from '../stores/notifications';
+import type { NetworkError } from './request-from-api';
+
+export function isNetworkError(
```
stevekinney, line 5: 1. We can probably break this out into its own utility.
2. We can probably add some unit tests.

src/lib/utilities/request-from-api.ts
```diff
@@ -77,7 +83,11 @@ export const requestFromAPI = async <T>(
       if (onError && isFunction(onError)) {
         onError({ status, statusText, body });
       } else {
-        throw new Error(`${status}: ${statusText}`);
+        throw {
+          statusCode: response.status,
+          statusText: response.statusText,
+          response,
+        } as NetworkError;
```
stevekinney, line 90: I think you can't catch a type, so I don't know if the `as` does anything. But, I could be wrong and this isn't a hill that I'm interested in dying on.

src/routes/__error.svelte
```diff
@@ -0,0 +1,31 @@
...
+  export let status;
+  let requestFromAPIError;
+  console.log(error);
+  try {
+    requestFromAPIError = JSON.parse(error.message);
+    if (isNetworkError(requestFromAPIError)) {
+      status = requestFromAPIError.statusCode;
+    }
+  } catch (e) {
+    //
```
stevekinney, line 27: Drop this?

src/routes/__layout.svelte
```diff
@@ -62,7 +63,13 @@
   <Header />
   <section id="content" class="h-full mx-10 mb-10 mt-8">
     <div class="flex flex-col h-full gap-4">
-      <slot />
+      <ErrorBoundary
+        onError={() => {
+          // console.log('OnError!', args);
```
stevekinney, line 68: 🤨

src/routes/client-error.svelte
```diff
@@ -0,0 +1,7 @@
+<script lang="ts">
+  throw {
+    statusCode: 404,
+    statusText: 'zomg unfound',
```
stevekinney, line 4: 🤨

src/routes/namespaces/[namespace]/workflows/[workflow]/__error.svelte
```diff
@@ -0,0 +1,37 @@
...
+  let requestFromAPIError;
+
+  try {
+    requestFromAPIError = JSON.parse(error.message);
+    if (isNetworkError(requestFromAPIError)) {
+      status = requestFromAPIError.statusCode;
+      console.log(status);
+    }
+  } catch (e) {
+    //
```
Comment (stevekinney) [inline src/routes/namespaces/[namespace]/workflows/[workflow]/__error.svelte:26]: Drop this?

src/routes/server-error.svelte
```diff
@@ -0,0 +1,9 @@
+<script lang="ts">
+  import type { NetworkError } from '$lib/utilities/request-from-api';
+
+  throw {
+    statusCode: 500,
+    statusText: "Shits' broke",
```
stevekinney, line 6: 🤨

src/lib/utilities/request-from-api.ts
GiantRobots: The type of this won't end up elsewhere, but it's here to type check the object we're throwing so someone in the future will get an error if they add a "NewKey" without updating the type interface.

src/lib/stores/error-store.ts
```diff
@@ -0,0 +1,4 @@
+import type { NetworkError } from '$lib/utilities/request-from-api';
+import { writable } from 'svelte/store';
+
+export const networkError = writable<NetworkError | null>(null);
```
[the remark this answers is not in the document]
GiantRobots: We subscribe to the store in the error boundary view to get app errors to render into the error boundary. Could probably come up with a better way but that's what I have for now.

src/lib/components/error-boundary.svelte
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    if (onError && theError) {
+      onError(theError);
+    }
+
+    if (isNetworkError($theError)) {
+      status = $theError.statusCode;
+    }
+  }
```
GiantRobots, line 28: @stevekinney This is where we check the error that is thrown into the error boundary or check the error store to react to an error event during runtime.

stevekinney (approved):

PR #260, opened 2022-02-12
Navigate to default namespace from settings

feedmeapples:
## What was changed

Starts navigating to the default namespace set through Settings 

## Why?

Allows to customize the default namespace to navigate to 

## Checklist

1. Closes 

2. How was this tested:

scenario A: Temporal service is offline (or throws errors)
1. Navigate to both http://localhost:3000/ and http://localhost:3000/namespaces
2. Expected: redirects http://localhost:3000/namespaces/default/workflows
3. Navigate to http://localhost:3000/namespaces/accounting/workflows
4. Expected: should stay on http://localhost:3000/namespaces/accounting/workflows

scenario B: defaultNamespace is set to empty string in settings ``
1. Navigate to both http://localhost:3000/ and http://localhost:3000/namespaces
2. Expected: redirects http://localhost:3000/namespaces/default/workflows
3. Navigate to http://localhost:3000/namespaces/accounting/workflows
4. Expected: should stay on http://localhost:3000/namespaces/accounting/workflows

Scenario C: defaultNamespace is set to `sales`
1. Navigate to both http://localhost:3000/ and http://localhost:3000/namespaces
2. Expected: redirects http://localhost:3000/namespaces/sales/workflows
3. Navigate to http://localhost:3000/namespaces/accounting/workflows
4. Expected: should stay on http://localhost:3000/namespaces/accounting/workflows

3. Any docs updates needed?


src/routes/__layout.svelte
```diff
@@ -23,7 +23,7 @@
 
     loadUser();
     loadCluster();
-    loadSettings();
+    await loadSettings();
```
GiantRobots, line 26: We probably only want to force this to be syncronous when we actually need the data so there should probably be a better mechanism to do that.

src/lib/services/settings-service.ts
```diff
@@ -11,5 +12,6 @@ export const fetchSettings = async (request = fetch): Promise<Settings> => {
     auth: {
       enabled: !!settings?.Auth?.Enabled,
     },
+    defaultNamespace: settings?.DefaultNamespace || 'default',
```
GiantRobots, line 15: If you make the default in writable null or undefined then you can use the ?? operator and preserve strings but your code should be correct in general.

src/routes/namespaces/index.svelte
```diff
@@ -1,5 +1,7 @@
-<script context="module" lang="ts">
-  export async function load() {
-    return { status: 302, redirect: '/namespaces/default' };
-  }
+<script lang="ts">
+  import { settings } from '$lib/stores/settings';
+  import { goto } from '$app/navigation';
+
+  const navigateTo = $settings.defaultNamespace;
+  goto(`/namespaces/${navigateTo}`);
```
GiantRobots, line 6: Does this add to the history?

ie.

Navigate to

namespaces ->
/namespaces/defaultNamespace

or will the history just be 
/namespaces/defaultNamespace

I think it should be the latter and we should double check.

If it's the first I think we can use the `replaceState` param on goto to support the second version.

src/lib/stores/settings.ts
```diff
@@ -6,6 +6,7 @@ export const settings = writable<Settings>({
   auth: {
```
GiantRobots, line 6: The alternative here is to make this `<Settings | null>` I like doing this because when strict error checking is on it makes it a bit easier to enforce that you're checking if settings is actually set when you need it as opposed to potentially getting an empty incorrect configuration. 

Just commenting you don't have to change if you don't like that approach.

GiantRobots (changes requested):
Fix the synchronous call in the __layout code.

src/lib/services/settings-service.ts
feedmeapples: addressed and made the goto navigation reactive until there is a value fetched

src/lib/stores/settings.ts
feedmeapples: done,
though since rn checking for null is not enforced in svelte files, this seems to result in potentially more errors if a dev forrgets to put a ?

src/routes/namespaces/index.svelte
feedmeapples: done

src/routes/__layout.svelte
feedmeapples: removed await and made the navigation reactively wait until there is a value

stevekinney (approved):

stevekinney:
Closes #265.

GiantRobots (approved):

PR #269, opened 2022-02-22
Beta badge

GiantRobots:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/lib/components/badge.svelte
+++ b/src/lib/components/badge.svelte
@@ -1,32 +1,28 @@
 <script lang="ts">
-  export let type: BadgeTypes;
-  type BadgeTypes = keyof typeof Badges;
+  type BadgeTypes = 'alpha' | 'beta';
 
-  // We need to keep the entire classname intact so tailiwnd
-  // doesn't auto purge it
-  const Badges = {
-    beta: `
-        bg-blue-100 
-        text-blue-700 
-        border-blue-700`,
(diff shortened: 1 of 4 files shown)


src/lib/components/badge.svelte
```diff
@@ -0,0 +1,32 @@
+<script lang="ts">
+  export let type: BadgeTypes;
+  type BadgeTypes = keyof typeof Badges;
+
+  // We need to keep the entire classname intact so tailiwnd
```
stevekinney, line 5: I might just use CSS classes here like we do in the rest of the application.

Also, there is a small typo.

stevekinney:
I think we can simplify this a bit:

```svelte
<script lang="ts">
  type BadgeTypes = 'alpha' | 'beta';

  export let type: BadgeTypes;
</script>

<div
  class="rounded-sm text-center inline-block px-1 ml-1 text-sm align-middle"
  class:alpha={type === 'alpha'}
  class:beta={type === 'beta'}
>
  <slot />
</div>

<style lang="postcss">
  .beta {
    @apply bg-blue-100  text-blue-700  border-blue-700;
  }

  .alpha {
    @apply bg-purple-100 text-purple-700 border-purple-700;
  }
</style>

```

stevekinney (approved):

PR #283, opened 2022-03-04
Support custom OIDC auth scenarios

feedmeapples:
## What was changed

Starts reading and setting additional auth options, ex. organization id and invitation code. Option names are controlled by [ui server config](https://github.com/temporalio/ui-server/blob/28e513f7ad770bde8cef290ebfe2b1764f323c22/config/development.yaml#L22). Values for these options may come from outside, such as from an invitation email

## Why?

Supports custom auth scenarios such as Auth0 invitation flow

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/routes/_header.svelte
+++ b/src/routes/_header.svelte
@@ -9,6 +9,18 @@
   import { user } from '$lib/stores/user';
 
   import NavigationLink from './_navigation-link.svelte';
+  import { page } from '$app/stores';
+
+  let loginUrl = '';
+  $: {
+    const query = $settings?.auth?.options
+      .filter((o) => !!$page.url.searchParams.get(o))
+      .map((o) => `&${o}=${$page.url.searchParams.get(o)}`)
+      .join();
+
+    loginUrl =
+      import.meta.env.VITE_API + '/auth/sso?' + encodeURIComponent(query);
+  }
 </script>
 
 <header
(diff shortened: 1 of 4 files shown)


src/routes/_header.svelte
```diff
@@ -55,8 +67,7 @@
       {:else}
         <button
           class="header-button"
-          on:click={() =>
-            window.location.assign(import.meta.env.VITE_API + '/auth/sso')}
+          on:click={() => window.location.assign(loginUrl)}
```
GiantRobots, line 70: Hmm this looks like it should be goto() not `window.location`

src/routes/_header.svelte
```diff
@@ -9,6 +9,18 @@
   import { user } from '$lib/stores/user';
 
   import NavigationLink from './_navigation-link.svelte';
+  import { page } from '$app/stores';
+
+  let loginUrl = '';
+  $: {
+    const query = $settings?.auth?.options
+      .filter((o) => !!$page.url.searchParams.get(o))
+      .map((o) => `&${o}=${$page.url.searchParams.get(o)}`)
```
GiantRobots, line 18: Would be nice to change o to either option or optionName

GiantRobots (approved):

PR #297, opened 2022-03-10
Fix namespace capture and default namespaces

GiantRobots:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/lib/components/select/namespace-select.svelte
+++ b/src/lib/components/select/namespace-select.svelte
@@ -11,7 +11,7 @@
     (namespace) => namespace.namespaceInfo.name,
   );
 
-  let selectedNamespace = $currentNamespace;
+  $: selectedNamespace = $currentNamespace;
 
   function switchNamespace() {
     goto('/namespaces/' + selectedNamespace);
--- a/src/lib/stores/namespace.ts
+++ b/src/lib/stores/namespace.ts
@@ -1,18 +1,21 @@
 import { browser } from '$app/env';
 import { page } from '$app/stores';
-import { derived } from 'svelte/store';
-import UrlPattern from 'url-pattern';
+import { namespaceUrlPattern } from '$lib/utilities/namespace-url-pattern';
+import { derived, get } from 'svelte/store';
+import { settings } from '$lib/stores/settings';
 
-const pattern = new UrlPattern('/namespaces/:namespace/*');
 const onlyInBrowser = <T>(fn: () => T): T => {
   if (browser) return fn();
 };
 
 export const namespace = derived(page, ($page) => {
-  const match = pattern.match($page.url.pathname);
+  const match = namespaceUrlPattern.match($page.url.pathname);
+
+  const defaultNamespace = get(settings).defaultNamespace;
+
   const namespace =
     onlyInBrowser(
-      () => localStorage.getItem('currentNamespace') || 'default',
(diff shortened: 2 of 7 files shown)


src/lib/components/select/namespace-select.svelte
```diff
@@ -11,7 +11,7 @@
     (namespace) => namespace.namespaceInfo.name,
   );
 
-  let selectedNamespace = $currentNamespace;
+  $: selectedNamespace = $currentNamespace;
```
stevekinney, line 14: What is this doing? The `$currentNamespace` should be reactive already.

src/lib/stores/namespace.ts
```diff
@@ -1,18 +1,21 @@
...
-const pattern = new UrlPattern('/namespaces/:namespace/*');
 const onlyInBrowser = <T>(fn: () => T): T => {
   if (browser) return fn();
 };
 
 export const namespace = derived(page, ($page) => {
-  const match = pattern.match($page.url.pathname);
+  const match = namespaceUrlPattern.match($page.url.pathname);
+
+  const defaultNamespace = get(settings).defaultNamespace;
```
stevekinney, line 14: Can't we do `$settings.defaultNamespace`?

src/lib/components/select/namespace-select.svelte
GiantRobots: Without
https://user-images.githubusercontent.com/1903839/157760417-3d0ada05-fc0a-4eaa-9c75-49b85370bfc0.mov

With

https://user-images.githubusercontent.com/1903839/157760522-baf493d2-ff82-477a-b95d-519548c51cfa.mov

src/lib/stores/namespace.ts
GiantRobots: Screen Shot 2022-03-10 at 2 51 15 PM

GiantRobots: It gets upset with me when I try. I'd love to use $ is there another way you know of to make that work?

GiantRobots: I think it's because it's in a ts file and Svelte reactivity isn't enabled in those, unless I'm horribly mistaken (I hope so)

src/lib/components/select/namespace-select.svelte
GiantRobots: For whatever reason when browsing between different "current" namespaces it's not reacting to the changes and that enabled it to switch in the dropdown.

src/routes/__layout.svelte
```diff
@@ -40,7 +40,7 @@
     loadCluster().catch(() =>
       notifications.add('error', 'Unable to fetch cluster info'),
     );
-    loadSettings();
+    await loadSettings();
```
stevekinney, line 43: Should all of this just be in the `load` function and not a store at all?

GiantRobots: One way or the other I 100% think this needs to be re-evaluated. I don't like how it works with the rest of the code and the stuff we're doing to make this work correctly in other parts of the app. 

This PR makes the cloud namespaces work so I think it's important to push it out independently of fixing that.

GiantRobots: Yeah thinking more on this it should definitely be hydrated in that load function and passed down to stuff and into context and we should use it there

PR #299, opened 2022-03-10
Rework settings access

GiantRobots:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


src/lib/components/select/namespace-select.svelte
```diff
@@ -7,9 +7,9 @@
   import { getAppContext } from '$lib/utilities/get-context';
   import { isCloud } from '$lib/utilities/env';
 
-  $: namespaces = (getAppContext('namespaces') ?? []).map(
-    (namespace) => namespace.namespaceInfo.name,
-  );
+  $: namespaces = (getAppContext('namespaces') ?? [])
+    .map((namespace) => namespace?.namespaceInfo?.name ?? undefined)
```
stevekinney, line 11: The `?? undefined` seems extraneous?

src/routes/index.svelte
```diff
@@ -1,12 +1,11 @@
...
   import { goto } from '$app/navigation';
   import { onMount } from 'svelte';
+  import { page } from '$app/stores';
 
   onMount(async () => {
-    await loadSettings();
-    if ($settings?.defaultNamespace) {
-      goto(`/namespaces/${$settings.defaultNamespace}`, { replaceState: true });
-    }
+    goto(`/namespaces/${$page.stuff.settings.defaultNamespace}`, {
```
stevekinney, line 7: This should be a `routeFor`.

src/routes/namespaces/index.svelte
```diff
@@ -1,12 +1,11 @@
...
   import { goto } from '$app/navigation';
+  import { page } from '$app/stores';
   import { onMount } from 'svelte';
 
   onMount(async () => {
-    await loadSettings();
-    if ($settings?.defaultNamespace) {
-      goto(`/namespaces/${$settings.defaultNamespace}`, { replaceState: true });
-    }
+    goto(`/namespaces/${$page.stuff.settings.defaultNamespace}`, {
```
stevekinney, line 7: This should be a `routeFor`.

src/lib/components/select/namespace-select.svelte
GiantRobots: It makes TS happy because that type is `string | null | undefined` so setting that to undefined and then filtering appeases the compiler

stevekinney (approved):

PR #301, opened 2022-03-11
Simplify event history

stevekinney:
This PR simplifies the event history routing to a single view and removes a bunch of complexity.

It also allows you to navigate to related links as you would in the compact view.


src/lib/components/input-and-result.svelte
```diff
@@ -4,7 +4,7 @@
 
   export let events: HistoryEventWithId[];
 
-  $: inputAndResult = getWorkflowStartedAndCompletedEvents(events);
+  let inputAndResult = getWorkflowStartedAndCompletedEvents(events);
```
GiantRobots, line 7: Was the reactivity statement here messing with the page or was this for another reason? Mostly curious for my own knowledge

src/lib/components/filter-checkbox.svelte
```diff
@@ -0,0 +1,32 @@
...
+  import { updateQueryParameters } from '$lib/utilities/update-query-parameters';
+
+  export let parameter: string;
+  export let name: string;
+  export let value: boolean;
+
+  let _value: boolean =
+    !!(parameter && $page.url.searchParams.get(parameter)) || value;
+
+  const id = `${parameter || name}-filter`;
```
GiantRobots, line 15: This falsiness check feels not great here. Code is solid but we should be careful going forward with how this component/API works.

src/lib/models/group-events/get-group-for-event.ts
```diff
@@ -0,0 +1,15 @@
...
+  event: HistoryEventWithId,
+  groups: CompactEventGroups,
+): CompactEventGroup => {
+  const eventId = event.id;
+
+  for (const group of groups) {
+    if (eventId === group.id) return group;
+    for (const ev of group.events.values()) {
+      if (ev.id === event.id) {
+        return group;
```
GiantRobots, line 11: A test or two to verify happy/sad paths on this would probably go a long way.

src/lib/utilities/route-for.ts
```diff
@@ -57,35 +50,11 @@ const routeForWorkflow = ({
...
-  eventId: string,
+  view?: 'full' | 'json',
 ): string => {
-  return `${routeForEventHistory(
-    parameters,
-    view,
-  )}/${eventType}-${eventId}/events/${activityId}`;
+  if (!view) return `${routeForWorkflow(parameters)}/history`;
+  if (view === 'full') return `${routeForWorkflow(parameters)}/history/summary`;
+  if (view === 'json') return `${routeForWorkflow(parameters)}/history/json`;
```
GiantRobots, line 57: Drop it like it's hot 🔥

src/lib/components/filter-checkbox.svelte
stevekinney: I don't plan on keeping this code for very long.

src/lib/models/group-events/get-group-for-event.ts
stevekinney: As we discussed, it had a test, but I added an additional test too.

GiantRobots (approved):

Issue #308, opened 2022-03-15
[0.4.0] Query link in history view shows 500 page


tsurdilo:
<img width="711" alt="Screen Shot 2022-03-14 at 9 26 25 PM" src="https://user-images.githubusercontent.com/119422/158287795-163b2343-a58f-4ee6-9425-b5c4f37018e8.png">

In this case for a running workflow when clicking on the "Query" link it shows a 500 error page:

<img width="652" alt="Screen Shot 2022-03-14 at 9 27 30 PM" src="https://user-images.githubusercontent.com/119422/158287855-252f205b-0858-4c58-9b3e-76d055ff4fee.png">

stevekinney:
I just recreated this. The "good" news is that it's the same issue as #291 in so far that it's not about the status of the workflow, but whether or not there is a worker available to query.

PR #328, opened 2022-03-15
Load queries at runtime and remove temporal-system from the list of namespaces

### LINKED ISSUE #291: Clicking on "Stack Trace" on a completed workflow does nothing
Screen Shot 2022-03-07 at 12 11 11 PM

trying to click on "Stack Trace" it does not change the blue underline to see that it did anything
could be some indicator that there is no stack trace for a closed workflow exec

stevekinney:
Fixes issue where the UI would crash if a worker was not running when fetching stack traces or queries.

Fixes #291 and #308.

### DIFF
--- a/src/lib/utilities/request-from-api.ts
+++ b/src/lib/utilities/request-from-api.ts
@@ -22,6 +22,7 @@ type RequestFromAPIOptions = {
   options?: Parameters<typeof fetch>[1];
   token?: string;
   onError?: ErrorCallback;
+  notifyOnError?: boolean;
   shouldRetry?: boolean;
   retryInterval?: number;
 };
@@ -53,6 +54,7 @@ export const requestFromAPI = async <T>(
     request = fetch,
     token,
     shouldRetry = false,
+    notifyOnError = true,
     onError,
     retryInterval = 5000,
   } = init;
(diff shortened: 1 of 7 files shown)


src/lib/utilities/request-from-api.ts
```diff
@@ -87,14 +89,18 @@ export const requestFromAPI = async <T>(
...
-    if (shouldRetry && retryCount > 0) {
-      return new Promise((resolve) => {
-        setTimeout(() => {
-          resolve(requestFromAPI(endpoint, init, retryCount - 1));
-        }, retryInterval);
-      });
+    if (notifyOnError) {
+      handleError(error);
+
+      if (shouldRetry && retryCount > 0) {
```
feedmeapples, line 95: why retrying depends on notifyOnError being true?

src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/stack-trace.svelte
```diff
@@ -30,37 +24,53 @@
...
+    stackTrace = getWorkflowStackTrace({
       workflow,
       namespace,
     });
 
-    isLoading = false;
-    currentdate = new Date();
+    stackTrace.then(() => {
+      currentdate = new Date();
+    });
```
Comment (GiantRobots) [inline src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/stack-trace.svelte:48]: This could be await if we update the methods to be async. But I don't think there's a lot of value in that. Just throwing it out there.

GiantRobots (approved):

PR #329, opened 2022-03-15
Hide null and undefined values in event details

stevekinney:


src/lib/components/event-details.svelte
```diff
@@ -10,22 +10,20 @@
...
-      {:else if typeof value === 'object'}
-        <CodeBlock content={value} />
-      {:else if value}
-        <p><span class="bg-gray-300 text-gray-700 px-2">{value}</span></p>
-      {:else}
-        <p class="text-gray-500">Undefined</p>
-      {/if}
-    </div>
-  </article>
+  {#if value === null || value === undefined}
```
GiantRobots, line 13: !== instead of === to hide these

GiantRobots (approved):

Issue #336, opened 2022-03-16
[0.4.0] ActivityTaskSchedule event - namespace is "undefined"


tsurdilo:
<img width="1365" alt="Screen Shot 2022-03-15 at 11 14 44 PM" src="https://user-images.githubusercontent.com/119422/158509680-6d4d64c5-03d1-4a45-a4f5-84ccc1dcfe8b.png">

Namespace here is "undefined", in v1 its blank. I think in both cases we should not display this property at all.

stevekinney:
Yea, I agree, I pulled out all of the `null` and `undefined` values in #329, but I wasn't totally sure what to do with empty strings.

@canvanooo Are you okay with me pulling it out?

Issue #335, opened 2022-03-16
[0.4.0] Don't display "0s" for timeouts


tsurdilo:
In event history view (when clicking on a particular event and viewing the table with its properties) in 0.4.0 events that are not set for example heartbeatTimeout are shown as "0s". It would be better not show this and leave them blank as its done in v1, for example:

<img width="907" alt="Screen Shot 2022-03-15 at 11 09 00 PM" src="https://user-images.githubusercontent.com/119422/158509162-b42e37d5-ee59-407f-9bbf-98bb86a548d0.png">
<img width="1362" alt="Screen Shot 2022-03-15 at 11 08 51 PM" src="https://user-images.githubusercontent.com/119422/158509163-cea269b0-7e97-47ba-9a49-f4e28b4369b1.png">

stevekinney:
@canvanooo Do you want blank values or just to hide them?

Issue #334, opened 2022-03-16
[0.4.0] History event Type confusing


tsurdilo:
When clicking on a history event, the event details show a "Type" property as first prop in the table. This prop is confusing, and think maybe we should not display it or if we want to display it, match it to the name of the event on the left hand side (the event name that user clicked on)

(click on this event for example:)
<img width="454" alt="Screen Shot 2022-03-15 at 11 02 24 PM" src="https://user-images.githubusercontent.com/119422/158508500-618cda56-1a98-4cfb-8e2c-4683ebda5be4.png">
(first prop)
<img width="712" alt="Screen Shot 2022-03-15 at 11 02 34 PM" src="https://user-images.githubusercontent.com/119422/158508498-32ec32ad-ec3b-42c7-96a6-befcdffdcf45.png">

The type prop is not present in v1:
<img width="1398" alt="Screen Shot 2022-03-15 at 11 02 06 PM" src="https://user-images.githubusercontent.com/119422/158508502-f0371a86-0e8f-45c4-ad33-d0a952cfcb6d.png">

Same applies for other events as well such as WorkflowTaskScheduled, et

PR #340, opened 2022-03-16
Remove additional from the event details table, closes #333, #334, #335, and #336

### LINKED ISSUE #333: [0.4.0] Small text change for input/result
Screen Shot 2022-03-15 at 10 57 24 PM

"Input" matches in both cases, then we have "Results" and "Result", would be nice to align them too as "Result" (no s) or call them in both cases Inputs / Results (with s)

stevekinney:
Remove empty strings and zero second values. Remove the `type` property we add when creating the `HistoryEventWithId` model. Remove heading from workflow inputs and results.

### DIFF
--- a/src/lib/components/event-details.svelte
+++ b/src/lib/components/event-details.svelte
@@ -3,6 +3,13 @@
 
   import CodeBlock from './code-block.svelte';
 
+  const shouldDisplay = (value: unknown): boolean => {
+    if (value === null) return false;
+    if (value === undefined) return false;
+    if (value === '') return false;
+    return true;
+  };
+
   export let attributes:
     | EventAttribute
     | PendingActivity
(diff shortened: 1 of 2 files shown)


GiantRobots (approved):

src/lib/components/event-details.svelte
```diff
@@ -3,25 +3,32 @@
 
   import CodeBlock from './code-block.svelte';
 
+  const shouldDisplay = (key: string, value: unknown): boolean => {
+    if (value === null) return false;
+    if (value === undefined) return false;
+    if (value === '') return false;
+    if (value === '0s') return false;
+    if (key === 'type') return false;
+    return true;
```
GiantRobots, line 12: If we need to expand this it may be useful to check for existence in an object instead of a series of if's totes fine right meow

GiantRobots (approved):

PR #348, opened 2022-03-16
Login page

GiantRobots:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


stevekinney (approved):

src/routes/login/index.svelte
```diff
@@ -0,0 +1,54 @@
...
+  <a href="/" class="flex my-4 max-h-8">
+    <img src="/logo.svg" alt="Temporal Logo" class="max-h-8" />
+  </a>
+</header>
+
+<section class="text-center my-[20vh]">
+  <h1 class="text-8xl font-semibold">Welcome back.</h1>
+  <p class="my-7">Lets get you signed in.</p>
+  <div class="mx-auto">
+    <Button
```
stevekinney, line 45: This should just be a link that looks like a button, right?

src/routes/_header.svelte
```diff
@@ -68,7 +58,11 @@
           />
         </button>
       {:else}
-        <button class="header-button" on:click={() => goto(loginUrl)}>
+        <button
```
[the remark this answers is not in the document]
GiantRobots: Actually this is super annoying to program as a link (runtime, settings funness) however we should have already redirected to the login page so Imma just drop this like it's hot.

src/lib/utilities/get-login-url.ts
```diff
@@ -0,0 +1,16 @@
+export const getLoginUrl = (
+  settings: Settings,
+  currentSearchParams: URLSearchParams,
+): string => {
+  const login = new URL('/auth/sso', settings.baseUrl);
+  const opts = settings.auth.options ?? [];
+
+  opts.forEach((option) => {
```
feedmeapples, line 8: after testing seems opts are not being added when making a request to `/auth/sso`

feedmeapples (changes requested):

src/lib/utilities/get-login-url.ts
feedmeapples: fixed in the last commit

feedmeapples (approved):

stevekinney (approved):

PR #352, opened 2022-03-17
Fall back to the first event when the current event is no longer viewable, closes #343

### LINKED ISSUE #343: [0.4.0] Summary event filtering not clearing properties
1. View workflow history page click on some event to show its props.
2. From filter dropdown select an event that is not in history, the "Summary" clears but the previously selected properties dont, that ends up in a view that looks like this:
Screen Shot 2022-03-16 at 11 32 44 AM

Then you can again filter by lets say "Timer" and you see two events and still old props on right:

Screen Shot 2022-03-16 at 11 37 19 AM

stevekinney:
- Adds logic for defaulting to the closest event when switching views and filters.
- Persists current view when it is still eligible to be viewed.


src/lib/components/toggle-button.svelte
```diff
@@ -11,17 +11,27 @@
   export let group = getAppContext('group');
   export let scale = 1;
   export let href = '#';
+  export let base = href;
+
+  console.log({ href, base });
```
feedmeapples, line 16: testing stuff?

src/lib/utilities/route-for.ts
```diff
@@ -1,137 +1,110 @@
...
   queue: string;
-} & NamespaceParameter;
+};
 
-export type RouteParameters = NamespaceParameter &
-  WorkflowParameters &
-  EventParameter &
-  ActivityParameter &
-  TaskQueueParameter;
+export const isEventView = (view: string): view is EventView => {
```
feedmeapples, line 12: was this meant to be used in `routeForEventHistory` ?

feedmeapples (approved):

PR #355, opened 2022-03-18
Include workflow signaled event under Signals filter

### LINKED ISSUE #342: [0.4.0] Event Filtering in summary view - Signals
Have a workflow that has a bunch of signals:

Screen Shot 2022-03-16 at 11 29 48 AM

Then I select "Signal" from the filter dropdown and nothing shows:

Screen Shot 2022-03-16 at 11 29 56 AM

feedmeapples:
## What was changed

Added WorfklowExecutionSignaled under Signal filters
image

## Why?

fixes #342 

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/lib/models/event-history/get-event-categorization.ts
+++ b/src/lib/models/event-history/get-event-categorization.ts
@@ -20,6 +20,7 @@ export const eventTypeCategorizations: Readonly<
 
   SignalExternalWorkflowExecutionFailed: 'signal',
   SignalExternalWorkflowExecutionInitiated: 'signal',
+  WorkflowExecutionSignaled: 'signal',
 
   TimerCanceled: 'timer',
   TimerFired: 'timer',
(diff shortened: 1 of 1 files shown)


src/lib/models/event-history/get-event-categorization.ts
```diff
@@ -20,6 +20,7 @@ export const eventTypeCategorizations: Readonly<
 
   SignalExternalWorkflowExecutionFailed: 'signal',
   SignalExternalWorkflowExecutionInitiated: 'signal',
+  WorkflowExecutionSignaled: 'signal',
```
feedmeapples, line 23: we may want to have a list of categories per event type?

stevekinney: Totally agree.

stevekinney (approved):

PR #362, opened 2022-03-21
Implement pagination store

stevekinney:
- [x] Abstracts pagination logic to a testable store covered with tests
- [x] Add hooks for updating query parameters
- [x] Reimplements pagination for event history
- [x] Refactors pagination on workflow list page

### DIFF
--- a/src/events.d.ts
+++ b/src/events.d.ts
@@ -165,3 +165,6 @@ type FetchEventsResponse = {
   events: HistoryEventWithId[];
   eventGroups: CompactEventGroups;
 };
+
+type IterableEvent = HistoryEventWithId | CompactEventGroup;
+type IterableEvents = IterableEvent[];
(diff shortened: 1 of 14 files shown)


src/events.d.ts
```diff
@@ -165,3 +165,6 @@ type FetchEventsResponse = {
   events: HistoryEventWithId[];
   eventGroups: CompactEventGroups;
 };
+
+type IterableEvent = HistoryEventWithId | CompactEventGroup;
```
Alex-Tideman, line 169: One other spot we could use the IterableEvent type is in getEventsInCategory() in get-event-categorization.ts.

stevekinney: Good catch! Just updated it!

Alex-Tideman (approved):

PR #370, opened 2022-03-23
Remove namespace context and store

stevekinney:
@feedmeapples Can you battle test this for me against Cloud?

- Removes store for namespace
- Removes namespace from application context


feedmeapples:
tested with our cloud, lg with the last commit

feedmeapples (approved):

PR #372, opened 2022-03-23
Update Cypress; add Github action

stevekinney:


.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,30 @@
...
+  pull_request:
+    branches: [main]
+
+jobs:
+  cypress-run:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Install dependencies
```
feedmeapples, line 15: might also need 

```
      - uses: actions/setup-node@v1
        with:
          node-version: 16
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,30 @@
...
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+jobs:
+  cypress-run:
+    runs-on: ubuntu-20.04
```
feedmeapples, line 11: can we run on `ubuntu-latest`?

stevekinney: Done!

cypress/integration/workflow-executions.spec.js
```diff
@@ -1,43 +1,86 @@
...
+
+      cy.wait('@workflows-api').then((interception) => {
+        expect(interception.request.url).to.contain(result);
+      });
+    });
+
+    it('should send the correct query for Workflow ID', () => {
+      const result = encodeURIComponent('WorkflowId="002c98_Running"');
+
+      cy.get('#workflow-id-filter').type('002c98_Running');
```
feedmeapples, line 57: how is this linked to an html element? didn't find workflow-id-filter

stevekinney: It's created by `FilterSearch`. It takes the query parameter and tacks on `-filter`.

.github/workflows/integration-tests.yml
feedmeapples: i would prefer to have an explicit node setup. I think on other platforms than ubuntu 20.04 this won't work

PR #383, opened 2022-03-24
Support redirecting back to the original UI page

feedmeapples:
## What was changed

Redirects users back to the original open UI page

## Why?

UX

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


src/lib/utilities/get-login-url.ts
```diff
@@ -1,6 +1,7 @@
 export const getLoginUrl = (
   settings: Settings,
   currentSearchParams: URLSearchParams,
+  redirect_url: string,
```
stevekinney, line 4: Nit: Why does this have an underscore? The convention is camelcase.

stevekinney: Also: Should just be optional with a `?` so that we don't have to pass `null` in constantly?

You could just do `redirectUrl = null` and get the same effect without _having_ to always passing `null` in.

src/lib/utilities/get-login-url.ts
```diff
@@ -1,6 +1,7 @@
 export const getLoginUrl = (
```
stevekinney, line 1: Hot take: Should this part of `route-for.ts`?

src/lib/utilities/get-login-url.ts
feedmeapples: after debugging various cases changed the logic that redirectUrl is now expected to come with `currentSearchParams: URLSearchParams`

Alex-Tideman (approved):

PR #393, opened 2022-03-30
Correct workflow syntax with API path

feedmeapples:
## What was changed

Corrected workflow related syntax and API

related: https://github.com/temporalio/ui-server/pull/124

## Why?

wrong syntax could confuse

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/lib/utilities/route-for-api.ts
+++ b/src/lib/utilities/route-for-api.ts
@@ -38,10 +38,10 @@ export function routeForApi(
     'task-queue': `/namespaces/${parameters?.namespace}/task-queues/${parameters?.queue}`,
     workflows: `/namespaces/${parameters?.namespace}/workflows`,
     'workflows.archived': `/namespaces/${parameters?.namespace}/workflows/archived`,
-    workflow: `/namespaces/${parameters?.namespace}/workflows/${parameters?.executionId}/executions/${parameters?.runId}`,
-    'workflow.terminate': `/namespaces/${parameters?.namespace}/workflows/${parameters?.executionId}/executions/${parameters?.runId}/terminate`,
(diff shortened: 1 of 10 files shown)


src/lib/utilities/route-for-api.ts
```diff
@@ -38,10 +38,10 @@ export function routeForApi(
     'task-queue': `/namespaces/${parameters?.namespace}/task-queues/${parameters?.queue}`,
     workflows: `/namespaces/${parameters?.namespace}/workflows`,
     'workflows.archived': `/namespaces/${parameters?.namespace}/workflows/archived`,
-    workflow: `/namespaces/${parameters?.namespace}/workflows/${parameters?.executionId}/executions/${parameters?.runId}`,
-    'workflow.terminate': `/namespaces/${parameters?.namespace}/workflows/${parameters?.executionId}/executions/${parameters?.runId}/terminate`,
-    events: `/namespaces/${parameters?.namespace}/workflows/${parameters?.executionId}/executions/${parameters?.runId}/events`,
-    query: `/namespaces/${parameters?.namespace}/workflows/${parameters?.executionId}/executions/${parameters?.runId}/query`,
+    workflow: `/namespaces/${parameters?.namespace}/workflows/${parameters?.workflowId}/runs/${parameters?.runId}`,
```
GiantRobots, line 41: Hmm we should probably modify the workflowId here to be urlEncoded

GiantRobots (approved):

src/lib/utilities/route-for-api.ts
feedmeapples: i did this in a separate stream. this also requires a server side change

Issue #392, opened 2022-03-30
[0.8.0] 404 for workflows that have a forward slash in the id


feedmeapples:
**Describe the bug**

Navigating to a workflow with a `/` in the ID results in a 404 page 

**To Reproduce**

1. Run a workflow with a `/` in the ID
2. Navigate to workflows page
3. Click on the workflow run
4. See 404 error

**Expected behavior**

Page should still open

**Screenshots**

![image](https://user-images.githubusercontent.com/11838981/160733735-07b0fde5-41d2-4689-9189-a3fbc86a91d6.png)

**Desktop (please complete the following information):**

- Version UI v0.8.0

**Additional context**

feedmeapples:
the frontend side fix is pretty straighforward https://github.com/temporalio/ui/pull/394

the ui-server side of the fix is not:
 - grpc-gateway v2 allows to address this with `WithUnescapingMode` https://pkg.go.dev/github.com/grpc-ecosystem/grpc-gateway/v2@v2.10.0/runtime?utm_source=gopls#WithUnescapingMode
However updating to grpc-gateway v2 introduces a breaking change incompatible with github.com/temporalio/gogo-protobuf/protoc-gen-gogoslick

- another option would be to move Workflow ID into search parameters for GET request. Then for consistency do the same for the rest of params and have all of them as search params. 
In this case the API could change 
- from `/namespaces/${nid}/workflows/${wid}/runs/${runId}/terminate`
- to `/workflows/terminate?namespaceId=${nid}&workflowId=${wid}&runId=${runId}`

 @GiantRobots @Alex-Tideman @stevekinney 
 
 what's your opinion about the latter?

stevekinney:
I don't hate the latter option.

feedmeapples:
later realized the latter won't fix the issue.

I'm considering migrating `WithUnescapingMode` from `grpc-gateway` `v2` into `v1`

Alex-Tideman:
@feedmeapples Any chance we could use something like https://pkg.go.dev/net/url#QueryEscape to do this instead of the grpc-gateway?

feedmeapples:
there have been more issues with other special characters https://github.com/temporalio/ui/pull/491

PR #394, opened 2022-03-31
Fix workflow page 404 by encoding ID and forward slash

feedmeapples:
## What was changed

* URI encodes workflow IDs in routing and GET paths

## Why?

Fixes 404 on workflows page when workflow ID contains a forward slash #392 

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


stevekinney:
This has a regression where workflows with `#` in the ID no longer get encoded correctly.

PR #396, opened 2022-04-01
Route for refactor

GiantRobots:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/lib/utilities/route-for-api.ts
+++ b/src/lib/utilities/route-for-api.ts
@@ -1,48 +1,67 @@
+import urlcat from 'urlcat';
+
 let base = (import.meta.env?.VITE_API as string) ?? process.env.VITE_API;
-if (base.endsWith('/')) base = base.slice(0, -1);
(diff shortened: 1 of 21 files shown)


stevekinney:
Can we conform to the same way we do [the client-side `routeFor`](https://github.com/temporalio/ui/blob/main/src/lib/utilities/route-for.ts)?

src/lib/utilities/route-for-api.ts
```diff
@@ -1,48 +1,67 @@
+import urlcat from 'urlcat';
+
 let base = (import.meta.env?.VITE_API as string) ?? process.env.VITE_API;
-if (base.endsWith('/')) base = base.slice(0, -1);
+base = `${base}/api/v1/`;
```
feedmeapples, line 4: could you add few tests for relative paths when `VITE_API` is set to empty string or `/`

feedmeapples (changes requested):

PR #417, opened 2022-04-11
Improve summaries; refactor views

stevekinney:
- Prefer failure messages when present
- In compact view: use the last event for the summary.
- Reduce duplication for viewing events.


src/lib/components/event/event-summary-row.svelte
```diff
@@ -1,13 +1,25 @@
...
+  export let expandAll = false;
+  export let compact = false;
+
+  let selectedId = event.id;
 
-  export let event: HistoryEventWithId;
-  export let expandAll: boolean = false;
+  let eventGroup = isEventGroup(event)
+    ? event
+    : getGroupForEvent(event, $page.stuff.eventGroups);
```
Alex-Tideman, line 19: Gonna need to do getGroupForEvent(event, $page?.stuff?.eventGroups ?? $importEventGroups) to make this refactor work with the importer since the event groups are set in a store and not on the load function (stuff)

src/lib/components/event/event-group-details.svelte
```diff
@@ -1,18 +1,22 @@
...
-            <span class="event-type" class:active={id === eventId}
+          <li
+            on:click|preventDefault|stopPropagation={() => {
+              selectedId = id;
+            }}
+          >
+            <span class="event-type" class:active={id === selectedId}
               >{eventInGroup.eventType}</span
             >
+            ({id})
```
Alex-Tideman, line 19: Don't think this needs to be changed, but I thought it was a count and not an id. Just food for thought

stevekinney: What if we put a “#” before it?

src/lib/components/event/event-summary-row.svelte
stevekinney: Ah, maybe we should just use context or pass it down instead.

Alex-Tideman: Yeah pass it on down to the component.

PR #422, opened 2022-04-12
Show version banner only if recommend > current version

feedmeapples:
## What was changed

added sanity checks before showing banner about new Temporal version

## Why?

removing reliance on Temporal service and Version info tool to recommend only newer versions

## Checklist

1. Closes 

2. How was this tested:

Tried on Temporal versions `1.14.2`, `1.15.2` and `1.16.0` and verified that banner is shown accordingly

3. Any docs updates needed?

### DIFF
--- a/src/lib/components/banner.svelte
+++ b/src/lib/components/banner.svelte
@@ -9,11 +9,32 @@
     Low: 'low',
   } as const;
 
+  const isNewer = (version1: string, version2: string) => {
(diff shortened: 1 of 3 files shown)


stevekinney (approved):
Maybe we bust that logic out into a unit test?

feedmeapples:
> Maybe we bust that logic out into a unit test?

done

Issue #429, opened 2022-04-13
[0.10.0] UI won't load if custom authorizer rejects anonymous access to ListNamespaces


jmcnevin:
**Describe the bug**
We are seeing a blank screen when loading the UI.  The network panel in the browser shows 401 responses to the backend for requests to `api/v1/namespaces?`.  We have the UI configured for OAuth SSO, but we never get to the point where the user can log in.  Our temporal server has a custom authorizer in place that does not permit anonymous requests to the ListNamespaces rpc operation.

**To Reproduce**
Put an authorizer in place that requires a bearer token from GRPC clients.
Configure the UI to use and SSO provider.

**Expected behavior**
UI should not attempt anonymous GRPC requests prior to the user logging in.

**Desktop (please complete the following information):**
- OS: MacOS
- Browser Firefox

feedmeapples:
thx for testing & reporting. Should be fixed in 0.10.1

PR #430, opened 2022-04-13
Fix unathorized error failing UI load

feedmeapples:
## What was changed

addressing UI blank page with requests failing with unauthorized errors 

## Why?

addressing #429 

## Checklist

1. Closes 

2. How was this tested:

Tested against Temporal Cloud and verified Login & Main functionality works as expected

3. Any docs updates needed?

### DIFF
--- a/src/routes/__layout.svelte
+++ b/src/routes/__layout.svelte
@@ -20,7 +20,7 @@
     let cluster: GetClusterInfoResponse;
     let namespacesResp: ListNamespacesResponse;
 
-    if (!settings.auth.enabled || user.email) {
+    if (!settings.auth.enabled || user?.email) {
       namespacesResp = await fetchNamespaces(settings, fetch);
 
       defaultNamespace = getDefaultNamespace({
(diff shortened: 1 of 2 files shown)


src/routes/__layout.svelte
```diff
@@ -10,22 +10,32 @@
...
+    let namespaces: ListNamespacesResponse;
 
-    const defaultNamespace = getDefaultNamespace({ namespaces, settings });
-    const user = await fetchUser(fetch);
-    const cluster = await fetchCluster(settings, fetch);
+    if (!settings.auth.enabled || user.email) {
+      namespaces = await fetchNamespaces(settings, fetch);
+
+      defaultNamespace = getDefaultNamespace({
+        namespaces: [],
```
Alex-Tideman, line 27: We still want to pass in the namespaces here instead of []

feedmeapples: good catch

feedmeapples:
working on adding cypress tests to catch that separately

stevekinney (approved):

PR #456, opened 2022-04-19
Truncate workflow id and type and add copyable tooltip

Alex-Tideman:
## What was changed
The workflow table can have different heights for rows based on the length of the workflow id and type. To make the table more consistent, this PR truncates the id and type based on the column width as well as adds a tooltip with the full id and type. You can click on the tooltip to quickly copy the id or type.

Screen Shot 2022-04-19 at 2 31 10 PM
Screen Shot 2022-04-19 at 2 31 30 PM

## Why?
Easier to view the table when all rows are the same height, but can also view full information and copy quickly if needed.

### DIFF
--- a/src/lib/components/copyable.svelte
+++ b/src/lib/components/copyable.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
   import Icon from 'svelte-fa';
   import { faCheck, faCopy } from '@fortawesome/free-solid-svg-icons';
+  import { noop } from 'svelte/internal';
 
   export let content: string;
   export let visible = false;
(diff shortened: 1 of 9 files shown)


src/lib/components/copyable.svelte
```diff
@@ -33,17 +35,37 @@
...
 </script>
 
-<div class="flex gap-2 items-center group {$$props['container-class']}">
-  <slot>
-    <span class={$$props.class} class:select-all={!$$slots.default}
-      >{content}</span
-    >
-  </slot>
-  <button on:click|preventDefault|stopPropagation={copy}>
+{#if clickAllToCopy}
```
stevekinney, line 38: I don't love the giant conditional, maybe two components is better? But, this is fine for now.

Alex-Tideman: Yeah I don't either, but the code is so similar felt like it should stay in same file. If only you could export multiple components in one file...

src/routes/namespaces/[namespace]/workflows/_workflows-summary-table.svelte
```diff
@@ -1,11 +1,46 @@
...
+    workflowIdColumnWidth,
+    workflowTypeColumnWidth,
+  } from '$lib/stores/column-width-store';
+
+  // Need to measure column width since we can't use a tooltip and use overflow-hidden for text overflow
+  let workflowIdWidth: number = 0;
+  let workflowTypeWidth: number = 0;
+
+  $: {
+    workflowIdColumnWidth.set(workflowIdWidth);
```
Comment (stevekinney) [inline src/routes/namespaces/[namespace]/workflows/_workflows-summary-table.svelte:12]: You don't need to do both the reactive statements and use the store. An `=` on a store prefixed with a `$` will do this automatically.

src/lib/components/copyable.svelte
```diff
@@ -33,17 +35,37 @@
...
+  <div
+    class="flex gap-2 items-center group {$$props['container-class']}"
+    on:click|preventDefault|stopPropagation={copy}
```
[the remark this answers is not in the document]
Alex-Tideman: Good call

src/routes/namespaces/[namespace]/workflows/_workflows-summary-table.svelte
Comment (Alex-Tideman) [reply src/routes/namespaces/[namespace]/workflows/_workflows-summary-table.svelte]: 👍

stevekinney (approved):

PR #469, opened 2022-04-21
Make banner component reusable

feedmeapples:
## What was changed

Decouples Banner.svelte from Temporal Version notification logic
Makes Banner component generic and reusable

## Why?

Will allow to differentiate on Banner implementations

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


src/lib/components/banner/banner-temporal-version.svelte
```diff
@@ -0,0 +1,28 @@
...
+
+  export let cluster: ClusterInformation;
+
+  const severities = {
+    High: 'high',
+    Medium: 'medium',
+    Low: 'low',
+  } as const;
+
+  $: recommended = cluster?.versionInfo?.recommended;
```
stevekinney, line 13: Do these all need to be reactive? Ideally a new `cluster` is being passed in, right?

src/lib/stores/banner.ts
```diff
@@ -1,7 +1,12 @@
+import { get } from 'svelte/store';
 import { persistStore } from '$lib/stores/persist-store';
 
-export const closedBannerId = persistStore('closedBannerId', null);
+export const closedBanners = persistStore('closedBanners', JSON.stringify([]));
```
stevekinney, line 4: Maybe `persistStore` should just handle stringifying and parsing since it's _always_ writing to `localStorage`?

src/lib/stores/banner.ts
```diff
@@ -1,7 +1,12 @@
...
 
 export const close = (bannerId: string): void => {
-  closedBannerId.set(bannerId);
+  let banners = JSON.parse(get(closedBanners)) ?? [];
+  banners = [...banners, bannerId];
+  banners = banners.slice(banners.length - 10, banners.length);
```
stevekinney, line 9: Why 10? Even if we have a good reason, we should at least store that value in a well-named variable that explains why.

src/lib/components/banner/banner.svelte
```diff
@@ -0,0 +1,37 @@
+<script lang="ts">
+  import { closedBanners, close } from '$lib/stores/banner';
+  export let key: string;
+  export let severity: 'high' | 'medium' | 'low';
+  export let message: string;
+  export let link: string;
+
+  $: show = !(JSON.parse($closedBanners) ?? []).includes(key);
```
stevekinney, line 8: There is a lot of logic to unpack here:

1. If it does _not_
2. include a key
3. from this parsed store
4. _or_ an empty array
5. show it

stevekinney: It seems like we could move this business logic out of the view. Let some regular JavaScript function or derived store figure out what to show and just show the banners.

src/lib/stores/banner.ts
feedmeapples: yea sg, making persistStore generic

feedmeapples: yea sg, made `persistStore` generic and handle serialization

src/lib/components/banner/banner-temporal-version.svelte
feedmeapples: yea no need anymore, updated. I think it used to come from the store.

src/lib/components/banner/banner.svelte
feedmeapples: wondering if this would be easier to read?
`$: hide = ($closedBanners ?? []).includes(key);`

about moving out the logic.. might go with the derived store, though what's the issue with having this in the component? seems pretty minimal

stevekinney (approved):

PR #477, opened 2022-04-25
Redirect to login if unauthorized

Alex-Tideman:
## What was changed
In the root layout, check for user. If they exist, continue on, otherwise redirect to login page.

## Why?
Better auth flow, don't allow users to view namespace information if not logged in.

### DIFF
--- a/src/routes/__layout-root.svelte
+++ b/src/routes/__layout-root.svelte
@@ -10,25 +10,29 @@
   import { fetchCluster } from '$lib/services/cluster-service';
   import { fetchNamespaces } from '$lib/services/namespaces-service';
   import { getDefaultNamespace } from '$lib/utilities/get-namespace';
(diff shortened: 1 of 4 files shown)


src/routes/__layout-root.svelte
```diff
@@ -10,26 +10,30 @@
...
-      namespacesResp = await fetchNamespaces(settings, fetch);
-
-      defaultNamespace = getDefaultNamespace({
-        namespaces: namespacesResp?.namespaces,
-        settings,
-      });
-      cluster = await fetchCluster(settings, fetch);
+    if (!isAuthorized(settings, user)) {
+      return {
+        status: 302,
```
Alex-Tideman, line 22: The question here is 302 to /login or a 404.

stevekinney (approved):

PR #480, opened 2022-04-25
Address bugs on workflow execution page

stevekinney:
- Address issue where the namespace dropdown and the workflows would get out of sync
- Remove flickering when updating the workflow executions list
- Add tests

### DIFF
--- a/src/lib/stores/workflows.ts
+++ b/src/lib/stores/workflows.ts
@@ -1,0 +1,41 @@
+import { Writable, writable } from 'svelte/store';
+import { page } from '$app/stores';
+
+import { fetchAllWorkflows } from '$lib/services/workflow-service';
+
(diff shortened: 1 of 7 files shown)


src/lib/stores/workflows.ts
```diff
@@ -0,0 +1,41 @@
...
+import type { StartStopNotifier } from 'svelte/store';
+
+type WorkflowStoreParameters = {
+  namespace: string;
+  query: string;
+};
+
+type EventualWorkflows = Eventual<CombinedWorkflowExecutionsResponse>;
+
+const updateWorkflows: StartStopNotifier<EventualWorkflows> = (set) => {
```
Alex-Tideman, line 16: I like this. One place to read why workflows are updated

Alex-Tideman (approved):

PR #484, opened 2022-04-25
Use native underlines for links

stevekinney:


src/lib/components/error.svelte
```diff
@@ -25,19 +26,19 @@
...
+    <Link
       href={currentLocation}
       on:click={() => {
         if (browser) {
           window.location.reload();
         }
       }}
       target="_self"
-      class="underline-offset-2 underline">Try a refresh</a
+      class="underline-offset-2 underline">Try a refresh</Link
```
GiantRobots, line 37: You could probably drop some of the underline classes here because they're defined in the component.

GiantRobots (approved):

Alex-Tideman (approved):

src/lib/components/error.svelte
stevekinney: Oh, good callout.

PR #489, opened 2022-04-26
Add reverse query param to event history route

Alex-Tideman:
## What was changed
Persist reverse sorting of event history to each subsequent workflow you visit. Also remove indenting of subrow events.

### DIFF
--- a/src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/history/index.svelte
+++ b/src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/history/index.svelte
@@ -4,13 +4,17 @@
   import { onMount } from 'svelte';
   import { routeForEventHistory } from '$lib/utilities/route-for';
   import { page } from '$app/stores';
+  import { eventFilterSort } from '$lib/stores/event-filters';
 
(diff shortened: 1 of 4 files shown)


stevekinney (approved):

src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/history/index.svelte
```diff
@@ -4,13 +4,17 @@
...
   import { page } from '$app/stores';
+  import { eventFilterSort } from '$lib/stores/event-filters';
 
   const { namespace, workflow, run } = $page.params;
 
   onMount(async () => {
+    const queryParams =
+      $eventFilterSort === 'reverse' ? { sort: 'reverse' } : undefined;
     goto(
       routeForEventHistory({
```
Comment (stevekinney) [inline src/routes/namespaces/[namespace]/workflows/[workflow]/[run]/history/index.svelte:15]: What about just using `updateQueryParams`? It seems like you're setting yourself up for pain by keeping this in a store _and_ trying to keep the query param in sync?

You could easily do something like `$page.url.searchParams.has('sort')` or `$page.url.searchParams.get('sort') === reverse`.

PR #491, opened 2022-04-26
Fix workflow page 404 on special characters in workflow ID

feedmeapples:
## What was changed

makes URI path encoding compatible with Svelte Kit's/vite's URI decoding which uses `decodeURI`

runes from https://github.com/minimaxir/big-list-of-naughty-strings/blob/master/blns.txt
image

## Why?

Resolves 404 on workflow page

## Checklist

1. Closes 

2. How was this tested:

Tested on all special characters

3. Any docs updates needed?

### DIFF
--- a/src/lib/utilities/route-for.ts
+++ b/src/lib/utilities/route-for.ts
@@ -1,5 +1,6 @@
 import { browser } from '$app/env';
 import { toURL } from '$lib/utilities/to-url';
+import { encodeURIForSvelte } from '$lib/utilities/encode-uri';
 
 type RouteParameters = {
   namespace: string;
@@ -62,7 +63,7 @@ export const routeForWorkflow = ({
   run,
   ...parameters
 }: WorkflowParameters): string => {
-  const wid = encodeURIComponent(workflow);
+  const wid = encodeURIForSvelte(workflow);
 
   return `${routeForWorkflows(parameters)}/${wid}/${run}`;
 };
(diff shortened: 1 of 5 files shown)


src/lib/utilities/route-for.ts
```diff
@@ -62,7 +63,7 @@ export const routeForWorkflow = ({
   run,
   ...parameters
 }: WorkflowParameters): string => {
-  const wid = encodeURIComponent(workflow);
+  const wid = encodeURIForSvelte(workflow);
```
stevekinney, line 66: This is fine for now, but maybe later we come up with a better name. 😂

stevekinney (approved):

src/lib/utilities/route-for.ts
feedmeapples: let's see maybe they resolve this issue by the time of v1.0.0 :) https://github.com/temporalio/ui/pull/491/files#diff-f21ba85e44db5049b5d73da8f05e545676388faba448b457f8eb61ace4a5edfcR2

feedmeapples: let's see maybe they resolve this issue by the time of v1.0.0 :) https://github.com/temporalio/ui/pull/491/files#diff-f21ba85e44db5049b5d73da8f05e545676388faba448b457f8eb61ace4a5edfcR2

PR #492, opened 2022-04-28
Remove SSR, fix onError listener and use correct websocket var for hasWebsocket check

Alex-Tideman:
## What was changed
Fixed three issues with data-converter websocket setup. 
## Why?
To get data-converter working.

## Test
Confirmed fixes with getting Data Converter Test using Go Encryption sample to successful convertPayload and show success status of data-converter.


GiantRobots (approved):
🐿

PR #495, opened 2022-04-28
Event History Styling

Alex-Tideman:
## What was changed
Moved all event details to full row below summary row. Updated certain information to show as plain text instead of badges, remove unnecessary keys in code blocks, move Retry Policies to their own lines.

Screen Shot 2022-04-28 at 9 51 49 AM

## Why?
Easier to read event details. Next up -> Grouping them with pills to view only certain information instead of seeing all of it.


stevekinney (approved):

stevekinney:
I'm not @canvanooo, but I'd say we should make the code blocks show up on a new line beneath "Input" (and its friends) and then just have it take up the full width or something.

PR #497, opened 2022-04-28
Add Remote Data Encoder support

Alex-Tideman:
## What was changed
There are now two ways do decode data from the UI:
1. Websocket to tctl plugin
2. Http request to codec endpoint

This PR splits the logic for those two decoding strategies by checking for the codec endpoint (either from settings or through the UI). This PR also adds UI to let a user set and clear both the codec endpoint and the websocket port. You can still use the data-converter/:port route (magic link from tctl) to set the port. However, since the codec endpoints will have http:// or http:// in them, it makes more sense to use a UI instead of a route.

Screen Shot 2022-05-03 at 12 53 25 PM

Screen Shot 2022-05-03 at 12 53 52 PM

## Why?
Allows decoding of payloads from a remote endpoint as well as tctl plugin.


src/lib/components/data-encoder-status.svelte
```diff
@@ -6,16 +6,30 @@
...
+    ? $lastDataConverterStatus === 'error'
+    : $lastDataEncoderStatus === 'error';
+  $: success = $dataConverterPort
+    ? $lastDataConverterStatus === 'success'
+    : $lastDataEncoderStatus === 'success';
 </script>
 
-{#if $dataConverterPort}
-  {#if $lastDataConverterStatus === 'notRequested'}
+{#if $dataEncoderEndpoint || $dataConverterPort}
```
GiantRobots, line 31: It would be nice if we had a derived store with the meaning of this like enabledRemoteDataEncoder

src/lib/components/data-encoder-status.svelte
```diff
@@ -6,16 +6,30 @@
...
   import Tooltip from '$lib/components/tooltip.svelte';
+  $: notRequested = $dataConverterPort
+    ? $lastDataConverterStatus === 'notRequested'
+    : $lastDataEncoderStatus === 'notRequested';
+  $: error = $dataConverterPort
+    ? $lastDataConverterStatus === 'error'
+    : $lastDataEncoderStatus === 'error';
+  $: success = $dataConverterPort
+    ? $lastDataConverterStatus === 'success'
+    : $lastDataEncoderStatus === 'success';
```
GiantRobots, line 28: Originally I had used booleans and then realized I was doing a lot of resets in the logic like if error success = false not requested = false error = true kind of stuff. But it may be more convenient if we moved back to that approach or used a derived store with the information involved.

src/lib/models/event-history/index.ts
```diff
@@ -12,7 +17,11 @@ export async function getEventAttributes(
   historyEvent: HistoryEvent,
 ): Promise<EventAttributesWithType> {
   const { key, attributes } = findAttributesAndKey(historyEvent);
-  const decodedAttributes = await convertPayloadToJson(attributes);
+  const port = get(dataConverterPort);
+
+  const decodedAttributes = port
+    ? await convertPayloadToJsonWithWebsocket(attributes)
+    : await convertPayloadToJsonWithCodec(attributes);
```
GiantRobots, line 24: What if the remote codec and the websocket are configured? Should we show an error? Or provide some ux to set it to the other option?

src/lib/utilities/decode-payload.ts
```diff
@@ -28,9 +32,47 @@ export function decodePayload(
...
+  if (anyAttributes?.queryResult) {
+    const queryResult = anyAttributes?.queryResult;
+
+    Object.entries(queryResult).forEach(([key, value]) => {
+      queryResult[key] = decodePayload(value);
+    });
+  }
+};
+
+export const convertPayloadToJsonWithCodec = async (
```
GiantRobots, line 74: I dig the separation of these concerns

src/lib/utilities/decode-payload.ts
```diff
@@ -69,42 +111,54 @@ export const convertPayloadToJson = async (
...
-    });
+    if (anyAttributes?.input?.payloads) {
+      anyAttributes.input.payloads = JSONPayload;
+    }
+    if (anyAttributes?.result?.payloads) {
+      anyAttributes.result.payloads = JSONPayload;
+    }
   }
 
+  decodePayloadAttributes(anyAttributes);
```
GiantRobots, line 161: It may be nice to move this outside of either of these so the payload would be base64 decoded independently of the logic to convert with a websocket or remote codec.

GiantRobots (approved):

src/lib/components/data-encoder-status.svelte
stevekinney: Co-signed.

src/lib/utilities/decode-payload.test.ts
```diff
@@ -142,3 +156,37 @@ describe(convertPayloadToJson, () => {
     expect(dataConverterStatus).toEqual('notRequested');
   });
 });
+
+describe(convertPayloadToJsonWithCodec, () => {
```
stevekinney, line 160: Should this be commented out?

stevekinney (approved):

src/lib/utilities/decode-payload.test.ts
Alex-Tideman: Nope, need to mock out fetch within that function.

src/lib/models/event-history/index.ts
Alex-Tideman: Good question. I'll ask Candace.

src/lib/components/data-encoder-status.svelte
Alex-Tideman: 👍

src/lib/components/data-encoder-status.svelte
Alex-Tideman: Tri-signed.

src/lib/services/data-encoder.ts
```diff
@@ -0,0 +1,28 @@
...
+} from '$lib/stores/data-encoder-config';
+import type { Payload } from '$types';
+
+export async function convertPayloadWithCodec(
+  payload: Payload,
+  endpoint: string,
+): Promise<Payload> {
+  const encoderResponse: Promise<Payload> = fetch(endpoint + '/decode', {
+    method: 'POST',
+    body: JSON.stringify(payload),
```
robholland, line 13: The API takes a `Payloads` object now, not a `Payload`.

robholland: (And also returns a `Payloads`)

PR #506, opened 2022-04-29
Implement client-side stores for event history

stevekinney:
Removes fetching of event in the load function (which was blocking) and moves the logic to a client-side store.

Still to do:

- [ ] Implement loading states
- [ ] Progressively fill in event table for large event payloads


Alex-Tideman:
I very much like this approach, it's going to make things like adding polling events much easier. Only issue I see right now are the url params being double encoded. Currently it fails to fetch any events and returns a 400 for any url with encoding. 

Fetching events on main:
http://localhost:8080/api/v1/namespaces/canary/workflows/temporal.canary.cron-workflow.sanity-2022-05-02T09%253A21%253A06-05%253A00%252Fworkflow.standard-visibility/runs/57d24543-2fae-4ec2-a5e6-d90ad9e2501a/events

Fetching events on this branch:
http://localhost:8080/api/v1/namespaces/canary/workflows/temporal.canary.cron-workflow.sanity-2022-05-02T09%25253A21%25253A06-05%25253A00%25252Fworkflow.advanced-visibility.scan/runs/2e4edffe-dcb5-4db1-b8e7-85cd54333b5b/events

stevekinney:
> I very much like this approach, it's going to make things like adding polling events much easier. Only issue I see right now are the url params being double encoded. Currently it fails to fetch any events and returns a 400 for any url with encoding.

I found the root cause, it looks like `$page.params` values are encoded by default, which makes sense. So, now I decode them when reading them and it works.

PR #509, opened 2022-05-03
Add tests for event history views and date formatting

feedmeapples:
## What was changed

Cypress tests for Events History:

- Timeline view
- Compact View
- Formatting of event start time

## Why?

ensuring history works as expected in cases

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/lib/components/dropdown-menu.svelte
+++ b/src/lib/components/dropdown-menu.svelte
@@ -41,11 +41,12 @@
   });
 </script>
 
-<div class="relative inline mx-2" bind:this={menu}>
+<div class="relative inline mx-2" bind:this={menu} data-cy={$$props.dataCy}>
   <IconButton
     icon={faCaretDown}
     on:click={() => (show = !show)}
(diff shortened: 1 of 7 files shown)


src/lib/components/dropdown-menu.svelte
```diff
@@ -41,11 +41,12 @@
   });
 </script>
 
-<div class="relative inline mx-2" bind:this={menu}>
+<div class="relative inline mx-2" bind:this={menu} data-cy={$$props.dataCy}>
```
stevekinney, line 44: You don't need to change this now, but you can also do `$$props['data-cy']` if you want to stay consistent with the resulting HTML.

stevekinney (approved):

Alex-Tideman (approved):

src/lib/components/dropdown-menu.svelte
feedmeapples: i've tried `$$props['data-cy']` at first and this resulted in an undefined

PR #511, opened 2022-05-03
Deduplicate initial fetches on workflow list

stevekinney:


src/lib/stores/workflows.ts
```diff
@@ -1,41 +1,73 @@
...
+  return parameters.subscribe(({ namespace, query }) => {
+    if (isNewRequest(namespace, query, previous)) {
+      withLoading(loading, async () => {
+        const { workflows } = await fetchAllWorkflows(namespace, { query });
+        if (workflows) {
+          set(workflows);
+        } else {
+          setTimeout(() => {
+            set(workflows);
+          }, 300);
```
Alex-Tideman, line 64: What exactly is going on here? Set workflows to undefined after 300?

Alex-Tideman (approved):

PR #513, opened 2022-05-04
Add tests for Events Filtering Sorting and Formatting

feedmeapples:
## What was changed

Cypress tests for Events Filtering, Sorting, Expand button 

## Why?

ensuring history works as expected

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


feedmeapples:
it fails because there is an actual UI issue with Elapsed Time formatting

PR #516, opened 2022-05-04
Fix error parsing queries from the Java SDK

stevekinney:
The Java SDK uses commas between the queries. This change solves for that.

### DIFF
--- a/src/lib/utilities/get-query-types-from-error.ts
+++ b/src/lib/utilities/get-query-types-from-error.ts
@@ -1,9 +1,17 @@
 export const getQueryTypesFromError = (message: string): string[] => {
+  console.log(message);
   const indexOfOpeningBracket = message.indexOf('[');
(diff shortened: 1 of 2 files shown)


src/lib/utilities/get-query-types-from-error.ts
```diff
@@ -1,9 +1,17 @@
 export const getQueryTypesFromError = (message: string): string[] => {
+  console.log(message);
```
Alex-Tideman, line 2: You want to keep this?

Alex-Tideman (approved):

src/lib/utilities/get-query-types-from-error.ts
feedmeapples, line 2: leftover?

stevekinney: Nope.

PR #532, opened 2022-05-12
Move query and searchType out of load

Alex-Tideman:
## What was changed
When there are no workflows found for a query and you try to toggle between basic and advanced search, it gets stuck. This fixes that issue by moving the query and searchType out of the load function.

https://user-images.githubusercontent.com/7967403/168147041-9ba0e43a-8436-4384-aa3a-cdf30d8d8a4f.mov


stevekinney (approved):
Of course it was the `load`.

PR #540, opened 2022-05-17
Failed state colors, fix compact view when sort is descending

Alex-Tideman:
## What was changed
Row highlight colors were added to other failed state workflows in both Timeline and Compact views:

Red: Failed or Timed Out event
Yellow: Canceled event
Pink: Terminated event

Also, this fixes a bug where the compact view was not building the full group when sort was descending because the order of events didn't create them correctly.

Screen Shot 2022-05-17 at 12 17 26 PM

Screen Shot 2022-05-17 at 12 17 19 PM

## Why?
Faster for our users to find the troublesome events.

### DIFF
--- a/src/lib/models/event-groups/get-event-in-group.ts
+++ b/src/lib/models/event-groups/get-event-in-group.ts
@@ -44,26 +44,40 @@ export const eventOrGroupIsFailureOrTimedOut = (
   return eventIsFailureOrTimedOut(event as WorkflowEvent);
 };
 
-const eventIsCanceledOrTerminated = (event) => {
+const eventIsCanceled = (event) => {
   return (
     isActivityTaskCanceledEvent(event) ||
     isTimerCanceledEvent(event) ||
-    isWorkflowExecutionTerminatedEvent(event) ||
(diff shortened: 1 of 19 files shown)


src/lib/models/event-groups/get-event-in-group.ts
```diff
@@ -1,15 +1,83 @@
...
+const eventIsCanceled = (event) => {
+  return (
+    isActivityTaskCanceledEvent(event) ||
+    isTimerCanceledEvent(event) ||
+    isWorkflowExecutionCanceledEvent(event) ||
+    isChildWorkflowExecutionCanceledEvent(event)
+  );
+};
+
+const groupIsCanceled = (eventGroup: EventGroup): boolean => {
```
stevekinney, line 56: I wonder if these are better just a methods on the object?

stevekinney (approved):

src/lib/models/event-groups/get-event-in-group.ts
Alex-Tideman: Good call, moved them to the EventGroup

PR #541, opened 2022-05-17
Show 404 for non-existent namespace, show API error message on error page

Alex-Tideman:
## What was changed
1. Show api error message on error page.
2. We redirect to defaultNamespace if a non-existent namespace is entered in URL. This now shows a 404 page with error message for better error handling. 

Screen Shot 2022-05-17 at 3 17 03 PM

Screen Shot 2022-05-17 at 3 16 26 PM

## Why?
More information to user to understand errors

## Todo

- [ ] Change style of error message?

### DIFF
--- a/src/lib/utilities/get-namespace.ts
+++ b/src/lib/utilities/get-namespace.ts
@@ -15,13 +15,12 @@ export const getNamespace = ({
   namespace,
   defaultNamespace,
   namespaces,
-}: GetNamespaceParameters): string => {
-  if (
-    namespace &&
-    namespaces &&
-    namespaces.find((ns) => ns.namespaceInfo.name === namespace)
-  ) {
-    return namespace;
+}: GetNamespaceParameters): string | undefined => {
+  if (namespace && namespaces) {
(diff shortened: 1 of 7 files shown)


src/lib/utilities/get-namespace.ts
```diff
@@ -15,13 +15,12 @@ export const getNamespace = ({
...
   namespaces,
-}: GetNamespaceParameters): string => {
-  if (
-    namespace &&
-    namespaces &&
-    namespaces.find((ns) => ns.namespaceInfo.name === namespace)
-  ) {
-    return namespace;
+}: GetNamespaceParameters): string | undefined => {
+  if (namespace && namespaces) {
```
stevekinney, line 19: What about…

```js
if (!namespace) return;
if (!namespaces) return;
return namespaces.find((ns) => ns.namespaceInfo.name === namespace);
```

stevekinney (approved):

src/lib/utilities/get-namespace.ts
stevekinney: That way it's easier to reason about the steps we take to make sure we should do the thing.

PR #546, opened 2022-05-22
Fix lint warning

lorensr:
## Description

Fixes this warning:

image

Also, if we don't like warnings passing CI, we could do `eslint --ignore-path .gitignore --max-warnings 0 .`


lorensr:
Underscore prefix is also understood by `svelte-check`, silencing this:

image

stevekinney (approved):

PR #547, opened 2022-05-22
GraphQL POC

lorensr:
## Description

POC for fetching data from our [GraphQL API](https://github.com/temporalio/graphql)! 😃 To try it out:

```
git clone https://github.com/temporalio/graphql.git
cd graphql
npm i
npm start
```

```
git clone https://github.com/temporalio/ui.git
cd ui
git checkout graphql
npm i
npm codegen
npm start
```

^ note the extra codegen step

Results in this beautiful UI. Candace would be so proud.

image

### Pros

Faster development speed due to:

- Less code
- Declarative data fetching: declare the data you want to use in your component, and it will appear 😄
- Better types
- Better tooling (Apollo dev tools, pictured below, and query editors like GraphiQL/Playground)
- Ability for FE devs to edit the server. Since it's in TS and is small—mostly these files:
  - server: https://github.com/temporalio/graphql/blob/main/src/index.ts
  - resolvers: https://github.com/temporalio/graphql/blob/main/src/resolvers.ts
  - schema: https://github.com/temporalio/graphql/blob/main/src/schema.graphql
- Ability to easily go serverless (the graphql server is a single endpoint, and can run in a single serverless function)
- Nice features of Apollo Client: polling, reactive global cache. Normalized, so if you query for a single workflow in the list on the workflow detail page, you get the cached data immediately (and if the server response is different, that data will update). 

image

### Cons

- Further server dev required
  - More queries required—currently only has this functionality: https://github.com/temporalio/graphql/blob/main/src/schema.graphql#L90-L102
  - Needs to also provide Payloads so that remote data converter can be used (currently all data is converted by the server with the default converter)
  - Auth (or we could keep using Go server for that)
  - Do we do data fetching during SSR? I haven't looked into how svelte kit SSR works, but it seems like `svelte-apollo` might not yet support that.
- FE migration time, although we could do incremental adoption, e.g. only use GraphQL API for new features for now, like start workflow.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ node_modules
 /build-local
 /build-cloud
 /coverage
+src/gql/
--- a/codegen.yml
+++ b/codegen.yml
@@ -1,0 +1,7 @@
+schema: 'http://localhost:4000'
+documents:
+  - 'src/**/*.svelte'
+  - '!src/gql/**/*'
+generates:
+  ./src/gql/:
+    preset: gql-tag-operations-preset
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,6 +15,7 @@
         "json-beautify": "^1.1.1",
         "just-debounce": "^1.1.0",
         "sanitize-html": "^2.6.1",
+        "svelte-apollo": "^0.5.0",
         "svelte-fa": "^2.4.0",
         "url-pattern": "^1.0.3",
         "websocket-as-promised": "^2.0.1"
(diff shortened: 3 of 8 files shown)


stevekinney:
I'm going to close this for now since `main` has diverged significantly right now.

lorensr:
Here are my more developed thoughts on the matter 😄

https://twitter.com/lorendsr/status/1597819111559507968

PR #557, opened 2022-05-27
Add `npm install` to 'Trying it out instruction'

AndreevDm:
Otherwise you will get
```bash
sh: 1: svelte-kit: not found
```

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


stevekinney (approved):
Good point. Thanks!

PR #598, opened 2022-06-15
Workflow Type click filter

Alex-Tideman:
## What was changed
In the workflows list, the workflow type was a link to the individual workflow page. This PR changes that behavior to now set the workflow type filter on the link click. If other filters are already applied, it will preserve those in addition to include the type. The rest of the row is still a link to navigate to individual workflow page. A new filter icon is added to the tooltip for the Workflow Type.

Screen Shot 2022-06-15 at 12 30 35 PM
Screen Shot 2022-06-15 at 12 30 26 PM

https://user-images.githubusercontent.com/7967403/173889951-a94cbede-c0ef-4e92-9766-84c01d7ca1db.mov

## Why?
Much quicker for users to filter by type by clicking instead of typing.

### DIFF
--- a/.env
+++ b/.env
@@ -1,2 +1,2 @@
-VITE_API="http://localhost:8080"
+VITE_API="http://localhost:8081"
 VITE_MODE="development"
--- a/.env
+++ b/.env
@@ -1,2 +1,2 @@
-VITE_API="http://localhost:8081"
+VITE_API="http://localhost:8080"
 VITE_MODE="development"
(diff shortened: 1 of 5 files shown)


rossedfort (approved):
lgtm, did we want to remove the filter if they click the workflow type while that filter is applied?

Alex-Tideman:
@rossedfort Great idea, I added a fill to the icon when the filter is applied, if you click it again it clears the filter and fill goes away

PR #610, opened 2022-06-21
Use global state as fallback from env var for base url

Alex-Tideman:
## What was changed
If no VITE_API env var is set, look for global state baseUrl


src/lib/utilities/route-for-api.ts
```diff
@@ -1,16 +1,11 @@
-// import { browser } from '$app/env';
-// import { namespaceUrlPattern } from './namespace-url-pattern';
-
-// const RealBaseUrl = `https://web.${
-//   namespaceUrlPattern.match(window.location.pathname)?.namespace
-// }.tmprl.cloud/api/v1/`;
-
-let base = (import.meta.env?.VITE_API as string) ?? process.env.VITE_API;
+let base =
+  import.meta.env?.VITE_API ??
```
GiantRobots, line 2: This will never resolve because of ?. When this packages does process.env.VITE_API resolve to ""?

I think globalThis needs to come first to have our correct functionality.

PR #611, opened 2022-06-21
improve pending activities summary and details

### LINKED ISSUE #465: Add ability to see pending activity failure details
If a workflow has pending activities, users can specify failure details in the failure that comes from the activity.
Currently (v1 and v2) these details are not shown under pending activities view, just the failure itself:

Screen Shot 2022-04-21 at 11 07 41 AM

when you click on that failure (in this case "My Custom Failure") the full JSON response does include the details:

Screen Shot 2022-04-21 at 11 09 16 AM

see the top-level applicationFailureInfo element (not the cause.applicationFailureInfo which for some reason does not include details).

Would be great to display these details in the pending activities view as well, as it helps users debug the failure. 

User slack thread: https://temporalio.slack.com/archives/CTRCR8RBP/p1650534760129989

rossedfort:
## What was changed
Improve layout and information architecture for pending activities. This includes the "summary" table on the workflow details page as well as the full table under the Pending Activities tab on the workflow details page.

### Pending Activities "Summary"
Before:
Screen Shot 2022-06-21 at 2 56 51 PM

After:
Screen Shot 2022-06-21 at 2 56 45 PM

### Pending Activities "Details"
Before:
Screen Shot 2022-06-21 at 3 00 05 PM

After:
Screen Shot 2022-06-21 at 3 00 08 PM

## Why?
Provide more information to users as well as more obvious indication of failed attempts.

## Checklist

1. Closes #465 

2. How was this tested:

3. Any docs updates needed?


rossedfort:
The Heartbeat Details and Last Failure code blocks are no longer each 50% width, but I think that's ok. I would rather them justify with the `Attempt` column.

Alex-Tideman:
Yeah I think that's fine, looks good to me.

PR #615, opened 2022-06-22
create a copyable input component

rossedfort:
## What was changed
Introduces a new `<Input />` component to holocene, supporting a dark/light variety, icons, and a copyable version.
Screen Shot 2022-06-22 at 12 12 54 PM

## Why?
We'll need a component that allows user to copy text from the ui

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


src/lib/holocene/input.svelte
```diff
@@ -0,0 +1,108 @@
...
+
+  let copied = false;
+  const handleCopy = () =>
+    navigator.clipboard
+      .writeText(value)
+      .then(() => {
+        copied = !copied;
+        setTimeout(() => (copied = false), 1000);
+      })
+      .catch((error) => console.error(error));
```
rossedfort, line 22: This code is copied across the app a few times, is there a better way to share singleton stateful utilities across components in svelte à la custom hooks in react?

rossedfort: This looks promising https://svelte.dev/repl/7580c4426c1947d8aa3d149a05bdc895?version=3.38.3

Alex-Tideman (approved):

PR #618, opened 2022-06-22
add vitest and set up some basic snapshot tests

rossedfort:
## What was changed
- add vitest
- move button to holocene
- write some snapshot tests for button
- add snapshot test workflow

## Why?
vitest > jest

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


src/lib/components/data-encoder-settings.svelte
```diff
@@ -5,7 +5,7 @@
   import { validateHttpOrHttps, validateHttps } from '$lib/utilities/is-http';
 
   import Modal from './modal.svelte';
-  import Button from './button.svelte';
```
GiantRobots, line 8: It may be worthwhile adding a `$holocene` alias because no one likes relative paths.

tsconfig.test.json
```diff
@@ -2,7 +2,7 @@
   "extends": "./tsconfig.json",
   "compilerOptions": {
     "isolatedModules": false,
-    "types": ["jest"]
+    "types": ["jest", "vitest/globals"]
```
GiantRobots, line 5: Hmm this may cause trouble for us. If we're switching to `vitest` we probably want to drop the jest global types as well.

stevekinney: Probably once we fully switch. Does this get rid of Jest completely or just _add_ Vitest?

rossedfort: +1, this PR only adds vitest so I think we'll need both for the time being. The current plan is to move away from Jest incrementally, but it might be worth investigating the effort to do it in one fell swoop.

GiantRobots: We can also import vitest components directly and avoid this situation as well, it has its own pluses and minuses. Positive no implicit imports. Negative more work to migrate in general

stevekinney (approved):

PR #628, opened 2022-06-27
Replace Jest with Vitest

stevekinney:
Replaces Jest with Vitest.

To do:

- [ ] Get type-checking working with Vitest globals or import functions.


vitest.config.ts
```diff
@@ -11,10 +11,14 @@ export default defineConfig({
...
       $holocene: path.resolve(__dirname, './src/lib/holocene/'),
+      $app: path.resolve(__dirname, './src/lib/svelte-mocks/app/'),
     },
   },
   test: {
     globals: true,
     environment: 'jsdom',
+    deps: {
+      inline: ['date-fns'],
+    },
```
rossedfort, line 21: I swear I tried this and it didn't work 😭 maybe it was some other error

PR #641, opened 2022-06-29
Better page transitions, add hash to href to always trigger page store updates

Alex-Tideman:
## What was changed
Quicker page transition and loading icon on workflow page. Add hash to url on update query params.

## Why?
Better UI for workflow page, fixes url query param updates for npm package.

### DIFF
--- a/src/lib/utilities/update-query-parameters.ts
+++ b/src/lib/utilities/update-query-parameters.ts
@@ -35,6 +35,7 @@ export const updateQueryParameters = async ({
   }
 
   if (browser && url.href !== window.location.href) {
+    // Needed to trigger store updates when used as npm package
     if (!url.href.endsWith('#')) {
       url.href = url.href + '#';
     }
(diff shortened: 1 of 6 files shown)


src/lib/utilities/update-query-parameters.ts
```diff
@@ -35,6 +35,10 @@ export const updateQueryParameters = async ({
   }
 
   if (browser && url.href !== window.location.href) {
+    // Needed to trigger store updates when used as npm package
+    if (!url.href.endsWith('#')) {
```
Alex-Tideman, line 39: This is gross, but this works well. It doesn't actually add a '#' in the url due to Svelte magic.

stevekinney: I hate it but I trust you.

stevekinney (approved):

rossedfort (approved):

PR #642, opened 2022-06-29
Add isCloud check for supportsReverseOrder

Alex-Tideman:
## What was changed
With the added temporalVersion check for the reverse call, cloud will always have reverse disabled since the cluster store is never set. I hard coded in 1.17.0 but definitely looking for suggestions here. Is it safe to assume cloud will always be above 1.16? Should I just do a isCloud check here?

## Why?
Add reverse events call for cloud

### DIFF
--- a/src/lib/services/cluster-service.ts
+++ b/src/lib/services/cluster-service.ts
@@ -7,10 +7,7 @@ export const fetchCluster = async (
   settings: Settings,
   request = fetch,
 ): Promise<GetClusterInfoResponse> => {
-  if (settings.runtimeEnvironment.isCloud) {
-    cluster.set({ serverVersion: '1.17.0' });
-    return;
-  }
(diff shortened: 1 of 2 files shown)


stevekinney (commented):
Ugh. Maybe in the `supportsReverseOrder` store we do an `isCloud` check?

The real answer is that we probably need a more well-thought out solution for a registry of what features we do and do not support in a given instance.

Even if you think about the other places we use `isCloud`, that's not really what we care about. We care if the UI has a particular feature.

In a future world, we might want to let users turn on and off features with environment variables.

Alex-Tideman:
Yeah I guess even when a UI feature is typed to the server version, it wouldn't be the worst to have an env flag for a feature and toggle on after we know they are on a version that supports it.

stevekinney (approved):

PR #646, opened 2022-06-30
Always show event history in ascending order in JSON view, closes #645

### LINKED ISSUE #645: Json History view is in reverse order
Json History view is in reverse order. Some users cut/paste it and try to run tests with WorkflowReplayer and it wont work.

Also the page title for json history says "Recent Events" but the json history view always shows the entire history, maybe this is confusing

When Downloading json history, it is also in reverse order which is not useful for users. If possible please show in right order.

stevekinney:
/cc @tsurdilo

### DIFF
--- a/src/lib/services/events-service.ts
+++ b/src/lib/services/events-service.ts
@@ -35,17 +35,14 @@ export const fetchRawEvents = async ({
   namespace,
   workflowId,
   runId,
-  sort,
   onStart,
   onUpdate,
   onComplete,
 }: FetchEventsParameters): Promise<HistoryEvent[]> => {
-  const endpoint = getEndpointForSortOrder(sort);
-
   const response = await paginated(
     async (token: string) => {
       return requestFromAPI<GetWorkflowExecutionHistoryResponse>(
-        routeForApi(endpoint, { namespace, workflowId, runId }),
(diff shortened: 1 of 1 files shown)


stevekinney:
@Alex-Tideman I just decided to lazily fetch the ascending events for this view since I suspect that it's not as frequently accessed as the others.

src/lib/pages/workflow-history-json.svelte
```diff
@@ -1,6 +1,22 @@
...
+
   import CodeBlock from '$lib/components/code-block.svelte';
-  import { events } from '$lib/stores/events';
+  import Loading from '$lib/components/loading.svelte';
+  import { fetchRawEvents } from '$lib/services/events-service';
+  import { decodeURIForSvelte } from '$lib/utilities/encode-uri';
+
+  const { namespace, workflow: workflowId, run: runId } = $page.params;
+  const events = fetchRawEvents({
+    namespace: decodeURIForSvelte(namespace),
```
stevekinney, line 11: It's kind of gross that we need to do this outside of the service itself. Why doesn't `events-service.ts` just handle this under the hood?

Alex-Tideman (approved):

PR #653, opened 2022-06-30
minor style/API changes for holocene components

rossedfort:
## What was changed

- add `type` prop to Button with default value 'button'
- add default empty string to `subtitle` prop on Accordion
- add 2 new icons and undo icon size changes from #651
- add `readonly` prop to DropdownButton with default value `false`
- fix some styles
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


src/lib/holocene/dropdown-button/index.svelte
```diff
@@ -5,31 +5,46 @@
 
   export let label: string;
   export let icon: IconName = undefined;
-  export let showDropdown: boolean = false;
+  export let readonly: boolean = false;
```
Alex-Tideman, line 8: When you set defaults like this, you don't need the type (i.e. it knows the type based on the default value). Just fyi so typescript warnings don't bark at ya.

Alex-Tideman (approved):

src/lib/holocene/dropdown-button/index.svelte
rossedfort: Nice ok, I'm used to explicitly typing everything. Surprised TS would yell at you for being safe 😅

PR #671, opened 2022-07-07
Update the namespace list to call the getNamespaceList function on lo…

GiantRobots:
…ad. This allows cloud-ui to only fetch namespaces once.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


stevekinney (approved):

Alex-Tideman:
Looks good. One thing we'll need to do in the future is make that NamespaceItem href dynamic depending on the page you are on (i.e. click when on a workflows page, go to workflows page for that namespace. click on a schedules page, go to schedules page for that namespace). But that's for another day.

PR #677, opened 2022-07-08
Add nested fields to find payloads

Alex-Tideman:
## What was changed
Add support for nested fields with payloads to decode more payloads

### DIFF
--- a/src/lib/utilities/decode-payload.ts
+++ b/src/lib/utilities/decode-payload.ts
@@ -83,27 +83,38 @@ export const decodePayloadAttributes = (
 };
 
 // List of fields with payloads
-const payloadFields: string[] = [
-  'input',
-  'result',
-  'details',
-  'heartbeatDetails',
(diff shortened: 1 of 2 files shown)


src/lib/utilities/decode-payload.ts
```diff
@@ -83,27 +83,42 @@ export const decodePayloadAttributes = (
...
+  ['details', 'version'],
+  ['details'],
+  ['heartbeatDetails'],
+  ['lastHeartbeatDetails'],
+  ['lastFailure'],
+  ['lastCompletionResult'],
+  ['queryArgs'],
+  ['answer'],
+  ['signalInput'],
 ];
```
stevekinney, line 102: This is a tiny little nit that you can ignore, but this _might_ make the TypeScript typings better and then you can drop the `string[][]` on L86.

```suggestion
] as const;
```

PR #678, opened 2022-07-11
Rework Code Blocks in Pending Activities

stevekinney:
Addresses an edge case where if a failure had long message, it would exceed the width of the container.

Before:

image

After:

Screen Shot 2022-07-11 at 11 00 54 AM


rossedfort (approved):
looks good, did you double check that the content flows well when you make the screen smaller? I think that's why I went with a grid layout.

PR #681, opened 2022-07-11
consolidate disparate tables across the app to use a single table

rossedfort:
## What was changed

adds a single `<Table />` component with support for both simple and fancy variants. Migrates existing tables to use the new table, except for the event history table. Gets rid of global table css and scopes styles to the component.
## Why?

For consistent table styles
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


cypress/integration/workflow-executions.spec.js
```diff
@@ -106,7 +106,7 @@ describe('Workflow Executions List', () => {
         );
 
         cy.get(
-          `[href="/namespaces/default/workflows/${workflowId}/${runId}"] > :nth-child(2) > .table-link`,
+          `[href="/namespaces/default/workflows/${workflowId}/${runId}"] > :nth-child(3) > .table-link`,
```
stevekinney, line 109: Maybe we add a better selector on to the table rows? We are already probably using some kind of index in the `{#each}` block in Svelte.

Alex-Tideman (approved):

PR #683, opened 2022-07-12
only push version bump for publish-package action, changes should already be committed

rossedfort:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


GiantRobots (approved):
Drop it like it's hot

PR #687, opened 2022-07-14
Show search attributes on the namespace settings page, fix CodeBlock

stevekinney:
- Adds support for displaying search attributes in the namespace settings page.
- Fixes a bug in the `CodeBlock` component where it might show stale data.
- Adds support for all operators in the tokenizer.
- Increases test coverage for `is.ts`.

### DIFF
--- a/src/lib/utilities/route-for-api.ts
+++ b/src/lib/utilities/route-for-api.ts
@@ -63,7 +63,9 @@ export function routeForApi(
   parameters: TaskQueueRouteParameters,
   shouldEncode?: boolean,
 ): string;
-export function routeForApi(route: ParameterlessAPIRoutePath): string;
+export function routeForApi(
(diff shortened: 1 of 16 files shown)


src/lib/utilities/route-for-api.ts
```diff
@@ -77,6 +79,7 @@ export function routeForApi(
     user: '/me',
     namespaces: '/namespaces',
     'task-queue': `/namespaces/${parameters?.namespace}/task-queues/${parameters?.queue}`,
+    'search-attributes': '/search-attributes',
```
Alex-Tideman, line 82: Are search attributes not per namespace?

Alex-Tideman: Nevermind, it's per cluster. 👍

Alex-Tideman (approved):

PR #691, opened 2022-07-14
Fix auth redirection to VITE_API address

feedmeapples:
## What was changed

Fixed auth redirection when VITE_API is set to a different origin

## Why?

Auth would always redirect to `window.location.origin` when authenticating instead of VITE_API

## Checklist

1. Closes 

2. How was this tested:

ui-server running under http://localhost:8088/ and ui under http://localhost:3000/
Verified that UI redirects to http://localhost:8088/auth/sso

added unit tests

3. Any docs updates needed?

### DIFF
--- a/src/lib/services/settings-service.ts
+++ b/src/lib/services/settings-service.ts
@@ -1,5 +1,6 @@
 import { browser } from '$app/env';
 import { settings } from '$lib/stores/settings';
+import { getApiOrigin } from '$lib/utilities/get-api-origin';
 import { getEnvironment } from '$lib/utilities/get-environment';
 import { requestFromAPI } from '$lib/utilities/request-from-api';
 import { routeForApi } from '$lib/utilities/route-for-api';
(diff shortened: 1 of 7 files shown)


GiantRobots (approved):

src/lib/services/settings-service.ts
```diff
@@ -21,7 +21,7 @@ export const fetchSettings = async (request = fetch): Promise<Settings> => {
       options: settingsResponse?.Auth?.Options,
     },
     baseUrl:
-      import.meta?.env?.VITE_API ?? browser ? window.location.origin : '',
+      import.meta.env.VITE_API ?? (browser ? window.location.origin : ''),
```
Alex-Tideman, line 24: If this returns as an empty string instead of undefined, we will need || instead of ??

feedmeapples: `VITE_API=""` currently means a relative path the same as "/" https://github.com/temporalio/ui/blob/3ab14b473a34ef593c874673326493218726d664/src/lib/utilities/route-for-api.ts#L17

Alex-Tideman (approved):

PR #707, opened 2022-07-19
Implement Chip Input

rossedfort:
## What was changed

- Better support for Label and/or Placeholder on existing `<Input />` component
- Move existing `<Input />` component to a sub-directory in holocene
- Add a `<ChipInput />` component
  - `,` or `Enter` are valid delimeters
  - `Backspace` removes chips (if not currently typing a new chip)
  - Can click `x` to remove chips

https://user-images.githubusercontent.com/11775628/179817709-b2a93555-f2e9-4c3d-a660-cc7a3903e164.mov

## Why?

Need it for user admin pages
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


src/lib/holocene/input/chip-input.svelte
```diff
@@ -0,0 +1,109 @@
...
+  export let hintText = '';
+  export let validator: (value: string) => boolean = () => true;
+
+  let chips: string[] = [];
+  let displayValue: string = '';
+
+  $: invalid = false;
+  $: {
+    invalid = chips.some((chip) => !validator(chip));
+    value = chips.join(',');
```
rossedfort, line 20: Wondering if callers should just bind to an array of strings instead of a comma separated string.

i.e.
`<ChipInput bind:chips={['a', 'b', 'c']} />`
instead of
`<ChipInput bind:value="a,b,c" />`

stevekinney (approved):

PR #716, opened 2022-07-20
add checkbox input

rossedfort:
## What was changed

- add checkbox input
- clean up some styles for other components
- v2.1.5
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/lib/holocene/navigation/full-nav.svelte
+++ b/src/lib/holocene/navigation/full-nav.svelte
@@ -128,14 +128,6 @@
       </NavRow>
     {:then user}
       {#if user?.email}
-        <NavRow {isCloud} link={linkList.settings}>
-          <Tooltip right hide={$navOpen} text="Settings">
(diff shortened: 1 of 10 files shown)


src/lib/holocene/navigation/full-nav.svelte
```diff
@@ -128,6 +128,14 @@
       </NavRow>
     {:then user}
       {#if user?.email}
+        <NavRow {isCloud} link={linkList.settings}>
```
Alex-Tideman, line 131: I wouldn't add this yet until we have a settings page in cloud. Also going to need a different guard for this since it's only for cloud.temporal.io

rossedfort: Oh ya, good call 👍

Alex-Tideman (approved):

PR #721, opened 2022-07-21
Added stacktrace-query fixtures from a windows machine

BashkirovN:
## Why?
Additional fixtures to support test coverage for windows paths


src/lib/utilities/stack-trace/get-file-paths-from-stack-trace.test.ts
```diff
@@ -0,0 +1,86 @@
+import { describe, expect, it } from 'vitest';
+import UnixTSStacktraceQueryJson from '../../../fixtures/stacktrace-query.ts-sdk.json';
```
stevekinney, line 2: We should be able to update these to `$fixtures/stacktrace-query.ts-sdk.json`.

BashkirovN: I've tried that. Even though the code compiles and works, VSCode gives a 'Cannot find module' error for this syntaxis.

stevekinney (approved):

PR #722, opened 2022-07-22
Move Tab to holocene, add icon for terminal

Alex-Tideman:
## What was changed
Move Tab to holocene, make active a prop instead of computed, add icon

## Why?
Allow Tab to be used in package


src/lib/holocene/icon/paths.ts
```diff
@@ -219,23 +219,23 @@ const globeIcon = {
   paths: [
     {
       d: 'M6 11C8.7614 11 11 8.7614 11 6C11 3.23857 8.7614 1 6 1C3.23857 1 1 3.23857 1 6C1 8.7614 3.23857 11 6 11Z',
-      stroke: '',
+      stroke: 'currentColor',
```
rossedfort, line 222: I think "currentcolor" is technically supposed to be all lowercase, but if this works, nbd.

https://developer.mozilla.org/en-US/docs/Web/CSS/color_value#currentcolor_keyword

rossedfort (approved):

src/lib/holocene/icon/paths.ts
Alex-Tideman: Yeah it works, but good call on checking.

PR #726, opened 2022-07-25
move existing modal component to holocene

rossedfort:
## What was changed

- moves the existing Modal component to Holocene and adds a Chapter in Fiction
- Also makes menu and menu-item styles consistent across selects/dropdown-buttons/split-buttons.
- Fixes a bug with indeterminate checkboxes
- Fixes a typo in file naming (primatives -> primitives)
- Disambiguate Cards and Alerts

### Modal
Screen Shot 2022-07-27 at 9 31 43 AM

### Card
Screen Shot 2022-07-27 at 9 30 31 AM

### Alert
Screen Shot 2022-07-27 at 9 29 37 AM

## Why?

- need a modal in cloud-ui
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


stevekinney (approved):

src/lib/holocene/select/option.svelte
```diff
@@ -18,7 +18,7 @@
 <script lang="ts">
   import { createEventDispatcher } from 'svelte';
   import Icon from '$holocene/icon/index.svelte';
-  import MenuItem from '$holocene/primatives/menu/menu-item.svelte';
+  import MenuItem from '$holocene/primitives/menu/menu-item.svelte';
```
GiantRobots, line 21: I prefer primat ives 🦍

src/lib/holocene/split-button.svelte
```diff
@@ -1,47 +1,56 @@
...
-    @apply cursor-not-allowed opacity-50;
-  }
+    :global(.segment) {
+      @apply relative flex w-fit items-center justify-center py-2 font-secondary text-sm;
+    }
 
-  :global(.split-button-menu) {
-    @apply flex min-w-max flex-col items-start gap-y-4 border-gray-300 p-3 px-6 font-secondary text-sm;
-  }
+    :global(.segment.left) {
```
GiantRobots, line 40: I was thinking on these you could do the terrible thing of :first-child and :last-child and then the css becomes automagic for any amount of containing elements without having to ho hum about what is the leftmost part of the button and the right most. This is totally fine too just a thought

rossedfort: Ah yeah, I think for split button this should be OK, but for something like the Toggle Buttons, we will definitely want to do that.
Screen Shot 2022-07-27 at 10 37 41 AM

GiantRobots: Yeah when I think about it those feel like they could/should be composed of the more primative 🦍 elements. But this approach is also fine for right meow

GiantRobots (approved):

PR #733, opened 2022-07-28
Add error state to Textarea

laurakwhit:
## What was changed

Adds an error state to the `<Textarea />` holocene component.

Screen Shot 2022-07-28 at 5 22 32 PM

## Why?

We'll need the ability to alert a user if there is an error (e.g. if the text is not in the right format).

## Checklist

1. Closes: n/a

2. How was this tested: n/a

3. Any docs updates needed? none

### DIFF
--- a/src/lib/holocene/textarea.svelte
+++ b/src/lib/holocene/textarea.svelte
@@ -3,12 +3,34 @@
   export let disabled: boolean = false;
   export let placeholder: string = '';
   export let rows: number = 5;
+  export let error: string = null;
+  let hasError = error || $$slots.error;
 </script>
 
 <textarea
   class="font-mono min-h-fit w-full rounded border border-gray-900 p-5 text-sm"
+  class:error={hasError}
   bind:value
   {disabled}
   {placeholder}
   {rows}
 />
+{#if hasError}
+  <div class="error-msg">
+    {#if $$slots.error}
+      <slot name="error" />
+    {:else}
+      <p>{error}</p>
+    {/if}
+  </div>
+{/if}
+
+<style lang="postcss">
+  .error {
+    @apply border-danger;
+  }
+
+  .error-msg {
+    @apply border-danger font-primary text-sm font-normal text-danger;
+  }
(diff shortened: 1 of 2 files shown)


src/lib/holocene/textarea.svelte
```diff
@@ -3,12 +3,34 @@
...
+  class:error={hasError}
   bind:value
   {disabled}
   {placeholder}
   {rows}
 />
+{#if hasError}
+  <div class="error-msg">
+    {#if $$slots.error}
+      <slot name="error" />
```
laurakwhit, line 21: Added this so we can create customized errors (e.g. add links).

src/routes/fiction/chapters/textarea.svelte
```diff
@@ -13,6 +13,12 @@
   props={{ disabled: true, value }}
 />
 
+<Chapter
+  description="A textarea with an error"
+  component={Textarea}
+  props={{ value, error: 'There was an error.' }}
+/>
+
```
laurakwhit, line 21: To view an example using `slot` add the following
```js
<Chapter description="A textarea with an error">
  <Textarea {value}><p slot="error">There was an error.</p></Textarea>
</Chapter>
```

src/lib/holocene/textarea.svelte
```diff
@@ -3,12 +3,34 @@
   export let disabled: boolean = false;
   export let placeholder: string = '';
   export let rows: number = 5;
+  export let error: string = null;
+  let hasError = error || $$slots.error;
```
stevekinney, line 7: Whoa, I dig this.

stevekinney (approved):

rossedfort (approved):
Nice 💪

src/lib/holocene/textarea.svelte
laurakwhit: This would have been cool, but apparently you can't wrap slots in if statements within the parent component like I was thinking you could 😞

PR #738, opened 2022-08-01
Check if workflow cannot terminate

Alex-Tideman:
## What was changed
Wrong boolean check on Terminate Workflow modal


stevekinney (approved):
Good catch. I got bitten by my own silly refactoring.

bkniffler:
Thanks for fixing, could we get this released?

PR #739, opened 2022-08-01
Add spellcheck to textarea

laurakwhit:
## What was changed

Adds the `spellcheck` attribute to the `textarea` holocene component.

## Why?

Always an optional `spellcheck` prop to be passed to `Textarea`.

## Checklist

1. Closes: n/a

2. How was this tested:
- [x] Verify adding `spellcheck` set to `true` or `false` works as expected

### DIFF
--- a/src/lib/holocene/textarea.svelte
+++ b/src/lib/holocene/textarea.svelte
@@ -5,6 +5,7 @@
   export let isValid: boolean = true;
   export let placeholder: string = '';
   export let rows: number = 5;
+  export let spellcheck: boolean = null;
   export let value: string;
   export let onBlur: (e: Event) => void = noop;
 </script>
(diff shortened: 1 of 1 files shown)


src/lib/holocene/textarea.svelte
```diff
@@ -5,6 +5,7 @@
   export let isValid: boolean = true;
   export let placeholder: string = '';
   export let rows: number = 5;
+  export let spellcheck: boolean = null;
```
rossedfort, line 8: For my own knowledge: why `null` here instead of just unset or false?

laurakwhit: > All other attributes are included unless their value is [nullish](https://developer.mozilla.org/en-US/docs/Glossary/Nullish) (null or undefined).

I set it to `null` thinking that `spellcheck` should just not be included unless it is defined ([doc link](https://svelte.dev/docs#template-syntax-attributes-and-props))

rossedfort (approved):

src/lib/holocene/textarea.svelte
rossedfort: Nice ok that makes sense!

PR #740, opened 2022-08-02
Update holocene tooltip

laurakwhit:
## What was changed

Adds additional positioning options for the holocene `tooltip` component (`bottomLeft` and `bottomRight`), as well as multiline support, and removes the css shadow.
Screen Shot 2022-08-02 at 11 20 05 AM
Screen Shot 2022-08-02 at 11 20 24 AM
Screen Shot 2022-08-02 at 11 20 34 AM

## Why?

This will allow support for new designs on disabled buttons.

## Checklist

1. Closes: n/a 

2. How was this tested: n/a

### DIFF
--- a/src/lib/holocene/tooltip.svelte
+++ b/src/lib/holocene/tooltip.svelte
@@ -14,6 +14,7 @@
   export let left = false;
   export let copyable = false;
   export let hide: boolean | null = false;
+  export let width: number = null;
 </script>
 
 {#if hide}
@@ -30,6 +31,7 @@
       class:bottomRight
       class:top
       class:topRight
+      style={width ? `white-space: pre-wrap; width: ${width}px;` : null}
     >
       <div class="inline-block rounded-lg bg-gray-800 px-2 py-2">
         {#if copyable}
--- a/src/lib/holocene/tooltip.svelte
+++ b/src/lib/holocene/tooltip.svelte
@@ -9,6 +9,8 @@
   export let topRight = false;
   export let right = false;
   export let bottom = false;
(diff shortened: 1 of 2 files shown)


src/lib/holocene/tooltip.svelte
```diff
@@ -54,7 +60,7 @@
 
 <style lang="postcss">
   .tooltip {
-    @apply invisible absolute left-0 top-0 z-50 inline-block translate-x-12 whitespace-nowrap text-xs opacity-0 shadow-md transition-all;
```
laurakwhit, line 57: I've removed the `shadow-md` as it was causing this behavior 👇 

| Existing | Updated|
| ------- | -------|
|Screen Shot 2022-08-01 at 4 31 55 PM|Screen Shot 2022-08-01 at 4 32 20 PM|

src/lib/holocene/tooltip.svelte
```diff
@@ -9,9 +9,12 @@
   export let topRight = false;
   export let right = false;
   export let bottom = false;
+  export let bottomLeft = false;
+  export let bottomRight = false;
   export let left = false;
   export let copyable = false;
   export let hide: boolean | null = false;
+  export let width: number = null;
```
Alex-Tideman, line 17: Just a small thing, if default is null then type should be number | null

src/lib/holocene/tooltip.svelte
Alex-Tideman: Good catch!

Alex-Tideman (approved):

PR #743, opened 2022-08-02
Add disabled state to Accordion

laurakwhit:
## What was changed

This adds a `disabled` attribute to the holocene `Accordion` component.
Untitled
## Why?

This allows the accordion to not be closed in cases it should not be navigated away from (e.g. there are errors on inputs within it).

## Checklist

1. Closes: n/a 

2. How was this tested:

- [ ] Verify there is a default cursor over the accordion row when it is `disabled`
- [ ] Verify the arrow icon is a lighter gray when the accordion is `disabled`
- [ ] Verify the accordion remains open when `disabled`

### DIFF
--- a/src/lib/holocene/accordion.svelte
+++ b/src/lib/holocene/accordion.svelte
@@ -1,11 +1,14 @@
 <script lang="ts">
+  import colors from 'tailwindcss/colors';
   import Icon from '$lib/holocene/icon/index.svelte';
   import type { IconName } from './icon/paths';
   export let title: string;
   export let subtitle: string = '';
   export let icon: IconName | undefined = undefined;
-
   export let open: boolean = false;
+  export let disabled: boolean = false;
+
+  $: open = disabled ? true : open;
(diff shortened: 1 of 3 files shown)


src/lib/holocene/accordion.svelte
```diff
@@ -24,7 +31,7 @@
         </h2>
         <Icon
           name={open ? 'caretUp' : 'caretDown'}
-          stroke="currentcolor"
+          stroke={disabled ? colors.gray[500] : 'currentcolor'}
```
rossedfort, line 34: Nice I like this, usually I just look up the hex value in our config and hard code it 😂

rossedfort (approved):
don't forget to bump the version in package.json if you want to publish 🚀

src/lib/holocene/accordion.svelte
laurakwhit: That way is probably more performant 😅

PR #745, opened 2022-08-03
Add locked state to Input

laurakwhit:
## What was changed

This PR adds a `locked` attribute to the holocene `Input` component. It also updates the styling for the `light` theme `copyable` input by merged styles into one `disabled` style.

| Designs | Existing | Final |
| ----- | ----- | ----- |
|Screen Shot 2022-08-03 at 10 41 24 AM|existing|suggested|

## Why?

We will need a `locked` option for inputs based on designs and both `copyable` and `locked` inputs are `disabled`.

## Checklist

1. Closes: n/a 

2. How was this tested:
- [ ] Verify the `copyable` input looks as expected for `light` and `dark` themes
- [ ] Verify the `locked` input looks as expected for `light` and `dark` themes

### DIFF
--- a/src/lib/holocene/input/input.svelte
+++ b/src/lib/holocene/input/input.svelte
@@ -16,10 +16,9 @@
   export let valid = true;
   export let hintText = '';
   export let maxLength = 0;
-  export let locked: boolean = false;
 
   const { copy, copied } = copyToClipboard(value);
-  $: disabled = disabled || copyable || locked;
+  $: disabled = disabled || copyable;
 </script>
 
 <div class={$$props.class}>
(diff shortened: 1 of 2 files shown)


src/lib/holocene/input/input.svelte
```diff
@@ -16,24 +16,26 @@
   export let valid = true;
   export let hintText = '';
   export let maxLength = 0;
+  export let locked: boolean = false;
```
laurakwhit, line 19: The designs show an `email` input which made me wonder, do we want a `type` attribute as well?

rossedfort: I don't know that we need an extra prop for this, I think we can use the existing `disabled` prop.

laurakwhit: Wasn't sure if we would have other possible `disabled` inputs, but if `locked` === `disabled` that makes sense. Updated ✅

rossedfort (approved):

PR #748, opened 2022-08-04
Cloud UI holocene updates

Alex-Tideman:
## What was changed
Add readOnly prop to Accordion to show static card with no caret. Reduce padding on MenuItem. Add pluralize utility

### DIFF
--- a/src/lib/holocene/accordion.svelte
+++ b/src/lib/holocene/accordion.svelte
@@ -7,6 +7,7 @@
   export let icon: IconName | undefined = undefined;
   export let open: boolean = false;
   export let disabled: boolean = false;
+  export let readOnly: boolean = false;
 
   $: open = disabled ? true : open;
 </script>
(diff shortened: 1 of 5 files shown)


src/lib/holocene/accordion.svelte
```diff
@@ -16,11 +17,11 @@
 >
   <div class="w-full">
     <div
-      class="accordion-open flex cursor-pointer flex-col"
+      class="accordion-open flex {!readOnly ? 'cursor-pointer' : ''} flex-col"
```
laurakwhit, line 20: Is this way of adding conditional classed preferred vs. doing something like `class:readOnly`?

src/lib/utilities/pluralize.ts
```diff
@@ -0,0 +1,6 @@
+export const pluralize = (word: string, count: number): string => {
```
laurakwhit, line 1: Are there any plans to add something like lodash in the future?

laurakwhit (approved):
Just a few questions r.e. my own curiosity, but lgtm 🚀

laurakwhit:
Also, would be great to add this to `Fiction` if you feel so inclined 🙇‍♀️ 

```ui
<Chapter description="A readOnly Accordion">
  <Accordion
    readOnly
    open
    title="Certificates"
    subtitle="Expires on Wed Feb 01, 2030"
  >
    <p>Accordion content here.</p>
  </Accordion>
</Chapter>
```

PR #749, opened 2022-08-04
Move link to holocene

laurakwhit:
## What was changed

This PR moves the `Link` component to holocene and updates the padding on the `Textarea` component.

## Why?

We will need access to `Link` in `cloud-ui` and the `Textarea` now more accurately reflects the designs.

## Checklist

1. Closes: n/a 

2. How was this tested: n/a


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@temporalio/ui",
-  "version": "2.1.13",
+  "version": "2.1.14",
```
Alex-Tideman, line 3: I just bumped to 2.1.14, let's make this 2.1.15

Alex-Tideman: Otherwise looks good!

laurakwhit: Is [this commit](https://github.com/temporalio/ui/pull/749/commits/78b170ca950e59f68c4ad341609c4ec2c5b0f477) autogenerated? I definitely don't remember adding it 😅

Alex-Tideman: Doesn't get autogenerated but main might have had that version already but it wasn't published yet

Alex-Tideman: The general rule of thumb we try to follow is if a PR has updates for the npm package, bump the version number, get a review, and if approved merge and then publish

Alex-Tideman (approved):

package.json
laurakwhit: Got it, thanks! I think will actually leave this for now and bump it in my next PR that will be up shortly.

PR #750, opened 2022-08-04
Add book icon

laurakwhit:
## What was changed

Adds a book icon to holocene.

Screen Shot 2022-08-04 at 2 44 35 PM

## Why?

For use in `cloud-ui` designs.

## Checklist

1. Closes: n/a 

2. How was this tested: n/a


src/lib/holocene/icon/paths.ts
```diff
@@ -22,6 +22,49 @@ const arrowUpIcon: Icon = {
   ],
 };
 
+const bookIcon: Icon = {
+  paths: [
+    {
+      d: 'M12 18.75V8.25C12 7.42157 12.6716 6.75 13.5 6.75H19.05C19.2985 6.75 19.5 6.95147 19.5 7.2V17.0357',
+      stroke: '#111827',
+      'stroke-width': '1.125',
+      'stroke-linecap': 'round',
```
rossedfort, line 31: `stroke-linecap: round` is already set on the `<path />` in the component itself, so we don't need to specify it here.

src/lib/holocene/icon/index.svelte
```diff
@@ -58,7 +68,7 @@
             {...path}
             stroke-linecap="round"
             stroke-linejoin="round"
-            stroke-width={strokeWidth}
```
laurakwhit, line 61: Added this as I noticed the workflow icon (and info icon) also had `stroke-width` defined in `paths.ts` 👇 

|Before | After|
| --  |--|
|Screen Shot 2022-08-04 at 3 22 18 PM|Screen Shot 2022-08-04 at 3 23 32 PM|

rossedfort (approved):

PR #758, opened 2022-08-08
Locked select menu

laurakwhit:
## What was changed

Holocene `Select` component now has a `disabled` state with a `lock` icon. 

| Designs | Disabled | 
| ---|---|
|Screen Shot 2022-08-08 at 1 23 40 PM|Screen Shot 2022-08-08 at 5 25 39 PM|

**Note:** This also reverts changes made in https://github.com/temporalio/ui/pull/754 to the `Dropdown` component, as we actually want the `locked` state on the `Select` and not the `Dropdown` component 🤦‍♀️ 
## Why?

To allow for accurate implementation of designs.
## Checklist

1. Closes: `Create “Locked” Select Menu`

2. How was this tested:

- [ ] Verify there is a lock icon and the button is `disabled` when `disabled` is set
- [ ] Verify a `readOnly` dropdown menu does **not** have a gray background

### DIFF
--- a/src/lib/holocene/select/select.svelte
+++ b/src/lib/holocene/select/select.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
   import { createEventDispatcher } from 'svelte';
+  import colors from 'tailwindcss/colors';
   import Icon from '$holocene/icon/index.svelte';
   import Menu from '$holocene/primitives/menu/menu.svelte';
   import Option, { isOption } from '$holocene/select/option.svelte';
@@ -16,6 +17,7 @@
   export let options: T[] | undefined = undefined;
   export let dark: boolean = false;
   export let placeholder = '';
+  export let disabled: boolean = false;
 
   let _value: OptionType['value'] | T;
   let _selected: string | T;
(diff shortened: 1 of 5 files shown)


src/lib/holocene/select/select.svelte
```diff
@@ -46,22 +48,28 @@
   {/if}
   <MenuContainer class="w-full">
     <MenuButton
-      class="select-input-container"
+      class="select-input-container {disabled ? 'disabled' : ''}"
```
Alex-Tideman, line 51: Best to do class:disabled when it's a boolean like this of either adding a class. Usually I prefer the inline when it's between two different class {disabled ? 'disabled' : 'enabled'}

laurakwhit: I did it this way so that the `disabled` class is only being passed down to the `MenuButton` for the `Select` component. 
The `DropdownButton` also uses the `MenuButton` but the `disabled` state (set for the `readOnly` option) does not have the `disabled` styling. Adding `class:disabled` results in 👇
| Before | After |
|---| ---|
|Screen Shot 2022-08-09 at 9 24 22 AM|Screen Shot 2022-08-09 at 9 25 51 AM|
 
Maybe we could consider passing a `readOnly` prop to the `MenuButton`?

stevekinney: Given the trade offs, that sounds reasonable.

Alex-Tideman (approved):

PR #761, opened 2022-08-10
WIP: Native select

Alex-Tideman:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/lib/holocene/select/native-select.svelte
+++ b/src/lib/holocene/select/native-select.svelte
@@ -1,15 +1,4 @@
 <script lang="ts">
-  import { createEventDispatcher } from 'svelte';
-  import colors from 'tailwindcss/colors';
-  import Icon from '$holocene/icon/index.svelte';
-  import Menu from '$holocene/primitives/menu/menu.svelte';
-  import Option, { isOption } from '$holocene/select/option.svelte';
-  import type { Option as OptionType } from '$holocene/select/option.svelte';
-  import MenuButton from '$holocene/primitives/menu/menu-button.svelte';
-  import MenuContainer from '$holocene/primitives/menu/menu-container.svelte';
-
-  let show = false;
-
   export let label = '';
   export let id: string;
   export let placeholder: string;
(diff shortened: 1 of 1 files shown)


src/lib/holocene/select/native-select.svelte
```diff
@@ -0,0 +1,16 @@
...
+  export let id: string;
+  export let placeholder: string;
+  export let value = '';
+  export let options: { label: string; value: string }[] = [];
+</script>
+
+{#if label}
+  <label for={id} class={$$props.labelClass}>{label}</label>
+{/if}
+<select {value} {id} class={$$props.class} {placeholder} on:select>
```
GiantRobots, line 12: Probably bind the value so we can just do `<NativeSelect bind:value={} />` so we get reactivity on the value up and down.

src/lib/holocene/select/native-select.svelte
```diff
@@ -0,0 +1,16 @@
...
+  {#each options as option}
+    <option value={option.value}>{label}</option>
```
GiantRobots, line 14: I think I would prefer this to just be a default slot and compose the options myself. That way we could expose more bind values and generally have more control over the sub components.

src/lib/holocene/select/native-select.svelte
GiantRobots: In the future maybe some default styling that works well enough to make it look like the non native counterpart for at least the initial view.

src/lib/holocene/select/native-select.svelte
```diff
@@ -0,0 +1,16 @@
+<script lang="ts">
+  export let label = '';
+  export let id: string;
+  export let placeholder: string;
+  export let value = '';
+  export let options: { label: string; value: string }[] = [];
+</script>
+
+{#if label}
+  <label for={id} class={$$props.labelClass}>{label}</label>
```
GiantRobots, line 10: If the label is wrapped around the select you can skip the ID because it's associated and majestical

PR #769, opened 2022-08-15
Autocomplete and select updates

rossedfort:
## What was changed

- change autocomplete to use menu primitives
- make Option interface generic
- support wider Options
- v2.1.23

Force an option's `label` to never break and apply `min-width: fit-content` so that the menu doesn't shrink smaller than the label.
Screen Shot 2022-08-15 at 12 21 16 PM

Should solve this in cloud-ui
Screen Shot 2022-08-15 at 12 23 34 PM

## Why?

Better dev experience for Selects and visual consistency for components with menus.
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@temporalio/ui",
-  "version": "2.1.22",
+  "version": "2.1.23",
```
Alex-Tideman, line 3: Will need to be .24, been publishing a lot today

Alex-Tideman (approved):

PR #790, opened 2022-08-25
Fix wrapping of workflow statuses

stevekinney:
### Before

Screen Shot 2022-08-25 at 9 36 41 AM

### After

Screen Shot 2022-08-25 at 9 39 16 AM


rossedfort (approved):
💯

PR #794, opened 2022-08-29
add select option group, refactor select component

rossedfort:
## What was changed

Add a select-option component
Screen Shot 2022-08-29 at 1 12 08 PM

Also refactors our `<Select />` component API to be more similar with a native HTML `<select />`.
changes include:
- Support binding to a value
- Remove dependency on a `on:select` change handler
- Removes `options` prop
- Adds an optional `onChange` handler
- Adds an optional `displayValue` prop that allows a non-string value to be mapped to a string.

before:
```jsx
  <Select
    label="Favorite Drink"
    placeholder="Favorite Drink"
    id="favorite-drink"
    options={drinkOptions}
    value={$favoriteDrink}
    on:select={(event) => favoriteDrink.set(event.detail.value)}
  />
```
after:
```jsx
<Select
  label="Favorite Drink"
  placeholder="Favorite Drink"
  id="favorite-drink"
  bind:value={favoriteDrink}
  onChange={optionallyDoSomeSideEffect}
>
  <Option value="coffee">Coffee</Option>
  <Option value="tea">Tea</Option>
  <Option value="milk">Milk</Option>
</Select>
```
## Why?

Nicer devx for Select menus.
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@temporalio/ui",
-  "version": "2.1.37",
+  "version": "2.1.38",
   "type": "module",
   "description": "Temporal.io UI",
   "keywords": [
--- a/src/lib/holocene/primitives/menu/menu-item.svelte
+++ b/src/lib/holocene/primitives/menu/menu-item.svelte
@@ -44,21 +44,19 @@
   .dark {
     @apply bg-primary text-white hover:bg-gray-800;
 
-    &.selected {
-      @apply bg-gray-800;
+    &.selected,
+    &.active {
+      @apply bg-gray-800 text-blue-100;
     }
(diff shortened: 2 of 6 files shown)


Alex-Tideman:
In our other selects, the designs always had a blue checkmark. Maybe ask Candace if that should be the default instead of grey?

Screen Shot 2022-08-29 at 12 22 11 PM

Alex-Tideman:
I think something's up when the default value is undefined.

Screen Shot 2022-08-29 at 2 26 12 PM

Alex-Tideman (approved):

PR #806, opened 2022-09-06
fix icons

rossedfort:
## What was changed

- create new simpler Icon component
- rename icon/index.svelte to icon/icon.svelte for better intellisense
- create a svelte component for every icon
- use new 24x24 px icons
- add a icon chapter

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


Alex-Tideman:
One thing I noticed is the filter icon doesn't fill when clicked (in the workflow list, click the filter icon on a row).

Everything else is looking good, might just want one pass with Candace to check on sizes (Namespace switch/nav looks a little small)

rossedfort:
> One thing I noticed is the filter icon doesn't fill when clicked (in the workflow list, click the filter icon on a row).

Ah great catch, I noticed this on Friday but forgot over the long weekend lol. Thanks.

Alex-Tideman (approved):

PR #815, opened 2022-09-08
Add action locations to Pagination 

laurakwhit:
## What was changed

Adds two additional locations for the `action` slot in the Holocene `Pagination` component, as well as updates existing location name.

| `action` is now `action-top-right` |   `action-bottom-right` |  `action-bottom-left` |
| --- | --- | --- |
|Screen Shot 2022-09-08 at 1 14 04 PM|Screen Shot 2022-09-08 at 1 13 33 PM|Screen Shot 2022-09-08 at 1 13 15 PM|

## Why?

For use in designs.

## Checklist

1. Closes: n/a 

2. How was this tested: n/a

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@temporalio/ui",
-  "version": "2.1.46",
+  "version": "2.1.47",
   "type": "module",
   "description": "Temporal.io UI",
   "keywords": [
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@temporalio/ui",
-  "version": "2.1.47",
+  "version": "2.1.48",
   "type": "module",
   "description": "Temporal.io UI",
   "keywords": [
--- a/src/lib/holocene/pagination.svelte
+++ b/src/lib/holocene/pagination.svelte
@@ -43,13 +43,16 @@
 <svelte:window bind:innerWidth={screenWidth} />
 
 <div class="pagination relative mb-8 flex flex-col gap-4">
-  <div class="flex justify-between">
-    <div class="flex items-center">
-      {#if updating}
-        <p class="mr-6 text-gray-600">Updating…</p>
-      {/if}
+  <div
+    class={`flex items-center ${
+      updating || $$slots['action-top-left'] ? 'justify-between' : 'justify-end'
+    }`}
+  >
+    {#if updating}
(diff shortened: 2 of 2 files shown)


Alex-Tideman:
Let' add a top-left action slot too

src/lib/holocene/pagination.svelte
```diff
@@ -82,38 +82,46 @@
           <Icon name="chevron-right" />
         </button>
       </div>
-      <slot name="action" />
+      <slot name="action-top-right" />
```
laurakwhit, line 85: > Let's add a top-left action slot too

@Alex-Tideman I didn't add one there as we currently have `Updating…` text in that area. I could just put the action to the left or right of it though 🤔 Any preference?

Screen Shot 2022-09-08 at 3 54 36 PM

laurakwhit: **Update:** I've adding it to the right of the text 👇 

| Updating | Not Updating |
| ---- | ---- |
| Screen Shot 2022-09-08 at 4 11 19 PM| Screen Shot 2022-09-08 at 4 09 20 PM|

Alex-Tideman: I think what we could do is:

      {#if updating}
        <p class="mr-6 text-gray-600">Updating…</p>
      {:else}
      <slot name="action-top-left" />
      {/if} 

See how that looks and if it's too much of a flash of content.

I think eventual end state of the updating state would be to use the a <Skeleton> in each row

laurakwhit: Done ✅ It's not too much of a flash, but I definitely don't love hiding and then showing the button. Might think about refactoring to `<Skeleton>` in another PR 🤔

Alex-Tideman: Let's show this to Candace in our meeting, see what she thinks.

laurakwhit: **Update:** We can stick with this solution for now.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@temporalio/ui",
-  "version": "2.1.46",
+  "version": "2.1.47",
```
Alex-Tideman, line 3: 2.1.48 (I snuck a published in)

Alex-Tideman (approved):

PR #825, opened 2022-09-12
Fix is unauthorized

Alex-Tideman:
## What was changed
Check only for statusCode on response for isUnauthorized and isForbidden.

## Why?
The isNetworkError was returning false because some response bodys are undefined:

```
    networkErr?.statusCode !== undefined &&
    networkErr?.statusText !== undefined &&
    networkErr?.response !== undefined

```

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/lib/utilities/handle-error.ts
+++ b/src/lib/utilities/handle-error.ts
@@ -52,10 +52,10 @@ export const handleUnauthorizedOrForbiddenError = (
   }
 };
 
-const isUnauthorized = (error: any): error is NetworkError => {
+const isUnauthorized = (error: any): boolean => {
   return error?.statusCode === 401;
 };
 
-const isForbidden = (error: any): error is NetworkError => {
+const isForbidden = (error: any): boolean => {
   return error?.statusCode === 403;
 };
--- a/src/lib/utilities/handle-error.ts
+++ b/src/lib/utilities/handle-error.ts
@@ -52,18 +52,10 @@ export const handleUnauthorizedOrForbiddenError = (
   }
 };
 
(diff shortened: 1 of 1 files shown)


stevekinney (approved):
What was the logic for `isNetworkError` and is it being used anywhere else?

We use it on a few __error layouts to show statusCode/statusText. I think we could probably keep it if we remove the response check:

export function isNetworkError(
  error: unknown | NetworkError,
): error is NetworkError {
  const networkErr = error as NetworkError;
  return (
    networkErr?.statusCode !== undefined &&
    networkErr?.statusText !== undefined &&
    networkErr?.response !== undefined
  );
}

Alex-Tideman:
We use it on a few __error layouts to show statusCode/statusText. I think we could probably keep it if we remove the response check:

```
export function isNetworkError(
  error: unknown | NetworkError,
): error is NetworkError {
  const networkErr = error as NetworkError;
  return (
    networkErr?.statusCode !== undefined &&
    networkErr?.statusText !== undefined &&
    networkErr?.response !== undefined
  );
}

```

Alex-Tideman:
@GiantRobots You remember if there was a specific reason for that response check?

feedmeapples:
> @GiantRobots You remember if there was a specific reason for that response check?

i think it partially comes from the APIs returning always 403 instead of 401?

according to another thread this is what the APIs should do 
image

GiantRobots (approved):

GiantRobots:
The response check in there was because that's how most errors looked when I was adding error boundaries and the like into everything. Other errors need to be surfaced in different ways 404's show on the screen, 500's screen etc some are toast's.

PR #826, opened 2022-09-12
Handle errors that have status prop instead of statusCode

feedmeapples:
## What was changed

Added `error.status` handling in addition to `error.statusCode`

## Why?

backends may fill out `status` instead of `statusCode`

## Checklist

1. Closes 

2. How was this tested:

unit tests +
manually throwing 401 in ui-server

3. Any docs updates needed?


Alex-Tideman (approved):

GiantRobots (approved):
We should probably normalize these somehow

PR #828, opened 2022-09-12
Show auth error message on Login screen

feedmeapples:
## What was changed

Shows error message on Login

image

## Why?

help users figure out what went wrong

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/lib/holocene/button.svelte
+++ b/src/lib/holocene/button.svelte
@@ -44,7 +44,7 @@
     <slot />
     {#if count > 0}
       <Badge
-        class="badge absolute top-0 right-0 origin-bottom-left origin-bottom-left translate-y-[-10px] translate-x-[10px]"
+        class="badge absolute top-0 right-0 origin-bottom-left translate-y-[-10px] translate-x-[10px]"
         type="count">{count}</Badge
       >
     {/if}
--- a/src/lib/utilities/handle-error.ts
+++ b/src/lib/utilities/handle-error.ts
@@ -7,18 +7,18 @@ import { routeForLoginPage } from './route-for';
 
 // This will eventually be expanded on.
 export const handleError = (
-  error: unknown,
+  error: any,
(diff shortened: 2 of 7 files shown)


src/lib/holocene/button.svelte
```diff
@@ -44,7 +44,7 @@
     <slot />
     {#if count > 0}
       <Badge
-        class="badge absolute top-0 right-0 origin-bottom-left origin-bottom-left translate-y-[-10px] translate-x-[10px]"
+        class="badge absolute top-0 right-0 origin-bottom-left translate-y-[-10px] translate-x-[10px]"
```
stevekinney, line 47: I don't love that we're hardcoding values here, but you didn't do that. My. bigger question is do we know _why_ we were using `origin-bottom-left` and are we sure it's not going to break something else?

src/lib/utilities/handle-error.ts
```diff
@@ -12,12 +13,12 @@ export const handleError = (
   isBrowser = browser,
 ): void => {
   if (isUnauthorized(error) && isBrowser) {
-    window.location.assign(routeForLoginPage());
+    window.location.assign(routeForLoginPage(error?.['message']));
```
stevekinney, line 16: Why not `error?.message`?

stevekinney (approved):

src/lib/holocene/button.svelte
Alex-Tideman: `origin-bottom-left` was a duplicate class in this case, I think the formatter caught it. It's the positioning of the count badge if you pass a count prop to a button.

src/lib/utilities/handle-error.ts
feedmeapples: changed the error type back to `any` so it allows to use `error?.message`

PR #830, opened 2022-09-13
Number input

Alex-Tideman:
## What was changed
Add a new NumberInput component with units and max props.

Screen Shot 2022-09-13 at 3 30 38 PM

## Why?
For numbers


rossedfort:
Looks good, although I'm wondering if we should just extend our existing `<Input />` to take a `type="number"` prop. What do you think?

Alex-Tideman:
So I thought about that too but Candace's designs keep it as a normal input (allows string). Generally input type="number" sucks and it's better to use a normal input.

rossedfort (approved):

PR #831, opened 2022-09-14
Store user session in browser

feedmeapples:
## What was changed

Moved user session to be stored client side

## Why?

- Allows to scale UI horizontally by making ui-server stateless

## Checklist

1. Closes 

2. How was this tested:

E2E: Verified that a user can login
Manual: Verified login and logout flows. Verified that API requests contain `authorization` and `authorization-extras` headers. Verified that codec requests contain `authorization` header 

3. Any docs updates needed?


rossedfort (approved):

src/lib/stores/events.ts
```diff
@@ -50,6 +50,8 @@ const runId = derived([page], ([$page]) => {
 
 const settings = derived([page], ([$page]) => $page.stuff.settings);
 
+const accessToken = derived([page], ([$page]) => $page.stuff.user?.accessToken);
```
GiantRobots, line 53: Since we're already using stuff everywhere it's cool to let this through but stuff is being deprecated and we probably want to find the better paradigm for it later later. (No work needed just opining)

src/lib/utilities/request-from-api.ts
```diff
@@ -136,6 +130,35 @@ const withSecurityOptions = (
...
+  options: RequestInit,
+  isBrowser = browser,
+): Promise<RequestInit> => {
+  if (getUser().accessToken) {
+    options.headers = await withBearerToken(
+      options?.headers,
+      () =>
+        new Promise((resolve) => {
+          resolve(getUser().accessToken);
+        }),
```
GiantRobots, line 143: I think you can just write that like this

```javascript
async () => getUser().accessToken
```

src/lib/utilities/request-from-api.ts
```diff
@@ -159,6 +182,27 @@ const withBearerToken = async (
...
+): HeadersInit => {
+  // At this point in the code path, headers will always be set.
+  /* c8 ignore next */
+  if (!headers) headers = {};
+  if (!isBrowser) return headers;
+
+  try {
+    if (idToken) {
+      headers['Authorization-Extras'] = idToken;
+    }
```
GiantRobots, line 198: Why does this need to be in a try catch?

src/lib/utilities/request-from-api.ts
```diff
@@ -159,6 +182,27 @@ const withBearerToken = async (
...
 
+const withIdToken = (
+  headers: HeadersInit,
+  idToken: string,
+  isBrowser = browser,
+): HeadersInit => {
+  // At this point in the code path, headers will always be set.
+  /* c8 ignore next */
+  if (!headers) headers = {};
+  if (!isBrowser) return headers;
```
GiantRobots, line 193: If this had `strict: true` in tsconfig this code could be made more obviously safe.


We can also default headers in the call with the `= {}` to skip the if (headers)  
```javascript
const withIdToken = (
	headers: HeadersInit = {},
	idToken: string,
	isBrowser = browser,
  )
```

GiantRobots (approved):

src/lib/stores/events.ts
feedmeapples: oh i would like to not use stuff and at least store instead, can change that in the PR

src/lib/utilities/request-from-api.ts
feedmeapples: oops

src/lib/services/workflow-service.ts
```diff
@@ -48,18 +48,13 @@ export const fetchAllWorkflows = async (
       `Error fetching workflows: ${err.status}: ${err.statusText}`;
   };
 
-  const handleError = () => {
```
Alex-Tideman, line 51: Is this case handled by ui-server now?

feedmeapples: when i tried setting a bad namespace and it would get processed by onError and not handleError

feedmeapples: also would it make sense to pass just a single error handler and process various cases in it instead of passing two error handlers. I think this would help understand the code easier vs trying to figure out the difference between `onError` and `handleError`

feedmeapples: image

feedmeapples: discussed directly, this change should be good

Then we also want to refactor some of error handling to completely remove the second error handler param

PR #836, opened 2022-09-15
Remove %sveltekit.assets%

Alex-Tideman:
## What was changed
Always get static assets from '/'

## Why?
Fix assets in build


feedmeapples (approved):

feedmeapples:
approving though this will break the Public Path feature.

To properly fix the issue we will want to upgrade to the latest sveltekit

PR #839, opened 2022-09-16
Add scroll to top button

laurakwhit:
## What was changed

This adds a button that appears on scroll and takes the user back to the top of the page or the top of a specified container.

| Default | Hover |
| ----- | ------ |
|Screen Shot 2022-09-16 at 4 05 18 PM|Screen Shot 2022-09-16 at 4 05 29 PM|

## Why?

For use in updated designs.

## Checklist

1. Closes: **Back to Top / Back to Bottom**  

2. How was this tested:

- [ ] Verify button shows up on scroll
- [ ] Verify button can take user to top of page
- [ ] Verify button can take user to top of a container
- [ ] Verify button can show up at a specified number of pixels

### DIFF
--- a/src/lib/holocene/scroll-to-top.svelte
+++ b/src/lib/holocene/scroll-to-top.svelte
@@ -29,7 +29,12 @@
 <div id="scroll-container" class={$$props.class}>
   <slot />
   <div class="back-to-top" class:hidden>
-    <Button icon="arrow-up" variant="secondary" on:click={scrollToTop} />
+    <Button
+      class="!py-1.5 !px-1"
+      icon="arrow-up"
+      variant="secondary"
+      on:click={scrollToTop}
+    />
   </div>
 </div>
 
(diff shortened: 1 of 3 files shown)


src/lib/holocene/scroll-to-top.svelte
```diff
@@ -0,0 +1,49 @@
...
+  }
+</script>
+
+<svelte:window on:scroll={handleOnScroll} />
+
+<div id="scroll-container" class={$$props.class}>
+  <slot />
+  <div class="back-to-top" class:hidden>
+    <Button
+      class="!py-1.5 !px-1"
```
Alex-Tideman, line 33: What does the ! in front of the class do? I've never seen that with Tailwind

laurakwhit: I'm using the [important modifier ](https://tailwindcss.com/docs/configuration#important-modifier) here because there is some variant styling (specifically `padding`) applied as `postcss` on the `button` component which overrides the `classes` being passed down to the component.

The other solution here would be to create a `icon` variant and add specific padding for it in the `postcss` on the `button` component, but I'm not so sure we want the same padding for all icons. Hence this hack 😅

Alex-Tideman (approved):

PR #840, opened 2022-09-19
CoreUser context

Alex-Tideman:
## What was changed
Define a CoreUser model that allows use to check permissions for user actions based on their authorization.

Add a check on the terminate button to see if the current user can terminate. If not, show a Tooltip and disable button.
Screen Shot 2022-09-19 at 9 47 11 AM

## Why?
Start framework for flexible authorization model on user actions.


rossedfort (approved):

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@temporalio/ui",
-  "version": "2.1.53",
+  "version": "2.1.54",
```
laurakwhit, line 3: [Looks](https://www.npmjs.com/package/@temporalio/ui) like we might want `2.1.55` here.

PR #841, opened 2022-09-19
Add scroll to top button

laurakwhit:
## What was changed

This adds a button that appears on scroll for **all** pages to take the user back to the top of the page.

## Why?

It was requested.

## Checklist

1. Closes: Back to Top / Back to Bottom 

2. How was this tested:

- [ ] Verify button shows up on scroll down
- [ ] Verify button does not show up when user is at the top of the page
- [ ] Verify, when clicked, page is taken back to the top
- [ ] Verify button shows up on all pages that can scroll
- [ ] Verify layout for all pages looks as expected

### DIFF
--- a/src/routes/__layout-root.svelte
+++ b/src/routes/__layout-root.svelte
@@ -63,6 +63,7 @@
   import Banners from '$lib/components/banner/banners.svelte';
   import { ErrorBoundary } from '$lib/components/error-boundary';
   import PageTitle from '$lib/holocene/page-title.svelte';
+  import ScrollToTop from '$lib/holocene/scroll-to-top.svelte';
 
   export let user: User;
   export let uiVersionInfo: UiVersionInfo;
@@ -81,5 +82,6 @@
         <slot />
       </ErrorBoundary>
     </div>
+    <ScrollToTop />
   </section>
 </div>
--- a/src/routes/__layout-root.svelte
+++ b/src/routes/__layout-root.svelte
@@ -69,14 +69,14 @@
 </script>
 
 <PageTitle />
-<div class="flex h-screen w-screen flex-row">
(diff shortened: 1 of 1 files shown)


src/routes/__layout-root.svelte
```diff
@@ -63,23 +63,25 @@
   import Banners from '$lib/components/banner/banners.svelte';
   import { ErrorBoundary } from '$lib/components/error-boundary';
   import PageTitle from '$lib/holocene/page-title.svelte';
+  import ScrollToTop from '$lib/holocene/scroll-to-top.svelte';
```
laurakwhit, line 66: Wondering if we should add this and make updates to `__layout-import.svelte` as well 🤔

src/routes/__layout-root.svelte
```diff
@@ -63,23 +63,25 @@
...
 
   export let user: User;
   export let uiVersionInfo: UiVersionInfo;
 </script>
 
 <PageTitle />
-<div class="flex h-screen w-screen flex-row">
+<div class="flex w-screen flex-row">
```
laurakwhit, line 73: The `h-screen` was messing with the `on:scroll` here, hence it's removal.

Alex-Tideman: I'm cool with this, but make sure to link with cloud ui and test it there to see if there's any wonkiness importing it in. That repo also uses several h-screen classes

laurakwhit: Does `cloud-ui` use this layout? I've also made [updates](https://github.com/temporalio/cloud-ui/pull/247) directly to the layout in `cloud-ui`.

Alex-Tideman (approved):

PR #842, opened 2022-09-20
Add current page to Pagination params

laurakwhit:
## What was changed

This adds the current `page` to query parameters for the `Pagination` component.

## Why?

The page url now has a page number for easy linking and quick access.

## Checklist

1. Closes: **Add current page to Pagination query parameters** 

2. How was this tested:

- [ ] Verify selecting `next` updates the page number in the search params correctly
- [ ] Verify selecting `previous` updates the page number in the search params correctly
- [ ] Verify change the page number in the search params takes the user to the correct page
- [ ] Verify defining a `startingIndex` for a `Pagination` component takes the user to the correct page and the page is listed in the search params

### DIFF
--- a/src/lib/holocene/pagination.svelte
+++ b/src/lib/holocene/pagination.svelte
@@ -6,6 +6,7 @@
     perPageFromSearchParameter,
     perPageOptions,
   } from '$lib/stores/pagination';
+  import { updateQueryParameters } from '$lib/utilities/update-query-parameters';
 
   import Icon from '$holocene/icon/icon.svelte';
 
@@ -36,6 +37,14 @@
       width = document.getElementById(floatId)?.clientWidth;
     }
   });
+
+  const handlePageChange = () => {
+    updateQueryParameters({
+      parameter: 'page',
+      value: $store.currentPage,
+      url: $page.url,
+    });
+  };
 
   $: floatStyle = getFloatStyle({ width, height, screenWidth });
 </script>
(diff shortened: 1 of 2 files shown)


src/lib/holocene/pagination.svelte
```diff
@@ -37,6 +38,14 @@
     }
   });
 
+  const handlePageChange = () => {
```
laurakwhit, line 41: Wondering if we want to also call `handlePageChange` `onMount` or not? It doesn't seem necessary if the first page is always going to be `1`.  However, if a `startingIndex` has been defined for the `Pagination` component the component loads to the corresponding page (which can be something other than `1`) and there is no `page` defined in the search params.

Alex-Tideman: What's the difference between the startingIndex and the page? Just trying to understand what we would using the startingIndex for

Alex-Tideman: Oh I see it's basically the same thing, it's just that you set the starting page with a prop. Yeah I think what you could do is an onMount and if startingIndex is not 0, update the currentPage and call handlePageChange.

laurakwhit: Updated ✅

Alex-Tideman (approved):

src/lib/holocene/pagination.svelte
Alex-Tideman: 👍

PR #847, opened 2022-09-20
NamespaceSwitcher update, small UI style updates for quality of life

Alex-Tideman:
## What was changed

- Make NamespaceSwitcher button look like a button
- Change nav expand/collapse to chevron to not confuse icon with logout
- Faster expand/collapse of nav
- Make DataConverter arrow colors darker to see on cloud
- Add hover state on modal

Screen Shot 2022-09-20 at 1 14 25 PM
 close icon

## Why?
Based on feedback


src/lib/holocene/modal.svelte
```diff
@@ -22,7 +22,10 @@
         class="float-right cursor-pointer p-6"
```
laurakwhit, line 22: Unrelated, but it might be nice to set `cursor-default` on the root `div` so it's not a pointer for the entire page/modal.

src/lib/holocene/navigation/_nav-container.svelte
```diff
@@ -32,7 +32,7 @@
       style="top: 52px;"
       on:click={toggleNav}
     >
-      <Icon name={$navOpen ? 'nav-collapse' : 'nav-expand'} />
+      <Icon name={$navOpen ? 'chevron-left' : 'chevron-right'} />
```
laurakwhit, line 35: Wondering if we want to remove `nav-collapse` and `nav-expand` icons to prevent any future confusion.

laurakwhit (approved):

src/lib/holocene/navigation/_nav-container.svelte
Alex-Tideman: Great idea, done

src/lib/holocene/modal.svelte
Alex-Tideman: Good call, added cursor-default on the top level div

src/lib/holocene/namespace-selector.svelte
```diff
@@ -2,19 +2,25 @@
...
-  data-cy="namespace-select-button"
->
-  <Icon name="namespace-switcher" class="mx-2 inline cursor-pointer" />
-  <FeatureTag feature="namespaceSelector" />
-</button>
+<Tooltip text="Select a namespace">
+  <button
+    class="relative"
+    on:click={toggleNamespaceSelector}
+    data-cy="namespace-select-button"
```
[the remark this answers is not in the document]
Alex-Tideman: Great call

PR #850, opened 2022-09-23
Update Textarea events

laurakwhit:
## What was changed

Added
*  `on:input`
*  `on:change`
*  `on:focus`
*  `on:blur`

to the `Textarea` component.

## Why?

So we can utalize more than just `onBlur`.

## Checklist

1. ClosesL n/a 

2. How was this tested: n/a

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@temporalio/ui",
-  "version": "2.1.58",
+  "version": "2.1.59",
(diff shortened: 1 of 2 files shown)


Alex-Tideman (approved):

Alex-Tideman:
We'll just want to make sure we replace the onBlur in cloud-ui when we bump the version

Issue #845, opened 2022-09-20
Add links to children inside StartChild events


lorensr:
**Is your feature request related to a problem? Please describe.**

User feedback, transitioning from v1 to v2:

> It seems we don't include links to child workflows in the workflow pill when checking parent workflow history. Is it intentional?

> I have seen the list of pending child workflow at the top of a running workflow but navigating a completed workflow is harder than in V1 because we lack those child workflow links.

![image](https://user-images.githubusercontent.com/251288/191322255-b3bf7b54-70db-4e08-8891-cef5f8f5cb4e.png)

> Actually if we could to step into a child workflow without expanding history events and clicking that Workflow pill would save us many clicks :smile:

**Describe the solution you'd like**

Add links to children inside non-expanded StartChild events

**Additional context**

Continuing from #672

PR #851, opened 2022-09-26
Link to child workflows from parent workflow histories

lminaudier:
Attempt to partially solve #845.

## What was changed

The PR wraps child workflow IDs and Run IDs displayed in workflow histories with links to the corresponding workflow pages. 

## Why?

This is very useful when debugging workflows in the UI as you don't have to go back to the workflow list view and filter by workflow ID. 

Ideally, we could jump to a child workflow in one click as highlighted in #845 instead of having to click on the `Workflow` pill, but his might need more design decision on how to render that out, so I haven't tried to implement it. 

## Checklist

1. This only partially implements #845 

2. How was this tested:

Only locally, running the UI with a demo workflow

image

image

3. Any docs updates needed?

No

--- 

My frontend, typescript and svelte knowledge is very limited and I am not really satisfied with how I had to pass down the entire event attributes to the `EventDetailsRowExpanded` component to be able to get both the workflow and run IDs. I am not sure there is a better way given the current implementation which is a simple for loop to "display that attributes list" without full context around the attributes.

### DIFF
--- a/src/lib/components/event/event-details-full.svelte
+++ b/src/lib/components/event/event-details-full.svelte
@@ -41,7 +41,7 @@
       />
       {#each Object.entries(attributes) as [key, value] (key)}
         {#if attributeGrouping[activePill]?.includes(key)}
-          <EventDetailsRowExpanded {key} {value} class="w-full" />
+          <EventDetailsRowExpanded {key} {value} {attributes} class="w-full" />
         {/if}
       {/each}
     </div>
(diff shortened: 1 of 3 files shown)


src/lib/components/event/event-details-full.svelte
```diff
@@ -41,7 +41,7 @@
       />
       {#each Object.entries(attributes) as [key, value] (key)}
         {#if attributeGrouping[activePill]?.includes(key)}
-          <EventDetailsRowExpanded {key} {value} class="w-full" />
+          <EventDetailsRowExpanded {key} {value} {attributes} class="w-full" />
```
lminaudier, line 44: How can we improve that to only forward what we need (i.e the workflow id + run id only) ?

Alex-Tideman: Thanks @lminaudier! There's no good way to do this currently, we are planning on breaking apart the rows to be more specific to the type of data it shows. But that will be a much bigger refactor and we'd like to get this improvement out. I'll push up a PR that has these changes plus showing the link in the collapses view.

PR #852, opened 2022-09-26
Show parent link in child workflow event history, add Loading spinner to event history

### LINKED ISSUE #845: Add links to children inside StartChild events

Alex-Tideman:
## What was changed
For child workflows, add a link to the workflowId and workflowRunId of the parent workflow in the event history. This also adds the workflowId as the preferred key to show on the collapsed row with a link. The links will open in new tab. The labels for all events are changed to p tags so the font is consistent for all data.

Included in this is also adding the Loading component to the event history when the event history is being loaded into the table. 

BeforeLink1
BeforeLink2
AfterLink1
AfterLink2

## Why?
Easier to jump to parent workflow, better UX.

1. Closes #845 

2. How was this tested:
All tests are passing, manually tested feed/compact of multiple workflows

## Contributors
@lminaudier

### DIFF
--- a/src/lib/components/event/event-details-full.svelte
+++ b/src/lib/components/event/event-details-full.svelte
@@ -41,7 +41,7 @@
       />
       {#each Object.entries(attributes) as [key, value] (key)}
         {#if attributeGrouping[activePill]?.includes(key)}
-          <EventDetailsRowExpanded {key} {value} class="w-full" />
+          <EventDetailsRowExpanded {key} {value} {attributes} class="w-full" />
         {/if}
       {/each}
     </div>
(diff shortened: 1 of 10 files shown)


stevekinney (approved):

lminaudier:
Awesome, thanks a lot!

PR #853, opened 2022-09-26
Decode pending activities

Alex-Tideman:
## What was changed
In the event history store, loop through pending activities and decode any payloads. Reuses all of the current decode logic.

Screen Shot 2022-09-26 at 12 17 23 PM
Screen Shot 2022-09-26 at 12 17 31 PM

## Why?
Show heartbeatDetails payloads plus any other future payload fields on Pending Activities.

### DIFF
--- a/src/lib/stores/workflow-run.ts
+++ b/src/lib/stores/workflow-run.ts
@@ -7,19 +7,22 @@ import { fetchWorkflow } from '$lib/services/workflow-service';
 import { getPollers } from '$lib/services/pollers-service';
 import type { GetPollersResponse } from '$lib/services/pollers-service';
 import { decodeURIForSvelte } from '$lib/utilities/encode-uri';
+import { decodePendingActivity } from '$lib/models/pending-activities';
 
 export const refresh = writable(0);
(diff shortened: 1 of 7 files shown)


src/lib/stores/workflow-run.ts
```diff
@@ -35,12 +38,26 @@ const updateWorkflowRun: StartStopNotifier<{
...
-  return parameters.subscribe(({ namespace, workflowId, runId }) => {
+  return parameters.subscribe(({ namespace, workflowId, runId, settings }) => {
     if (namespace && workflowId && runId) {
       withLoading(loading, updating, async () => {
         const workflow = await fetchWorkflow({ namespace, workflowId, runId });
         const { taskQueue } = workflow;
         const workers = await getPollers({ queue: taskQueue, namespace });
+
+        // Decode pending activities
+        const pendingActivities = workflow?.pendingActivities ?? [];
```
stevekinney, line 49: Given that we do this for events in the model, I wonder if this belongs in a pending activities model as well. It would probably be easier to test.

stevekinney (approved):

PR #854, opened 2022-09-27
Fix `window.fetch` warning

lorensr:


README.md
```diff
@@ -30,6 +30,8 @@ Developing the UI has the same prequisites as trying it out. Once you've created
 pnpm start
 ```
 
+and open [`localhost:3000`](http://localhost:3000).
```
lorensr, line 33: Added because the 127.0.0.1 address displayed in the terminal results in CORS errors:

image

And I wasn't able to get vite to change it. I tried `--host localhost` and this:

image
image

src/routes/__layout-root.svelte
```diff
@@ -37,7 +37,7 @@
 
     const cluster: GetClusterInfoResponse = await fetchCluster(settings, fetch);
 
-    fetchSearchAttributes(settings);
+    fetchSearchAttributes(settings, fetch);
```
lorensr, line 40: Resolves:

image

stevekinney: If you want to break this out into it’s own PR, I’m comfortable merging this in while Candace reviews the design/copy change.

lorensr: @stevekinney  talked to Candace and removed the copy, so I think this is ready for merging ☺️

Alex-Tideman (approved):

PR #855, opened 2022-09-27
Use a promise for determining base url for api

Alex-Tideman:
## What was changed
Remove string template for building the base url for the api url and instead check for GetNamespaces() promise. This requires all routeForApi()'s to be promises. In OSS, this will be resolved immediately since globalThis?.GetNamespaces() is not defined and will continue to use getApiOrigin().  

## Why?
Because the url isn't guaranteed to be well named/consistent, we want to return a promise that resolves the api url for a namespace. 

## Checklist

2. How was this tested:
Updated all unit tests for routeForApi to pass, all other tests pass. Manually tested all parts of app.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@temporalio/ui",
-  "version": "2.1.59",
+  "version": "2.1.60",
   "type": "module",
   "description": "Temporal.io UI",
   "keywords": [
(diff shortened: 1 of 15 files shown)


rossedfort:
Idea that I'm not attached to because you already made a lot of changes: `requestFromApi` takes a `Promise<string>` for the route and awaits the promise itself instead of each of the callers.

PR #860, opened 2022-09-30
Fix dropdown button icon

rossedfort:
## What was changed

Abstract the open/close chevron icon into the menu-button component so that I could add the `pointer-events-none` class in 1 place instead of many. Adding this class means clicks from this element will be ignored (and captured by the parent) in opening a dropdown menu via a dropdown button primitive. 

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


src/lib/holocene/primitives/menu/menu-button.svelte
```diff
@@ -29,6 +31,12 @@
   {disabled}
 >
   <slot />
+  {#if hasIndicator}
```
laurakwhit, line 34: Wondering if we might want to keep the prop `readonly` (instead of `hasIndicator`) and use it with disabled here
```suggestion
  {#if !readonly && !disabled}
```
That way it's a little clearer what states the icons should/shouldn't be shown in and we don't have to pass an additional prop based on disabled which can already be passed down (e.g. `hasIndicator={!disabled}`). 

Or even take it one step further and just do this
```suggestion
  {#if !disabled}
```
since I'm not sure there is a case where we want the icons when its disabled or we don't want the icons and it's not effectively disabled 🤔

rossedfort: That make a lot of sense, there is one case however where this button can be enabled yet we don't show a caret: https://holocene.preview.thundergun.io/story/src-lib-holocene-dropdown-menu-story-svelte?variantId=_default

Maybe the issue here is that `<DropdownMenu />` shouldn't be using the `<MenuButton />` primitive though.

laurakwhit (approved):

PR #870, opened 2022-10-05
Encode task queue URLs in routes

wyattanderson:
## What was changed

This change ensures task queue names are URL-encoded.

## Why?

If a task queue name contains special characters (especially `/`), clicking a link to said task queue will result in a 404 as the route doesn't match.

## Checklist

1. Closes 

2. How was this tested:

- Tested with a local `temporalite` instance with a task queue containing slashes. Verified clicking the link works as expected.
- Added unit test coverage for this edge case.

3. Any docs updates needed?

N/A


Alex-Tideman (approved):
Thanks!

PR #878, opened 2022-10-13
POC of prop forwarding for holocene components

rossedfort:
## What was changed

Proof of concept/RFC on how to implement type-safe prop forwarding for holocene components.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


src/lib/holocene/accordion.svelte
```diff
@@ -2,6 +2,21 @@
   import Icon from '$holocene/icon/icon.svelte';
   import Badge from '$holocene/badge.svelte';
   import type { IconName } from './icon/paths';
+  import { omit } from '$lib/utilities/omit';
+
+  type $$Props = ComponentProps<
```
rossedfort, line 7: I'm not sure where the official docs live but this is implemented according to [this](https://github.com/dummdidumm/rfcs/blob/ts-typedefs-within-svelte-components/text/ts-typing-props-slots-events.md#typing-props) RFC.

Alex-Tideman:
So here's an idea. Might not be great, but I'm lazy and also would keep $$Props types and the svelte file props from drifting. Have a script that would look at the first html element to get the HTMLDivElement type, and then look at any line that starts with export let and grab all the props with type.

rossedfort:
> So here's an idea. Might not be great, but I'm lazy and also would keep $$Props types and the svelte file props from drifting. Have a script that would look at the first html element to get the HTMLDivElement type, and then look at any line that starts with export let and grab all the props with type.

That's an interesting idea 🤔 I agree, having types defined in both `$$Props` and each `export let prop` is not ideal.

src/lib/holocene/button.svelte
```diff
@@ -19,22 +36,20 @@
...
     on:click
-    class="button {variant} {classes}"
+    class="button {variant} {$$props.class}"
     class:selected={active}
     class:large
     class:thin
-    data-cy={dataCy}
     {type}
     {disabled}
+    {...omit($$restProps, 'class')}
```
laurakwhit, line 52: Is there currently a use case where `$$restProps` is needed? 

Just thinking from a design systems standpoint I'm not sure if we want to be passing down unknown attributes. If we need something more extensible it seems like that would indicate we haven't thought through a potential variant or we should be building a custom component 🤔 

> It shares the same optimisation problems as $$props, and is likewise not recommended.

On the other hand... We're already using `$$props` and I know this would likely be useful 😅 Just don't know if it's the _right_ thing to do.

One alternative we could explore would be using objects as props, especially if we want more flexibility for customization of child elements.
```suggestion
export let buttonProps = {}
<div>
  <button
    on:click
    ...
    {...buttonProps}
    ...
```

rossedfort: > Is there currently a use case where $$restProps is needed?

This came about because I wanted to write a playwright test in cloud-ui, but couldn't add a `data` attribute or even an `id` to a `<Button />` without first going to add that new prop to the component in Holocene, publishing a new package version, updating the package in cloud-ui, etc. That led to this idea of enabling all Holocene components to accept any valid HTML attribute as a prop. I've seen design systems work similarly to this, because you wouldn't want to maintain a list of valid HTML Attributes for every component in your component library. For example, if we waned Holocene to support `aria` attributes on every component, specifying that for each component would be cumbersome. 

Another pattern I've seen is to create a `<Box />` component, or "primitive", from which all other components are made. The `<Box />` component would take an `as` prop representing the HTML Element it actually renders, so `<Box as="a" />` would render an anchor tag, <Box as="div" />` would render a `<div />` etc. Then this `<Box />` component would maintain this list of valid HTML Attributes and do the prop forwarding itself.

laurakwhit: > couldn't add a data attribute or even an id to a <Button /> without first going to add that new prop to the component in Holocene

Ah I see. And yeah that isn't ideal 😞 The `<Box />` is an interesting pattern 🤔 
I'll all for giving `$$restProps` a try! Although, I wonder if there's a better approach to this pattern at least...
```js
omit($$restProps, 'class')
```
I'll noodle on it a bit 🍜 Thanks for putting this POC together!

rossedfort: @laurakwhit yeah, that's the part of this implementation I'm least OK with, but without that, if a class is passed to a component, it will overwrite our inline component classes entirely :/

PR #879, opened 2022-10-13
Decode queries and stack traces, use json-bigint to support BigInt

Alex-Tideman:
## What was changed
Pass query api results into the data encoder (which includes stack trace query). Use json-bigint to support BigInt types to prevent corruption of very large numbers with JSON.parse/stringify.

### Before (rounding large numbers)
Screen Shot 2022-10-13 at 1 28 45 PM

### After (exact large numbers)
Screen Shot 2022-10-13 at 1 29 08 PM

## Why?
Keep quality of data, decode queries when using data encoder.


Alex-Tideman:
Want to get some opinions on using json-bigint (https://github.com/sidorares/json-bigint). There are no great solutions for this problem until JSON.parse supports BigInt

rossedfort (approved):

PR #880, opened 2022-10-13
Auto-refresh of workflow and event history every 15 seconds

### LINKED ISSUE #862: Refresh button on detail page

Jumping back to the main page to reload, or using force-reload in the browser, to get updated details is a pain.

It would be nice, when looking at the workflow execution details, if it would provide a refresh button similar to what is on the main page.  This prevents having to jump around to get up to date info on currently running tasks.

An auto-refresh when not "completed" may be nice too...

## Describe alternatives you've considered

None.

## Additional context

None.

Alex-Tideman:
## What was changed
Auto-refresh workflow with event history every 15 seconds if the workflow is running. Auto-refresh is defaulted off. Once toggled on, it will immediately fetch workflow and event history and then set the refresh interval for 15 seconds.

Screen Shot 2022-10-17 at 12 18 21 PM
Screen Shot 2022-10-17 at 12 18 51 PM

## Why?
Provide liveliness of workflow data.

Closes #862


src/lib/layouts/workflow-history-layout.svelte
```diff
@@ -39,7 +41,18 @@
     run: workflow.runId,
   };
 
+  $: {
```
laurakwhit, line 44: Could this just be?
```suggestion
  $: {
    clearInterval(refreshInterval);
    if ($workflowRun.workflow.isRunning) {
      // Auto-refresh of 15 seconds
      refreshInterval = setInterval(refreshWorkflows, 15000);
    }
  }
```

laurakwhit (approved):

src/lib/layouts/workflow-history-layout.svelte
```diff
@@ -39,7 +41,18 @@
...
+    if ($workflowRun.workflow.isRunning) {
+      clearInterval(refreshInterval);
+      // Auto-refresh of 15 seconds
+      refreshInterval = setInterval(() => ($refresh = Date.now()), 15000);
```
StarpTech, line 48: Hi, I recommend making this optional. It can be very annoying because the UI has components with states.

Alex-Tideman: Yup will be an opt-in feature, have the designs for it now and it will be included in this PR.

src/lib/layouts/workflow-history-layout.svelte
Alex-Tideman: 💯

src/lib/layouts/workflow-history-layout.svelte
StarpTech: Awesome

rossedfort (approved):

PR #883, opened 2022-10-14
Fix task queues fetched twice on workflow

laurakwhit:
## What was changed

I've removed the `<slot />` in favor of conditionally rendering different layouts based on the `eventViewType`. This doesn't seem like the ideal solution, as we also still need to render them under `compact`, `feed`, and `json` in `src/routes/.../history` 🤔 
## Why?

When a user clicks on a workflow in the workflow list and navigates to it the task-queues are fetched twice.

This is because `fetchWorkflows` in [subscribe](https://github.com/temporalio/ui/blob/main/src/lib/stores/workflow-run.ts#L47) is being called twice in the workflow run store. This is due to the components (e.g. `WorkflowHistoryFeed`) being passed [here](https://github.com/temporalio/ui/blob/main/src/routes/namespaces/%5Bnamespace%5D/workflows/%5Bworkflow%5D/%5Brun%5D/history/__layout.svelte#L9) (or more specifically [here](https://github.com/temporalio/ui/blob/main/src/lib/layouts/workflow-history-layout.svelte#L156)) rendering twice when the page is navigated to.

I don't know why they're doing that 😅 Maybe a bug? Maybe a symptom of how we're doing routes and layouts here? Will continue to investigate... 

## Checklist

1. Closes: n/a 

2. How was this tested:

- [ ] When navigating to a workflow from the workflow list, verify task-queue are fetched only **once**
- [ ] Verify navigating to the following works as expected
   - `history`
   - `history/feed`
   - `history/compact`
   - `history/json`

### DIFF
--- a/src/lib/layouts/workflow-history-compact.svelte
+++ b/src/lib/layouts/workflow-history-compact.svelte
@@ -1,14 +1,8 @@
 <script lang="ts">
-  import { page } from '$app/stores';
-
   import EventSummary from '$lib/components/event/event-summary.svelte';
-  import PageTitle from '$lib/holocene/page-title.svelte';
   import { ascendingEventGroups, loading } from '$lib/stores/events';
-
-  const workflow = $page.params?.workflow;
 </script>
 
-<PageTitle title={`Workflow History | ${workflow}`} url={$page.url.href} />
 <EventSummary
   items={$ascendingEventGroups}
   groups={$ascendingEventGroups}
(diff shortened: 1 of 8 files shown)


laurakwhit:
Closing this for now. May revisit if this is still an issue after SvelteKit upgrade.

laurakwhit:
This has been fixed in https://github.com/temporalio/ui/pull/984 🎉

PR #894, opened 2022-10-20
Namespace Selector

Alex-Tideman:
## What was changed
For self-hosted users, if a user logs in and is not authorized to view the default namespace (either default from settings or first namespace in namespace list api call), the user is immediately kicked back out to login even if if they have access to other namespaces.

This updates the code to check if a user has access to the lasted used namespace or default namespace:
- Show Loading screen while checking
- Determine initial namespace to redirect to. If user has last used namespace, go to that one, other use the default namespace logic (default namespace from settings or first namespace in namespace list)
- If they have access to the namespace, redirect to the workflows page.
- If they do not have access, redirect to /select-namespace page.
- From /select-namespace, they can view and filter namespaces. On click, check for access. If they have access to the namespace, redirect to the workflows page. If not, show toast message.

Screen Shot 2022-10-20 at 1 54 52 PM
 
## Why?
Prevent users from being stuck in auth logout flow when they don't have access to the default namespace.


rossedfort (approved):
Looks good to me, however we had a [design](https://www.figma.com/file/8dSoDU2sWeQf7CgZzTqzeW/Cloud-Account-Management?node-id=2174%3A25233) for a similar case in cloud-ui, wonder if we should use this modal as opposed to a full page?

Alex-Tideman:
Yea I talked to her about that, rather have a specific page instead of a root route modal. This is more short-term fix to problem, a full landing page will be in the near future.

PR #899, opened 2022-10-25
Correctly format retention periods

### LINKED ISSUE #885: The UI shows an incorrect value for retention period when it is higher than 30 days
**Describe the bug**
The UI shows an incorrect value for retention period when the namespace's retention period is set higher than 30 days. 

**To Reproduce**

Use tctl to set the namespace's retention period, higher than 30 days
then navigate to http://localhost:8080/namespaces/default

- after running `tctl namespace update --retention 33` the UI shows 2 days

Screenshot 2022-10-17 at 12 14 42

- after running `tctl namespace update --retention 30` the UI shows 30 days

stevekinney:
Closes #885 and adds unit tests.

### DIFF
--- a/src/lib/utilities/format-date.test.ts
+++ b/src/lib/utilities/format-date.test.ts
@@ -145,13 +145,17 @@ describe('fromSecondsToDaysOrHours', () => {
     const seconds = '432000s';
     expect(fromSecondsToDaysOrHours(seconds)).toBe('5 days');
   });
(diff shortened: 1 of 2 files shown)


Alex-Tideman:
Good catch, seems like that durationToString for days is scoped to max out for days in a month

Alex-Tideman (approved):

PR #900, opened 2022-10-25
Parent workflow links

### LINKED ISSUE #864: [Bug] Parent Id link is not correct when the parent workflow link is in different namespace.
### What are you really trying to do?

Create a child workflow in a different namespace. Go to the childworkflow in UI. The link to parent workflow in `parent Id` is not correct.

Expected:
`/namespaces/<PARENT_NAMESPACE>/workflows/<PARENT_WORKFLOW_ID>/<PARENT_RUN_ID>`

Actual:
`/namespaces/<CHILD_NAMESPACE>/workflows/<PARENT_WORKFLOW_ID>/<PARENT_RUN_ID>`

### Environment/Versions
ui-server version 2.6.2

### Additional context

image

Alex-Tideman:
## What was changed
Add links to parent workflow executions with lookup to parent namespace.

Also fixes issue with execution/task-queue links on goto erroring out (no workflow or namspace params)

## Why?
Better navigation to parent workflows

Fixes #864


src/lib/components/event/event-details-row-expanded.svelte
```diff
@@ -40,7 +40,7 @@
         {inline}
       />
     </div>
-  {:else if shouldDisplayAsExecutionLink(key)}
+  {:else if shouldDisplayAsExecutionLink(key) && namespace && workflow}
```
laurakwhit, line 43: What do you think about passing `$page.params` to `shouldDisplayAsExecutionLink` and checking for `namespace` and `workflow` there? It seems like they're needed wherever this is currently being called and I see you've done something similar with `shouldDisplayParentWorkflowLink`.

laurakwhit (approved):
💥 🚀

src/lib/components/event/event-details-row.svelte
```diff
@@ -35,7 +35,7 @@
       {format(key)}
     </p>
     <CodeBlock content={getCodeBlockValue(value)} class="w-[95%]" {inline} />
-  {:else if shouldDisplayAsExecutionLink(key)}
+  {:else if shouldDisplayAsExecutionLink(key) && workflow && namespace}
```
stevekinney, line 38: Maybe we need a new utility at this point? Or maybe `shouldDisplayAsExecutionLink` needs more optional arguments?

src/lib/components/workflow/workflow-detail.svelte
```diff
@@ -8,7 +8,7 @@
   export let textSize = 'md';
 </script>
 
-<article class="flex gap-2 text-{textSize}">
+<article class="flex gap-2 text-{textSize}" data-cy={$$props['data-cy']}>
```
stevekinney, line 11: Not a blocker or anything, but maybe we just spread the props at this point?

src/lib/layouts/workflow-history-layout.svelte
```diff
@@ -22,6 +22,11 @@
 
   const { namespace } = $page.params;
   const { workflow, workers } = $workflowRun;
+  const namespaces = $page.stuff?.namespaces;
+
+  $: parentNamespace = namespaces?.find(
```
stevekinney, line 27: I'd break this out and wrap it in a unit test. If you count every `?.` as a conditional, there are lots of fun ways things could go wrong.

src/lib/utilities/get-single-attribute-for-event.ts
```diff
@@ -108,6 +109,28 @@ export const shouldDisplayChildWorkflowLink = (
   return false;
 };
 
+const keysWithParentExecutionLinks = [
+  'parentWorkflowExecutionWorkflowId',
+  'parentWorkflowExecutionRunId',
+] as const;
+
+// For linking to a parent workflow
```
stevekinney, line 117: This comment is probably unnecessary given the very clear name of the function. I think if we're going to add comments, we might as well use JSDoc syntax so that Intellisense picks it up, right?

PR #903, opened 2022-10-26
API Pagination Holocene component

laurakwhit:
## What was changed

Adds an API Pagination component to Holocene that fetches data based on page size.

Screen Shot 2022-10-27 at 12 39 00 PM

## Why?

Allows us to fetch data as needed.

## Checklist

1. Closes: n/a 

2. How was this tested: n/a

### DIFF
--- a/src/lib/holocene/api-pagination.svelte
+++ b/src/lib/holocene/api-pagination.svelte
@@ -87,18 +87,15 @@
         {/if}
       </p>
       <nav
-        class="flex items-center justify-start gap-8 text-sm md:justify-end md:text-base"
+        class="flex items-center justify-start gap-4 text-sm md:justify-end md:text-base"
       >
-        <div class="flex items-center justify-center gap-2">
-          <p class="w-fit text-right">Per Page</p>
-          <FilterSelect
-            label="Per Page"
-            parameter={$store.key}
-            value={String($store.pageSize)}
-            options={perPageOptions($store.pageSize)}
-          />
-        </div>
-        <div class="flex items-center justify-center gap-6">
+        <FilterSelect
+          label="Per Page"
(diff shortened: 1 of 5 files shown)


Alex-Tideman:
If you want to pay for my prior sins, we should add some unit tests for that api-pagination.ts file. pagination.ts has a bunch of them if you need a starting point.

src/lib/holocene/api-pagination.svelte
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  $: {
+    onPageSizeChange(pageSize);
+  }
+
+  $: isEmpty = $store.items.length === 0 && !$store.loading;
+</script>
+
+{#if isUnauthorizedError(error) && $$slots.unauthorized}
```
laurakwhit, line 69: This doesn't seem super extensible 😞 Would welcome any thoughts on a better way to go about this.

src/lib/stores/api-pagination.ts
```diff
@@ -0,0 +1,136 @@
...
+    items: any[],
+    store: PaginationItems,
+  ) => {
+    const _store = { ...store };
+    _store.items = items;
+    _store.updating = false;
+    _store.loading = false;
+    _store.hasNext = Boolean(nextToken);
+    _store.index = store.nextIndex;
+    _store.indexTokens = { ...store.indexTokens };
```
laurakwhit, line 93: Added this since the `indexTokens` on `initialState` were being overwritten.
```js
    _store.indexTokens = { ...store.indexTokens };
```

rossedfort (approved):
Looks good, I think it would behoove us to see if there's more opportunity to share code between this component and the base `<Pagination />` component some time in the future.

PR #904, opened 2022-10-27
Create fetching for individual namespace

stevekinney:


src/api.d.ts
```diff
@@ -1,7 +1,8 @@
 type WorkflowsAPIRoutePath =
   | 'workflows'
   | 'workflows.archived'
-  | 'workflows.count';
+  | 'workflows.count'
+  | 'namespace';
```
Alex-Tideman, line 5: Tiny nit but I don't think a single namespace api belong with the WorkflowsAPIRoutePath. I'm guessing they have the same parameters but probably best to have a separate NamespaceAPIRoutePath

Alex-Tideman (approved):

PR #907, opened 2022-10-28
Remove fetch namespaces on load

laurakwhit:
## What was changed

Removes `namespaces` from `$page.stuff` and instead fetches them as needed.

## Why?

So we aren't blocking `load` on the root layout waiting for namespaces.

## Checklist

1. Closes: DT-37 

2. How was this tested: TBD


src/lib/components/namespace-list.svelte
```diff
@@ -1,19 +1,79 @@
...
+  import { lastUsedNamespace } from '$lib/stores/namespaces';
+  import { workflowSorts, workflowFilters } from '$lib/stores/filters';
+  import { goto } from '$app/navigation';
+  import {
+    routeForWorkflows,
+    routeForSchedules,
+  } from '$lib/utilities/route-for';
+  import { fetchNamespaces } from '$lib/services/namespaces-service';
 
-  export let getNamespaceList: () => Promise<NamespaceItem[]> = null;
```
Alex-Tideman, line 10: We'll need to rethink this a bit since cloud relies on passing in GetNamespaceList to FullNav and passing this down to namespace-list.

laurakwhit: I've added back `getNamespaceList` and the other props being passed to `Navigation` ✅ 

One thing I noticed is that we are currently passing `$lastUsedNamespace` (a `string`) to `activeNamespace` (a `Namespace`) in cloud currently which seems incorrect 🤔

PR #912, opened 2022-10-31
WIP - bulk terminate

rossedfort:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


src/lib/services/terminate-service.ts
```diff
@@ -24,3 +25,42 @@ export async function terminateWorkflow({
...
+
+  return await requestFromAPI<null>(route, {
+    options: {
+      method: 'POST',
+      body: stringifyWithBigInt({
+        jobId: uuidv4(),
+        namespace,
+        visibilityQuery: query,
+        reason,
+        terminationOperation: {},
```
rossedfort, line 60: Idk if we need to include `details` or `identity` here, but it seems to work fine without it.

Alex-Tideman (approved):

stevekinney (approved):

src/lib/services/terminate-service.ts
Alex-Tideman: We don't need to, but down the road we might send 'Web UI' or something as identity for all APIs. But not for this PR

src/routes/namespaces/[namespace]/workflows/index@root.svelte
```diff
@@ -3,10 +3,13 @@
...
+  import { isVersionNewer } from '$lib/utilities/version-check';
 </script>
 
 <PageTitle
   title={`Workflows | ${$page.params.namespace}`}
   url={$page.url.href}
 />
-<Workflows />
+
+<Workflows bulkActionsEnabled={isVersionNewer($temporalVersion, '1.18.0')} />
```
Comment (rossedfort) [inline src/routes/namespaces/[namespace]/workflows/index@root.svelte:15]: @Alex-Tideman do we need to just set this to `false` for this PR? I think we're good to release this properly once ui-server is updated. I think it's a fairly polished experience, except maybe for large batches.

Comment (Alex-Tideman) [reply src/routes/namespaces/[namespace]/workflows/index@root.svelte:15]: I think we should set this to false on the page itself (right now export let bulkActionsEnabled = true). Because this should get deployed with the ui-server change when it's merged, but if we update package.json version, cloud will get this but the ui-server won't have been deployed yet

Alex-Tideman (approved):

PR #915, opened 2022-11-02
Remove aliases to date-fns methods

stevekinney:
Removes renaming functions form `date-fns`.


laurakwhit:
Went with https://github.com/temporalio/ui/pull/917 instead.

PR #921, opened 2022-11-03
DT-50 - modal and skeleton updates.

rossedfort:
## What was changed

- Allow pressing "Escape" to close a modal
- Remove the `height` class from Skeleton
- Bump version to 2.1.81
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/lib/holocene/modal.svelte
+++ b/src/lib/holocene/modal.svelte
@@ -11,11 +11,14 @@
   export let confirmDisabled: boolean = false;
   export let large: boolean = false;
 
-  const dispatch = createEventDispatcher();
+  const dispatch = createEventDispatcher<{
+    cancelModal: undefined;
+    confirmModal: undefined;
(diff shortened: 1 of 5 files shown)


src/lib/holocene/modal.svelte
```diff
@@ -11,14 +11,24 @@
...
   export let large: boolean = false;
 
-  const dispatch = createEventDispatcher();
+  const dispatch = createEventDispatcher<{
+    cancelModal: undefined;
+    confirmModal: undefined;
+  }>();
+
+  const handleKeyUp = (event: KeyboardEvent) => {
+    if (event.key === 'Escape') {
```
laurakwhit, line 20: 🥳 Wondering if we want to make clicking the overlay close the modal as a well?

laurakwhit (approved):

PR #923, opened 2022-11-03
Fix event history sort order stickiness

Alex-Tideman:
## What was changed
When navigating away from the workflow details page, the $eventSortParam store was always being set to 'descending', thus always setting eventSortOrder to 'descending'. Removing that store dependencies fixes the issue and keeps the correct behavior.

### DIFF
--- a/src/lib/stores/event-view.ts
+++ b/src/lib/stores/event-view.ts
@@ -48,13 +48,10 @@ export const supportsReverseOrder = derived(
 export const eventSortOrder: Readable<EventSortOrder> = derived(
   [eventFilterSort, supportsReverseOrder],
   ([$eventFilterSort, $supportsReverseOrder]) => {
-    let sortOrder: EventSortOrder;
+    let sortOrder: EventSortOrder = 'ascending';
     if ($supportsReverseOrder) {
       sortOrder = $eventFilterSort;
-    } else {
-      sortOrder = 'ascending';
     }
-
     return sortOrder;
   },
 );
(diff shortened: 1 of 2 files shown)


src/lib/stores/event-view.ts
```diff
@@ -55,14 +46,11 @@ export const supportsReverseOrder = derived(
 );
 
 export const eventSortOrder: Readable<EventSortOrder> = derived(
-  [eventFilterSort, supportsReverseOrder, eventSortParam],
-  ([$eventFilterSort, $supportsReverseOrder, $eventSortParam]) => {
-    let sortOrder: EventSortOrder;
+  [eventFilterSort, supportsReverseOrder],
+  ([$eventFilterSort, $supportsReverseOrder]) => {
+    let sortOrder: EventSortOrder = 'ascending';
```
stevekinney, line 51: Shouldn't we default to "descending" and does this totally drop support for the query param?

Alex-Tideman: No we should default to 'ascending' since not all versions support reverse order. If it does support reverse order, the eventFilterSort store is defaulted to 'descending' so first time users will still get reverse order if the version supports it. The sort change button updates the query param, this eventSortParam store was only used in this derived store (I'm not totally sure why we even had it in the first place, guessing a sveltekit issue)

stevekinney: Yea, we definitely need to rethink our state management.

stevekinney (approved):

Issue #909, opened 2022-10-29
Improve keyboard experience in Data Encoder modal


lorensr:
## Is your feature request related to a problem? Please describe.

Here, it's hard to get focus to the inputs using my keyboard. And when I enter text and hit the return key, nothing happens.

![image](https://user-images.githubusercontent.com/251288/198752098-82f34cf3-08ed-4b4b-bfff-0e1bf90c15aa.png)

## Describe the solution you'd like

- Once modal is opened, focus Remote Codec Endpoint input field. (Or alternatively, make it selected when I hit `Tab` once.)
- When I hit `return`, perform the `Set`  button action
- When I Set a change, maybe close the nodal, or change `Cancel` button to `Done`

PR #925, opened 2022-11-03
Improve keyboard experience in Data Encoder modal

laurakwhit:
## What was changed

* Hitting return/enter on inputs sets endpoint/port
* Once modal is opened, focus goes to first focusable element after hitting tab
* `Cancel` button text has been changed to `Done`
## Checklist

1. Closes https://github.com/temporalio/ui/issues/909 

2. How was this tested: 

- [ ] Verify hitting return/enter on inputs sets endpoint/port
- [ ] Verify once modal is opened, focus goes to first focusable element after hitting tab

### DIFF
--- a/src/lib/holocene/data-encoder-settings.svelte
+++ b/src/lib/holocene/data-encoder-settings.svelte
@@ -36,6 +36,10 @@
     }
   };
 
+  const onPortSet = () => {
+    $dataConverterPort = port;
+  };
+
   const onEndpointClear = () => {
     endpoint = '';
     $dataEncoderEndpoint = null;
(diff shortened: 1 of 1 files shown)


src/lib/holocene/data-encoder-settings.svelte
```diff
@@ -52,23 +56,28 @@
   hideConfirm
   on:cancelModal={onCancel}
   on:confirmModal={onCancel}
+  cancelText="Done"
 >
-  <h3 slot="title" data-cy="data-encoder-title">Data Encoder</h3>
+  <h3 slot="title" data-cy="data-encoder-title" tabindex={0} autofocus>
```
laurakwhit, line 61: Ideally (instead of using `tabindex` and `autofocus` here) we might want to look into updating our Holocene modal component to take focus when opened and have a focus trap.

GiantRobots: I like that suggestion a lot can we make a Jira ticket for it and throw it in next sprint?

laurakwhit: Added ✅

GiantRobots (approved):

PR #930, opened 2022-11-07
Add Local Activity support and show activity type for Local Activities in event history

Alex-Tideman:
## What was changed
LocalActivity is a special marker and has an activity type associated with it. This change labels any localActivity markers as 'Local Activity' and finds the activity type from the details data payload.

**Before**
Screen Shot 2022-11-07 at 2 32 32 PM

**After**
Screen Shot 2022-11-07 at 2 32 48 PM

Screen Shot 2022-11-07 at 2 42 25 PM

## Why?
Show specific information for a LocalActivity

## Checklist

- [x] Confirm with Dmitry


stevekinney (approved):

Spikhalskiy (approved):
Thank you Alex for improving this!

PR #936, opened 2022-11-08
enable and production-ize bulk termination

rossedfort:
## What was changed

A few fixes before releasing bulk terminate
- close pagination loophole that allows users to specify page sizes that aren't supported
- enables bulk termination in the ui for clusters that support advanced visibility _and_ are running temporal server version >1.18.0
- adds cy tests
- fixes bulk action button for small screens
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


Alex-Tideman:
Let's add some Cypress tests for this feature

Alex-Tideman:
This is looking good! I think the one other valuable Cypress test to add here is for non-elasticsearch, assert the selectable table isn't shown.

Alex-Tideman (approved):

PR #944, opened 2022-11-14
add ability to cancel workflows from workflow details page

### LINKED ISSUE #802: Add Cancel button
**Is your feature request related to a problem? Please describe.**

A user is using the Terminate button because it's the only way to (via the UI) stop a Workflow. I think we want to generally encourage people to use Cancel over Terminate.

**Describe the solution you'd like**

Add a Cancel button and make it appear the default/recommended way to stop the Workflow.

rossedfort:
## What was changed

- Add a split button to the workflow header with cancel being the primary action - hide the terminate button under the menu options
- Refactor toaster component to support toasts in different position on the page, eg. bottom right, bottom left, and top left.
- Show an alert on the Workflow Details page for workflows that are currently in the process of being cancelled.
## Why?

To allow users to cancel workflows from the UI
## Checklist

1. Closes #802

2. How was this tested:

Tested locally against docker-compose and canary 
3. Any docs updates needed?


src/lib/components/workflow-actions.svelte
```diff
@@ -0,0 +1,147 @@
...
+
+      setTimeout(() => {
+        toaster.push({
+          message: 'Canceled workflow.',
+          yPosition: 'bottom',
+        });
+        showCancellationConfirmation = false;
+        loading = false;
+        $refresh = Date.now();
+      }, 1000);
```
rossedfort, line 74: Just calling out this atrocity, but it works in my testing. Should we poll instead?

cypress/integration/workflow-termination.spec.js
```diff
@@ -49,6 +50,6 @@ describe('Terminate Button is Shown', () => {
     cy.wait('@settings-api');
     cy.wait('@workflow-api');
 
-    cy.get('[data-cy="terminate-button"]').should('not.exist');
+    cy.get('[aria-controls="workflow-actions"]').should('not.exist');
```
Alex-Tideman, line 53: I think it'd be worth it to add the modal part of the flow to these tests. Click button, assert modal is there with right title / confirm button text and input for reason for terminate.

src/lib/components/workflow-actions.svelte
Alex-Tideman: So the one thing I think worth check with OSS Server team is if a cancel request could be blocked or delayed by a current activity. I had a workflow with an WorkflowTaskFailed (see recording), and when I cancel it doesn't cancel (still running). I don't get any error from the cancel request, but maybe it does cancel just not until some task finishes.

Alex-Tideman (approved):

Alex-Tideman (approved):

PR #956, opened 2022-11-21
[DT-149] Fix for invite user issue with email addition auto scrolling

laurakwhit:
## What was changed

* Adds auto scroll so on the input (and cursor) is visible after chips are added or the input is selected.
* Capitalizes "N" for Namespace on the Namespace switcher and elsewhere

| Tooltip | Switcher | Welcome page |
| -- | -- | -- |
|Screen Shot 2022-11-21 at 2 13 26 PM|Screen Shot 2022-11-21 at 2 13 39 PM|Screen Shot 2022-11-21 at 2 14 58 PM|

* Capitalizes "N" for Namespace on the Namespace switcher
## Checklist

1. Closes: `DT-149` and `DT-151` 

2. How was this tested:

- [ ] Verify that after typing out multiple tags (past the min height), and another tag, is added that the container scrolls and the input and cursor are visible
- [ ] Verify after pasting multiple tags (past the min height) that the container scrolls and the input and cursor are visible
- [ ] Verify scrolling up and removing a tag via `x` that the container does not scroll 
- [ ] Verify if there are multiple tags (past the min height) and the input is selected, the container scrolls and the input and cursor are visible
- [ ] Verify input still works as expected
---
- [ ] Verify `Namespace` is capitalized on the Namespace switcher
- [ ] Verify `Namespace` is capitalized on the welcome page

### DIFF
--- a/cypress/integration/namespaces.spec.js
+++ b/cypress/integration/namespaces.spec.js
@@ -74,12 +74,12 @@ describe('Namespace Select', () => {
 
   it('have the correct namespaces in the dropdown when using navigation header', () => {
     cy.get('@namespace-select-button').click({ wait: 1000 });
-    cy.get('.prose > .text-2xl').contains('Select a Namespace');
+    cy.get('h2').contains('Select a Namespace');
   });
 
   it('navigates to the correct namespaces in the dropdown when using navigation header', () => {
     cy.get('@namespace-select-button').click({ wait: 1000 });
-    cy.get('.prose > .text-2xl').contains('Select a Namespace');
+    cy.get('h2').contains('Select a Namespace');
     cy.get('[data-cy="namespace-list"] > :nth-child(2)').click();
     cy.get('[data-cy="namespace-name"]').contains(namespaces[1]);
   });
--- a/cypress/integration/namespaces.spec.js
+++ b/cypress/integration/namespaces.spec.js
@@ -74,12 +74,12 @@ describe('Namespace Select', () => {
 
   it('have the correct namespaces in the dropdown when using navigation header', () => {
     cy.get('@namespace-select-button').click({ wait: 1000 });
-    cy.get('.prose > .text-2xl').contains('Select a namespace');
+    cy.get('.prose > .text-2xl').contains('Select a Namespace');
   });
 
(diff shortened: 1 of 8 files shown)


cypress/integration/namespaces.spec.js
```diff
@@ -74,12 +74,12 @@ describe('Namespace Select', () => {
 
   it('have the correct namespaces in the dropdown when using navigation header', () => {
     cy.get('@namespace-select-button').click({ wait: 1000 });
-    cy.get('.prose > .text-2xl').contains('Select a namespace');
+    cy.get('.prose > .text-2xl').contains('Select a Namespace');
```
rossedfort, line 77: Maybe toss a better selector on this element if we are relying on it for a test - although I'm not sure why we are even testing copy 🤔

rossedfort (commented):
Couple of small suggestions, but looks good otherwise!

src/lib/holocene/input/chip-input.svelte
```diff
@@ -10,13 +12,39 @@
   export let required = false;
   export let hintText = '';
   export let validator: (value: string) => boolean = () => true;
+  const values = writable<string[]>(chips);
```
rossedfort, line 15: What's the motivation behind making this a store?

laurakwhit: I wanted some way to compare the previous values to the new values to determine if `shouldScrollToInput` after the input updates. Open to ideas if there's a better way to go about this!

cypress/integration/namespaces.spec.js
```diff
@@ -74,12 +74,12 @@ describe('Namespace Select', () => {
 
   it('have the correct namespaces in the dropdown when using navigation header', () => {
     cy.get('@namespace-select-button').click({ wait: 1000 });
-    cy.get('.prose > .text-2xl').contains('Select a Namespace');
+    cy.get('h2').contains('Select a Namespace');
```
rossedfort, line 77: If we put another `<h2>` on this page, this test will break. I had something in mind more like adding a `data-cy` attribute on the h2 element, then using that in the test.

Something like `cy.get('[data-cy="namespace-selector-heading"]').contains('Select a Namespace');`

rossedfort (approved):

PR #966, opened 2022-11-29
DT-131 - improve keyboard navigation for modals

rossedfort:
## What was changed

- trap focus within modals by cycling to the previous element if the first/last focusable element is focused
- works for forward and backward navigation, i.e. `tab` and `shift+tab`

https://user-images.githubusercontent.com/11775628/204651804-9798489b-0eca-43cc-a733-42d0031fb246.mov

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/lib/holocene/modal.svelte
+++ b/src/lib/holocene/modal.svelte
@@ -15,6 +15,7 @@
   let closeButton: HTMLButtonElement;
   let cancelButton: Button;
   let confirmButton: Button;
+  let modalElement: HTMLDivElement;
 
   const dispatch = createEventDispatcher<{
     cancelModal: undefined;
@@ -49,20 +50,20 @@
   };
 
   $: {
-    if (open && closeButton) {
-      closeButton.focus({ focusVisible: true } as any);
+    if (open && modalElement) {
+      modalElement.focus();
     }
   }
 </script>
 
 <svelte:window on:keydown={handleKeyboardNavigation} />
 {#if open}
-  <div class="modal">
+  <div bind:this={modalElement} class="modal">
     <div on:click={cancelModal} class="overlay" />
     <div
       class="body"
       class:large
-      aria-modal="true"
+      tabindex="-1"
       role="alertdialog"
       aria-labelledby="modal-title"
       aria-describedby="modal-description"
--- a/src/lib/holocene/modal.svelte
+++ b/src/lib/holocene/modal.svelte
@@ -12,9 +12,6 @@
   export let large: boolean = false;
   export let loading: boolean = false;
 
-  let closeButton: HTMLButtonElement;
-  let cancelButton: Button;
-  let confirmButton: Button;
   let modalElement: HTMLDivElement;
 
   const dispatch = createEventDispatcher<{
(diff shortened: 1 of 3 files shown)


src/lib/holocene/modal.svelte
```diff
@@ -21,45 +25,82 @@
...
+          event.preventDefault();
+        }
+      } else if (document.activeElement === confirmButton.buttonElement) {
+        closeButton.focus();
+        event.preventDefault();
+      }
     }
   };
+
+  $: {
```
laurakwhit, line 51: I think we can utilize `bind:this` and focus the modal instead of the button (which will make the screen reader a little bit easier to understand and avoids `autofocus`).
```suggestion
  $: {
    if (open && modal) {
      modal.focus();
    }
  }
```

 See additional comment below 👇

src/lib/holocene/modal.svelte
```diff
@@ -21,45 +25,82 @@
...
-<svelte:window on:keyup={handleKeyUp} />
+<svelte:window on:keydown={handleKeyboardNavigation} />
 {#if open}
   <div class="modal">
     <div on:click={cancelModal} class="overlay" />
-    <div class="body" class:large>
+    <div
+      class="body"
+      class:large
+      aria-modal="true"
```
laurakwhit, line 65:
```suggestion
      tabindex="-1"
      bind:this={modal} // add a modal (HTMLDivElement) variable
      role="dialog"
      aria-label={ariaLabel} // add an ariaLabel prop
```

src/lib/holocene/modal.svelte
```diff
@@ -21,45 +26,82 @@
...
+    }
+
+    if (event.key === 'Tab') {
+      if (event.shiftKey) {
+        if (document.activeElement === closeButton) {
+          confirmButton.focus();
+          event.preventDefault();
+        }
+      } else if (document.activeElement === confirmButton.buttonElement) {
+        closeButton.focus();
```
laurakwhit, line 46: There's an edge case here where if `loading` is true the focus can escape the modal. 

What do you think about a more generic check for a first and last focusable element? e.g.

```js
    const focusable = modalElement.querySelectorAll('button');
    const firstFocusable = focusable[0];
    const lastFocusable = focusable[focusable.length - 1];
    if (event.key === 'Tab') {
      if (event.shiftKey) {
        if (document.activeElement === firstFocusable) {
          lastFocusable.focus();
          event.preventDefault();
        }
      } else if (document.activeElement === lastFocusable) {
        firstFocusable.focus();
        event.preventDefault();
      }
    }
```

This means we could also remove the `focus` export on the `Button` component.

src/lib/holocene/modal.svelte
```diff
@@ -21,45 +26,82 @@
...
 </script>
 
-<svelte:window on:keyup={handleKeyUp} />
+<svelte:window on:keydown={handleKeyboardNavigation} />
 {#if open}
-  <div class="modal">
+  <div bind:this={modalElement} class="modal">
     <div on:click={cancelModal} class="overlay" />
-    <div class="body" class:large>
+    <div
```
laurakwhit, line 63: We actually want the `bind:this={modalElement}` on this `div` as it has the modal contents and doesn't include the overlay.

laurakwhit (changes requested):

src/lib/holocene/modal.svelte
rossedfort: Ah nice catch. Yeah I had initially thought of a more generic solution like that, but I didn't love the idea of using a `querySelectorAll` to get all the focusable elements, I thought using `bind` with only the elements we wanted focusable was safer. That said, I don't know why we wouldn't want any button in a modal to _not_ be focusable, so I think I like this solution better 👍

src/lib/holocene/modal.svelte
```diff
@@ -21,32 +23,70 @@
...
+      class:large
+      tabindex="-1"
+      role="alertdialog"
+      aria-labelledby="modal-title"
+      aria-describedby="modal-description"
+    >
       {#if !loading}
-        <div class="float-right p-6" on:click={cancelModal}>
+        <button
+          aria-label="Close Modal"
```
laurakwhit, line 74:
```suggestion
          aria-label={cancelText}
```

laurakwhit (approved):
One last little suggestion, but otherwise lgtm :shipit:

PR #970, opened 2022-12-02
forward props for Accordion

rossedfort:
## What was changed

Implement type-safe prop forwarding for `<Accordion />`
- also make Accordions expandable via keyboard

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


src/lib/holocene/accordion.svelte
```diff
@@ -2,65 +2,79 @@
...
-        {#if !readOnly}
-          <Icon
-            name={open ? 'chevron-up' : 'chevron-down'}
-            class={disabled ? 'text-gray-500' : 'text-primary'}
-          />
-        {/if}
+  <button
+    id="{id}-trigger"
+    aria-expanded={open}
+    aria-controls="{id}-content"
```
laurakwhit, line 45: We could use `title` here instead, but there is the risk that it won't be unique. We might want to think about creating a generic util to generate a unique id for use cases such as this.

```suggestion
    aria-controls="{title}-content"
```

Another approach could be to rework this to be a `<li>` inside of a `<ul>` and use the index + title for the id.

Ultimately, I think `aria-expanded` is the most important addition here 😅

laurakwhit (approved):
✨ 🪗 ✨

PR #971, opened 2022-12-05
Clear workflowQuery when no workflowSort and workflowFilters

Alex-Tideman:
## What was changed
When a single filter was applied (i.e. WorkflowType), and then cleared from the filter input and not the advanced search input, the workflowQuery would not get cleared and so when a user clicked on a workflow and then Back to Workflows, it would apply the filter even after being cleared. This would happen only if a single filter was applied. 

This PR fixes that behavior and clears the workflowQuery in the case when no workflow filters or sorts are applied.

### DIFF
--- a/src/lib/pages/workflows-with-new-search.svelte
+++ b/src/lib/pages/workflows-with-new-search.svelte
@@ -56,6 +56,12 @@
     }
   }
 
+  $: {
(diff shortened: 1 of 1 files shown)


rossedfort (approved):

rossedfort:
Just want to make sure this works with batch terminate/cancel too

PR #981, opened 2022-12-07
[DT-190] Reveal all workflow ID

laurakwhit:
## What was changed

Wraps `Workflow ID` instead of truncating it in the table on the `Recent Workflow` page so it is always visible.

## Checklist

1. Closes: `DT-190` 

2. How was this tested:

- [ ] Verify workflow summary with filters is responsive
- [ ] Verify workflow summary with filters and bulk actions is responsive
- [ ] Verify workflow summary (without filters) is responsive

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@temporalio/ui",
-  "version": "2.1.90",
+  "version": "2.1.91",
   "type": "module",
   "description": "Temporal.io UI",
   "keywords": [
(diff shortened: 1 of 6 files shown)


Alex-Tideman:
Looks much better! Only thing I would suggest is giving the workflow ID column more width when the date fields are removed (i think md and lg break points, sm looks good).

Screen Shot 2022-12-08 at 8 09 24 AM

Alex-Tideman:
One more thing, we'll want more width on the 'Select all' column on sm.

Screen Shot 2022-12-08 at 8 16 06 AM

Alex-Tideman:
One last final thing, I think we still want the awesome gradient for the row on hover.

Alex-Tideman (approved):

PR #997, opened 2022-12-14
Event History API Pagination and Keyboard Navigation

Alex-Tideman:
## What was changed
This PR is a major change to how we fetch the event history. Previously, we fetched the entire history and used client-side pagination. This could take a very long time to fetch for large event histories, especially if they had a lot of payloads as well.

This PR introduces API pagination for the event history as well as keyboard navigation to control the pagination. We fetch the start and end of the event history to get input/result, and then then api-pagination component controls fetching the event history pages. The arrow keys jump pages and row for easily navigating between events.

Screen Shot 2022-12-15 at 12 15 25 PM

Screen Shot 2022-12-15 at 12 15 35 PM

## Why?
Drastically improves the performance of fetching the event history as well as better UX for navigating event history.

## Checklist

- [ ] Fix cypress tests with event-history api mocks


src/lib/components/event/event-shortcut-keys.svelte
```diff
@@ -0,0 +1,82 @@
+<script lang="ts">
+  import Shortcut from '$lib/holocene/keyboard-shortcut/shortcut.svelte';
+  export let open = false;
+  export let compact: boolean;
+</script>
+
+{#if open}
+  <div class="modal">
```
rossedfort, line 8: Why not use holocene's modal?

src/lib/components/event/event-summary.svelte
```diff
@@ -1,49 +1,104 @@
...
+<EventShortcutKeys open={showShortcuts} {compact} />
+{#key [$eventFilterSort, category, $refresh]}
+  <ApiPagination
     let:visibleItems
-    let:initialItem
+    let:updating
+    let:activeIndex
+    let:setActiveIndex
+    onFetch={fetchEvents}
+    onError={(error) => console.error(error)}
```
rossedfort, line 80: Is there anything else we want to do with this error?

src/lib/holocene/api-pagination.svelte
```diff
@@ -1,73 +1,127 @@
...
+        break;
+      case 'ArrowDown':
+        if (shifted && onShiftDown) {
+          onShiftDown(event);
+          store.reset();
+          initalDataFetch();
+        } else {
+          store.nextRow();
+        }
+        break;
```
rossedfort, line 96: Saw a nice comment during the demo about using vim keys, i.e. `h`, `j`, `k`, and `l`, maybe worth adding those?

src/lib/holocene/api-pagination.svelte
```diff
@@ -78,100 +132,151 @@
...
           <button
             class="caret"
             disabled={!$store.hasPrevious}
             on:click={store.previousPage}
           >
-            <Icon name="chevron-left" />
+            <span
+              class="arrow arrow-left"
+              class:arrow-left-disabled={!$store.hasPrevious}
+            />
```
rossedfort, line 220: Why move away from the chevrons? Alternatively, could we use the `<Shortcut />` component here?

src/lib/holocene/keyboard-shortcut/arrow-down.svelte
```diff
@@ -0,0 +1,11 @@
+<svg
+  class="h-4 w-4"
+  aria-hidden="true"
+  fill="currentColor"
+  xmlns="http://www.w3.org/2000/svg"
+  viewBox="0 0 320 512"
+  ><path
+    d="M310.6 246.6l-127.1 128C176.4 380.9 168.2 384 160 384s-16.38-3.125-22.63-9.375l-127.1-128C.2244 237.5-2.516 223.7 2.438 211.8S19.07 192 32 192h255.1c12.94 0 24.62 7.781 29.58 19.75S319.8 237.5 310.6 246.6z"
+  /></svg
+>
```
rossedfort, line 10: I think we should add these to our `Icon` component, then this component just becomes

```
<Icon name="keyboard-arrow-down" />
<span class="sr-only">Arrow key down</span>
```
Probably not super likely we need these keyboard arrow icons outside of this context (e.g. without the label), but we might.

src/lib/services/events-service.ts
```diff
@@ -52,11 +52,118 @@ export const fetchRawEvents = async ({
...
+    {
+      request: fetch,
+      params: { maximumPageSize: '20' },
+    },
+  );
+
+  return response.history.events;
+};
+
+export const fetchStartAndEndEvents = async (
```
rossedfort, line 78: Might be worth adding some tests for these more complex API interactions

rossedfort (commented):
Looks awesome, just some general code organization improvements I think we could make.

src/lib/components/event/event-shortcut-keys.svelte
Alex-Tideman: Because the Modal event listeners get overridden by the api-pagination event listeners. (i.e. Close / x doesn't work). Also, doesn't need button. But also... waiting on Ash's designs for this so it will change. Just wanted something out there for now. 

Not sure if 'Space' will be the key to trigger open/close either.

src/lib/components/event/event-summary.svelte
Alex-Tideman: Yes. Totally forgot about this. Good catch.

src/lib/holocene/api-pagination.svelte
Alex-Tideman: Good call, easy to add that.

src/lib/holocene/api-pagination.svelte
Alex-Tideman: I wanted something that looked like the arrow keys. Let me look into using the Shortcut

src/lib/holocene/keyboard-shortcut/arrow-down.svelte
Alex-Tideman: Yup I should do that.

src/lib/services/events-service.ts
```diff
@@ -52,11 +52,118 @@ export const fetchRawEvents = async ({
...
+    namespace,
+    workflowId,
+    runId,
+  });
+
+  const response = await requestFromAPI<GetWorkflowExecutionHistoryResponse>(
+    route,
+    {
+      request: fetch,
+      params: { maximumPageSize: '20' },
```
Alex-Tideman, line 71: Is 20 a good number here? In reality it will return 40ish events (the maximumPageSize is the batch size, the number of events is unknown). We could probably make this 10 or lower

src/lib/components/event/event-shortcut-keys.svelte
rossedfort: Hmm I see. Maybe we do away with `Space`  and just have an Icon button next to the pagination controls?

Screen Shot 2022-12-15 at 11 57 40 AM

As of right now, you'd have to know the 'Space' keyboard shortcut to see the pagination keyboard shortcuts

src/lib/components/event/event-summary.svelte
stevekinney: Hot take, we should make our own `log` function that proxies to `console.log` in development but logs to Sentry or _something_ in production.

src/lib/holocene/api-pagination.svelte
rossedfort: Oh also, I think you can git rid of `shifted` and just use `event.shiftKey` https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/shiftKey

src/lib/components/event/event-summary.svelte
```diff
@@ -1,49 +1,104 @@
...
+    {total}
   >
-    <EventSummaryTable
-      updating={$updating}
-      {compact}
-      on:expandAll={handleExpandChange}
-    >
-      {#each visibleItems as event (`${event.id}-${event.timestamp}`)}
+    <EventSummaryTable {updating} {compact} on:expandAll={handleExpandChange}>
+      {@const events = compact ? groupEvents(visibleItems) : visibleItems}
```
stevekinney, line 88: I am leery of doing stuff like assigning variables in the view layer.

src/lib/components/event/event-summary.svelte
```diff
@@ -1,49 +1,104 @@
...
+      {#each events as event, index (`${event.id}-${event.timestamp}`)}
```
stevekinney, line 89: Should this be something we do on the model when we generate it? Even if we call it like `idWithTimestap` or something like that. Do we only want the `timestamp` to make it unique? If so, maybe we should just use `uuid` instead in the model.

src/lib/components/event/event-summary.svelte
```diff
@@ -1,49 +1,104 @@
...
         <EventSummaryRow
           {event}
-          {groups}
           {compact}
-          expandAll={$expandAllEvents === 'true'}
-          {initialItem}
           {visibleItems}
+          expandAll={$expandAllEvents === 'true'}
```
stevekinney, line 94: My guts days `$expandAllEvents` should be coursing strings to booleans upon assignment and we shouldn't have to do stuff like this in the view.

src/lib/components/event/event-summary.svelte
```diff
@@ -1,49 +1,104 @@
...
+          initialItem={$eventHistory?.start?.[0]}
```
stevekinney, line 95: See all of my other comments about how we should create these values in the `<script>` area (the controller) or in the model.

src/lib/holocene/api-pagination.svelte
```diff
@@ -1,73 +1,127 @@
...
+        onError(error);
+      }
+    } else {
+      store.nextPage();
     }
   }
 
-  $: {
-    onPageChange(nextIndex);
+  async function handleKeydown(event: KeyboardEvent) {
```
stevekinney, line 67: Do we have unit tests for this? I ask because I don't see it exported. It's kind of weird that it's in a `.svelte` file at all. I'd rather see us use dependency injection and pull this out completely. This is kind of like what I said to @rossedfort in I think the signal PR.

src/lib/layouts/workflow-header.svelte
```diff
@@ -81,7 +81,7 @@
   $: cancelInProgress =
     $workflowRun?.workflow?.status === 'Running' &&
     !$updating &&
-    $eventHistory.events.some(
+    $eventHistory.end.some(
```
stevekinney, line 84: This should probably be a method or a helper function and we can test it.

src/lib/layouts/workflow-header.svelte
```diff
@@ -144,7 +144,9 @@
         href={routeForEventHistory({
           ...routeParameters,
         })}
-        amount={workflow?.historyEvents}
+        amount={$eventHistory.end[0]?.id
```
stevekinney, line 147: This check should probably be checked in the component that receives `amount`.

src/lib/components/event/event-shortcut-keys.svelte
Alex-Tideman: Yeah for sure need a prompt or something to tell them how to view it. I'll ask Ash what they have in mind.

src/lib/stores/api-pagination.ts
```diff
@@ -1,140 +1,217 @@
...
 
 export type PaginationStore = PaginationMethods & Readable<PaginationItems>;
 
-export function createPaginationStore(): PaginationStore {
-  const paginationStore = writable(initialStore);
+export function createPaginationStore(
+  pageSizeOptions: string[] = options,
+): PaginationStore {
+  // TODO, use first option in pageSizeOptions for pageSize instead of defaultItemsPerPage
+  const initialPageSize = parseInt(pageSizeOptions?.[0]) ?? defaultItemsPerPage;
```
stevekinney, line 65: What happens if `parseInt` gives us `NaN`?  I know we're all _real_ excited about `??`, but we might be getting carried away.

```js
NaN ?? 'lol' === NaN
NaN || 'lol' === 'lol'
```

But again, my general advice around pulling this logic out into smaller helper functions that we can test applies here.

src/lib/holocene/api-pagination.svelte
Alex-Tideman: Yup that absolutely works and is so much better.

src/lib/stores/api-pagination.ts
```diff
@@ -1,140 +1,217 @@
...
-
-        if ($pagination.items.length < $pageSize) {
-          return $pagination.index * $pageSize + $pagination.items.length;
-        }
-
-        return $pagination.index * $pageSize + $pageSize;
-      };
+      const visibleItems =
+        $pagination.indexData[$pagination.index]?.items ?? [];
+      const indexStart = $pagination.indexData[$pagination.index]?.start ?? 0;
```
stevekinney, line 82: Can we just put these methods on `$pagination`?

src/lib/stores/api-pagination.ts
```diff
@@ -1,140 +1,217 @@
...
-      if (store.nextIndex === 0) _store.hasPrevious = false;
+    const currentIndex = _store.index;
+    const store = {
+      ..._store,
+      indexData: { ..._store.indexData },
+      hasNext: Boolean(nextToken),
+      updating: false,
+      loading: false,
+    };
+    // Return early if page does not have any items
```
stevekinney, line 112: We can probably remove this comment.

src/lib/components/event/event-date-filter.svelte
```diff
@@ -58,7 +60,7 @@
   };
 
   $: value =
-    $eventSortOrder === 'descending' &&
+    (compact ? true : $eventSortOrder === 'descending') &&
```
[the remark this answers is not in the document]
Alex-Tideman: 100% agree with this and all below comments. I'll make those changes. Appreciate the attention to detail.

PR #1007, opened 2022-12-16
[DT-139] Add link to child workflows for StartChildWorkflowExecutionInitiated

laurakwhit:
## What was changed

* ~Adds link on `Workflow ID` in `StartChildWorkflowExecutionInitiated` event details~ (see https://github.com/temporalio/ui/pull/1007#discussion_r1061979857)
* Renames `Workflow Execution Workflow ID` and `Workflow Execution Run ID` in` ChildWorkflowExecutionStarted` event details to `Workflow ID` and `Run ID`.

|Before | After |
|--|--|
|Screenshot 2022-12-16 at 12 17 27 PM|Screenshot 2022-12-16 at 12 18 02 PM|
|Screenshot 2022-12-16 at 12 17 04 PM|Screenshot 2022-12-16 at 12 13 05 PM|

## Checklist

1. Closes: `DT-139` 

2. How was this tested:

- [x] Verify `Workflow Execution Workflow ID` and `Workflow Execution Run ID` are now `Workflow ID` and `Run ID`
- [ ] ~Verify there is a link on the `Workflow ID` in `StartChildWorkflowExecutionInitiated` event details~
- [ ] ~Verify `Child Workflow Execution Run ID` does not appear in `StartChildWorkflowExecutionInitiated` event details~


Alex-Tideman:
Let's add some unit tests but otherwise looks good. I think down the road we'll want to make a more generic approach to getting details of an event based on another event.

src/lib/components/event/event-summary-row.svelte
```diff
@@ -58,8 +58,28 @@
       });
       timeDiffChange = timeDiff ? `(${descending ? '-' : '+'}${timeDiff})` : '';
     }
+
```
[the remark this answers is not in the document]
laurakwhit: Updated ✅

src/lib/utilities/get-single-attribute-for-event.ts
```diff
@@ -188,3 +201,38 @@ export const getSingleAttributeForEvent = (
 
   return getSummaryAttribute(event);
 };
+
+export const checkForChildWorkflowExecutionAndAddRunIdAttribute = ({
```
laurakwhit, line 205: A few things to consider...
1) The util `checkForChildWorkflowExecutionAndAddRunIdAttribute` mutates the `currentEvent`. This is the same behavior as before, but now I'm wondering if we want to make a copy and return the `currentEvent` from the util instead 🤔 
2) There are instances where the `StartChildWorkflowExecutionInitiated` event is not immediately followed by a `ChildWorkflowExecutionStarted` event, which means that this logic doesn't work and the `Workflow Id` will not have a link. The logic gets increasingly complex if I try to iterate through events to account for this, but I think it might be possible. Maybe by looking for a `ChildWorkflowExecutionStarted`  event with a matching workflow id? Curious if you think it's worth investigating...

laurakwhit: **Update:** We have decided not to add the child execution link to the `Workflow ID` on the `StartChildWorkflowExecutionInitiated` event. 

The `StartChildWorkflowExecutionInitiated` event does not have the run ID needed to create the link. The run ID is created and added to the workflow execution in the following `ChildWorkflowExecutionStarted` event. 

Rather than loop through following events (which may not all be fetched in the UI) to find the `ChildWorkflowExecutionStarted` event with the run ID needed for the link on the `StartChildWorkflowExecutionInitiated` event, we have determined it just makes sense to only have a link when the run ID exists for that event.

Alex-Tideman (approved):

PR #1028, opened 2023-01-06
minor changes made to links

DJSanti:
## What was changed
Several legacy links were updated to point to the correct documentation.

## Why?
Docs team recently made significant changes to content. Updating links fixes what may have broke due to being moved or erased.


src/lib/pages/workflow-stack-trace.svelte
```diff
@@ -81,7 +81,7 @@
       {#if workflow.isRunning && workers?.pollers?.length === 0}
         <p>
           To enable <Link
-            href="https://docs.temporal.io/concepts/what-is-a-query/#stack-trace-query"
+            href="https://docs.temporal.io/workflows#query"
```
flossypurse, line 84: There is actually a `#stack-trace-query` anchor
```suggestion
            href="https://docs.temporal.io/workflows#stack-trace-query"
```

flossypurse (approved):

PR #1031, opened 2023-01-09
API Pagination fixes

Alex-Tideman:
## What was changed
1. Add back `empty` slot to support custom no results pages and default 'No items' text if no visibleItems
2. Trigger resetPageSize and initialPageFetch when the page size changes.

## Why?
Support custom no items UI and fix bug when picking a new page size in the dropdown doesn't trigger a refetch with the page size.

### DIFF
--- a/src/lib/holocene/api-pagination.svelte
+++ b/src/lib/holocene/api-pagination.svelte
@@ -1,4 +1,5 @@
 <script lang="ts">
+  import { page } from '$app/stores';
   import Alert from '$lib/holocene/alert.svelte';
   import FilterSelect from '$lib/holocene/select/filter-select.svelte';
   import SkeletonTable from '$lib/holocene/skeleton/table.svelte';
@@ -30,9 +31,20 @@
     if (error) error = undefined;
   }
 
+  $: isEmpty = $store.visibleItems.length === 0 && !$store.loading;
+  $: pageSizeChange =
+    !$store.loading && $store.pageSize !== $store.previousPageSize;
+
   onMount(() => {
     initalDataFetch();
   });
+
+  $: {
+    if (pageSizeChange) {
+      store.resetPageSize($store.pageSize);
+      initalDataFetch();
+    }
+  }
 
   async function initalDataFetch() {
     const fetchData: PaginatedRequest = await onFetch();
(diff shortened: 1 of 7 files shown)


rossedfort (approved):

src/lib/holocene/api-pagination.svelte
```diff
@@ -30,10 +30,21 @@
     if (error) error = undefined;
   }
 
+  $: isEmpty = $store.visibleItems.length === 0 && !$store.loading;
+  $: pageSizeChange =
```
laurakwhit, line 34: Running into a few things:
Screenshot 2023-01-09 at 10 29 51 AM

1. On a running workflow the number of `visibleItems` is increasing, but the total remains the same and it gets into a weird state such as in the screenshot above where there are 19 events, but it still says the total is 13.
2. If I pass in `pageSizeOptions` and change the number it will trigger a refetch, but it always just fetches **all** of the events. Do we want to update `fetchPaginatedEvents` to account for a page size? I realize it's probably fine to just have the default 100 for now, but not sure if that will change in the future.
3. There is an unused `supportsReverseOrder` import in `events-service.ts`.

laurakwhit (commented):
Changes look good! Just a few other things I noticed that might be worth a look.

stevekinney (approved):

src/lib/holocene/api-pagination.svelte
Alex-Tideman: Great finds!

1. Because we have to fetch the start and end events for input/result/total event count plus the first page of events, there can be a mismatch on the total count due to timing. I.e start/end call resolves with 13 events, first page resolves with 19 events. I'll see if we can do something about that, but in a separate PR.
2. Because the event history can never guarantee exactly how many events can be returned (it's batched), we don't use a pageSize option in getPaginatedEvents (_pageSize = 100 is there to appease the TS gods, but it's not actually used in the requestFromAPI).
3. Removed

laurakwhit: > Because the event history can never guarantee exactly how many events can be returned 

Ah I forgot about this fun quirk 🤦‍♀️

PR #1033, opened 2023-01-09
Use $lib/holocene instead of $holocene

Alex-Tideman:
## What was changed
Update references of $holocene to $lib/holocene if it's a holocene file.

## Why?
Fix importing of $holocene alias for npm package


rossedfort (approved):
Maybe we can dig into why it's not working another time, but I don't want to hold up merging the upgrade 👍

Alex-Tideman:
For sure, I'm going to beta publish this first and see if it actually fixes the problem before merging

src/lib/holocene/accordion.svelte
```diff
@@ -1,6 +1,6 @@
 <script lang="ts">
-  import Icon from '$holocene/icon/icon.svelte';
-  import Badge from '$holocene/badge.svelte';
+  import Icon from '$lib/holocene/icon/icon.svelte';
```
laurakwhit, line 2: Have we looked into using [this](https://kit.svelte.dev/docs/configuration#alias) instead? Seems like it would create both the vite and tsconfig settings, but unsure if those settings are the issue here.

Alex-Tideman: Yeah we have alias config in both ui/cloud-ui.

Alex-Tideman:
So I removed all the $holocene alias references and that fixes the imports. I'm not heartbroken about removing this for now since it's just saving a few keystrokes ($lib/holocene -> $holocene) and the auto-importer usually does this for now. Once we get this in, we can experiment with a single instance of an alias and figure out the right config for it.

src/lib/holocene/accordion.svelte
laurakwhit: But in `vite.config.ts` and not [svelte.config.ts](https://kit.svelte.dev/docs/configuration#alias). Was wondering if there was something with the `tsconfig` that could be magically fixed by adding to `svelte.config.ts` instead.

PR #1038, opened 2023-01-10
Remove API Pagination for Event History

Alex-Tideman:
## What was changed

- Remove API pagination for event history and load full history in background.
- Add keyboard navigation to pagination
- Remove supportsReverseOrder check (we assume now everyone is on 1.16+)

## Why?
Because we need all the events in the event history for compact view and for visualizations, load the start and end events to show input/results and initial events in table while the full event history loads in background.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


rossedfort (dismissed):

rossedfort (approved):

src/lib/holocene/pagination.svelte
```diff
@@ -76,9 +77,42 @@
   };
 
   $: floatStyle = getFloatStyle({ width, height, screenWidth });
+
+  async function handleKeydown(event: KeyboardEvent) {
```
laurakwhit, line 81: One odd edge case is if you select the dropdown for page size, select a size, and then attempt to use the row shortcuts it goes up/down a row and then opens up the dropdown again 🔽 

Screenshot 2023-01-12 at 1 10 58 PM

src/lib/layouts/workflow-history-layout.svelte
```diff
@@ -17,48 +21,73 @@
...
+      workflowId,
+      runId,
+      settings,
+      accessToken: user?.accessToken,
+    });
+    $eventHistory = events;
+  };
+
+  const onViewClick = (view: EventView) => {
+    if ($page.url.searchParams.get('page')) {
```
laurakwhit, line 55: I think we might want to do something similar for page size. Currently when you select a new page size the page remains the same and you can end up in a weird state (specifically in the `Compact` view) 🔽 

Screenshot 2023-01-12 at 1 14 22 PM

or with an incorrect url in the `History` view.

src/lib/pages/workflow-history-feed.svelte
```diff
@@ -2,10 +2,40 @@
...
+  import { page } from '$app/stores';
+  import { fetchAllEvents } from '$lib/services/events-service';
+  import { eventFilterSort, type EventSortOrder } from '$lib/stores/event-view';
+  import { refresh } from '$lib/stores/workflow-run';
+
+  let events: CommonHistoryEvent[] = [];
+
+  $: namespace = $page.params.namespace;
+  $: workflowId = $page.params.workflow;
+  $: runId = $page.params.run;
```
laurakwhit, line 14: There are couple places where we could do this instead 🔽 
```suggestion
  $: ({ namespace, workflow: workflowId, run: runId } = $page.params);
```

laurakwhit (commented):
Haven't looked too closely at all of the code, but I have been trying to break things 😅 
Found a few things that might be worth looking into...

src/lib/holocene/pagination.svelte
Alex-Tideman: Good find! That is weird

PR #1048, opened 2023-01-17
DT-186 - set up service for completing and failing an activity

rossedfort:
## What was changed

Add a service and functions for completing and failing an activity
- depends on https://github.com/temporalio/ui-server/pull/271
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


src/types/index.ts
```diff
@@ -27,6 +27,14 @@ export type PatchScheduleRequest =
   temporal.api.workflowservice.v1.IPatchScheduleRequest;
 export type UpdateScheduleRequest =
   temporal.api.workflowservice.v1.IUpdateScheduleRequest;
+export type FailActivityTaskRequest =
```
stevekinney, line 30: This file is auto-generated, so we might want to put this in a `.d.ts` file elsewhere.

Alex-Tideman (approved):

PR #1050, opened 2023-01-18
Include credentials with the data codec decode request

christopherb-stripe:
## What was changed

The [`{credentials: 'include'}`](https://developer.mozilla.org/en-US/docs/Web/API/fetch#credentials) is added to the `fetch` request which performs the data conversions.

## Why?

Slack message: https://temporalio.slack.com/archives/C011AAP3GFP/p1671643822741769

We don't use the authorization header to authenticate users internally but instead use a browser cookie. Without this option the `POST` request doesn't contain this cookie (as the request is cross-origin) and so is rejected before it can reach the data converter.

## Checklist

_How was this tested?_

I've added a unit test for this but admittedly was not able to test this end-to-end due to limitations that mean that it's difficult for us to deploy a Docker image that isn't from one of your releases.

_Any docs updates needed?_

I don't believe so.

### DIFF
--- a/src/lib/services/data-encoder.ts
+++ b/src/lib/services/data-encoder.ts
@@ -39,6 +39,7 @@ export async function convertPayloadsWithCodec({
   const encoderResponse: Promise<Payloads> = fetch(endpoint + '/decode', {
     headers,
     method: 'POST',
+    credentials: 'include',
     body: stringifyWithBigInt(payloads),
   })
     .then((r) => r.json())
--- a/src/lib/utilities/decode-payload.test.ts
+++ b/src/lib/utilities/decode-payload.test.ts
@@ -276,6 +276,28 @@ describe('convertPayloadToJsonWithCodec', () => {
     const dataConverterStatus = get(lastDataEncoderStatus);
     expect(dataConverterStatus).toEqual('notRequested');
   });
+  it('Should include credentials with the request for cookie based authentication of data converters', async () => {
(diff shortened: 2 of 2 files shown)


christopherb-stripe:
I think Stripe may already have a CLA signed for all employees but I'll check tomorrow.

stevekinney:
> I think Stripe may already have a CLA signed for all employee contributions but I'll check tomorrow.

Cool. Let us know. You might have to just click and sign because technology is hard. But, if that's not a thing you feel comfortable doing, I'll talk to our security team figure out how to move forward.

src/lib/services/data-encoder.ts
```diff
@@ -39,6 +39,7 @@ export async function convertPayloadsWithCodec({
   const encoderResponse: Promise<Payloads> = fetch(endpoint + '/decode', {
     headers,
     method: 'POST',
+    credentials: 'include',
```
stevekinney, line 42: @robholland and/or @feedmeapples You both have a better sense of all of the use cases. Should we do this by default or make it an environment variable.

robholland: I believe this is fine without configuration. Unlike the security token setup credentials would only be sent to domain names which had set credentials, so this won't increase risk imho.

christopherb-stripe:
Added a new commit which fixes the formatting of the test file.

robholland (approved):

stevekinney (approved):

PR #1051, opened 2023-01-18
Event History and Workflow Run store refactor

Alex-Tideman:
## What was changed
Simplify the approach to the workflowRun and eventHistory stores, remove complexity of listening to parameter changes and only fetch in layout. Move workflow-run-layout to the top level +layout.svelte file so all routes under /run use the layout and only fetch the workflow once.

## Why?
Simplifies the code, removes duplicate layouts and refetching data on every tab change.

## Checklist

- [x] Figure out client action prop passing

1. Closes 

2. How was this tested:

3. Any docs updates needed?


rossedfort (approved):
💯

PR #1061, opened 2023-01-19
Fix lint issue in handleErrors; add type casting to `has`

stevekinney:
- Uses `unknown` instead of any in `handleErrors`.
- `has` now lets TypeScript know that an object has a property that it just confirmed existed.

### DIFF
--- a/src/lib/utilities/handle-error.ts
+++ b/src/lib/utilities/handle-error.ts
@@ -70,11 +70,11 @@ const hasStatusCode = (
   statusCode: number | string,
 ): error is TemporalAPIError => {
   if (has(error, 'statusCode')) {
-    return error?.statusCode === statusCode;
+    return error.statusCode === statusCode;
   }
 
   if (has(error, 'status')) {
-    return error?.status === statusCode;
+    return error.status === statusCode;
   }
 
   return false;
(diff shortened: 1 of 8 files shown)


src/lib/utilities/handle-error.ts
```diff
@@ -2,24 +2,32 @@ import { browser } from '$app/environment';
...
 
-// This will eventually be expanded on.
 export const handleError = (
-  error: any,
+  error: unknown,
   toasts = toaster,
   errors = networkError,
   isBrowser = browser,
 ): void => {
+  console.log({ error });
```
laurakwhit, line 16: So we want this 🪵 here?

rossedfort, line 16: 🚮

stevekinney: Nope

src/lib/utilities/handle-error.ts
```diff
@@ -58,10 +57,25 @@ export const handleUnauthorizedOrForbiddenError = (
...
 };
 
-export const isForbidden = (error: any): boolean => {
-  return error?.statusCode === 403 || error?.status === 403;
+const hasStatusCode = (
+  error: unknown,
+  statusCode: number | string,
+): error is TemporalAPIError => {
+  if (has(error, 'statusCode')) {
+    return error?.statusCode === statusCode;
```
rossedfort, line 73: Slight nit: this (and the one below in L77) shouldn't be needed if we're already checking if 'statusCode' is in error, right?

rossedfort (approved):

PR #1065, opened 2023-01-20
DT-233 and DT-237 - fix keyboard a11y warnings in holocene

rossedfort:
## What was changed

- in places where we had an accessibility warning, a keyup event listener was added which either calls a `click` handler (if provided), dispatches a click event, or simply captures the event when it's unused.
- spruce up vertical nav so it's navigable via keyboard, also fixes focus borders of nav rows
- remove hamburger header and navigation header
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


src/lib/holocene/navigation/drawer.svelte
```diff
@@ -28,8 +29,10 @@
   on:click-outside={onClose}
 >
   <div class="relative">
-    <div class="absolute right-0 top-0 cursor-pointer" on:click={onClose}>
-      <Icon name="close" />
+    <div class="absolute right-0 top-0">
+      <IconButton on:click={onClose}>
```
laurakwhit, line 33: Could this be shortened to ⬇️ ?
```suggestion
      <IconButton on:click={onClose} icon="close" />
```

src/lib/holocene/data-encoder-status.svelte
```diff
@@ -16,46 +16,58 @@
   {#if $dataEncoder?.hasNotRequested}
     <Tooltip right text={'Data encoder is configured'}>
```
laurakwhit, line 17: If we use `NavTooltip` here instead then it will follow the same hide/show behavior as the other navigation links.

```suggestion
    <NavTooltip right text={'Data encoder is configured'}>
```

src/lib/holocene/navigation/nav-row.svelte
```diff
@@ -4,28 +4,34 @@
...
-    </a>
-  {:else}
+{#if handleClick !== null}
+  <button on:click={handleClick} class="{classes} text-left">
     <slot />
-  {/if}
-</li>
+  </button>
+{:else}
+  <li
```
laurakwhit, line 16: What do you think about moving the `button` inside the `li` element here?
```suggestion
  <li
  class={link ? '' : classes}
  class:noFilter
  class:wrap
  data-cy={$$props['data-cy']}
>
  {#if handleClick !== null}
    <button on:click={handleClick} class="{classes} text-left">
      <slot />
    </button>
  {:else if link && externalLink}
    <a href={link} rel="noreferrer" target="_blank" class={classes}>
      <slot />
    </a>
  {:else if link}
    <a href={link} class={classes}>
      <slot />
    </a>
  {:else}
    <slot />
  {/if}
</li>
```

laurakwhit (commented):
So many nice little improvements 🎉

src/lib/holocene/data-encoder-status.svelte
rossedfort: I did try this and it introduced some very strange animation behavior that I couldn't pin down. Additionally, I thought that it might be nice to show the tooltip regardless of the nav state because it provides additional context on whether data-encoder is configured.

src/lib/holocene/navigation/nav-row.svelte
rossedfort: Similar to using NavTooltip in data-encoder-status I think this caused some unintended behavior with the nav animation. I'll double check though, there might be an easy fix 👍

rossedfort: This works great, I'll make this change 🎉

PR #1069, opened 2023-01-23
[DT-231] Ensure all page headings use semantic markup

laurakwhit:
## What was changed

* Updated heading levels on various pages
* Made the pending activities details a list
* Refactored the workers list to user the `Table` component
* Added optional title to icons

## Why?

To ensure page headings (and other elements) use semantic markup and are navigable via an AT.

## Checklist

1. Closes: `DT-231` 

2. How was this tested:

- [ ] Verify page headings still appear as expected
- [ ] Verify the details list under workflow > pending activities appears as expected
- [ ] Verify the workers list under workflow > workers appears as expected


src/lib/components/workers-list.svelte
```diff
@@ -5,57 +5,66 @@
...
           {#if poller.taskQueueTypes.includes('WORKFLOW')}
-            <Icon name="checkmark" class="text-blue-700" />
+            <Icon
+              name="checkmark"
+              class="m-auto text-blue-700"
+              title="exists"
+            />
           {:else}
-            <Icon name="close" class="text-primary" />
+            <Icon name="close" class="m-auto text-primary" title="none" />
```
laurakwhit, line 50: Are there better/more descriptive titles we could use here besides `none` and `exists`?

stevekinney (approved):

src/lib/components/workers-list.svelte
laurakwhit: **Update:** I've just gone with `yes` and `no`.

PR #1080, opened 2023-01-25
Fix First on Relationships

laurakwhit:
## What was changed

Account for various event types and use the combined attributes to get the `first`, `previous`, and `next` run IDs.

## Why?

E.g. the `previous` event might not be `WorkflowExecutionStarted` and therefore `workflowExecutionStartedEventAttributes` won't work.

## Checklist

1. Closes: n/a 

2. How was this tested:

- [ ] Verify `first`, `previous`, and `next` show up as expected under `Workflow details`  > `Relationships`

### DIFF
--- a/src/lib/components/workflow/workflow-relationships.svelte
+++ b/src/lib/components/workflow/workflow-relationships.svelte
@@ -2,38 +2,27 @@
   import { page } from '$app/stores';
   import { routeForEventHistory } from '$lib/utilities/route-for';
   import { workflowRun } from '$lib/stores/workflow-run';
-  import { eventHistory } from '$lib/stores/events';
 
   import Accordion from '$lib/holocene/accordion.svelte';
   import Badge from '$lib/holocene/badge.svelte';
   import ChildWorkflowsTable from '$lib/components/workflow/child-workflows-table.svelte';
   import WorkflowDetail from '$lib/components/workflow/workflow-detail.svelte';
-  import type { CombinedAttributes } from '$lib/utilities/format-event-attributes';
 
-  type WorkflowEventWithAttributes = WorkflowEvent & {
-    attributes: CombinedAttributes;
-  };
+  export let hasChildren: boolean;
+  export let hasRelationships: boolean;
+  export let first: string;
+  export let parent: WorkflowIdentifier;
+  export let next: string;
+  export let previous: string;
 
-  $: children = $workflowRun.workflow?.pendingChildren.length;
-  $: parent = $workflowRun.workflow?.parent;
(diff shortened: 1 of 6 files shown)


src/lib/components/workflow/workflow-relationships.svelte
```diff
@@ -8,23 +8,23 @@
   import Badge from '$lib/holocene/badge.svelte';
   import ChildWorkflowsTable from '$lib/components/workflow/child-workflows-table.svelte';
   import WorkflowDetail from '$lib/components/workflow/workflow-detail.svelte';
+  import type { CombinedAttributes } from '$lib/utilities/format-event-attributes';
+
+  type WorkflowEventWithAttributes = WorkflowEvent & {
```
laurakwhit, line 13: Would the better solution be to update `attributes` on `WorkflowEvent` in `events.d.ts`? I'm not sure `EventAttribute` is always accurate 🤔

rossedfort: The workflow event types are tricky, I'm not sure I have the expertise to recommend the best approach, but I do think we should avoid casting those `lastEvent` and `firstEvent` reactive variables if possible. We may be able to use the `EventWithAttributes` type and/or use some of the [type guards](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) in `is-event-type.ts`. @stevekinney would prob have the best advice here 🙂

rossedfort: I'm thinking something like:

```svelte
let firstEvent: unknown
$: firstEvent = $eventHistory.start[0];
if (isWorkflowExecutionStartedEvent(firstEvent)) {
  // do the thing with firstEvent types as `WorkflowExecutionStartedEvent`
}
```

Idk if that's actually helpful but just some thoughts 😅

stevekinney: Yea, that casting is opting out of a lot of type safety. Also: Do you want the _first_ event or do you want the `WorkflowExecutionStarted` event? (Yes, I get that the first event probably is the `WorkflowExecutionStarted`, but that's more of a coincidence than your intention in the code.

stevekinney: This is alway _way_ too much logic to live in the view layer.

rossedfort (approved):
Looks great 💯  couldn't hurt to throw a few unit tests on `get-workflow-relationships` if you're up for it

laurakwhit:
@rossedfort good idea! Will add a follow up task to add some tests 📝

PR #1082, opened 2023-01-26
Break out utility function for finding relationships

stevekinney:


src/lib/components/workflow/workflow-relationships.svelte
```diff
@@ -2,38 +2,27 @@
...
   $: ({ workflow, namespace } = $page.params);
 </script>
 
 <Accordion title="Relationships" icon="relationship">
   <div slot="summary" class="hidden flex-row gap-2 lg:flex">
     <Badge type={parent ? 'purple' : 'gray'}>{parent ? 1 : 0} Parent</Badge>
-    <Badge type={children ? 'purple' : 'gray'}
-      >{children} Pending Children</Badge
+    <Badge type={hasChildren ? 'purple' : 'gray'}
+      >{hasChildren} Pending Children</Badge
```
laurakwhit, line 25: Maybe we want a `numOfChildren` variable instead? Otherwise we should update this to 👇 
```suggestion
      >{$workflowRun.workflow.pendingChildren.length} Pending Children</Badge
```
so we have a number here 

Screenshot 2023-01-26 at 2 37 34 PM

PR #1083, opened 2023-01-26
use adapter-static for local builds

### LINKED ISSUE #1081: UI fails to load when navigating directly to a page more than one directory deep
**Describe the bug**

As far as I can tell, starting with 2.10.0, assets in the returned HTML are requested from `./_app` instead of `/_app`. When a user navigates directly to a route that is more than one directory deep, the assets fail to load because the browser requests them relative to the page URL.

For instance, if a user clicks an external link like `http://localhost:8080/namespaces/default/workflows` or refreshes the browser while on this page, the browser tries to request assets from `http://localhost:8080/namespaces/default/_app/immutable/start-ca935cf5.js` instead of `http://localhost:8080/_app/immutable/start-ca935cf5.js`, and the page never loads.

**To Reproduce**

1. Start a local cluster with `docker-compose up`
2. Navigate to http://localhost:8080/, which the app will rewrite to http://localhost:8080/namespaces/default/workflows
3. Refresh the page
4. The page fails to load

rossedfort:
## What was changed

- Use @sveltejs/adapter-static for local builds to fix an issue  with relative paths in index.html that we not resolved properly when hosted via ui-server
## Why?

Attempt to fix an issue reported by customers
## Checklist

1. Closes 
closes #1081

2. How was this tested:

3. Any docs updates needed?


laurakwhit (approved):

rossedfort:
> I would leave build dir and move it to `build` for both adapters

This is a good idea for the record, but I can't get it to work right now, so will make this change in a follow up PR.

PR #1087, opened 2023-01-27
[DT-232] Add keyboard events to elements with click events in core UI

laurakwhit:
## What was changed

* Changed `div` to `button` if there was an `on:click`
* Conditionally render `Copyable` as a button if clickable
* Fixed styling on the Select Namespace page
* Updated the namespace list links to be the entire clickable list item

## Why?

Removes all non-interactive elements with an `on:click`.

## Checklist

1. Closes: `DT-232` 

2. How was this tested:

- [ ] Verify the following show up as expected in the UI
   - `Copyable` components
   - `event-category-filter.svelte`
   - `event-date-filter.svelte`
   - `event-history-timeline-container.svelte`
   - `date-time-format.svelte`
   - `namespace-list.svelte`
   - `schedule-advanced-settings.svelte`
   - `dropdown-filter/conditional.svelte`
   - `dropdown-filter/sort.svelte`
   - `workflows-summary-row-with-filters.svelte`
   - `workflows-summary-row.svelte
   - `Select Namespace` page


src/lib/components/copyable.svelte
```diff
@@ -11,20 +10,37 @@
...
-    >
-  </slot>
-  <button on:click={(e) => copy(e, content)}>
-    <Icon
-      name={$copied ? 'checkmark' : 'copy'}
-      stroke={color}
-      class={`${visible ? 'visible' : 'invisible group-hover:visible'}`}
-    />
+{#if clickAllToCopy}
+  <button
```
laurakwhit, line 14: Might there be a way we could conditionally add an `on:click`? 

If so, instead of this conditional with repeated code, we could so something like `<svelete:element this={clickAllToCopy ? 'button' : 'div'} />` ✨

src/lib/components/filters/date-time-format.svelte
```diff
@@ -5,7 +5,7 @@
   import {
```
laurakwhit, line 5: Is this file even needed anymore?

src/lib/components/workflow/dropdown-filter/conditional.svelte
```diff
@@ -3,22 +3,22 @@
 </script>
```
laurakwhit, line 3: Also, this file. Can it be removed entirely?

src/lib/components/copyable.svelte
stevekinney: We should do that… but maybe as a different PR?

src/routes/(app)/select-namespace/+page.svelte
```diff
@@ -73,8 +80,20 @@
 </div>
 
 <style lang="postcss">
+  .search {
+    @apply mb-5 flex rounded-full border-[3px] border-gray-900 bg-white p-1 pr-4;
```
stevekinney, line 84: I don't love the hard-coded values here. Should we just add to the values in Tailwind? That said, it's not a blocker at this moment.

laurakwhit: Extended our `tailwind.config` to include 
```js
      borderWidth: {
        '3': '3px',
      },
```

PR #1089, opened 2023-01-30
Merge ui-server into ui repo

robholland:
## What was changed
Merged the ui-server application code into this repo. Existing ui code is shifted down a directory into `ui`.

## Why?
To make it more obvious to users where the code for the UI lives and simplify build pipelines.

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

NOTES:

1. This must be released as 2.11.x, skipping quite a few versions from the current UI version. This is to ensure a sane versioning setup for consumers of the ui-server go module and the docker builds which were previously using ui-servers versioning (currently 2.10.x). The release notes should explain the jump.

TODO:

- [ ] Figure out what the README situation should be
- [x] Make sure docker secrets are set for ui repo as they were for ui-server
- [x] Carefully check github actions with team, ensure docker builds and releases have the versioning we expect
- [ ] Confirm the github actions I have not ported over from ui-server are not required

Note: Vercel will fail to deploy this because I have moved the ui files down a directory. I can't adjust the vercel config without breaking deploys for main, so it'll have to fail on this PR until we merge this and adjust vercel config.


stevekinney:
Do the following need to be in the top level?

- `docker`
- `proto`
- `plugins`
- `third_party`

It feels like all of the server-related stuff should live in the self-contained `/server` directory?

Especially especially since stuff like `package.json` got bumped down a level.

robholland:
I appreciate the desire, especially given we're merging into the ui repo rather than ui-server, but I think we'd be fighting uphill to do that. The go server embeds the ui rather than the other way around. I can try having `server` and `ui` top level directories with everything contained within them if you'd strongly prefer that.

robholland:
Note that the `server` directory here isn't a top-level container for all the ui-server code, but an go organisational folder as per the current `ui-server` code, any top level folders in this PR are because that's how it exists in the `ui-server` repo.

robholland:
Closing in favour of #1112

PR #1091, opened 2023-01-30
Fix branch names for e2e tests.

robholland:
## What was changed
Fixed the branch names for which the e2e integration tests workflow should trigger.

## Why?
It was still set the branch name for the PR which added the test. It needs to run on main.


robholland:
Fixed in #1096

PR #1101, opened 2023-02-01
Get namespace webUrl from page data

Alex-Tideman:
## What was changed
Get the namespace webUrl from page data instead of the cached GetNamespaces(). Removes need to await apiRoute since the webUrl will be in load function in top level layout so will be available in the page data.

Also as part of this, moved an Enum out of $types to models to fix linking ui.

## Why?
Pass webUrl to core.


src/lib/models/workflow-actions.ts
```diff
@@ -0,0 +1,5 @@
+export enum ResetType {
+  FirstWorkflowTask = 0,
+  LastWorkflowTask = 1,
+  EventId = 2,
+}
```
GiantRobots, line 5: Okay, but I'm just saying Enums are still the worst :P

src/lib/utilities/route-for-api.ts
```diff
@@ -11,14 +13,15 @@ const replaceNamespaceInApiUrl = (
...
-  if (globalThis?.GetNamespaces && namespace) {
-    const namespaces = await globalThis.GetNamespaces();
-    const configNamespace = namespaces?.find((n) => n.namespace === namespace);
+  // Use GetNamespace webUrl or fallback env var
+  const apiUrl = webUrl || globalThis?.AppConfig?.apiUrl;
+  if (apiUrl && namespace) {
     baseUrl =
-      configNamespace?.webUri ??
+      webUrl ||
       replaceNamespaceInApiUrl(globalThis?.AppConfig?.apiUrl, namespace);
```
GiantRobots, line 25: Some kind of logging here would probably be a good idea so we can indicate when the fallback api is being used.

GiantRobots (approved):

src/lib/utilities/route-for-api.ts
Alex-Tideman: 👍 Made this logic easier to grok and added a console.warn if using the fallback

PR #1109, opened 2023-02-03
Preserve 'constructor' keyword in json-bigint and add console.warn on decode error

Alex-Tideman:
## What was changed
If payload objects included the key `constructor`, parsing would fail. Preserve that keyword and include unit test. Add console.warn on failed parsing.

Tested with supplied payloads, successfully parsed.

### DIFF
--- a/src/lib/utilities/decode-payload.test.ts
+++ b/src/lib/utilities/decode-payload.test.ts
@@ -83,7 +83,17 @@ const JsonObjectEncoded = {
   data: 'eyAiVHJhbnNmb3JtZXIiOiAiT3B0aW11c1ByaW1lIiB9',
 };
 
+const JsonObjectEncodedWithConstructor = {
+  metadata: {
+    encoding: 'anNvbi9wbGFpbg==',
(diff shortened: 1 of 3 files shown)


src/lib/utilities/decode-payload.test.ts
```diff
@@ -106,6 +116,9 @@ describe('decodePayload', () => {
   it('Should decode a json payload with encoding json/plain', () => {
     expect(decodePayload(JsonObjectEncoded)).toEqual(JsonObjectDecoded);
   });
+  it('Should decode a json payload with constructor keyworkd with encoding json/plain', () => {
```
rossedfort, line 119: Should this be `json/protobuf` ? I thought that's the specific encoding that caused the bug. Or was it the same for all encodings?

rossedfort (approved):

src/lib/utilities/decode-payload.test.ts
Alex-Tideman: We have tests for json/protobuf. We decode anything that is json/....

PR #1112, opened 2023-02-03
Import ui-server.

robholland:
## What was changed
Merged `ui-server` code into the `ui` repo, including some refactorings.

## Why?
To make it easier for contributors and simplify builds and actions.

## TODO

1. [ ] Adjust server ui and server README as required
2. [x] Find a way to test the docker cloud build action without pushing potentially broken images
3. [x] Investigate whether server directory needs to be excluded for npm package
4. [x] Ensure assets are present for go import of the server module (for cli/temporalite)

## Checklist

1. Closes 

5. How was this tested:

6. Any docs updates needed?


server/.gitignore
```diff
@@ -0,0 +1,4 @@
+ui-server
+api
+openapi/assets/temporal
+ui/assets
```
feedmeapples, line 4: if UI assets are ignored will we still be able to serve the UI when /server is consumed as a module? e.g from Temporal CLI

Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+RUN npm install -g pnpm
+
+COPY package.json pnpm-lock.yaml ./
+RUN pnpm install
+
+COPY .babelrc .npmrc *.json *.yaml *.cjs *.js *.ts ./
+COPY src src
+
+ARG TEMPORAL_PUBLIC_PATH
+ARG UI_TARGET=local
```
feedmeapples, line 19: for consistency should we rename the arg TEMPORAL_PUBLIC_PATH to `UI_PUBLIC_PATH`?

server/config/base.yaml
```diff
@@ -0,0 +1,2 @@
+temporalGrpcAddress: 127.0.0.1:7233
+port: 8080
```
feedmeapples, line 2: since the UI assets now target :8233 in local dev mode let's also change this port number to 8233 so we don't have to modify each time we want to run the ui server from code

feedmeapples (approved):

stevekinney (approved):

PR #1113, opened 2023-02-03
Revert "Include credentials with the data codec decode request"

rossedfort:
Reverts temporalio/ui#1050

### DIFF
--- a/src/lib/services/data-encoder.ts
+++ b/src/lib/services/data-encoder.ts
@@ -39,7 +39,6 @@ export async function convertPayloadsWithCodec({
   const encoderResponse: Promise<Payloads> = fetch(endpoint + '/decode', {
     headers,
     method: 'POST',
-    credentials: 'include',
     body: stringifyWithBigInt(payloads),
   })
     .then((r) => r.json())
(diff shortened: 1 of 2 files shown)


rossedfort:
@christopherb-stripe - we noticed #1050 introduced an issue with existing customers:
```
The value of the ‘Access-Control-Allow-Origin’ header in the response must not be the wildcard ‘*’ when the request’s credentials mode is ‘include’
```

Unfortunately we're going to have to revert #1050 out of main as we have several customers who are affected by this change. We'll work toward a more nuanced solution for all auth models starting next week.

GiantRobots (approved):

christopherb-stripe:
@rossedfort I'm sorry about that - thanks for letting me know.

Would the future approach be something like exposing the included credentials conditionally based on a feature flag on the server which could be set by an environment variable or something like that?

rossedfort:
@christopherb-stripe no worries!

> Would the future approach be something like exposing the included credentials conditionally based on a feature flag on the server which could be set by an environment variable or something like that?

Yes, I think that sounds like the most likely solution.

PR #1118, opened 2023-02-07
Provide an option to send cross-origin credentials.

robholland:
## What was changed
Added an option to send credentials to the remote codec server.

## Why?
Some users would like to make use of cookies for authentication to their remote codec server. However CORS would refuse requests that claim they'd like to send credentials if the user's remote codec server does not expect them. Make this an option for users who need it.

This code is based on #1050 with code added to make it optional.

Screen Shot 2023-03-27 at 3 28 03 PM

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


Alex-Tideman:
@robholland Would we want to show this option for cloud.temporal.io?

robholland:
Yes, it would be applicable to both cloud and OSS installs.

src/lib/services/settings-service.ts
```diff
@@ -25,6 +25,7 @@ export const fetchSettings = async (request = fetch): Promise<Settings> => {
     codec: {
       endpoint: settingsResponse?.Codec?.Endpoint,
       passAccessToken: settingsResponse?.Codec?.PassAccessToken,
+      passCredentials: settingsResponse?.Codec?.PassCredentials,
```
Alex-Tideman, line 28: Are we adding this to the settings api through an env var?

Alex-Tideman:
I know Candace has updates to the modal itself, I'm going to run this new option by her and then update the modal in this PR.

src/lib/services/settings-service.ts
robholland: Yeah we'll want to do that like we do for the access token.

Alex-Tideman (approved):

PR #1121, opened 2023-02-07
Cache dependencies && update Node version

stevekinney:
1. Updates Github Actions to cache dependencies and also bumps the version of Node used to the latest LTS.
2. It also refactors the listing, unit test, type checking, and snapshot tests into one action where each job is run in parallel.
3. Cancels all workflows if a new commit is pushed to that branch.
4. Caches pnpm dependencies


.github/workflows/integration-tests.yml
```diff
@@ -5,19 +5,21 @@ on:
...
     steps:
-      - name: Checkout
-        uses: actions/checkout@v2
-      - name: Setup
-        uses: pnpm/action-setup@v2.2.2
-        with:
-          version: 7
-      - name: Install dependencies
-        run: pnpm install
+      - uses: actions/checkout@v3
```
rossedfort, line 20: Looks like a dupe checkout could occur here

rossedfort (commented):
Same comment applies to multiple files - can we optimize further by removing the `actions/checkout@v3` steps where we're already using the `checkout-and-setup` action you made?

```
    steps:
      - uses: actions/checkout@v3
      - name: Checkout and Setup Node
        uses: ./.github/actions/checkout-and-setup
```

rossedfort (approved):
⏩

PR #1126, opened 2023-02-08
[DT-301] - use batch API with list of workflow-executions

rossedfort:
## What was changed

- switch from constructing an imperformant `OR` query for batch operations to using the new `executions` field: https://github.com/temporalio/api/commit/aaedf8bd08a8bdaf4febe92dea30eed5b6217973
- also remove stale code for deprecated batch operation endpoints
## Why?

- For better performance in the API when creating a large batch operation
## Checklist

1. Closes 

2. How was this tested:

- tested locally against docker-compose
3. Any docs updates needed?


src/lib/services/batch-service.ts
```diff
@@ -43,55 +29,63 @@ export async function bulkTerminateByIDs({
...
+  return terminateWorkflows({ namespace, query, reason });
 }
 
 export async function bulkCancelByIDs({
   namespace,
   workflows,
   reason,
 }: CreateBatchOperationWithIDsOptions): Promise<string> {
-  return cancelWorkflows(namespace, queryFromWorkflows(workflows), reason);
+  const executions = workflows.map(({ runId, id }) => ({
```
laurakwhit, line 52: Wondering if we want to pull this out into a function (similar to how we had `queryFromWorkflows`)?

laurakwhit (approved):
Code lgtm 🚀 Will defer to you and @Alex-Tideman r.e. version compatibility.

Alex-Tideman:
@rossedfort I'm thinking we (unfortunately) need to add version check for 1.19.3 (or whatever this was released in) so older versions use the query

rossedfort:
> @rossedfort I'm thinking we (unfortunately) need to add version check for 1.19.3 (or whatever this was released in) so older versions use the query

Ah yeah, good catch 👍

Alex-Tideman (approved):

stevekinney (approved):

PR #1135, opened 2023-02-10
Use a Vite plugin for starting up a Temporal Server

stevekinney:
- Uses the [TypeScript SDK's testing server](http://github.com/temporalio/sdk-typescript/tree/main/packages/testing) to spin up a server when the UI development server starts up.
- Switches the Vite configuration toy TypeScript.
- Removes the script that downloads Temporal CLI.

### DIFF
--- a/.env
+++ b/.env
@@ -1,4 +1,3 @@
-VITE_API="http://localhost:7000"
-VITE_TEMPORAL_PORT="6000"
-VITE_UI_SERVER_PORT="7000"
+VITE_TEMPORAL_PORT="7233"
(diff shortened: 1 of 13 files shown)


rossedfort:
@stevekinney this looks good, but do we need each of those 4 `.env` files? I don't see where they are referenced.

stevekinney:
> @stevekinney this looks good, but do we need each of those 4 `.env` files? I don't see where they are referenced.

Vite uses them under the hood based on the mode (`development`, `testing`, etc.) that it's set to.

rossedfort (approved):

PR #1150, opened 2023-02-14
ARIA updates

laurakwhit:
## What was changed

* Uses semantic HTML whenever possible (instead of landmarks)
* Fixed UI spacing for version number in nav
* Added `aria-label` to pagination buttons
* Made label for `simple-select` available to screen readers (e.g. workflow filter buttons)
* Added role and type search to search inputs
* Added roles for `Alert` components
* Added `role="log"` for the `Toaster` component

## Checklist

1. Closes: `DT-230` 

2. How was this tested:

* Go to `System Settings` > `Accessibility` > `VoiceOver` on mac to test with screen reader
* CTRL+Option+U and use left/right arrow keys to find `Landmarks`

### DIFF
--- a/src/lib/components/workflow/workflow-filters.svelte
+++ b/src/lib/components/workflow/workflow-filters.svelte
@@ -70,7 +70,7 @@
   }, 300);
 </script>
 
-<section class="flex flex-col">
+<dev class="flex flex-col">
   <p class="pb-2 text-right text-xs">
     {#if searchType === 'advanced'}
       <a
@@ -146,4 +146,4 @@
       </Select>
     </div>
   {/if}
-</section>
+</dev>
--- a/src/lib/components/workflow/workflow-filters.svelte
+++ b/src/lib/components/workflow/workflow-filters.svelte
@@ -100,9 +100,13 @@
       on:submit={updateQuery}
     />
   {:else}
-    <div class="grid grid-cols-1 gap-4 sm:grid-cols-2 md:grid-cols-5">
+    <div
+      role="search"
+      class="grid grid-cols-1 gap-4 sm:grid-cols-2 md:grid-cols-5"
+    >
       <Input
         icon="search"
+        type="search"
         id="workflow-id-filter"
         placeholder="Workflow ID"
         bind:value={parameters.workflowId}
@@ -110,6 +114,7 @@
       />
       <Input
         icon="search"
+        type="search"
         id="workflow-type-filter"
         placeholder="Workflow Type"
         bind:value={parameters.workflowType}
--- a/src/lib/components/workflow/workflow-filters.svelte
+++ b/src/lib/components/workflow/workflow-filters.svelte
@@ -117,7 +117,7 @@
       />
       <Select
         id="time-range-filter"
-        placeholder="Time Range"
+        label="Time Range"
         bind:value={parameters.timeRange}
         on:change={handleParameterChange}
       >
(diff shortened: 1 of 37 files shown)


src/lib/components/workflow/workflow-filters.svelte
```diff
@@ -70,7 +70,7 @@
   }, 300);
 </script>
 
-<section class="flex flex-col">
+<dev class="flex flex-col">
```
rossedfort, line 73: `dev` or `div` ? 🤔

src/lib/holocene/alert.svelte
```diff
@@ -1,18 +1,26 @@
...
+    'data-testid'?: string;
+  }
+
   export let intent: 'warning' | 'error' | 'success' | 'info';
   export let title: string;
   export let icon: IconName = null;
   export let bold = false;
+
+  let className = '';
+  export { className as class };
```
rossedfort, line 20: I like this cause then we don't have to `omit($$restProps, 'class')`, nice!

rossedfort (approved):
looks great!

src/lib/holocene/alert.svelte
```diff
@@ -1,18 +1,26 @@
...
 </script>
 
-<div
-  class="alert {intent} {$$props.class}"
-  class:bold
-  data-testid={$$props.testId}
->
+<div class="alert {intent} {className}" class:bold {...$$restProps}>
```
rossedfort, line 23: Should we just add `role="alert"` here so we don't even have to think about adding it when we use an Alert?

rossedfort: Or at least make it a required prop?

laurakwhit: I was wondering about that... Most `Alert`s will need a role, but there are are cases in the UI currently where the `Alert` holds information that doesn't change and therefore isn't really an `alert` or a `status`. I opted to make `role` optional because of those cases.

Maybe we want to create another component not called `Alert` for those cases though? Called `Banner` or something? Or maybe we leave it as is for now and iterate on it in the future 😅

src/lib/holocene/autocomplete.svelte
```diff
@@ -25,7 +41,11 @@
       );
 </script>
 
-<div class="relative" use:clickOutside on:click-outside={() => (show = false)}>
+<div
+  class="relative {className}"
+  use:clickOutside
+  on:click-outside={() => (show = false)}
+>
   <Input
```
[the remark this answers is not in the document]
laurakwhit: Hmm I'm thinking **not** purely for stylistic reasons since the icon is interchangeable and isn't always a search icon.

src/lib/holocene/alert.svelte
rossedfort: Hm I see. Yeah sounds like things that we are currently using an Alert for should be maybe not be Alerts. I just would hate to have to keep going back and fixing a11y errors, ideally in using the holocene components we get good a11y baked in.

src/lib/holocene/autocomplete.svelte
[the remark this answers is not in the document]
rossedfort: Makes sense, I just thought an autocomplete would only really be used in a search context. But if not, this is fine 👍

src/lib/holocene/alert.svelte
laurakwhit: > ideally in using the holocene components we get good a11y baked in

100% agree. I'm going to update `Alert` to default to `role="alert"` and set `role="status"` on ones that don't need it.

PR #1154, opened 2023-02-15
Bump x/net for https://github.com/temporalio/ui/issues/1145

robholland:
## What was changed
Bump x/net version.

## Why?
https://github.com/temporalio/ui/issues/1145

## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?


robholland:
We'll need SDK to update their dep also, we won't get a clean run until we're using a bumped SDK version.

feedmeapples (approved):

PR #1155, opened 2023-02-15
Update supportsAdvancedVisibility to check for v1.20

Alex-Tideman:
## What was changed
Updated supportsAdvancedVisibility to check for 1.20 or newer version of Temporal. This will allow mysql/postgres visibility stores to use new filters.

If supportsAdvancedVisibility, update the DateTime query to use > instead of BETWEEN logic.

Disabled sorting UI for Start/Close Time if not using elastic search for now.

## Why?
Temporal v1.20 updated mysql/postgres visibility stores to have (almost) parity with elasticsearch. This PR lets users use our new filter workflow list page if using 1.20.

2. How was this tested:

- [x] Updated unit tests
- [x] New cypress tests for mysql only visibility store on 1.20
- [x] Tested locally with 1.20 with mysql/postgres and with elasticsearch
- [x] Tested locally with 1.19 with mysql/postgres and with elasticsearch
- [x] Tested locally on cloud


src/lib/utilities/advanced-visibility-enabled.ts
```diff
@@ -1,3 +1,20 @@
...
+  return (
+    cluster?.visibilityStore?.includes('elasticsearch') ||
+    isVersionNewer(version, '1.19')
+  );
+};
+
+// Add isVersionNewer check when we know when orderBy is supported
+export const advancedVisibilityEnabledWithOrderBy = (
+  cluster: ClusterInformation,
+  version: string,
```
stevekinney, line 16: It probably makes sense to get rid of this argument if it's not used?

src/lib/utilities/version-check.test.ts
```diff
@@ -21,6 +22,7 @@ describe('isRecommendedVersionNewer', () => {
     expect(isVersionNewer('xxx', '1.01.1')).toBe(false);
     expect(isVersionNewer('1.1', '1.01.1')).toBe(false);
     expect(isVersionNewer('1.01', '1.02')).toBe(false);
+    expect(isVersionNewer('1.19.3', '1.19')).toBe(false);
```
stevekinney, line 25: Isn't `1.19.3` newer than `1.19` though?

src/lib/utilities/advanced-visibility-enabled.ts
```diff
@@ -1,3 +1,20 @@
...
+) => {
   return cluster?.visibilityStore?.includes('elasticsearch');
+  // || isVersionNewer(version, '1.20')
```
stevekinney, line 19: Might as well get rid of this line.

Alex-Tideman: 👍

src/lib/utilities/version-check.test.ts
Alex-Tideman: No not if you don't provide a patch. So if you want to check if it's => 1.20, you do isVersionNewer(version, '1.19').

stevekinney (approved):

PR #1162, opened 2023-02-16
DT-384 - improve modal component to support form input

rossedfort:
## What was changed

- add `hasInput` prop to modals - if true, wrap content and buttons in a `<form />` and use `on:submit`
- fix keyboard nav for modals with inputs and/or disabled buttons

## Why?

- better keyboard a11y
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/lib/holocene/modal.svelte
+++ b/src/lib/holocene/modal.svelte
@@ -1,17 +1,34 @@
 <script lang="ts">
-  import Icon from '$lib/holocene/icon/icon.svelte';
-  import { createEventDispatcher } from 'svelte';
-  import Button from '$lib/holocene/button.svelte';
+  import type { HTMLAttributes } from 'svelte/elements';
   import { noop } from 'svelte/internal';
+  import { createEventDispatcher, type ComponentProps } from 'svelte';
+  import Icon from '$lib/holocene/icon/icon.svelte';
+  import Button from '$lib/holocene/button.svelte';
 
-  export let open: boolean = false;
+  interface $$Props extends HTMLAttributes<HTMLDivElement> {
+    open: boolean;
+    hideConfirm?: boolean;
+    confirmText?: string;
+    cancelText?: string;
+    confirmType?: ComponentProps<Button>['variant'];
+    confirmDisabled?: boolean;
+    large?: boolean;
+    loading?: boolean;
+    hasInput?: boolean;
+  }
+
+  export let open: boolean;
   export let hideConfirm: boolean = false;
   export let confirmText: string = 'Confirm';
   export let cancelText: string = 'Cancel';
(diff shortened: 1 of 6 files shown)


rossedfort:
I considered breaking the buttons out of Modal into a new component - `<ModalActions />`, so that we could compose the form ourselves, like

```
<Modal>
  <h3 slot="title">Title</h3>
  <form slot="content">
    <Input />
    <ModalActions type="submit" />
  </form>
</Modal>
```
but that would introduce a fair amount of breaking changes.

@Alex-Tideman @laurakwhit @stevekinney 
Thoughts?

src/lib/holocene/modal.svelte
```diff
@@ -91,29 +120,35 @@
...
-      <div class="flex items-center justify-end space-x-2 p-6">
-        <Button
-          thin
-          variant="secondary"
-          disabled={loading}
-          on:click={cancelModal}>{cancelText}</Button
-        >
-        {#if !hideConfirm}
+      <svelte:element
+        this={hasInput ? 'form' : 'div'}
```
laurakwhit, line 124: Do the actions need to be inside the `<form/>`? Could we just pass the form into the `<slot/>` for the modal? 
e.g.
```js
<Modal on:confirm={handleConfirm}>
    <form id="example-form" on:submit={handleConfirm}>
      <input/>
   </form>
</Modal>
```

And then have some way of making the confirm button `type="submit"` and have a reference to the form's `id`?
E.g.
```js
<Button type={formId ? "submit" : "button"} form={formId} />`

File: src/lib/holocene/modal.svelte [hunk 2]
```diff
@@ -67,12 +95,13 @@
     />
     <div
       bind:this={modalElement}
-      class="body"
+      class="body {className}"
       class:large
       tabindex="-1"
       role="alertdialog"
       aria-labelledby="modal-title"
       aria-describedby="modal-description"
```
laurakwhit, line 103: Looks like we could also add an [aria-modal="true](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-modal)" here.

src/lib/holocene/modal.svelte
rossedfort: Oh nice I didn't know this was a thing. I think we definitely should have done it this way if I hadn't done a large refactor 😂 Now `<Modal />` uses a `<dialog />` element which always contains a `<form />`, and the form always contains the buttons. I implemented it largely according to [this example](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#simple_example).

stevekinney (approved):

PR #1163, opened 2023-02-16
Walk back the changes to CODEOWNERS

stevekinney:


Alex-Tideman:
FYI I fixed this. You can leave the CODEOWNERS with all the individuals, but now their approvals isn't required. So it will auto-assign those peeps but not require their approval.

Alex-Tideman:
I unchecked this:

Screen Shot 2023-02-16 at 3 33 07 PM

Issue #1171, opened 2023-02-20
Payload decoding fails on large decimal number


mjameswh:
**Describe the bug**

A workflow's history contains some JSON payload (json/plain) and some field inside that payload is a decimal number of at least 15 characters. When inspecting that workflow in UI, the payload does not get decoded, and a warning message is displayed in the browser console, such as:

```
Could not parse payload:  SyntaxError: Cannot convert 0.2345678901234568 to a BigInt (at parse-with-big-int-f9a90488.js)
```

**To Reproduce**

Run this workflow, then open that workflow's in UI:

```
async myWorkflow() {
  return { result: 0.2345678901234568 };
}
```

**Additional context**

This is due to a [bug in json-bigint](https://github.com/sidorares/json-bigint/issues/73), which has been fixed months ago in that project's master branch. However, no NPM release has been published in years. Given that json-bigint is MIT licensed, it might make sense to fork it.

PR #1178, opened 2023-02-23
Fix error with large decimals in parseWithBigInt

Alex-Tideman:
## What was changed
If JSONBigNative.parse() fails, catch and return JSON.parse()

## Why?
A bug in json-bigint throws an error on decimals over 15 digits. This catches the error and falls back to JSON.parse()

Unit tests were added

Fixes issue #1171

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -103,7 +103,7 @@ specifiers:
   zx: ^7.1.1
 
 dependencies:
-  '@codemirror/autocomplete': 6.4.0_czcfkg2f66rxeiodoti7r2gulu
+  '@codemirror/autocomplete': 6.4.0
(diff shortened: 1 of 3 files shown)


mjameswh:
Just noting that a single JSON document can technically contain _both_ a large decimal number AND a big integer. In that case, big integer will loose precision.

But I don't think this is that much of an issue.

Alex-Tideman:
Yeah good call out. I want to get this fix out now and then work on a fork that handles both.

rossedfort (approved):

duongleh:
hi @Alex-Tideman have you resolved the issue with displaying big integer precision? We encountered a case in version 2.39.0 where the decoded data of a big integer is displayed incorrectly.

PR #1179, opened 2023-02-23
Fix default option behaviour

GiantRobots:
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,13 @@
     "temporal",
     "ui"
   ],
+  "typesVersions": {
+    ">4.0": {
+      "*": [
+        "./lib/*"
+      ]
+    }
+  },
   "license": "MIT",
   "homepage": "https://github.com/temporalio/ui/tree/main",
   "scripts": {
@@ -35,7 +42,7 @@
     "preview:local": "VITE_TEMPORAL_UI_BUILD_TARGET=local vite preview --port 3000",
     "preview:docker": "VITE_API=http://localhost:8080 VITE_TEMPORAL_UI_BUILD_TARGET=local vite preview --port 3000",
     "preview:cloud": "VITE_TEMPORAL_UI_BUILD_TARGET=cloud vite preview --port 3000",
-    "package": "svelte-package && rsync -r ./package/lib/ ./package/ && rm -rf ./package/lib",
+    "package": "svelte-package",
     "package:patch": "pnpm version patch && svelte-package",
     "package:minor": "pnpm version minor && svelte-package",
     "package:major": "pnpm version major && svelte-package",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -35,11 +42,11 @@
     "preview:local": "VITE_TEMPORAL_UI_BUILD_TARGET=local vite preview --port 3000",
     "preview:docker": "VITE_API=http://localhost:8080 VITE_TEMPORAL_UI_BUILD_TARGET=local vite preview --port 3000",
     "preview:cloud": "VITE_TEMPORAL_UI_BUILD_TARGET=cloud vite preview --port 3000",
-    "package": "svelte-package && rsync -r ./package/lib/ ./package/ && rm -rf ./package/lib",
+    "package": "svelte-package",
     "package:patch": "pnpm version patch && svelte-package",
     "package:minor": "pnpm version minor && svelte-package",
     "package:major": "pnpm version major && svelte-package",
-    "story:dev": "histoire dev",
+    "story:dev": "histoire dev --port 3030",
```
rossedfort, line 49: why the port change?

src/lib/holocene/primitives/menu/menu.svelte
```diff
@@ -7,20 +7,25 @@
...
+>
+  <slot />
+</ul>
 
 <style lang="postcss">
+  /* We should probably have a nonVisible class here that allows screen readers to see the content.
+  Which might be an option on the menu (for things like split buttons) */
+  .hide {
+    display: none;
+  }
```
rossedfort, line 27: maybe `sr-only` ? https://tailwindcss.com/docs/screen-readers

rossedfort (approved):
Looks good

src/lib/holocene/select/select.story.svelte
```diff
@@ -79,6 +79,24 @@
...
+      <OptionGroup label="Reptiles">
+        <Option value={4}>Snake</Option>
+        <Option value={5}>Lizard</Option>
+      </OptionGroup>
+      <OptionGroup label="Amphibians">
+        <Option value={6}>Frog</Option>
+        <Option value={7}>Salamander</Option>
+      </OptionGroup>
+    </Select>
+  </Hst.Variant>
```
rossedfort, line 98: Probably can 🚮

package.json
GiantRobots: I was on an airplane and the console outputted so much stuff I couldn't find the port it was running on and I didn't have internet so I just set it I'll revert.

src/lib/holocene/select/select.story.svelte
GiantRobots: This is essentially a test case to verify that the changes we implemented work. I think I'd prefer to keep it around.

rossedfort: Oh I didn't notice these `<Option />`'s had `value`'s, sounds good 👍

PR #1181, opened 2023-02-23
Expose CSRF cookie insecure setting as Docker env

feedmeapples:
## What was changed

Exposed `cookieInsecure` as Docker env variable

## Why?

Users who secure their UI deployments in non-HTTPS manner may want to disable secure cookie so CSRF cookie is sent over HTTP 

## Checklist

1. Closes 

2. How was this tested:

Changed the value in development.yaml and verified it is being set

3. Any docs updates needed?

### DIFF
--- a/server/docker/config_template.yaml
+++ b/server/docker/config_template.yaml
@@ -16,7 +16,7 @@ workflowSignalDisabled: {{ default .Env.TEMPORAL_WORKFLOW_SIGNAL_DISABLED "false
 workflowResetDisabled: {{ default .Env.TEMPORAL_WORKFLOW_RESET_DISABLED "false" }}
 batchActionsDisabled: {{ default .Env.TEMPORAL_BATCH_ACTIONS_DISABLED "false" }}
 cors:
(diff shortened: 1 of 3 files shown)


server/docker/config_template.yaml
```diff
@@ -16,6 +16,7 @@ workflowSignalDisabled: {{ default .Env.TEMPORAL_WORKFLOW_SIGNAL_DISABLED "false
 workflowResetDisabled: {{ default .Env.TEMPORAL_WORKFLOW_RESET_DISABLED "false" }}
 batchActionsDisabled: {{ default .Env.TEMPORAL_BATCH_ACTIONS_DISABLED "false" }}
 cors:
+  cookieInsecure: {{ default .Env.TEMPORAL_COOKIE_INSECURE "false" }}
```
robholland, line 19: I think better to include either `CORS` or `CSRF` in the env var name.

robholland (changes requested):

robholland (approved):

PR #1184, opened 2023-02-24
handle uncaught error in batch actions

rossedfort:
## What was changed

Wrap both terminate and cancel by IDs and query in a try/catch to ensure we are catching all errors.
## Why?

To fix the case where bulk terminating/cancelling workflows would throw an uncaught error
## Checklist

1. Closes 

2. How was this tested:

3. Any docs updates needed?

### DIFF
--- a/src/lib/pages/workflows-with-new-search.svelte
+++ b/src/lib/pages/workflows-with-new-search.svelte
@@ -92,7 +92,6 @@
 
   const resetPageToDefaultState = () => {
     terminating = false;
-
     $workflowFilters = [];
     $workflowSorts = [];
     updateQueryParameters({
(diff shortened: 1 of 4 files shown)


src/lib/pages/workflows-with-new-search.svelte
```diff
@@ -136,20 +135,21 @@
...
-        query: batchOperationQuery,
-      });
-      toaster.push({
-        message: 'The batch terminate request is processing in the background.',
-        id: 'batch-terminate-success-toast',
-      });
-    } else {
-      terminating = true;
-      try {
+    try {
```
laurakwhit, line 138: There might be an opportunity to reuse some of this logic in a reusable function e.g.

```js
  const batchAction = async (
    event: CustomEvent<{ reason: string }>,
    action: 'cancel' | 'terminate',
  ) => {
    const { namespace } = $page.params;
    const { reason } = event.detail;

    try {
      if (allSelected) {
        if (action === 'cancel') {
          await batchCancelByQuery({
            namespace,
            reason,
            query: batchOperationQuery,
          });
        }
        if (action === 'terminate') {
          await batchTerminateByQuery({
            namespace,
            reason,
            query: batchOperationQuery,
          });
        }
        toaster.push({
          message: `The batch ${action} request is processing in the background.`,
          id: `batch-${action}-success-toast`,
        });
      } else {
        const jobId = await bulkCancelByIDs({
          namespace,
          reason,
          workflows: cancelableWorkflows,
        });
        const workflows = await pollBatchOperation({
          namespace,
          jobId,
        });
        const COMPLETED_ACTION = {
          cancel: 'cancelled',
          terminate: 'terminated',
        };
        toaster.push({
          message: `Successfully ${COMPLETED_ACTION[action]} ${workflows} workflows.`,
          id: `batch-${action}-success-toast`,
        });
      }
    } catch (error) {
      toaster.push({
        message: `Unable to ${action} Workflows.`,
        variant: 'error',
      });
    } finally {
      resetPageToDefaultState();
    }
  };
```

But maybe something a little cleaner than that 😅

Alex-Tideman (approved):

PR #1185, opened 2023-02-27
Make batch action checkboxes less error prone

rossedfort:
Use new checkbox hover state to make the checkboxes in the Recent Workflows table rows easier to click. Also fix checkboxes to support `bind:group`

- [x] Does this change require a design review?
- Yes, worked with Candace on the checkbox hover state
- [x] Does this change require manual testing?
- Yes, I test locally but wouldn't mind another pair of eyes testing the UX of the checkboxes

### DIFF
--- a/src/lib/pages/workflows-with-new-search.svelte
+++ b/src/lib/pages/workflows-with-new-search.svelte
@@ -181,10 +181,6 @@
       selectedWorkflows[workflow.runId] && workflow.status === 'Running',
   );
 
-  $: {
-    console.log($workflowCount);
-  }
-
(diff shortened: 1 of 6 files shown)


Alex-Tideman (approved):

src/lib/pages/workflows-with-new-search.svelte
```diff
@@ -183,6 +181,10 @@
       selectedWorkflows[workflow.runId] && workflow.status === 'Running',
   );
 
+  $: {
+    console.log($workflowCount);
+  }
+
```
Alex-Tideman, line 187: 🗑️

PR #1187, opened 2023-02-27
implement err handling on Modals

rossedfort:
Allow parent component of Modals to set an error on the modal.
- note, this prevents the Modal's default behavior of closing when the submit/confirm button is clicked
Screen Shot 2023-03-01 at 10 04 48 AM

- [x] Does this change require a design review?
- [x] Does this change require manual testing?

### DIFF
--- a/src/lib/components/workflow-actions.svelte
+++ b/src/lib/components/workflow-actions.svelte
@@ -69,6 +69,7 @@
   };
 
   const handleSuccessfulTermination = async () => {
+    terminateConfirmationModal.close();
     $refresh = Date.now();
     toaster.push({
       id: 'workflow-termination-success-toast',
@@ -76,9 +77,11 @@
     });
   };
 
-  const handleTerminationError = () => {
+  const handleTerminationError = (error: NetworkError) => {
     reason = '';
-    toaster.push({ message: 'Cannot terminate workflow.', variant: 'error' });
+    terminateConfirmationModal.setError(
+      error?.message ?? 'An unknown error occurred.',
+    );
   };
 
   const terminate = () => {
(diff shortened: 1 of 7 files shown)


Alex-Tideman:
I'm good with this if @canvanooo is.

Alex-Tideman (approved):

src/lib/holocene/modal.svelte
```diff
@@ -44,10 +54,12 @@
 
   const handleCancel = () => {
     modalElement.close();
```
laurakwhit, line 56: Could we just call `close()` here instead?
```suggestion
    close();
```

src/lib/holocene/modal.svelte
```diff
@@ -44,10 +54,12 @@
...
+    error = '';
     dispatch('cancelModal');
   };
 
   const confirmModal = () => {
+    error = '';
```
laurakwhit, line 62: It seems like this might be redundant in cases where we call `modal.close()` when handling confirm. Maybe we can handle the close here instead?

rossedfort: @laurakwhit that's how the modal worked previous to this change, but now we want to keep the modal open if an error occurs in the request that's triggered when the modal is confirmed/submitted, so the modal can't really close itself unless it has knowledge of whether the request succeeds.

laurakwhit: I guess I was thinking we could somehow have a try/catch here as well, but it's definitely fine as is! Just trying to think through if there is a cleaner way to handle this...

src/lib/components/workflow-actions.svelte
```diff
@@ -100,17 +103,17 @@
         workflowId: workflow.id,
         runId: workflow.runId,
       });
+      cancelConfirmationModal.close();
```
Alex-Tideman, line 106: Should this be terminateConfirmationModal.close()?

rossedfort: No, I think the diff is just misleading.

Alex-Tideman: ah yup you're right

PR #1194, opened 2023-03-02
DT-551 - wrap decoding the workflow query in a try/catch

### LINKED ISSUE #1030: JS error with queries using %
**Describe the bug**

If I use Advance Search and type a query with % the query (request to the server) is never executed and I see a JS error in the console. 

Screenshot 2023-01-09 at 15 14 05

UI v 2.9.0

rossedfort:
## Description & motivation 💭 
This PR wraps a `decodeURIComponent` call in a try/catch to solve a bug when a user enters a query with a "%" character. This should be a non-breaking change as all queries that do need to be decoded should still function properly.

n/a
### Design Considerations 🎨 
n/a
## Testing 🧪 
This should be covered by existing advanced search e2e and unit tests, as well as my own manual testing.
### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

n/a
### Draft Checklist 
n/a
### Merge Checklist 
n/a
### Issue(s) closed 
closes #1030 
### Any docs updates needed? 
No

### DIFF
--- a/src/lib/services/workflow-service.ts
+++ b/src/lib/services/workflow-service.ts
@@ -114,9 +114,14 @@ export const fetchAllWorkflows = async (
   request = fetch,
   archived = false,
 ): Promise<CombinedWorkflowExecutionsResponse> => {
-  const query = decodeURIComponent(
-    parameters.query || toListWorkflowQuery(parameters, archived),
-  );
+  const rawQuery =
+    parameters.query || toListWorkflowQuery(parameters, archived);
+  let query: string;
+  try {
+    query = decodeURIComponent(rawQuery);
+  } catch {
+    query = rawQuery;
+  }
 
   const endpoint: ValidWorkflowEndpoints = archived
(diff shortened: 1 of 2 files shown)


src/lib/services/workflow-service.ts
```diff
@@ -114,9 +114,14 @@ export const fetchAllWorkflows = async (
   request = fetch,
   archived = false,
 ): Promise<CombinedWorkflowExecutionsResponse> => {
-  const query = decodeURIComponent(
-    parameters.query || toListWorkflowQuery(parameters, archived),
-  );
+  const rawQuery =
```
stevekinney, line 117: Can we add some unit tests around the query in question?

rossedfort: Yes, good call.

Alex-Tideman (approved):

PR #1195, opened 2023-03-03
Make date/time filter sticky and reapply workflow list filters when switching namespaces

laurakwhit:
## Description 💭 
Date/time filter (e.g. last `15 Minutes`, `All Time`, etc.) is persisted on the user's device so that when they access Temporal it applies the filter on the `Recent Workflows` page.

All filters on the `Recent Workflows` page. (e.g. `ExecutionStatus`, `WorkflowType`, `StartTime`, etc.) are persisted when switching namespaces.

### Design Considerations 🎨 
- [ ] Do we actually want to reapply workflow list filters when switching namespaces? It's possible that those applied filters will not return any results.
- [ ] Should we alert the user that the date/time filter is persisted? Could be confusing since none of the other filters are. Maybe we make it an opt-in feature? 

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added 
- [ ] Unit tests added

On the `Recent Workflows` page with advanced visibility and the new search enabled
- Select a date/time from the dropdown
   - Navigate away and then navigate to back to `/workflows` 
      - [ ] Verify the date/time filter is still applied
   - Log out, log in again, and then navigate to `/workflows` 
      - [ ] Verify the date/time filter is still applied
   - Change namespaces
      - [ ] Verify the date/time filter is still applied

- Navigate to a link with a date/time `query` that does not match one of the available dropdown options (e.g. `20 minutes`)
   - [ ] Verify the date/time filter says `Custom`
   - Navigate to `/workflows` without a query
      - [ ]  Verify the date/time from the link is not automatically applied
   - Navigate back to the link
   - Click on `Custom`
      - [ ] Verify the `Custom` query is not immediately reapplied and the inputs are blank
   - Add new `Custom` start and end times and click `Apply`
      - [ ] Verify the query is updated

- Select various filters/sorts
   - Change namespaces
      - [ ] Verify the filters/sorts are applied
   - Change some filters/sorts
      - [ ] Verify the filters/sorts are updated for the current namespace 

- Navigate to the workflow details page
   - [ ] Verify the `Back to Workflows` works as expected

- Click on the workflow nav links while on the `Recent Workflows` page
   - [ ] Verify all filters/sorts are still applied
- Click away from `Recent Workflows` and then on the workflow nav links 
   - [ ] Verify only the persisted filters are applied
---
On the `Recent Workflows` page **without** advanced visibility (and without the new search)
- [ ] Verify changing filters/sorts works as expected 
- [ ] Verify changing namespaces works as expected 
   - No filters/sorts will be applied
- [ ] Add a date/time filter and navigate to `/workflows`
   - The date/time filter will not be persisted

### Draft Checklist 
- [ ] Review design considerations

### Issue(s) closed: 
* `DT-24` 
* `DT-40`

### DIFF
--- a/src/lib/components/namespace-list.svelte
+++ b/src/lib/components/namespace-list.svelte
@@ -1,14 +1,17 @@
 <script lang="ts">
+  import { createEventDispatcher, onMount } from 'svelte';
   import { page } from '$app/stores';
 
+  import { workflowsQuery, workflowsSearch } from '$lib/stores/workflows';
+  import { routeForWorkflowsWithQuery } from '$lib/utilities/route-for';
+  import { toListWorkflowQuery } from '$lib/utilities/query/list-workflow-query';
+
+  import EmptyState from '$lib/holocene/empty-state.svelte';
   import Icon from '$lib/holocene/icon/icon.svelte';
 
-  import { onMount } from 'svelte';
-  import EmptyState from '$lib/holocene/empty-state.svelte';
-  import { createEventDispatcher } from 'svelte';
-
   export let getNamespaceList: () => Promise<NamespaceItem[]> = null;
-
+  const { parameters, searchType } = $workflowsSearch;
+  const query = toListWorkflowQuery(parameters);
   let namespaceList = null;
   let searchField: HTMLInputElement = null;
 
@@ -74,7 +77,11 @@
           class="first:rounded-t-xl first:border-t-3 last:rounded-b-xl last:border-b-3 border-b border-l-3 border-r-3 border-gray-900 flex border-collapse gap-2 hover:bg-gradient-to-br from-blue-100 to-purple-100 cursor-pointer"
         >
           <a
-            href={namespace.href(namespace.namespace)}
+            href={routeForWorkflowsWithQuery({
+              namespace: namespace.namespace,
+              query: $workflowsQuery || query,
+              search: searchType,
+            })}
             class="w-full flex p-3"
             class:active={namespace.namespace === $page.params?.namespace}
           >
--- a/src/lib/components/workflow/dropdown-filter/workflow-datetime-filter.svelte
+++ b/src/lib/components/workflow/dropdown-filter/workflow-datetime-filter.svelte
@@ -86,14 +86,18 @@
 
   const onChange = (_value: string) => {
     value = _value;
+
+    if (value === 'Custom') {
+      custom = true;
+      return;
+    }
+
     if (value === 'All Time') {
       $workflowFilters = [...getOtherFilters($workflowFilters)];
       $persistedWorkflowFilters = [
         ...getOtherFilters($persistedWorkflowFilters),
       ];
       custom = false;
-    } else if (value === 'Custom') {
-      custom = true;
     } else {
       const filter = {
         attribute: timeField,
--- a/src/lib/components/workflow/dropdown-filter/workflow-datetime-filter.svelte
+++ b/src/lib/components/workflow/dropdown-filter/workflow-datetime-filter.svelte
@@ -1,4 +1,5 @@
 <script lang="ts">
+  import { onMount } from 'svelte';
   import { timeFormat } from '$lib/stores/time-format';
   import { capitalize } from '$lib/utilities/format-camel-case';
   import {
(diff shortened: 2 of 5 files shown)


src/lib/components/namespace-list.svelte
```diff
@@ -74,7 +77,11 @@
           class="first:rounded-t-xl first:border-t-3 last:rounded-b-xl last:border-b-3 border-b border-l-3 border-r-3 border-gray-900 flex border-collapse gap-2 hover:bg-gradient-to-br from-blue-100 to-purple-100 cursor-pointer"
         >
           <a
-            href={namespace.href(namespace.namespace)}
+            href={routeForWorkflowsWithQuery({
```
laurakwhit, line 80: I've opted to add the query here instead of doing ` updateQueryParamsFromFilter` on a namespace change as `$page.params.namespace` is getting updated quite a bit and it felt excessive to be constantly updating the query params all the time.

src/lib/components/workflow/dropdown-filter/workflow-datetime-filter.svelte
```diff
@@ -80,7 +98,12 @@
         operator: '',
         parenthesis: '',
       };
-      $workflowFilters = [...getOtherFilters(), filter];
+      $workflowFilters = [...getOtherFilters($workflowFilters), filter];
+      $persistedWorkflowFilters = [
```
laurakwhit, line 102: **Note:** This seems like the most simple solution for the time being in order to only persist the filters we actually want (and not ones that are applied from a query on a link), but open to ideas about better ways to update `persistedWorkflowFilters` based on `workflowFilters` without repeating logic 💡

src/lib/components/namespace-list.svelte
Alex-Tideman: Yup agree this is the right move.

src/lib/components/workflow/dropdown-filter/workflow-datetime-filter.svelte
Alex-Tideman: I think this is totally fine for now, but when we implement our actions/reducers, we'll want to consolidate this logic under a single action. This will be hard to maintain and grok all the places we need to set these filters and how/when to keep them in sync.

Alex-Tideman:
I think the only thing you missed @laurakwhit is when you click the Workflows button in the nav. It clears the filter query but the filter UIs are still applied. So either we need to clear all filters on Workflow nav click, or apply them to the query. I'd say we probably should apply them, but also think that clicking that button would start a clean slate might be good. Open to debate.

src/lib/pages/workflows-with-new-search.svelte
```diff
@@ -79,6 +79,12 @@
     }
   });
 
+  afterUpdate(() => {
+    if (!query && ($workflowFilters.length || $workflowSorts.length)) {
```
laurakwhit, line 83: > either we need to clear all filters on Workflow nav click, or apply them to the query

@Alex-Tideman thanks for catching this! I've opted to apply the filters/sorts when the user clicks on the `Workflows` or logo nav links  **while** they are on the `Recent Workflows` page. If they click away (e.g. to `Schedules`) and then click back to `Workflows` only the persisted filters will be applied.

Alex-Tideman: I haven't used afterUpdate before, just calling out to check if this could get called immediately after the onMount that has very similar code.

Alex-Tideman: I think there's some wonkiness here. Sometime when I click the nav icon, i get the initial queries, but then a couple of seconds later I get a bunch more query api calls

laurakwhit: > this could get called immediately after the onMount that has very similar code.

I added that conditional (`if (!query)`) to check if there already is a query so that the url isn't updated again, but the `afterUpdate` is being called after `onMount`.

> couple of seconds later I get a bunch more query api calls
There will be at least two api calls (one without the query and one with). Maybe some room for improvement there... Are you seeing additional ones? And are they related to `count`?

PR #1202, opened 2023-03-07
DT-543 - add side modal with drag & drop interface for customizing workflow table columns

rossedfort:
## Description & motivation 💭 
This PR adds an `<OrderableList />` component which is needed to allow users to customize the columns in the workflows table. `<OrderableList />` takes an array of `OrderableItems`, which have a `label`, `key`, and `locked` boolean. When an item is `locked`, it can't be moved, nor can other items be dragged to its position. 

|designs|implementation|
|---|---|
|Screen Shot 2023-03-07 at 1 44 00 PM|Screen Shot 2023-03-07 at 1 47 49 PM|

### Design Considerations 🎨 
- [ ] Need to check if using the `lock` icon is OK - original designs specify a `pin` icon.
- [ ] Need to check if hiding the vertical chevrons and hyphen icons on `locked` items is OK.

### How was this tested 👻 
- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Draft Checklist 
- [x] Check on design changes
- [x] ~~Add an environment feature flag to disable this feature in the UI~~
### Merge Checklist 
- [x] ~~Get feedback from team on a slightly different implementation~~
- [x] a11y review - how do we make drag & drop accessible?


src/lib/holocene/orderable-list.svelte
```diff
@@ -14,26 +12,32 @@
...
   export let items: OrderableItem[];
-  export let readonly: boolean = false;
-  export let removeItem: (index: number) => void = noop;
-  export let addItem: (index: number) => void = noop;
+  export let availableItems: OrderableItem[] = [];
+
+  const addItem = (index: number) => {
+    let tempAvailableItems = [...availableItems];
+    items = [...items, ...tempAvailableItems.splice(index, 1)];
+    availableItems = tempAvailableItems;
```
Alex-Tideman, line 26: Just a tiny nit, from a readability perspective it's hard to grok that items = [...items, ...tempAvailableItems.splice(index, 1)] is mutating tempAvailableItems as well for the next line assignment. Maybe (??) better like:


```
    let tempAvailableItems = [...availableItems].splice(index, 1);
    items = [...items, ...tempAvailableItems];
    availableItems = tempAvailableItems;

File: src/lib/holocene/orderable-list.svelte [hunk 2]
```diff
@@ -14,26 +12,32 @@
...
+  };
+
+  const removeItem = (index: number) => {
+    let tempItems = [...items];
+    availableItems = [...tempItems.splice(index, 1), ...availableItems];
+    items = tempItems;
```
Alex-Tideman, line 32: Same as above

Alex-Tideman:
🔥 This is awesome. Only things I think to check out is the margins/spacing on the drawer (screenshot below for the hotkeys drawer) and the position of the close icon (maybe it should always be in top right corner of drawer and a little bigger)

Screen Shot 2023-03-08 at 8 53 01 AM

Screen Shot 2023-03-08 at 8 52 52 AM

src/lib/holocene/orderable-list.svelte
rossedfort: Hm actually this may not work, because `.splice()` mutates the array but also returns the item(s) removed, so `tempAvailableItems` would just be an array of the 1 item that was spliced. I'll see if there's a more readable way to do this, we basically are just moving an item from one array to the other.

rossedfort: How about

```typescript
  const addItem = (index: number) => {
    let tempAvailableItems = [...availableItems];
    const [removedItem] = tempAvailableItems.splice(index, 1);
    items = [...items, removedItem];
    availableItems = tempAvailableItems;
  };
```

Alex-Tideman: 👍

src/lib/holocene/drawer.svelte
```diff
@@ -2,54 +2,103 @@
...
   export let onClick: () => void;
+
+  const flyParams = {
+    duration: 500,
+    ...(position === 'bottom' ? { y: 200 } : { x: 200 }),
+  };
 </script>
 
 {#if open}
   <div
```
laurakwhit, line 20: What do you think about making this an `<aside/>`?

laurakwhit (approved):

PR #1204, opened 2023-03-08
Sync to cloud branch for ui-server with cloud assets.

robholland:
## Description & motivation 💭 

Sync the cloud build of ui-server to a cloud branch. This will be used to build and publish docker images from the ui-server repo like we do for the main verison.

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -5,9 +5,6 @@ FROM ${BASE_BUILDER_IMAGE} AS server-builder
 
 WORKDIR /home/server-builder
 
-COPY Makefile ./
-RUN make install-utils
-
 COPY go.mod go.sum ./
 RUN go mod download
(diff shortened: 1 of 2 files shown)


server/Dockerfile
```diff
@@ -5,9 +5,6 @@ FROM ${BASE_BUILDER_IMAGE} AS server-builder
 
 WORKDIR /home/server-builder
 
-COPY Makefile ./
```
robholland, line 8: This isn't actually needed in the ui-server version of the Dockerfile, GRPC files are already built before we sync so we don't need any utils to build them.

feedmeapples (approved):

PR #1207, opened 2023-03-09
Allow dynamic table columns in table

laurakwhit:
## Description & motivation 💭 
Fixes styling for the `Table` column so that instead of hard coding columns in `Table` components we can allow for dynamic columns that can be an arbitrary array of columns. 

n/a
### Design Considerations 🎨 
n/a
### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

- [ ] Verify the following tables look as expected and are responsive
   - `Recent Events`
      - `History` view (with rows expanded)
      - `Compact` view (with rows expanded)
      -  Empty state
   - `Workers`
   - `Pending Activities`
   - `Schedules`
   - `Namespaces`
   - `Recent Workflows`
      - With new search
         - Empty state
         - With rows expanded
         - With bulk actions enabled/disabled/allSelected
      - With old search
         - Empty state
         - With rows expanded
### Draft Checklist 
n/a

### Merge Checklist 
- [ ] Have at least one other person complete manual testing
### Issue(s) closed 
* `DT-554`

### Any docs updates needed?  
n/a

### DIFF
--- a/src/lib/components/event/event-empty-row.svelte
+++ b/src/lib/components/event/event-empty-row.svelte
@@ -6,7 +6,7 @@
 </script>
 
 <tr class="row">
-  <td class="table-cell" colspan="6">
+  <td colspan="6">
     {#if loading}
       <Loading />
     {:else}
--- a/src/lib/components/workflow/workflows-summary-row-with-filters.svelte
+++ b/src/lib/components/workflow/workflows-summary-row-with-filters.svelte
@@ -157,10 +157,10 @@
   }
 
   .time-cell {
-    @apply hidden xl:table-cell;
+    @apply max-xl:hidden;
   }
 
   .inline-time-cell {
-    @apply hidden md:block xl:hidden;
+    @apply xl:hidden;
   }
 </style>
--- a/src/lib/components/workflow/workflows-summary-table-with-filters.svelte
+++ b/src/lib/components/workflow/workflows-summary-table-with-filters.svelte
@@ -106,11 +106,11 @@
         {/if}
       </th>
       {#if showBulkActions}
-        <th class="overflow-visible whitespace-nowrap" colspan={columns.length}>
+        <th class="overflow-visible whitespace-nowrap">
           {#if allSelected}
-            <span class="font-semibold"
-              >All {filteredWorkflowCount} selected</span
-            >
+            <span class="font-semibold">
+              All {filteredWorkflowCount} selected
+            </span>
           {:else}
             <span class="font-semibold">{selectedWorkflowsCount} selected</span>
             <span>
@@ -141,6 +141,13 @@
             {/if}
           </div>
         </th>
+        {#each Array(columns.length - 1) as _, index}
+          {@const className =
+            index === columns.length - 2 || index === columns.length - 3
+              ? 'max-xl:hidden'
+              : ''}
+          <th class={className} />
+        {/each}
       {:else}
         <th class="w-32">
           <ExecutionStatusDropdownFilter />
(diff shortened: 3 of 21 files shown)


src/lib/components/event/event-empty-row.svelte
```diff
@@ -6,7 +6,7 @@
 </script>
 
 <tr class="row">
-  <td class="table-cell" colspan="6">
+  <td colspan="6">
```
laurakwhit, line 9: I have removed classes like `table-cell`, etc. since they are default styling for the element.

src/lib/components/workflow/workflows-summary-row-with-filters.svelte
```diff
@@ -23,14 +24,63 @@
...
+  export let columns: string[];
 
   $: href = routeForEventHistory({
     namespace,
     workflow: workflow.id,
     run: workflow.runId,
   });
 
-  let showFilterCopy = false;
+  $: data = [
```
laurakwhit, line 35: Definitely open for discussion on potentially better ways we can handle dynamically loading table cells and headers 💬

src/routes/(app)/namespaces/[namespace]/workflows/+page.svelte
```diff
@@ -6,6 +6,14 @@
 
   import AdvancedVisibilityGuard from '$lib/components/advanced-visibility-guard.svelte';
   import WorkflowsWithNewSearch from '$lib/pages/workflows-with-new-search.svelte';
+
+  const columns = [
```
Comment (laurakwhit) [inline src/routes/(app)/namespaces/[namespace]/workflows/+page.svelte:10]: I've temporarily added this `columns` variable so that we can calculate `colspan` for the tables.

src/lib/components/workflow/workflows-summary-table-with-filters.svelte
```diff
@@ -140,35 +141,28 @@
             {/if}
           </div>
         </th>
-        <th class="table-cell md:w-auto" />
-        <th class="table-cell xl:w-60" />
-        <th class="hidden xl:table-cell xl:w-60" />
-        <th class="hidden xl:table-cell xl:w-60" />
+        {#each Array(columns.length - 1) as _, index}
```
laurakwhit, line 144: I've opted for this hack rather than updating the `colspan` based on a breakpoint since that would require either creating a reusable `BulkActions` component or duplicating code here.

We should probably rethink this once we decide what responsiveness will look like on the new workflows table.

src/lib/pages/workflows-with-new-search.svelte
```diff
@@ -42,6 +42,8 @@
     $supportsAdvancedVisibility,
   );
 
+  export let columns: string[];
```
rossedfort, line 45: @laurakwhit can we just define `columns` here? I don't think we need this to be a prop. The we can remove this change: https://github.com/temporalio/cloud-ui/pull/412/files#diff-7ccc1372d570c8ec2ce0c7fffb604f961c6c63ebe1ce97e4028f65533ac02fb7R6-R11

I'm changing how this will work in my new branch anyway. Since users will be able to configure the columns (and we'll persist that configuration), we can access the configuration via a store in any component where we need it.

laurakwhit: I had defined it higher up so we could pass it to the old search page to determine the `colspan`. But if we're not going to support the old search/table we can just hard code the `colspan` there and define it here.

I won't touch this here so I don't mess with your WIP branch, but feel free to change it there!

PR #1211, opened 2023-03-14
On network error, set report to false on error to prevent Sentry errors

Alex-Tideman:
## Description & motivation 💭 
We purposefully throw in network errors in our code, but we don't want to report them when being used in Cloud with Sentry. This sets the error.report to false for Sentry to ignore.

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/utilities/handle-error.ts
+++ b/src/lib/utilities/handle-error.ts
@@ -36,6 +36,7 @@ export const handleError = (
     });
     // Re-throw error to prevent other code from attempting to render
     errors.set(error);
+    (error as any).report = false;
     throw error;
   }
 };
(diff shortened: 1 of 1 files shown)


src/lib/utilities/handle-error.ts
```diff
@@ -36,6 +36,7 @@ export const handleError = (
     });
     // Re-throw error to prevent other code from attempting to render
     errors.set(error);
+    (error as any).report = false;
```
stevekinney, line 39: I'd recommend maybe we get a little tighter (e.g. `error as { report: boolean }`).

Alex-Tideman: 👍

PR #1212, opened 2023-03-15
Automatically strip leading/trailing whitespace on ChipInput

laurakwhit:
## Description & motivation 💭 
When entering chips with validation (e.g. email addresses when inviting a user) any extra space at the beginning or end is causing a validation error. 

This PR adds `trim()` to remove the whitespace from both ends for entered and pasted values.

n/a
### Design Considerations 🎨 
n/a
### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

- [ ] Verify entering an email with a space in front and/or on the end is valid
- [ ] Verify pasting a list of emails with a `, ` between is valid and works as expected

### Draft Checklist 
n/a
### Merge Checklist 
n/a
### Issue(s) closed 
* `DT-564`
### Any docs updates needed? 
n/a

### DIFF
--- a/src/lib/holocene/input/chip-input.svelte
+++ b/src/lib/holocene/input/chip-input.svelte
@@ -42,9 +42,10 @@
   });
 
   const handleKeydown = (e: KeyboardEvent) => {
-    if ((e.key === ',' || e.key === 'Enter') && displayValue !== '') {
+    const value = displayValue.trim();
+    if ((e.key === ',' || e.key === 'Enter') && value !== '') {
       e.preventDefault();
-      values.update((previous) => [...previous, displayValue]);
+      values.update((previous) => [...previous, value]);
       displayValue = '';
     }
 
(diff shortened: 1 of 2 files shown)


src/lib/holocene/input/chip-input.svelte
```diff
@@ -62,12 +63,16 @@
   const handlePaste = (e: ClipboardEvent) => {
     e.preventDefault();
     const clipboardContents = e.clipboardData.getData('text/plain');
-    values.update((previous) => [...previous, ...clipboardContents.split(',')]);
+    values.update((previous) => [
+      ...previous,
+      ...clipboardContents.split(',').map((content) => content.trim()),
```
rossedfort, line 68: Maybe we `.trim()` clipboardContents` too?

laurakwhit: I think we should be covered with just `content.trim()` as any leading/trailing whitespaces will just be added to the first and last "content".

rossedfort: Oh yeah, very true.

rossedfort (approved):

PR #1218, opened 2023-03-16
DT-67 - improve reason field on batch operation modal

rossedfort:
## Description & motivation 💭 
This improves the UX in the batch operation confirmation modal, specifically: 
- make reason optional
- provide placeholder with user email (if available) and Web UI context
- use placeholder if no reason is entered, and append placeholder if custom reason is entered

This is so that users don't have to enter a reason even though the API requires it. It also has the benefit of adding some more meaningful context to batch operations.

Screen Shot 2023-03-16 at 11 05 24 AM
Screen Shot 2023-03-16 at 11 05 31 AM
Screen Shot 2023-03-16 at 11 06 04 AM

### Design Considerations 🎨 
N/a
### How was this tested 👻 
- [x] Manual testing
- [x] E2E tests added
- [ ] Unit tests added

- with the Web UI running against temporal cli version >= 0.6.0, open the recent workflows page, select some workflows, and click the "Request Cancellation" or "Terminate" buttons in the table header
- Ensure the correct placeholder information is in the "reason" Input.
- Ensure that the reason is properly forwarded to the API when left blank or when a custom reason is entered.
### Issue(s) closed 
DT-67
## Docs
N/a

### DIFF
--- a/src/lib/components/workflow/batch-operation-confirmation-modal.svelte
+++ b/src/lib/components/workflow/batch-operation-confirmation-modal.svelte
@@ -3,7 +3,11 @@
   import Modal from '$lib/holocene/modal.svelte';
   import { pluralize } from '$lib/utilities/pluralize';
   import Input from '$lib/holocene/input/input.svelte';
-  export let action: 'Terminate' | 'Cancel';
+  import { authUser } from '$lib/stores/auth-user';
+
+  type Action = 'Terminate' | 'Cancel';
+
+  export let action: Action;
   export let allSelected: boolean;
   export let actionableWorkflowsLength: number;
   export let query: string;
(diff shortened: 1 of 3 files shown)


src/lib/components/workflow/batch-operation-confirmation-modal.svelte
```diff
@@ -17,25 +21,31 @@
     confirm: { reason: string };
   }>();
 
+  $: confirmText = action === 'Cancel' ? 'Confirm' : action;
+  $: placeholder = `${pastTense(action)} from Web UI${
+    $authUser?.email ? ` by ${$authUser.email}` : ''
```
Alex-Tideman, line 26: What about the case when it's local development and there is no authUser?

rossedfort:
> Do we want to append that message when there is a custom reason? Not sure if there's a hard requirement around that or not but I'd think some users would want to be able to set the message to exactly what they want.

Yeah great point, Matt noted in the JIRA ticket for this that we should allow users to customize this message, just a matter of whether we want to do it now or later. I'm game for doing it now, but it would add a bit of scope and this smaller change may not make it for next release.

src/lib/components/workflow/batch-operation-confirmation-modal.svelte
rossedfort: I think this should be handled with my ternary: `$authUser?.email ? ` by ${$authUser.email}` : ''`, If there is no email, the message will just be `(Terminated|Canceled) from Web UI`. Should we add an IP address or something?

Alex-Tideman: durrr my bad

Alex-Tideman (approved):

PR #1219, opened 2023-03-16
Fix back button not working as expected with recent workflows

laurakwhit:
## Description & motivation 💭 
When paginating through pages and/or adding search queries on `Recent Workflows`
* clicking "back" in the browser will take you to the previous page or query 
* clicking `Back to Workflows` on the workflow detail page will take you back to the most recent workflows page with the most recent queries

This makes it easier to navigate between workflows or investigate multiple workflows.

n/a
### Design Considerations 🎨 
n/a
### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

- Go to `Namespaces` > then `Recent Workflows` > then enter a query 
   - [x] Verify clicking the back btn in the browser takes you back to `Recent Workflows` without the query 
- Go to `Recent Workflows` > then enter a query and go to another page > go to the detail page for a workflow
   - [x] Verify clicking the `Back to Workflows` returns to the current `Recent Workflows` page with the query applied
- Go to `Recent Workflows` > then enter a query and change the number per page > go to the detail page for a workflow
   - [x] Verify clicking the `Back to Workflows` returns to the current `Recent Workflows` page with the query applied and the same number per page
- Go to `Recent Workflows` > then enter a query > click on `Workflows` (it should clear the query) > go to the detail page for a workflow
   - [x] Verify clicking the `Back to Workflows` returns to the current `Recent Workflows` page with no query applied
- Add filters & sorts to `Recent Workflows`
   - [x] Verify clicking the workflows icon in the nav clears all filters and sorts in the UI
   - [x] Verify clicking the logo in the nav clears all filters and sorts in the UI 
- Select a workflow filter type by clicking on the filter button > deselect and then click on the filter button for another type  > deselect and then click on the filter button for another type
   - [ ] Verify the value in the input for the `Type` dropdown is correct 
- Select a workflow filter id by clicking on the filter button > deselect and then click on the filter button for another id > deselect and then click on the filter button for another id
   - [ ] Verify the value in the input for the `ID` dropdown is correct 
### Draft Checklist 
n/a
### Merge Checklist 
n/a
### Issue(s) closed 
* `DT-3`
* `DT-291`
### Any docs updates needed? 
n/a

### DIFF
--- a/src/lib/components/workflow/dropdown-filter/workflow-type.svelte
+++ b/src/lib/components/workflow/dropdown-filter/workflow-type.svelte
@@ -33,7 +33,7 @@
 
   function handleShowInput(event: CustomEvent) {
     const show = event.detail.show;
-    if (show && typeFilter && !value) {
+    if (show && typeFilter) {
       value = typeFilter.value;
     } else if (show && !typeFilter && value) {
       value = '';
--- a/src/lib/layouts/workflow-header.svelte
+++ b/src/lib/layouts/workflow-header.svelte
@@ -3,11 +3,7 @@
   import { fly } from 'svelte/transition';
 
   import { autoRefreshWorkflow } from '$lib/stores/event-view';
-  import {
-    workflowsQuery,
-    workflowsSearch,
-    workflowsPage,
-  } from '$lib/stores/workflows';
+  import { workflowsSearchParams } from '$lib/stores/workflows';
   import { refresh, workflowRun } from '$lib/stores/workflow-run';
   import { eventHistory } from '$lib/stores/events';
 
@@ -17,9 +13,8 @@
     routeForStackTrace,
     routeForWorkers,
     routeForWorkflowQuery,
-    routeForWorkflowsWithQuery,
+    routeForWorkflows,
   } from '$lib/utilities/route-for';
-  import { toListWorkflowQuery } from '$lib/utilities/query/list-workflow-query';
 
   import Badge from '$lib/holocene/badge.svelte';
   import Copyable from '$lib/components/copyable.svelte';
@@ -48,9 +43,6 @@
     workflow: workflow?.id,
     run: workflow?.runId,
   };
-
-  const { parameters, searchType } = $workflowsSearch;
-  const query = toListWorkflowQuery(parameters);
 
   $: isRunning = $workflowRun?.workflow?.isRunning;
   $: activitiesCanceled = ['Terminated', 'TimedOut', 'Canceled'].includes(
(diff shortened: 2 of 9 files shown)


src/lib/layouts/workflow-header.svelte
```diff
@@ -99,6 +103,7 @@
         namespace,
         query: $workflowsQuery || query,
         search: searchType,
+        page: $workflowsPage,
```
laurakwhit, line 106: I've opted to just keep track of the page in state rather then use `history.back()` here since the user could get to the workflows detail page from a link and we want the `Back to Workflows` button to take us to the last workflows page the user was on instead of potentially to another url entirely.

Alex-Tideman:
I think the thing missing here is clearing the page param. When I do a query, page to page=2, then click an execution, then go back, page=2 is perserved. However when i clear the query or click the workflow button, then go to an execution and then go back, page=2 is still perserved.

Alex-Tideman:
The one issue that I still see if if you apply some filters and click the Workflow icon in the nav, the query is cleared but the UI still shows the filters applied.

src/lib/stores/filters.ts
```diff
@@ -2,7 +2,35 @@ import type {
...
+const parameters = derived([query], ([$query]) => {
+  return {
+    query: $query,
+  };
+});
+
+const updateFilters: StartStopNotifier<WorkflowFilter[]> = (set) => {
+  return parameters.subscribe(({ query }) => {
+    if (!query && get(workflowFilters).length) {
+      // Clear filters if there is no query
```
laurakwhit, line 20: > the query is cleared but the UI still shows the filters applied

Looks like this is an existing bug 😞 I've added some updates to the `filters` store to clear filters and sorts if no query exists in the page url. 

This should cover both use cases of clicking the workflows icon and clicking the logo for recent workflows.

Alex-Tideman: 👍 nice

Alex-Tideman (approved):

src/lib/components/workflow/dropdown-filter/workflow-type.svelte
```diff
@@ -33,7 +33,7 @@
 
   function handleShowInput(event: CustomEvent) {
     const show = event.detail.show;
-    if (show && typeFilter && !value) {
+    if (show && typeFilter) {
```
Alex-Tideman, line 36: I think we need to do the same for workflow-id

PR #1229, opened 2023-03-22
DT-545 - add/remove/reorder workflow table columns

rossedfort:
### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


Alex-Tideman:
Doing a quick check, two little UI things stand out.
1. Need to scroll to get the triple dot options on the table. Need to pin that to top right (or move it elsewhere) so it's always accessible.
2. The draggable column is a little wonky. I can't keep my cursor on the line to drag it.

Otherwise this is looking really good.

rossedfort:
> 1. Need to scroll to get the triple dot options on the table. Need to pin that to top right (or move it elsewhere) so it's always accessible.
I checked with Candito on this and she said it was OK as is, but I agree I think it'd be better to have it pinned. I'll see if I can get that working without too much effort.
> 2. The draggable column is a little wonky. I can't keep my cursor on the line to drag it.
I'm gonna compare to other resizable tables and see if there's some improvements we can make. I tried adding a hover state but it's tricky cause the line is an `::after` psuedoelement

src/lib/components/workflow/workflows-summary-configurable-table/workflows-summary-configurable-table.svelte
```diff
@@ -0,0 +1,471 @@
...
+    <table class="workflow-summary-table">
+      <thead>
+        <tr class="workflow-summary-header-row">
+          {#each otherColumns as column}
+            <WorkflowsSummaryTableHeaderCell {column} {sortDisabled} />
+          {/each}
+          <th
+            class="px-2 h-10 w-12 text-right sticky right-0 flex items-center justify-center bg-primary"
+          >
+            <IconButton
```
laurakwhit, line 259: Might we want to make the ellipsis button always hug the right? It floats to the left if there aren't enough headers.

Screenshot 2023-03-27 at 3 49 25 PM
Screenshot 2023-03-27 at 3 57 26 PM

src/lib/components/workflow/workflows-summary-configurable-table/workflows-summary-configurable-table.svelte
```diff
@@ -0,0 +1,471 @@
...
+  <div
+    class="workflow-summary-table-wrapper pinned"
+    class:batch-actions-enabled={$supportsBulkActions}
+    class:batch-actions-visible={showBulkActions}
+    class:no-columns-pinned={pinnedColumns.length === 0}
+    bind:this={resizableContainer}
+    style="width:{resizableContainerWidth}px;"
+    on:mousedown|stopPropagation|preventDefault={handleMouseDown}
+  >
+    <table class="workflow-summary-table pinned">
```
laurakwhit, line 155: It gets a little wonky on smaller screen sizes ⬇️ 

Screenshot 2023-03-27 at 3 52 23 PM

src/lib/components/workflow/workflows-summary-configurable-table/workflows-summary-configurable-table.svelte
```diff
@@ -0,0 +1,471 @@
...
+    ><Icon class="inline" name="hyphen" /><span
+      >), Workflow<br />Headings to personalize the Workflow List Table.</span
+    >
+  </svelte:fragment>
+  <div class="flex flex-col gap-4">
+    <OrderableList>
+      <svelte:fragment slot="heading">
+        Workflow Headings <span class="font-normal">(in view)</span>
+      </svelte:fragment>
+      {#each $workflowTableColumns as { label, pinned }, index (label)}
```
laurakwhit, line 321: Wondering if we want an empty state row here instead of it just becoming a collapsed border/line ⬇️ 

Screenshot 2023-03-27 at 3 57 02 PM

src/lib/components/workflow/workflows-summary-configurable-table/workflows-summary-table-body-cell.svelte
```diff
@@ -0,0 +1,90 @@
...
+{:else if label === 'Type' || label === 'Workflow ID'}
+  <td
+    class="workflows-summary-table-body-cell relative"
+    on:mouseover={showFilterOrCopy}
+    on:focus={showFilterOrCopy}
+    on:mouseleave={hideFilterOrCopy}
+    on:blur={hideFilterOrCopy}
+  >
+    <span class="table-link">{cellContent}</span>
+    <FilterOrCopyButtons
```
laurakwhit, line 73: These are easily hidden from view in pinned columns... Wondering if we should put them at the beginning of the table cell instead? 

Screenshot 2023-03-27 at 3 53 24 PM
Screenshot 2023-03-27 at 3 53 38 PM

laurakwhit (commented):
It's all coming together, so exciting 🎉 

Noticed a few UI things on the first pass, nothing major though!

src/lib/components/workflow/workflows-summary-configurable-table/workflows-summary-table-body-cell.svelte
Alex-Tideman: I'm still not totally sold on the resizable columns. If you wanted more space to see the other columns, shouldn't you remove a pinned column? Like in the above case, move Workflow ID out of the pinned columns.

rossedfort: Yeah, I think the resizable pinned section might be more complicated than it's worth. Removing it 👍

rossedfort: removed in [296e2d6](https://github.com/temporalio/ui/pull/1229/commits/296e2d663ea227a8f2918439e86c6deb0b06a9e3)

src/lib/components/workflow/workflows-summary-configurable-table/workflows-summary-configurable-table.svelte
rossedfort: Good call, added an empty state in [bf62d26](https://github.com/temporalio/ui/pull/1229/commits/bf62d2663408730ce68ceb5055b418c4441dcadb)

src/lib/components/workflow/workflows-summary-configurable-table/workflows-summary-configurable-table.svelte
rossedfort: Nice catch, should be fixed: [77446df](https://github.com/temporalio/ui/pull/1229/commits/77446df02b0101b19d63bd49364dfd5627d6e925)

src/lib/components/workflow/workflows-summary-configurable-table/workflows-summary-configurable-table.svelte
rossedfort: Yes, good catch, should be fixed: [296e2d6](https://github.com/temporalio/ui/pull/1229/commits/296e2d663ea227a8f2918439e86c6deb0b06a9e3)

src/lib/components/workflow/workflows-summary-configurable-table/workflows-summary-configurable-table.svelte
```diff
@@ -0,0 +1,468 @@
...
+          >No Custom Search Attributes</OrderableListItem
+        >
+      {/each}
+    </OrderableList>
+  </div>
+</Drawer>
+
+<style lang="postcss">
+  .workflow-summary-tables-wrapper {
+    @apply relative flex flex-row w-full rounded-lg border-primary border-2 overflow-scroll;
```
laurakwhit, line 369: Still not 100% sure if this is what we want here. Users still can't scroll ALL of the table on small screens (only the pinned section), but really not sure who would even be using that table on small screens anyways so not too worried about it.

src/lib/components/workflow/workflows-summary-configurable-table/workflows-summary-configurable-table.svelte
```diff
@@ -0,0 +1,468 @@
...
+      &:not(.pinned) {
+        :global(td) {
+          @apply last:rounded-br-lg;
+        }
+      }
+    }
+  }
+
+  .workflow-summary-configurable-row:hover {
+    @apply bg-gradient-to-br from-blue-100 to-purple-100 bg-fixed;
```
laurakwhit, line 462: It might be nice to have hover for the whole row (both pinned section and the rest of the table), but it might not be worth the hassle 😅

src/lib/components/workflow/workflows-summary-configurable-table/workflows-summary-configurable-table.svelte
```diff
@@ -0,0 +1,468 @@
...
+  }
+
+  .workflow-summary-table-wrapper {
+    @apply relative flex bg-white;
+
+    &.pinned {
+      @apply rounded-l-lg;
+
+      &::after {
+        @apply absolute right-0 content-[''] bg-primary w-[3px] h-full;
```
laurakwhit, line 379: Is this intentional or do we want `2px` here? Or maybe `3px` on the table border to match the other `Table` components?

src/lib/components/workflow/workflows-summary-configurable-table/workflows-summary-configurable-table.svelte
```diff
@@ -0,0 +1,468 @@
...
+                    error={$workflowError}
+                  />
+                {/if}
+              </td>
+            </tr>
+          {/each}
+        {:else}
+          <tr>
+            <td>
+              <EmptyState title="No column headers are in view">
```
laurakwhit, line 273: Maybe we want some padding here so the text isn't so close to the edge? 

Screenshot 2023-03-29 at 3 11 53 PM

src/lib/components/workflow/workflows-summary-configurable-table/workflows-summary-configurable-table.svelte
```diff
@@ -0,0 +1,468 @@
...
+                <td>
+                  <Checkbox
+                    hoverable
+                    bind:group={selectedWorkflows}
+                    value={workflow}
+                    disabled={allSelected}
+                  />
+                </td>
+              {/if}
+              {#each pinnedColumns as column}
```
laurakwhit, line 212: If there are no pinned columns and the user drags/clicks a column to the top it will automatically be pinned. However, if there are no visible columns and a user adds a a column it will **not** be automatically pinned.

Just wanted to confirm this is expected behavior 📌

laurakwhit (approved):
Few small comments, but I think we can address some things in a follow-up if needed.

Let's get this table out there 🚀 ✨

src/lib/components/workflow/workflows-summary-configurable-table/workflows-summary-configurable-table.svelte
rossedfort: Hah yeah I noticed this as well, but I'm not sure it's possible with css alone and could involve some hacky JS considering we would have to apply the bg gradient to rows in 2 separate tables.

src/lib/components/workflow/workflows-summary-configurable-table/workflows-summary-configurable-table.svelte
rossedfort: Yeah agreed that this is less than ideal. I'm struggling to come up with a solution that works elegantly across all screen sizes 😕

src/lib/components/workflow/workflows-summary-configurable-table/workflows-summary-configurable-table.svelte
rossedfort: That said, I almost prefer the separate table row hover states cause it's a further indication that one section is "pinned" while the other can be scrolled 🤷

PR #1238, opened 2023-03-24
DRAFT: Add option to include headers for codec server

Alex-Tideman:
## Description & motivation 💭 
Allow customization of headers for the codec server by injecting a script onMount and setting globalThis

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

### DIFF
--- a/src/lib/services/data-encoder.ts
+++ b/src/lib/services/data-encoder.ts
@@ -7,6 +7,41 @@ import { stringifyWithBigInt } from '$lib/utilities/parse-with-big-int';
 
 import type { Payloads } from '$types';
 
+export function setCodecServerHeaders(
+  namespace: string,
+  settings: Settings,
+  accessToken: string,
+) {
+  let validationFailed = false;
+  const endpoint = settings?.codec?.endpoint;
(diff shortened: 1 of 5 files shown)


src/lib/services/data-encoder.ts
```diff
@@ -7,34 +7,60 @@ import { stringifyWithBigInt } from '$lib/utilities/parse-with-big-int';
...
+  accessToken: string,
+) {
+  let validationFailed = false;
   const endpoint = settings?.codec?.endpoint;
   const passAccessToken = settings?.codec?.passAccessToken;
 
   const headers = {
     'Content-Type': 'application/json',
     'X-Namespace': namespace,
+    credentials: 'same-origin',
```
GiantRobots, line 22: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS

I think if this is set at all it preflights the request which was the breaking change last time so we should triple check that setting this doesn’t break everything again

Alex-Tideman: Aight ill remove this

PR #1243, opened 2023-03-27
Remove key block for event-summary

Alex-Tideman:
## Description & motivation 💭 
With the SvelteKit upgrade, using the {#key [$eventFilterSort, category]} block on the event-summary caused rendering issues. This PR removes the key block, and the event history renders correctly on filter changes.

**BEFORE**
Screen Shot 2023-03-27 at 11 25 32 AM

**AFTER**
Screen Shot 2023-03-27 at 11 26 15 AM

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/event/event-summary.svelte
+++ b/src/lib/components/event/event-summary.svelte
@@ -77,30 +77,28 @@
   $: updating = currentEvents.length && !fullHistory.length;
 </script>
 
-{#key [$eventFilterSort, category]}
-  <Pagination
-    floatId="event-view-toggle"
-    {items}
-    {updating}
-    let:visibleItems
-    let:activeRowIndex
-    let:setActiveRowIndex
-    aria-label="recent events"
-  >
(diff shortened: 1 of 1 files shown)


src/lib/components/event/event-summary.svelte
```diff
@@ -77,30 +77,28 @@
...
-          active={activeRowIndex === index}
-          onRowClick={() => setActiveRowIndex(index)}
-        />
-      {:else}
-        <EventEmptyRow {loading} />
-      {/each}
-    </EventSummaryTable>
-  </Pagination>
-{/key}
+<Pagination
```
laurakwhit, line 80: Looks like filtering by `Workflow Events` is not working as expected 😞 The first time a filter is selected it works, but any ones after are not updating events.

| List | Filter |
|--|--|
|Screenshot 2023-03-27 at 10 33 56 AM|Screenshot 2023-03-27 at 10 34 11 AM|

laurakwhit (changes requested):

PR #1245, opened 2023-03-27
Add TEMPORAL_CODEC_INCLUDE_CREDENTIALS env var to pass credentials header

Alex-Tideman:
## Description & motivation 💭 
Add TEMPORAL_CODEC_INCLUDE_CREDENTIALS env var to allow passing credentials: 'include" header in codec server api requests.

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/server/docker/config_template.yaml
+++ b/server/docker/config_template.yaml
@@ -45,5 +45,6 @@ auth:
 codec:
   endpoint: {{ default .Env.TEMPORAL_CODEC_ENDPOINT "" }}
   passAccessToken: {{ default .Env.TEMPORAL_CODEC_PASS_ACCESS_TOKEN "false" }}
+  includeCredentials: {{ default .Env.TEMPORAL_CODEC_INCLUDE_CREDENTIALS "false" }}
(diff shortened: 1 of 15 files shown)


server/docker/config_template.yaml
```diff
@@ -45,5 +45,6 @@ auth:
 codec:
   endpoint: {{ default .Env.TEMPORAL_CODEC_ENDPOINT "" }}
   passAccessToken: {{ default .Env.TEMPORAL_CODEC_PASS_ACCESS_TOKEN "false" }}
+  includeCredentials: {{ default .Env.TEMPORAL_CODEC_INCLUDE_CREDENTIALS "false" }}
```
GiantRobots, line 48: I don't quite agree with supporting this everywhere maybe we can add something like "UNSUPPORTED" Or "DANGEROUS"?

Alex-Tideman: I'll let @robholland and @feedmeapples weigh in on the naming for this env var

PR #1257, opened 2023-03-29
DT-646 - add Reapply Type to Reset

### LINKED ISSUE #1240: Add Reapply Type selector to Reset Workflow pop-up

We sometimes want to reset a workflow and replay it up to some specified event, but without applying signals that happened after this event (which seems to be the default). The reason is that the signal payload might be invalid and subsequent workflow tasks that read this signal fail.

Add Reapply Type selector to the Reset Workflow pop-up (equivalent to the `tctl workflow reset --reapply-type` command line option)

## Describe alternatives you've considered

Use tctl

## Additional context

-

rossedfort:
## Description & motivation 💭 
This PR adds a new Select menu in the Workflow Reset form that allows users to specify a Reset Reapply Type. The valid options are "All" (which is the default), "Signal", and "None". This controls which event types are reapplied in the new workflow after the reset point.

see https://docs.temporal.io/tctl-v1/workflow#--reset_reapply_type for more

Screen Shot 2023-03-29 at 3 05 43 PM

### Design Considerations 🎨 
N/a - approved by @canvanooo

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

1. On the workflow details page for a running workflow, open the dropdown menu of the workflow actions split button, and click "Reset". 
2. Confirm the "Reapply Type" select menu is visible, and that changing it has an effect on the reapply type that is sent in the request to reset the workflow when the form is submitted.

### Issue(s) closed 
closes #1240 
## Docs
Docs Team have been informed of the changes


src/lib/components/workflow/workflow-reset-form.svelte
```diff
@@ -3,11 +3,14 @@
...
   import Select from '$lib/holocene/select/select.svelte';
   import { ResetType } from '$lib/models/workflow-actions';
+  import { ResetReapplyType } from '$types';
 
   const DEFAULT_RESET_ID_HINT_TEXT =
     'Only Workflow Events of type WorkflowTaskCompleted, WorkflowTaskFailed, or WorkflowTaskTimeout are supported.';
 
   export let resetType: ResetType;
+  export let resetReapplyType: ResetReapplyType;
+  ResetReapplyType.RESET_REAPPLY_TYPE_UNSPECIFIED;
```
Alex-Tideman, line 13: Is this supposed to be hanging out here?

src/lib/models/workflow-actions.ts
```diff
@@ -3,3 +3,9 @@ export enum ResetType {
   LastWorkflowTask = 1,
   EventId = 2,
 }
+
+export enum ResetReapplyType {
```
laurakwhit, line 7: Instead of adding an enum here, what do you think about using the proto definitions directly? E.g. 

```js
  import { temporal } from '@temporalio/proto';
  temporal.api.enums.v1.ResetReapplyType
```

laurakwhit (approved):
One comment below, but otherwise lgtm 🚀 

#### Testing 🧪
- [x] Confirm the "Reapply Type" select menu is visible, and that changing it has an effect on the reapply type that is sent in the request to reset the workflow when the form is submitted.

PR #1259, opened 2023-03-30
Fix a visual bug with configurable table + batch actions

rossedfort:
### before 🤢 
Screen Shot 2023-03-30 at 12 02 56 PM

### after 🤩 
Screen Shot 2023-03-30 at 12 13 36 PM

### Design Considerations 🎨 
N/a
### How was this tested 👻 
Also tested when batch actions are not supported
- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

## Checklists
N/a
## Docs
N/a


Alex-Tideman (approved):

src/lib/components/workflow/workflows-summary-configurable-table/workflows-summary-configurable-table.svelte
```diff
@@ -393,10 +385,6 @@
         &:not(.batch-actions-enabled) {
```
laurakwhit, line 385: Looks like we can still select in this state 😞 

Screenshot 2023-03-30 at 12 29 56 PM

laurakwhit (approved):

PR #1278, opened 2023-04-11
Lint holocene

stevekinney:
- Opts all Holocene components into linting.
- Fixes all lint errors in Holocene.

### DIFF
--- a/src/lib/holocene/button.svelte
+++ b/src/lib/holocene/button.svelte
@@ -3,7 +3,7 @@
   import type { IconName } from '$lib/holocene/icon/paths';
   import Badge from '$lib/holocene/badge.svelte';
 
-  export let disabled: boolean = false;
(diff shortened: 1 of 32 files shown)


src/lib/holocene/button.svelte
```diff
@@ -3,30 +3,30 @@
...
-  export let count: number = 0;
-  export let type: string = 'button';
-  export let unround: boolean = false;
-  export let unroundRight: boolean = false;
-  export let unroundLeft: boolean = false;
+  export let count = 0;
+  export let type = 'button';
+  export let unround = false;
+  export let unroundRight = false;
+  export let unroundLeft = false;
```
rossedfort, line 29: I kind of like explicit types even when they can be inferred, but alas

rossedfort (approved):

PR #1289, opened 2023-04-13
Add esm-env and replace browser imports

laurakwhit:
## Description & motivation 💭 

Screenshot 2023-04-13 at 5 47 43 PM

Instead of using `import { browser } from '$app/environment'` which is tied to SvelteKit, use `import { BROWSER } from 'esm-env'` so packages can work with any Svelte project.


rossedfort (approved):
lgtm, can we add a lint rule to prevent us from importing from `$app/environment` ? Or will there be an obvious error/failure if we do?

laurakwhit:
It will not currently error if we use `$app/environment`. We could definitely add a lint rule with a suggestion. Do we have any precedent for custom lint rules?

rossedfort:
lgtm 💯

.eslintrc.cjs
```diff
@@ -68,5 +68,16 @@ module.exports = {
       'error',
       { assertionStyle: 'as' },
     ],
+    '@typescript-eslint/no-restricted-imports': [
+      'error',
+      {
+        paths: [
+          {
+            name: '$app/environment',
+            message: 'Please use esm-env instead.',
```
Alex-Tideman, line 77: 🪄

PR #1300, opened 2023-04-17
DT-680 - improve accessibility of Tab component(s)

rossedfort:
## Description & motivation 💭 
Rewrites `<Tab />` component, and adds `<Tabs />`, `<TabList />`, and `<TabPanel />` components, which properly implement the `tab`, `tablist`, and `tabpanel` roles: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role#example

Also abstracts the logic of hiding/showing the `<TabPanel />` into context, inspired by https://svelte.dev/repl/8e68120858e5322272dc9136c4bb79cc?version=3.5.1
N/A
### Design Considerations 🎨 
N/A
### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

## Checklists
N/A
### Issue(s) closed 
N/A
## Docs
N/A


src/lib/holocene/tab/tab.svelte
```diff
@@ -0,0 +1,71 @@
...
+
+  const { registerTab, selectTab, activeTab } = getContext<TabContext>(TABS);
+
+  registerTab(id);
+
+  $: isActive = href ? active : $activeTab === id;
+
+  const handleClick = () => {
+    selectTab(id);
+    if (href) goto(href);
```
rossedfort, line 32: Wondering how people feel about using `goto` in a holocene component. I think I prefer it over a component that renders either an `<a />` or a `<button />`, but it still doesn't seem great.

Alex-Tideman: I think it's fine to use goto here

Alex-Tideman:
I'm assuming we'll need to have the corresponding cloud PR for this refactor too

rossedfort:
> I'm assuming we'll need to have the corresponding cloud PR for this refactor too

Yep, good call, luckily there are only 2 `<Tab />`'s in cloud 👍

Alex-Tideman (approved):

src/lib/holocene/tab/tab.svelte
```diff
@@ -0,0 +1,71 @@
...
+    href?: string;
+    panelId?: string;
+    disabled?: boolean;
+    active?: boolean;
+  }
+
+  export let label: string;
+  export let id: string;
+  export let href = '';
+  export let panelId = null;
```
laurakwhit, line 20: No action item/solution here that I can think of atm... Just wish there was a way we could ensure `panelId`/`tabId` was the same as the `id` in `tab-panel`/`tab` 🤔

laurakwhit (commented):
Can't get histoire to load in order to try out `tab.story.svelte` 😞 but other than that lgtm 🚀

src/lib/holocene/tab/tab.svelte
rossedfort: This was actually caused by importing `goto` in the Tab component. I've refactored it in [e8c35b9](https://github.com/temporalio/ui/pull/1300/commits/e8c35b9883f81a72dfdd2c286dbe9503aa71a417), but using `<svelte:element />` instead of an if/else

rossedfort:
> Can't get histoire to load in order to try out `tab.story.svelte` 😞 but other than that lgtm 🚀

This was actually caused by importing goto in the Tab component. I've refactored it in [e8c35b9](https://github.com/temporalio/ui/pull/1300/commits/e8c35b9883f81a72dfdd2c286dbe9503aa71a417), but using <svelte:element /> instead of an if/else

src/lib/holocene/tab/tab.svelte
```diff
@@ -0,0 +1,76 @@
...
+
+  $: isActive = isNull(active) ? $activeTab === id : active;
+
+  const handleClick = () => {
+    selectTab(id);
+  };
+</script>
+
+<svelte:element
+  this={href ? 'a' : 'button'}
```
laurakwhit, line 40: 💅

laurakwhit (approved):

PR #1321, opened 2023-04-20
fix some small visual bugs with configurable table

rossedfort:
## Description & motivation 💭 
fix some small visual bugs with configurable table
- properly display batch actions when there are no columns pinned
- hide resize handle when batch actions are disabled and there are no pinned columns
#### before
Screen Shot 2023-04-20 at 3 45 14 PM
Screen Shot 2023-04-20 at 3 45 38 PM

#### after
Screen Shot 2023-04-20 at 3 44 52 PM
Screen Shot 2023-04-20 at 3 45 56 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

Run oss locally and visually verify changes
## Checklists
N/a
### Issue(s) closed 
N/a
## Docs
N/a

### DIFF
--- a/src/lib/components/workflow/workflows-summary-configurable-table/workflows-summary-configurable-table.svelte
+++ b/src/lib/components/workflow/workflows-summary-configurable-table/workflows-summary-configurable-table.svelte
@@ -257,6 +257,7 @@
   </div>
   <div
     class="resizer"
+    class:batch-actions-enabled={$supportsBulkActions}
     class:no-columns-pinned={pinnedColumns.length === 0}
     class:batch-actions-visible={showBulkActions}
     on:mousedown|stopPropagation|preventDefault={handleMouseDown}
@@ -420,6 +421,10 @@
     @apply z-10 bg-primary w-0 border-r-[3px] border-primary cursor-col-resize;
   }
 
+  .resizer.no-columns-pinned:not(.batch-actions-enabled) {
+    @apply hidden;
+  }
+
   .resizer.no-columns-pinned,
   .resizer.batch-actions-visible {
     @apply pointer-events-none;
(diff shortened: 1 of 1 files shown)


src/lib/components/workflow/workflows-summary-configurable-table/workflows-summary-configurable-table.svelte
```diff
@@ -420,6 +421,10 @@
     @apply z-10 bg-primary w-0 border-r-[3px] border-primary cursor-col-resize;
   }
 
+  .resizer.no-columns-pinned:not(.batch-actions-enabled) {
```
laurakwhit, line 424: I would say just hide the resizer when batch actions are enabled too
```suggestion
  .resizer.no-columns-pinned{
```

but I think the hover state is maybe a bit of a blocker for now. What do you think?

Screenshot 2023-04-20 at 4 12 01 PM

src/lib/components/workflow/workflows-summary-configurable-table/workflows-summary-configurable-table.svelte
```diff
@@ -444,7 +449,7 @@
   }
 
   .workflow-summary-table-wrapper.pinned.no-columns-pinned.batch-actions-visible {
```
laurakwhit, line 451: We might want to consider a solution for this overlap in the future

Screenshot 2023-04-20 at 4 05 36 PM

laurakwhit (approved):

src/lib/components/workflow/workflows-summary-configurable-table/workflows-summary-configurable-table.svelte
rossedfort: Hm yeah I think just leaving the divider/resizer is OK for now, but I agree that it'd be nicer to hide it an fix the hover state in this case.

src/lib/components/workflow/workflows-summary-configurable-table/workflows-summary-configurable-table.svelte
rossedfort: I think this is OK given the complexity of this table. Ideally we have less stuff in the table headers IMO 😂

PR #1322, opened 2023-04-20
Make Codeblocks async by default or serially if they want

GiantRobots:
## Description & motivation 💭 
Move prism css/cjs files to static directory so we can import them with script tags in app.html. This allows us to use async functionality with web workers in prism because script tags have reference to its own filename while imports don't.

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Merge Checklist 

- [ ] Test with public path feature


rossedfort (dismissed):

rossedfort:
lgtm pending type error fix

Alex-Tideman:
The element doesn't get highlighted when async=true. We need to include a web worker file. I'll look into this more today.

stevekinney (approved):

PR #1325, opened 2023-04-21
Add VITE_APP_API_URL

laurakwhit:
## Description & motivation 💭 
Add additional env. var (`VITE_APP_API_URL`) option for endpoint in `getApiOrigin`.

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


Alex-Tideman (approved):

Alex-Tideman (changes requested):
We shouldn't need this code. The api from cloud should come from the namespace api response.

PR #1330, opened 2023-04-24
Make Start/End Time Filter Sticky

laurakwhit:
## Description & motivation 💭 
Time preference (e.g. last `15 Minutes`, `3 Days`, `All Time`…) is persisted so that when accessing Temporal or switching Namespaces, it applies the time preference on the `Recent Workflows` page.

### How was this tested 👻 

- [x] Manual testing
- [x] E2E tests added
- [ ] Unit tests added

- [ ] Verify time filter is persisted when
   - changing namespaces
   - navigating to recent workflows 
- [ ] Verify the time filter is properly applied in the 
   - time filter dropdown
   - manual search input
   - url
- [ ] Verify the time filter is **not** applied if the user
   - enters a custom manual search
   - navigates to a workflow url with a query
  
### Issue(s) closed 
* `DT-14`

### DIFF
--- a/src/lib/components/workflow/dropdown-filter/workflow-datetime-filter.svelte
+++ b/src/lib/components/workflow/dropdown-filter/workflow-datetime-filter.svelte
@@ -15,7 +15,11 @@
   import Select from '$lib/holocene/select/select.svelte';
   import Option from '$lib/holocene/select/option.svelte';
   import SimpleSplitButton from '$lib/holocene/simple-split-button.svelte';
-  import { workflowFilters, workflowSorts } from '$lib/stores/filters';
+  import {
+    persistedTimeFilter,
+    workflowFilters,
+    workflowSorts,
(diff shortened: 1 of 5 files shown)


src/lib/components/workflow/dropdown-filter/workflow-datetime-filter.svelte
```diff
@@ -53,9 +57,18 @@
       value = 'All Time';
       timeField = 'StartTime';
     } else {
-      value = custom ? 'Custom' : timeFilter.value;
+      value =
+        custom || !columnOrderedDurations.includes(timeFilter?.value)
```
laurakwhit, line 61: Currently, if a user enters a search or the url has a query that isn't included in `columnOrderedDuration` the value shows up as `undefined` in the dropdown. This change defaults it to `Custom`.

stevekinney (approved):

PR #1339, opened 2023-05-01
Remove migration notice from migrated README

jdocherty:
## Description & motivation 💭 

The README.md for ui-server had some left over text that lead to me recursively clicking a link a couple of times.

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


robholland:
The problem is this README gets copied to the ui-server repo, where that link makes sense... Not sure what to do there...

feedmeapples (approved):

feedmeapples:
> The problem is this README gets copied to the ui-server repo, where that link makes sense... Not sure what to do there...

I think it should be fine as long as we mention "For contributions follow this" and a link to the UI repo

PR #1351, opened 2023-05-08
DT-512 - add proper keyboard interactivity to Select component

rossedfort:
## Description & motivation 💭 

- Refactor `<Select />` to not use the Menu primitives so that we can better customize the behavior to be the same as that of a native html `<select />` element.
  - Add a `:focus` state on the select dropdown
  - Add a `:focus` state on the select options
  - `Space` or `ArrowDown` opens the select dropdown and focuses the selected option if it exists, otherwise focuses the first option
  - `ArrowUp` and `ArrowDown` focus the option above and below the current option respectively
  - `Enter` or `Space` select the focused option
  - `Escape` closes the select dropdown
- Refactor Menu components (`<MenuContainer />`, `<MenuButton />`, `<Menu />`, `<MenuItem />`) to use context for opening/closing the menu, which removes the dependency on the `trigger-menu` directive, which had some buggy behavior. Additionally, a better focus state was added to `<Menu />` and `<MenuItem />`.
- Switch `<WorkflowDatetimeFilter />` from using a `<Select />` to using Menu components.
- Remove `<SimpleSplitButton />` and fix an improper use of `<SplitButton />` in schedules-view.

https://user-images.githubusercontent.com/11775628/236952001-ca8531e4-064e-42e1-a9ba-e225fba03c63.mov

### Design Considerations 🎨 

- Need sign-off on focus states as there are no designs

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


rossedfort:
> The datetime select for workflow filter is no bueno now. Totally fine if this needs to be it's own component or if we should keep the ability to have a custom width.

Gah, nice catch. I wonder if a better approach would be for `<Select />` always to fill the width of its parent and just make sure the parent here has a width. Not sure I love passing a tailwind class to Select to set its width.

rossedfort:
@Alex-Tideman on second thought, it doesn't seem like this should be a  Select  but should actually be a Dropdown Menu

Alex-Tideman:
Two small things we can review tomorrow: 

1. When you pick a custom date in the workflow date filter, the menu closes.
2. Namespace select keyboard (up/down arrows and tabbing) doesn't work.

Alex-Tideman (approved):

PR #1352, opened 2023-05-09
Allow decoding event history for download and for JSON view

Alex-Tideman:
## Description & motivation 💭 
Add env var TEMPORAL_CODEC_DECODE_EVENT_HISTORY_DOWNLOAD which allows users to download the event history decoded instead of in the raw format (encoded). This env var is set to true in the development.yaml but false in config-template.yaml.

Add a toggle switch to view event history decoded in the JSON view. Default is set to view decoded.

Screen Shot 2023-05-09 at 9 41 25 AM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/pages/workflow-history-json.svelte
+++ b/src/lib/pages/workflow-history-json.svelte
@@ -36,7 +36,7 @@
     <label
       for="decode-event-history"
       class="flex items-center gap-4 font-secondary text-sm"
-      >View Decoded Event History<ToggleSwitch
+      >View decoded event history<ToggleSwitch
         id="decode-event-history"
         bind:checked={decodeEventHistory}
         data-testid="decode-event-history-toggle"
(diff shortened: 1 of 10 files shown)


src/lib/pages/workflow-history-json.svelte
```diff
@@ -3,21 +3,45 @@
...
 </script>
 
 {#await events}
   <Loading />
 {:then events}
+  <div class="flex justify-end">
+    <label
+      for="decode-event-history"
+      class="flex items-center gap-4 font-secondary text-sm"
+      >View Decoded Event History<ToggleSwitch
```
laurakwhit, line 39: Not sure if this is what we want here, but noticed elsewhere (e.g. `Auto refresh`) we lowercase following words. 

```suggestion
      >View decoded event history<ToggleSwitch
```

laurakwhit (approved):
- [x] Manual testing

PR #1362, opened 2023-05-12
DT-824 - set up i18n

rossedfort:
## Description & motivation 💭 
This PR lays a foundation for internationalization of the UI using `i18next`.

- add `i18next`, `i18next-http-backend`, and `i18next-browser-languagedetector` dependencies
- add the `i18n.init()` script to `src/routes/+layout.ts` and set the `i18n` store on `$page.data`
- add a `<Translate />` component
- add a script to generate JSON from TypeScript locale files so that we can have strongly typed keys

N/a
### Design Considerations 🎨 
N/a
### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

## Checklists
N/a
## Docs
N/a


scripts/generate-locales.ts
```diff
@@ -0,0 +1,118 @@
+import ts from 'typescript';
+import path, { relative } from 'path';
+import { writeFile, mkdir, stat } from 'fs/promises';
+import { glob, chalk } from 'zx';
+
+const SRC_DIR = path.resolve('./src/lib/i18n/locales');
+const DEST_DIR = path.resolve('./static/i18n/locales');
+
+const logAndExit = (msg: string, logError: (msg: string) => void) => {
```
stevekinney, line 9: We copied this from the Temporal CLI download script right? I wonder if we should break this out and put it… somewhere? (Not for this PR, of course.)

scripts/generate-locales.ts
```diff
@@ -0,0 +1,118 @@
...
+  logError(chalk.red(msg));
+
+  if (logError === console.error) {
+    process.exit(0);
+  }
+};
+
+const relativeTo = (path: string) => {
```
stevekinney, line 17: It seems a bit much, but I'm into it.

scripts/generate-locales.ts
```diff
@@ -0,0 +1,118 @@
...
+    const checker = program.getTypeChecker();
+
+    if (!sourceFile) {
+      logAndExit(
+        `Error retrieving source file information for ${source}`,
+        logError,
+      );
+    }
+
+    const sourceFileSymbol = checker.getSymbolAtLocation(sourceFile!);
```
stevekinney, line 41: Why the `!`?

scripts/generate-locales.ts
```diff
@@ -0,0 +1,118 @@
...
+    const defaultExport = checker.tryGetMemberInModuleExports(
+      'default',
+      sourceFileSymbol!,
+    );
+
+    if (!defaultExport) {
+      logAndExit(`No default export in source: ${source}`, logError);
+    }
+
+    const jsObject: Record<string, string> = {};
```
stevekinney, line 56: I hate the name of this object, but naming stuff is hard.

scripts/generate-locales.ts
```diff
@@ -0,0 +1,118 @@
...
+    if (defaultExport?.declarations && defaultExport.declarations[0]) {
+      defaultExport.declarations[0].forEachChild((child) => {
+        if (child.kind === ts.SyntaxKind.ObjectLiteralExpression) {
+          child.forEachChild((grandchild) => {
+            if (grandchild.kind === ts.SyntaxKind.PropertyAssignment) {
+              let key = '';
+              let value = '';
+              const [keyNode, valueNode] = [
```
stevekinney, line 64: I don't feel strongly about this. But if you're going to pop it into an array just to destructure it and that takes four lines, maybe you just do two variable assignments?

src/lib/i18n/keys.ts
```diff
@@ -0,0 +1,24 @@
...
+import type DEKeys from './locales/de/common';
+
+/**
+ * https://www.i18next.com/translation-function/plurals#singular-plural
+ * when translating strings that could be singular or plural, i.e. "0 Apples" or "1 Apple"
+ * i18next expects the keys to be suffixed with `_zero`, `_one`, or `_other` for 0, 1, or n > 1 items respectively.
+ * If more suffixes are needed, i.e. `_few`, add them here.
+ */
+type WithoutPluralSufix<T> = T extends
+  | `${infer P}_zero`
```
stevekinney, line 12: `*slow clap*`

scripts/generate-locales.ts
rossedfort: No, it's slightly different because this script can be run via `pnpm` or as a vite plugin during the build/serve commands. Vite plugins' Context have an `error` function: https://rollupjs.org/plugin-development/#this-error so I'm passing it through to here instead of using `console.error` when running as a vite plugin. This way any vite cleanup stuff can still happen normally instead of immediately exiting the process.

scripts/generate-locales.ts
rossedfort: Lol, I think I wrote it before I knew it was going to be a one liner.

scripts/generate-locales.ts
rossedfort: Telling the TypeScript compiler that we do have a defined `sourceFile` here, because we are checking for its truthiness in the lines above. https://github.com/temporalio/ui/pull/1362/files#diff-c113f7c7730c2f6cf15e9c61f2bb2be52ea8921825a0dd8f1aac12eed4f75a22R34-R39

I could refactor the code to avoid these assertions, but then I think it becomes slightly less readable.

scripts/generate-locales.ts
rossedfort: I can do better 😂

scripts/generate-locales.ts
rossedfort: Actually I can just add some `return`'s so that the compiler is happy, will do that.

PR #1366, opened 2023-05-15
DT-796 - reset batch checkboxes when workflows list is updating

rossedfort:
#### before

https://github.com/temporalio/ui/assets/11775628/bfc906f5-89ac-4c16-8ddc-d8d4cf9e2fc8
#### after

https://github.com/temporalio/ui/assets/11775628/3d45158d-f544-4e9a-96d4-1675e374d523

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

- Select the Checkbox in the workflows list table header
- trigger any update to the workflows list, i.e. new query, refresh, new filter.
- ensure batch action checkboxes are reset

### DIFF
--- a/src/lib/pages/workflows-with-new-search.svelte
+++ b/src/lib/pages/workflows-with-new-search.svelte
@@ -218,7 +218,7 @@
   );
 
   $: {
-    if ($workflowFilters || ($pageSelected && $updating)) {
+    if ($workflowFilters || $updating) {
       resetSelection();
     }
   }
--- a/src/lib/pages/workflows-with-new-search.svelte
+++ b/src/lib/pages/workflows-with-new-search.svelte
@@ -218,7 +218,7 @@
   );
 
   $: {
-    if ($workflowFilters || (pageSelected && $updating)) {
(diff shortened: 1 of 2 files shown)


src/lib/pages/workflows-with-new-search.svelte
```diff
@@ -218,7 +218,7 @@
   );
 
   $: {
-    if ($workflowFilters) {
+    if ($workflowFilters || (pageSelected && $updating)) {
```
laurakwhit, line 221: I think we want the store value (`$pageSelected`) here
```suggestion
    if ($workflowFilters || ($pageSelected && $updating)) {
```

laurakwhit (changes requested):

src/lib/pages/workflows-with-new-search.svelte
rossedfort: Ah good catch! Yeah we want `$pageSelected`, it's just a small optimization so that we only reset the checkboxes if they're checked in the first place.

PR #1367, opened 2023-05-15
Add a label to namespace search

stevekinney:
Adds a label to the namespace search


rossedfort:
I think we should switch this to use a Holocene `<Input />`, then when we do the work to enforce labels, we can make the necessary updates here.

rossedfort (approved):

PR #1371, opened 2023-05-16
DT-859 - replace eventId text input with selectable list

rossedfort:
## Description & motivation 💭 
Since we can calculate a list of ID's that are valid points to reset a workflow to, we should show this list to users when they are resetting a workflow instead of presenting them with a free text input.

https://github.com/temporalio/ui/assets/11775628/a0c206ac-5dd3-4ec8-8207-f90ce4c2631f

### Design Considerations 🎨 
- Need to confirm radio button UI w/ @tigernaut 
### How was this tested 👻 

- [x] Manual testing
- [x] E2E tests added
- [ ] Unit tests added

### Merge Checklist 
- Confirm radio button design


cypress/integration/workflow-actions.spec.js
```diff
@@ -96,47 +96,11 @@ describe('Workflow Actions', () => {
...
-    });
-
     it('to an arbitrary workflow task if the workflow is running, and the action is enabled', () => {
       cy.get('#workflow-actions-menu-button').click();
       cy.get('#workflow-actions-menu >> [data-testid="reset-button"]').click();
-      cy.get('[data-testid="workflow-reset-type-select"]').click();
-      cy.get(
-        '#reset-type-select-options > li[role="option"]:nth-child(3)',
-      ).click();
+      cy.get('#reset-event-5').click();
```
stevekinney, line 102: Not a blocker, but how do you feel about this value being hard-coded?

src/lib/components/workflow-actions.svelte
```diff
@@ -200,14 +187,23 @@
       return 'Resetting workflows is not enabled, please contact your administrator for assistance.';
     if (resetEnabled && workflow?.pendingChildren?.length > 0)
       return 'Cannot reset workflows with pending children.';
+    if (
```
stevekinney, line 190: I would say that this entire set of conditions should be broken out into a utility function and unit tested. We can (and should) make that a task.

src/lib/components/workflow-actions.svelte
```diff
@@ -200,14 +187,23 @@
...
+      resetEnabled &&
+      workflow?.pendingChildren?.length === 0 &&
+      $resetEvents.length === 0
+    )
+      return 'Cannot reset workflows without WorkflowTaskStarted, WorkflowTaskCompleted, or WorkflowTaskTimedOut events.';
   };
 
   $: workflowActions = [
```
stevekinney, line 198: I feel the same way about this.

src/lib/components/workflow/workflow-reset-form.svelte
```diff
@@ -2,35 +2,18 @@
...
-      value: ResetType.LastWorkflowTask,
-      label: 'Last workflow task',
-    },
-    {
-      value: ResetType.EventId,
-      label: 'Event ID',
-    },
-  ];
+  $: {
+    if ($resetEvents && $resetEvents.length > 0) {
```
stevekinney, line 13: Could `resetEvents` provide this? Why does this need to happen here?

src/lib/components/workflow/workflow-reset-form.svelte
```diff
@@ -46,36 +29,36 @@
...
-    {#each resetTypes as { value, label }}
-      <Option {value}>
-        {label}
-      </Option>
+  <ul class="w-full rounded border border-primary max-h-40 overflow-y-scroll">
+    {#each $resetEvents as event}
+      <li
+        class="w-full h-10 border-b border-primary last-of-type:border-none hover:bg-gradient-to-br from-blue-100 to-purple-100 bg-fixed first-of-type:rounded-t last-of-type:rounded-b"
+      >
+        <label
```
stevekinney, line 40: I love the wrapping an `<input>` in a `<label>` tag as much as the next person, but we should just keep them in parallel and use the `for` attribute and an `id` on the input.

cypress/integration/workflow-actions.spec.js
rossedfort: Yeah it's not ideal, but it's based on `event.id` our JSON mocks, so I think it's tolerable. I can still use a less brittle selector here though, so I'll change it 👍

src/lib/components/workflow-actions.svelte
rossedfort: Made a task, DT-969

src/lib/components/workflow-actions.svelte
rossedfort: Made a task, DT-969

src/lib/components/workflow/workflow-reset-form.svelte
rossedfort: This is setting the default value on the prop `eventId` if there is a valid event that can be used as a reset point. The reason I do it this way is because this component may mount before `resetEvents` is populated with the valid events for a given workflow, so we need to check to make sure we have events before getting the ID of the first one. I can try to bake it into the store, but I think it makes sense here.

rossedfort: I guess there's no real point in having a default value here, we should probably just disable the submit button until an event is selected.

src/lib/components/workflow/workflow-reset-form.svelte
rossedfort: https://www.w3.org/TR/2008/WD-WCAG20-TECHS-20080430/H44.html

> The HTML and XHTML specifications allow both implicit and explicit labels. However, some assistive technologies do not correctly handle implicit labels

I think maybe the solution here is keep the wrapping label but also add a `for` attr

stevekinney: Yea, and with the `for` attribute, you also get the ability to click on the label—that's what one me over eventually.

stevekinney (approved):

PR #1381, opened 2023-05-22
[DT-1011] Fix cannot read properties of null (reading 'close')

laurakwhit:
## Description & motivation 💭 
`pauseConfirmationModal.close()` was being called after an `await unpauseSchedule` or `await pauseSchedule`.

A user could navigate away during this time causing `pauseConfirmationModal` to be `undefined`.

`fetchSchedule` is also being called, which causes the page to reload and therefore makes `pauseConfirmationModal.close()` unnecessary. 

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

- [ ] Verify (with throttling) there is no error if selecting `pause`/`unpause` and immediately navigating to a different page 
- [ ] Verify the modal does not stay open after selecting `pause`/`unpause` and the schedule is updated as expected

### Issue(s) closed 
* `DT-1011`

### DIFF
--- a/src/lib/pages/schedule-view.svelte
+++ b/src/lib/pages/schedule-view.svelte
@@ -78,7 +78,6 @@
           reason,
         });
     scheduleFetch = fetchSchedule(parameters, fetch);
-    pauseConfirmationModal.close();
     reason = '';
   };
 </script>


rossedfort (approved):
lgtm, although I _think_ the modal is closing not because the `fetchSchedule` request, but because we are using `formmethod="dialog"` on the form inside the modal, which closes the dialog by default when the form is submitted.

See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/submit#formmethod and https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog for reference

laurakwhit:
@rossedfort I actually looked into this, so I'm glad someone else thought of it! We have a `preventDefault` on the `submit` for the modal `form` the prevents it from closing by default. It's closing because we're using `await scheduleFetch` to conditionally render the UI.

PR #1384, opened 2023-05-23
[DT-1014] Fix cannot read properties of null (reading 'setError')

laurakwhit:
## Description & motivation 💭 
`modal.setError()` is being called after an async function. A user can navigate away during this time causing the modal to be `undefined`. 

This PR adds optional chaining to check if the modal is `undefined`.

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
* `DT-1014`

### DIFF
--- a/src/lib/components/workflow-actions.svelte
+++ b/src/lib/components/workflow-actions.svelte
@@ -70,7 +70,7 @@
   };
 
   const handleSuccessfulTermination = async () => {
-    terminateConfirmationModal.close();
+    terminateConfirmationModal?.close();
     $refresh = Date.now();
     toaster.push({
       id: 'workflow-termination-success-toast',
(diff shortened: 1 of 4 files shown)


src/lib/components/workflow-actions.svelte
```diff
@@ -80,7 +80,7 @@
 
   const handleTerminationError = (error: NetworkError) => {
     reason = '';
-    terminateConfirmationModal.setError(
+    terminateConfirmationModal?.setError(
```
laurakwhit, line 83: Thoughts on using a pattern of only setting an error on the modal if it exists, otherwise adding a notification?

rossedfort (approved):

PR #1385, opened 2023-05-23
use parent namespace name in link to workflow run

rossedfort:
## Description & motivation 💭 

in #1350  and #1354 we missed a link in the Parent Workflows table in the Relationships section of the Workflow Details page

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/workflow/parent-workflow-table.svelte
+++ b/src/lib/components/workflow/parent-workflow-table.svelte
@@ -37,7 +37,7 @@
       <Link
         newTab
         href={routeForEventHistory({
-          namespace,
+          namespace: parentNamespaceName ?? namespace,
           workflow: parent.workflowId,
           run: parent.runId,
         })}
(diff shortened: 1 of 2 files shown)


src/lib/components/workflow/parent-workflow-table.svelte
```diff
@@ -37,7 +37,7 @@
       <Link
         newTab
         href={routeForEventHistory({
-          namespace,
+          namespace: parentNamespaceName ?? namespace,
```
laurakwhit, line 40: Related, I think we need a `namespace: child.namespace` here ⬇️ 
https://github.com/temporalio/ui/blob/e781ade6f23344fbaa1260e5e5503e1d1fc07546/src/lib/components/workflow/child-workflows-table.svelte#L73

laurakwhit (changes requested):

PR #1390, opened 2023-05-25
[DT-6] Add Local Activity to event history filter

laurakwhit:
## Description & motivation 💭 
* Adds a `Local Activity` to 
   * the `Workflow Events` filter in event `History` view
   * the `Event Type` filter in event `Compact` view
* Clears the filter when switching from `History` view to `Compact` view if the filter option does not exist

| History view | Compact view |
|--|--|
|Screenshot 2023-05-25 at 5 45 39 PM|Screenshot 2023-05-25 at 5 45 49 PM|

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests ~added~ updated

- [ ] Verify the `Local Activity` option* works as expected in the 
   - `History` view
   - `Compact` view
- [ ] Verify `LocalActivity` events also show up when filtering for `Marker` events in the `History` view
- [ ] Verify when changing from the `History` to `Compact` view
   - If the filter option exists it is persisted
   - If the filter option **does not** exist the filter is cleared 

*A sample with `LocalActivity` events can be found [here](https://github.com/temporalio/samples-go/tree/main/greetingslocal).
### Issue(s) closed 
* `DT-6`

### DIFF
--- a/src/lib/components/event/event-category-filter.svelte
+++ b/src/lib/components/event/event-category-filter.svelte
@@ -32,8 +32,8 @@
       value: _value,
       url: $page.url,
     }).then((v) => {
-      const hasOption = options.find(({ option }) => option === v);
-      _value = hasOption ? v?.toString() ?? null : null;
+      const visibleOption = options.find(({ option }) => option === v);
+      _value = visibleOption?.option?.toString() ?? null;
       $eventCategoryFilter = _value;
     });
   }
--- a/src/lib/components/event/event-category-filter.svelte
+++ b/src/lib/components/event/event-category-filter.svelte
@@ -32,7 +32,8 @@
       value: _value,
       url: $page.url,
     }).then((v) => {
-      _value = v?.toString() ?? null;
+      const hasOption = options.find(({ option }) => option === v);
+      _value = hasOption ? v?.toString() ?? null : null;
       $eventCategoryFilter = _value;
     });
   }
--- a/src/lib/models/event-history/get-event-categorization.test.ts
+++ b/src/lib/models/event-history/get-event-categorization.test.ts
@@ -88,6 +88,10 @@ describe('Event Category Data Structures', () => {
           "option": "command",
         },
         {
+          "label": "Local Activity",
+          "option": "local-activity",
+        },
+        {
           "color": "#EC4899",
           "label": "Marker",
           "option": "marker",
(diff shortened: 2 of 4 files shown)


src/lib/models/event-history/get-event-categorization.test.ts
```diff
@@ -142,7 +150,10 @@ describe('Event Category Data Structures', () => {
   });
 });
 
-const categories: Record<EventTypeCategory, EventType[]> = {
+const categories: Record<
+  Exclude<EventTypeCategory, 'local-activity'>,
```
laurakwhit, line 154: I've opted not to touch any of the logic in `getEventCategory`, etc. (and just omit the category here), but if we want to be categorizing events as `LocalActivity`, outside of just filtering for them, we may want to revisit this.

src/lib/models/event-history/get-event-categorization.ts
```diff
@@ -4,72 +4,75 @@ import type {
   IterableEvent,
 } from '$lib/types/events';
 
+type Categories = typeof CATEGORIES;
+export type EventTypeCategory = Categories[keyof Categories];
+
+export const CATEGORIES = {
```
laurakwhit, line 10: Changed this to make it easer to use auto-complete, but I'm by no means attached to this format at all if there are objections 😅

src/lib/models/event-history/get-event-categorization.ts
```diff
@@ -78,34 +81,40 @@ export type EventTypeOption = {
...
+  CATEGORIES.LOCAL_ACTIVITY,
+  CATEGORIES.SIGNAL,
+  CATEGORIES.TIMER,
 ];
 export const compactEventTypeOptions: EventTypeOption[] =
   allEventTypeOptions.filter(({ option }) =>
     compactEventTypes.includes(option),
   );
 
 const timelineEventTypes: EventTypeCategory[] = [
```
laurakwhit, line 111: Wondering if we want to add `Update` and `Local Activity` here as well...

src/lib/components/event/event-category-filter.svelte
```diff
@@ -32,7 +32,8 @@
       value: _value,
       url: $page.url,
     }).then((v) => {
-      _value = v?.toString() ?? null;
+      const hasOption = options.find(({ option }) => option === v);
+      _value = hasOption ? v?.toString() ?? null : null;
```
Alex-Tideman, line 36: Slightly confusing here... Why not use hasOption.toString()?

laurakwhit: Definitely not ideal with that nested logic 😅 Updated to 👇

```js
      const visibleOption = options.find(({ option }) => option === v);
      _value = visibleOption?.option?.toString() ?? null;
```

Alex-Tideman:
Looks like you gotta update that unit test for those new options.

Alex-Tideman (approved):

PR #1399, opened 2023-05-31
[DT-1051] Add identity and reason to Terminate in the UI

laurakwhit:
## Description & motivation 💭 
* Adds `from the Web UI` (with translation) to `Terminate` and `Reset` actions in the UI
* Adds user email as `identity` for `Terminate` action

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

- [ ] Verify the `Reason` under the event in event history includes `from the Web UI` for 
   - `Terminate` a workflow
   - `Reset` a workflow
- [ ] If there is a logged in user, verify the user email is included as the `Identity` under the event in event history
### Issue(s) closed 
* `DT-1049`
* `DT-1050`


src/lib/components/workflow-actions.svelte
```diff
@@ -53,6 +55,29 @@
   $: terminateEnabled = workflowTerminateEnabled($page.data.settings);
   $: resetEnabled = workflowResetEnabled($page.data.settings);
 
+  const formatReason = ({
```
Alex-Tideman, line 58: Seems like a nice little function that could benefit from a unit test.

src/lib/i18n/locales/en/workflow.ts
```diff
@@ -0,0 +1,9 @@
+export const Namespace = 'workflow' as const;
+
+export const Strings = {
+  terminated: 'Terminated',
+  reset: 'Reset',
+  'workflow-action-placeholder': '{{action}} from the Web UI',
+  'workflow-action-placeholder-by-email':
+    '{{action}} from the Web UI by {{email}}',
+} as const;
```
rossedfort, line 9: I think we already have these strings in the `workflows` namespace, I can't think of any reason we wouldn't want to use those. Was there a reason you made a new namespace?

rossedfort: I think we can put `Reset` in workflows, I think batch reset is coming soon, so it will make sense anyway

src/lib/components/workflow-actions.svelte
```diff
@@ -90,7 +93,12 @@
     terminateWorkflow({
       workflow,
       namespace,
-      reason,
+      reason: formatReason({
+        action: Action.Terminate,
+        reason,
+        email: $authUser.email,
+      }),
+      identity: $authUser.email,
```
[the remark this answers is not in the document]
rossedfort: In the `operation` part of the request, it looks like you can

https://github.com/temporalio/api/blob/master/temporal/api/workflowservice/v1/request_response.proto#L1210-L1231
https://github.com/temporalio/api/blob/master/temporal/api/batch/v1/message.proto#L56-L61

Although I think the requirement was just to do this for terminating individual workflows, right?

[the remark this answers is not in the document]
laurakwhit: I've gone ahead and added `identity` to `terminateWorkflows` as there may be some value to having it in on the batch system workflow (it just won't show in the UI like it does for individually terminated workflows).

rossedfort (approved):
💯

PR #1400, opened 2023-05-31
Remove duplicate authorization header forwarding

### LINKED ISSUE #1365: Authorization header duplicated in grpc header when making authorization enabled
**Describe the bug**
The authorization header is duplicated when making request from ui-server to temporal-server. I have added an additional log line in temporal-server `interceptor.go` to reveal the duplication. I chanced upon this error when running this set up in my company's infrastructure which might include internal proxies that might turn this into an issue.

**To Reproduce**

1. Run UI locally with authorization enabled.
2. Run curl command `curl localhost:8081/api/v1/namespaces -H 'Authorization: Bearer token'`
3. Check header metadata in interceptor at temporal-server

**Expected behavior**
Only one value of authorization header.

**Screenshots**
curl_command
intercepto_code_log
interceptor_logs

mlline00:
## Description & motivation 💭
As desicribed and demonstrated in #1365, the current ui-server middleware injects duplicate authorization metadata into frontend grpc requests. The behavior is caused by two pieces of middleware adding the authorization header from the incoming API request to the outgoing grpc request metadata:

1. The grpc-gateway [here](https://github.com/temporalio/grpc-gateway/blob/ea4029e1bec8672694afe8cabc98e450414dee9b/runtime/context.go#L123-L125)
2. The header forwarding metadataAnnotator injected by the ui-server [here](https://github.com/temporalio/ui/blob/d77b9231562fb00e19c915dc241dd1140a96b332/server/server/server.go#L60)

This PR prefers the grpc-gateway handle the authorization header and configures the ui-server metadataAnnotator accordingly.

One small adjustment to the repro steps called out in #1365, auth does not need to be enabled to observe this issue. The middleware configuration causing this behavior is not externally configurable. Any frontend grpc request generated by ui-sever will have two authorization entries in the metadata if the incoming API request includes an authorization header.

N/A

### Design Considerations 🎨
N/A

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

1. While debugging in ui-server mode, set a breakpoint at [grpc-gateway/runtime/context.go:168](https://github.com/temporalio/grpc-gateway/blob/ea4029e1bec8672694afe8cabc98e450414dee9b/runtime/context.go#L168)
2. Make a request (**_which includes an authorization header_**) to one of the API endpoints 
    * `/api/namespaces/v1?`, for example
3. Inspect `md` when stopped at the breakpoint set in step <span>#</span>1
    * With the current code, notice authorization has two elements. This are the duplicate authorization entries.
image
    * With this MR, authorization has a single element
image

## Checklists
N/A
### Issue(s) closed
closes #1365 

### Any docs updates needed?
N/A

### DIFF
--- a/server/server/server.go
+++ b/server/server/server.go
@@ -57,6 +57,7 @@ func NewServer(opts ...server_options.ServerOption) *Server {
 	authMiddleware := server_options.WithAPIMiddleware(([]api.Middleware{
 		headers.WithForwardHeaders(
 			[]string{
+				// NOTE: Authorization header is forwarded by grpc-gateway
 				auth.AuthorizationExtrasHeader,
 			}),
 	}))
--- a/server/server/server.go
+++ b/server/server/server.go
@@ -57,7 +57,6 @@ func NewServer(opts ...server_options.ServerOption) *Server {
 	authMiddleware := server_options.WithAPIMiddleware(([]api.Middleware{
 		headers.WithForwardHeaders(
 			[]string{
-				echo.HeaderAuthorization,
 				auth.AuthorizationExtrasHeader,
 			}),
 	}))


server/server/server.go
```diff
@@ -57,7 +57,6 @@ func NewServer(opts ...server_options.ServerOption) *Server {
 	authMiddleware := server_options.WithAPIMiddleware(([]api.Middleware{
 		headers.WithForwardHeaders(
 			[]string{
-				echo.HeaderAuthorization,
```
feedmeapples, line 60: let's add a comment here clarifying that the authorization header is forwarded by grpc-gateway. Ideally we would  use `runtime.NewServeMux(runtime.WithOutgoingHeaderMatcher` though this is not a priority now since the forwarding already works

feedmeapples:
thank you for the PR

server/server/server.go
feedmeapples, line 60: let's add a comment here clarifying that the authorization header is forwarded by grpc-gateway. Ideally we would  use `runtime.NewServeMux(runtime.WithOutgoingHeaderMatcher..` though this is not a priority now since the forwarding already works

feedmeapples (approved):

PR #1401, opened 2023-06-01
DT-610 - consolidate playwright e2e and integration tests

rossedfort:
## Description & motivation 💭 
This PR merges the existing `/e2e` directory with the `/tests` directory. Also rewrite some Go utils for running the tests in TypeScript. 

Notable Changes
* No more top level `e2e/` directory with its own package.json, playwright.config, go.mod, etc.
* 1 playwright config lives at the root of the repo, all playwright tests live in `tests/`
* e2e tests (which are run with real workflows, codec server and temporal server) live in `tests/e2e` and can be ran with `pnpm test:e2e`
* integration tests (which are run with mock workflows) live in `tests/integration` and can be ran with `pnpm test:integration`
* the codec server and workflows can also be run (while running `pnpm dev`), with `pnpm codec-server` and `pnpm workflows`
* disambiguate `scripts/` from `utilities/`
  *  `scripts/` is a collection of files with no top-level exports that can be run via the command line, i.e. `esno scripts/run-temporal-server.ts`
  * `utilities/` is a collection of files with exports that can be imported into other modules, i.e.
* add a `temporal/` directory for all things temporal-sdk related.
n/a
### Design Considerations 🎨 
n/a
### How was this tested 👻 
n/a
- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

n/a
## Checklists
n/a
## Docs
n/a


.env.test.integration
```diff
@@ -1,3 +1,5 @@
+export PW_EXPERIMENTAL_TS_ESM=1
+export PW_MODE=integration
```
stevekinney, line 2: Do we need `export`? Should we use it consistently?

stevekinney (approved):

.env.test.integration
rossedfort: I think if we use dotenv, i.e. `require('dotenv').config()` we don't, but if we `. ./.env` from the command line we do. Maybe I'll just use `dotenv` in my scripts and nuke the exports

PR #1404, opened 2023-06-02
Add pathForApi util

laurakwhit:
## Description & motivation 💭 
Pulls out logic for getting the path for the api into it's own util. This way, in cases where page data has not been loaded yet (e.g. in a load function), the base url can be passed in.

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/services/workflow-service.ts
```diff
@@ -171,13 +171,14 @@ export const fetchAllWorkflows = async (
 };
 
 export const fetchWorkflowForRunId = async (
-  parameters: { namespace: string; workflowId: string },
+  parameters: { namespace: string; workflowId: string; url?: string },
```
stevekinney, line 174: I guess we _could_ just make this a type, right?

```ts
type WorflowAPIPath = { namespace: string; workflowId: string; baseUrl?: string };

export const fetchWorkflowForRunId = async ({}) => {
  namespace,
  workflowId,
  baseUrl
}, request = fetch): Promise => { /* … */ };
```

src/lib/utilities/route-for-api.ts
```diff
@@ -35,7 +35,7 @@ const replaceNamespaceInApiUrl = (
   return '';
 };
 
-const base = (namespace?: string): string => {
+export const base = (namespace?: string): string => {
   let baseUrl = '';
   const webUrl: string | undefined = get(page).data?.webUrl;
```
stevekinney, line 40: I'm tempted to grab it from the `window` object so that it works anywhere, but I'm not sure what kind of access Vite/SvelteKit gives us off the top of my head. I _do_ think we do it elsewhere in the codebase, so there is likely prior art.

src/lib/utilities/route-for-api.ts
```diff
@@ -57,13 +57,19 @@ const base = (namespace?: string): string => {
...
 };
 
-const withBase = (endpoint: string, namespace?: string): string => {
+const getPath = (endpoint: string): string => {
   if (endpoint.startsWith('/')) endpoint = endpoint.slice(1);
+  return `/api/v1/${endpoint}`;
+};
+
+const withBase = (path: string, namespace?: string): string => {
   const baseUrl = base(namespace);
```
stevekinney, line 66: I don't see `base` in _this_ PR, but it feels like it could just do all of this logic?

src/lib/utilities/route-for-api.ts
```diff
@@ -80,6 +86,44 @@ const encode = (parameters: APIRouteParameters): APIRouteParameters => {
   );
 };
 
+export function pathForApi(
```
stevekinney, line 89: I know I wrote this originally, but it feels like it's getting out of control. 🤣

stevekinney (commented):
The code itself looks totally cool—nitpicks aside, that said, I would like to make sure of this is covered in unit tests (almost exhaustively) because I think this is due up for a refactor and rethinking and we want to make sure we don't break anything.

stevekinney (approved):

src/lib/utilities/route-for-api.ts
laurakwhit: Kept it separate so that I could use `base` in `fetchWorkflowForRunId` [here](https://github.com/temporalio/ui/pull/1404/files#diff-bec5e9fc4492624aa6906c9c6b0567006dbe866be6d0eda43af03204166e0101R179).

Alex-Tideman (approved):

PR #1405, opened 2023-06-02
Add filename linting

stevekinney:


.lintstagedrc.json
```diff
@@ -1,5 +1,6 @@
 {
   "*.{ts,js}": ["pnpm lint:fix", "prettier --write"],
```
laurakwhit, line 2: Is this redundant now with the addition of `"*": ["pnpm lint:fix", "prettier --write"]`?

.ls-lint.yml
```diff
@@ -0,0 +1,13 @@
...
+  .d.ts: kebab-case
+  .svelte: regex:[+_]?[a-z0-9\-\.]+
+
+ignore:
+  - .git
+  - .histoire
+  - .svelte-kit
+  - .vercel
+  - build
+  - node_modules
```
laurakwhit, line 13: Wondering if we want to add the following 
```suggestion
  - node_modules
  - dist
  - server
```

And maybe even `playwright-report` for when running the command locally.

PR #1412, opened 2023-06-05
Add initial test for accessibility

stevekinney:
Spoiler alert: It fails.

This change also adds unique IDs to all modals as that was a low-hanging fruit.


rossedfort (approved):
lgtm pending passing tests

PR #1417, opened 2023-06-07
[DT-971] Migrate Cypress stack-trace.spec.js tests to Playwright

laurakwhit:
### How was this tested 👻 

- [ ] Manual testing
- [x] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
* `DT-971`


tests/test-utilities/mocks/query-stack-trace-error.ts
```diff
@@ -0,0 +1,22 @@
...
+    payloads: [
+      {
+        an: 'error',
+      },
+    ],
+  },
+  queryRejected: null,
+};
+
+export default async function mockQueryApiWithStackTraceError(page: Page) {
```
laurakwhit, line 15: @rossedfort is this `mocks` folder where you were thinking to add individual mock api utils? If not, I'll create a new one!

rossedfort: Yep this looks good! FWIW - we could set a workflow up that returns this data when queried, but I think this is fine for now

tests/e2e/stack-trace.spec.js
rossedfort, line 1: I just realized, this file should be `.ts`

rossedfort (approved):

PR #1427, opened 2023-06-09
Remove use of remote/local codec endpoint

lorensr:
## Description & motivation 💭 

Issues with current:
- Jonathan requested UI align with docs: "Codec Server" over "Remote Codec"
- In development, an endpoint is often running locally, so "remote" may be confusing
- An endpoint can also be remote, so a "local setting" endpoint may be confusing, and also seems less precise to me than "browser setting"

Solution:
- "Codec Server"
- "Codec Server endpoint"
- "browser override"

Before:

image

After:

image

## Notes

This part of the Cloud UI would also need to be updated to match:

image

## TODO

If we like this change, I'll fix the tests


src/lib/holocene/data-encoder-settings.svelte
```diff
@@ -108,8 +108,8 @@
               id="use-configuration-endpoint-radio"
               data-testid="use-configuration-endpoint-input"
             />
-            Uses {namespaceOrCluster}-level settings, where available.
-            Otherwise, use my local setting.
+            Use {namespaceOrCluster}-level settings, where available.
+            Otherwise, use my browser setting.
```
lorensr, line 112: Is this actually how the feature works? If yes, we should remove the cases of "override" in copy.

Alex-Tideman: Removed override in copy.

Alex-Tideman (approved):

src/lib/holocene/data-encoder-settings.svelte
```diff
@@ -84,14 +84,20 @@
...
-        The Codec Server uses a remote codec endpoint to decrypt your payloads,
-        ensuring that Temporal never sees your data. A remote codec endpoint can
-        be set at the {namespaceOrCluster}-level, or locally in your browser.
+        A <a
+          rel="noreferrer"
+          target="_blank"
+          href="https://docs.temporal.io/dataconversion#codec-server"
+          class="text-blue-700 underline decoration-blue-700">Codec Server</a
+        >
+        decodes your payloads, ensuring that Temporal never sees your data. A Codec
```
lorensr, line 93:
```suggestion
        decodes your data. A Codec
```

@Alex-Tideman @aarohib because decoding may not decrypt?

PR #1444, opened 2023-06-20
Add TEMPORAL_HIDE_WORKFLOW_QUERY_ERRORS env var to hide workflow query errors from server

Alex-Tideman:
## Description & motivation 💭 
Users want the ability to hide error messages from the server in case they contain internal wording/messages.

Screen Shot 2023-06-20 at 9 23 36 AM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/stores/workflows.ts
+++ b/src/lib/stores/workflows.ts
@@ -1,5 +1,6 @@
 import { derived, readable, writable } from 'svelte/store';
 import { page } from '$app/stores';
+import { translate } from '$lib/i18n/translate';
 
 import {
   fetchAllWorkflows,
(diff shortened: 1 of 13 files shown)


src/lib/stores/workflows.ts
```diff
@@ -46,7 +68,13 @@ const updateWorkflows: StartStopNotifier<WorkflowExecution[]> = (set) => {
         }
 
         if (error) {
-          workflowError.set(error);
+          if (hideWorkflowQueryErrors) {
+            workflowError.set(
+              'A error has occurred while querying for Workflows.',
```
laurakwhit, line 73: Do we want to go ahead and add a `translate` here?

laurakwhit (commented):
Wondering about adding `hideWorkflowQueryErrors` to the default state [here](https://github.com/temporalio/ui/blob/main/src/lib/stores/settings.ts#L4) and [here.](https://github.com/temporalio/ui/blob/main/tests/test-utilities/mocks/settings.ts#L4)

src/lib/stores/workflows.ts
Alex-Tideman: Yup good call, added it.

PR #1445, opened 2023-06-20
Add Timeline view with labs guard

Alex-Tideman:
## Description & motivation 💭 
Add new Timeline view hidden behind labs flag.

To turn on, set labs=true as a query param on the Workflow Details page. That will set a localStorage flag to true so you don't have to keep adding the param. To turn off, set labs=false.

Screen Shot 2023-06-20 at 9 57 42 AM

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/stores/labs-mode.ts
```diff
@@ -0,0 +1,3 @@
+import { persistStore } from './persist-store';
+
+export const labsMode = persistStore('labsMode', false, true);
```
laurakwhit, line 3: Might be nice if we could do something like this 

```js
const labsParam = derived([page], ([$page]) =>
  $page.url.searchParams.get('labs'),
);

const updateLabsMode: StartStopNotifier<boolean> = (set) => {
  return labsParam.subscribe((labs) => {
    if (labs === 'true') {
      set(true);
    } else if (labs === 'false') {
      set(false);
    }
  });
};
```
with the `persistStore` so we don't have to be on the same page where the `LabsModeGuard` is to update the setting.

laurakwhit (approved):

PR #1446, opened 2023-06-20
Add scheduleToFrequency formatting with tests

Alex-Tideman:
## Description & motivation 💭 
This PR adds logic to read a StructuredCalendar and return a human readable string. This is useful for Schedules not created in the UI since we did some magic in the comment field in the UI to make it a nice label.

Unit tests added.

Screen Shot 2023-06-20 at 4 23 23 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

### DIFF
--- a/src/lib/components/schedule/schedule-frequency.svelte
+++ b/src/lib/components/schedule/schedule-frequency.svelte
@@ -2,6 +2,7 @@
   import { intervalToComment } from '$lib/utilities/schedule-comment-formatting';
   import type { IntervalSpec } from '$types';
   import type { StructuredCalendar } from '$lib/types/schedule';
+  import { structuredCalendarToFrequency } from '$lib/utilities/schedule-frequency-formatting';
 
   export let calendar: StructuredCalendar | undefined = undefined;
   export let interval: IntervalSpec | undefined = undefined;
@@ -13,7 +14,7 @@
 <div class="flex flex-col {$$props.class}">
   {#if calendar}
     <p data-testid="schedule-calendar-frequency">
-      {calendar?.comment ?? ''} UTC
+      {structuredCalendarToFrequency(calendar)}
     </p>
   {:else}
     <p data-testid="schedule-interval-frequency">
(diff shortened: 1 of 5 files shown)


src/lib/utilities/schedule-frequency-formatting.ts
```diff
@@ -0,0 +1,209 @@
...
+// Examples of output
+// Wednesday at 06:30pm UTC
+// Sunday, Monday, Tuesday, Wednesday, Thursday, Friday at 12:00pm UTC
+// Every 1, 15, 31 of the month at 12:00pm UTC
+// Every 1, 15, 31 of January, March, April, June, July, September, December at 12:00pm UTC
+
+const isDailyForTheMonth = (calendar: StructuredCalendar) => {
+  return (
+    calendar.dayOfMonth?.length === 1 &&
+    calendar.dayOfMonth[0]?.start === 1 &&
```
laurakwhit, line 13: Looking at the protos it seems like all the properties are optional and can be `null` for  `StructuredCalendar`.  If so, we might want this instead
```suggestion
    calendar.dayOfMonth?.[0]?.start === 1 &&
```

Although, TypeScript seems fine with it so maybe I'm misinterpreting 🤔

src/lib/components/schedule/schedule-frequency.svelte
```diff
@@ -13,7 +14,7 @@
 <div class="flex flex-col {$$props.class}">
   {#if calendar}
     <p data-testid="schedule-calendar-frequency">
-      {calendar?.comment ?? ''} UTC
+      {structuredCalendarToFrequency(calendar) || calendar?.comment || ''}
```
rossedfort, line 17: Could we just move the `|| calendar?.comment || ''` logic to `structuredCalendarToFrequency` ?

rossedfort (approved):
lgtm, might consider consolidating more of that template logic into the util you wrote though

src/lib/components/schedule/schedule-frequency.svelte
Alex-Tideman: good call

src/lib/utilities/schedule-frequency-formatting.ts
Alex-Tideman: I have the calendar.dayOfMonth?.length === 1 check before that line

laurakwhit: 🤦‍♀️ that you do.

PR #1455, opened 2023-06-22
DT-1017 - support unique workflow table columns per namespace

rossedfort:
## Description & motivation 💭 
Previously on the workflows list page, we would only fetch global search attributes to render the available table columns for the workflows summary table. With the change introduced in #1414, we fetch namespace specific search attributes any time the namespace path parameter changes. This allows us to let users configure their table columns on a per-namespace basis.

https://github.com/temporalio/ui/assets/11775628/91218af7-bba4-4d36-9fd9-4bd368cc336e

### Design Considerations 🎨 
N/a
### How was this tested 👻 
This is covered by existing e2e and unit tests, as well as manual testing
- [x] Manual testing
- [x] Existing E2E tests
- [x] Existing Unit tests 

- On the workflows list page, add/remove/reorder/pin/unpin columns on one namespace and ensure that the columns do not change for a second namespace and vice versa.
- Ensure only custom search attributes for a given namespace show in the side panel for managing columns.
## Checklists
N/a
## Docs
N/a


Alex-Tideman:
This is awesome. I think I only saw one instance when I was testing that the columns got reset to original default state. It was when I first pulled down, was on default, didn't change anything, went to canary, changed columns, went back to default, changed columns, then went back to canary (columns weren't preserved in change). But I'll keep testing to make sure I can repro.

package.json
```diff
@@ -48,8 +48,8 @@
     "test:coverage": "TZ=UTC vitest run --coverage",
     "test:e2e": "PW_MODE=e2e playwright test tests/e2e",
     "test:e2e:run": "pnpm test:e2e --ui",
-    "test:integration": "PW_MODE=integration playwright test tests/integration",
-    "test:integration:run": "pnpm test:integration --ui",
+    "test:integration:run": "PW_MODE=integration playwright test tests/integration",
```
Alex-Tideman, line 51: Was this meant to be switched? Is :run the UI mode or no?

src/lib/stores/workflow-table-columns.ts
```diff
@@ -70,67 +76,119 @@ const DEFAULT_AVAILABLE_COLUMNS: WorkflowHeader[] = [
   { label: 'Task Queue', pinned: false },
 ];
 
-const workflowTableColumns = persistStore<WorkflowHeader[]>(
-  'workflow-table-columns',
-  DEFAULT_COLUMNS,
+const getDefaultColumns = (): WorkflowHeader[] => {
```
Alex-Tideman, line 79: Nice!

tests/test-utilities/mock-apis.ts
```diff
@@ -26,9 +60,66 @@ export const mockWorkflowsApis = (page: Page) => {
     mockGlobalApis(page),
     mockWorkflowsApi(page),
     mockSearchAttributesApi(page),
+    mockWorkflowsCountApi(page),
   ]);
 };
 
 export const mockNamespaceApis = (page: Page) => {
```
Alex-Tideman, line 67: These are some solid utils.

rossedfort: Thanks, I'm hoping we can continue to follow this pattern where we just compose a bunch of mock calls together for each page we need to test. Then you can still import a mock API fn and call it individually if you need to override something.

tests/test-utilities/mocks/cluster.ts
```diff
@@ -1,6 +1,8 @@
 import type { Page } from '@playwright/test';
 import type { GetClusterInfoResponse as Cluster } from '$src/lib/types';
 
+export const CLUSTER_API = '**/api/v1/cluster?';
+
 const MOCK_CLUSTER: Cluster = {
```
Alex-Tideman, line 6: One thing that is going to be a struggle is remembering the default cluster server version (1.19.3) vs. what you need to pass in for the version for mockClusterApi. Like do we keep updating the default cluster to most recent or most stable (1.20 vs. 1.21) or do we keep it here forever and explicitly pass in newer versions. I think the latter, but it might be a bit of a headache down the road. Just something to think about.

Alex-Tideman:
Not sure if this is expected, but when I run Playwright locally all the workflows-list.spect.ts tests have the Server Failed to Respond error in them. One test failed on the initial run, then they all passed after that. But seems like those should be getting back mocked workflows.

Screen Shot 2023-06-28 at 9 15 23 AM

package.json
Alex-Tideman: 👍

tests/test-utilities/mocks/cluster.ts
rossedfort: Yeah that's a good point. It's possible we'd even want to run the same tests against diff cluster versions, so passing in an override will probably be the way to go.

Alex-Tideman (approved):

PR #1456, opened 2023-06-22
DT-826 - add i18n strings for event summary table

rossedfort:
## Description & motivation 💭 
- add i18n strings for event summary table (this does not add i18n for event attributes, they are returned from the server and formatted to be displayed in the UI)
- also remove i18n from page.data - we don't need it, and it prevents us from using the `t` function in non-svelte components

N/a
### Design Considerations 🎨 
N/a
### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

Load the workflow details -> history page aka event summary, and ensure no missing strings
## Checklists
N/a
## Docs
N/a


src/lib/i18n/locales/en/events.ts
```diff
@@ -0,0 +1,31 @@
...
+  'category-update': 'Update',
+  'category-workflow': 'Workflow',
+  'attribute-group-activity': 'Activity',
+  'attribute-group-parent': 'Parent',
+  'attribute-group-retry-policy': 'Retry Policy',
+  'attribute-group-schedule': 'Schedule',
+  'attribute-group-search-attributes': 'Search Attributes',
+  'attribute-group-summary': 'Summary',
+  'attribute-group-task-queue': 'Task Queue',
+  'attribute-group-workflow': 'Workflow',
```
laurakwhit, line 30: Can we combine `attribute-group-workflow` with `category-workflow` (line 22) since they are both "Workflow"? Or maybe something to add to `common.ts` instead?

src/lib/utilities/format-event-attributes.ts
```diff
@@ -35,8 +36,8 @@ const keysToExpand: Readonly<Set<string>> = new Set([
 
 const keysToFormat: Readonly<Set<string>> = new Set(['maximumAttempts']);
 
-export const UnlimitedAttempts = 'Unlimited';
-export const NoExpiration = 'No Expiration';
+export const UnlimitedAttempts = 'unlimited';
+export const NoExpiration = 'no-expiration';
```
laurakwhit, line 40: Do these need a `translate`?
```suggestion
export const UnlimitedAttempts = translate('unlimited');
export const NoExpiration = translate('no-expiration');
```

src/lib/components/event/event-category-filter.svelte
```diff
@@ -59,7 +62,7 @@
           data-testid="event-category-option"
           on:click={() => onOptionClick(option)}
         >
-          {label}
+          {translate('events', label)}
```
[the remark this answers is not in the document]
rossedfort: This is the label for each option (which the diff doesn't show), not the label that's defined up in line 19 for the overall dropdown menu, so this is correct.

src/lib/i18n/locales/en/events.ts
rossedfort: Yeah I think this is a good use case for the `common` namespace 👍

src/lib/utilities/format-event-attributes.ts
rossedfort: Unfortunately we can't use the `translate` function outside of a svelte component, because our `i18n.init()`  occurs in the root layout, and these files are evaluated before that point. I can try to come up with a solution for this in a future PR.

src/lib/i18n/locales/en/events.ts
rossedfort: Actually I'm going to leave this as is. In the template file, we call `translate('events', label)`, where `label` maps to these i18n keys. If we use a common key here, we'd have to add some logic to pass in the correct namespace, which seems overkill to reduce a little duplication.

src/lib/utilities/format-event-attributes.ts
laurakwhit: Ah sounds good. Do we want to leave these as is for now then? So it doesn't read as `no-expiration` in the UI?

laurakwhit (approved):
One comment (https://github.com/temporalio/ui/pull/1456#discussion_r1240332854) but otherwise lgtm 🚀

src/lib/utilities/format-event-attributes.ts
rossedfort: Ahh yes, when I realized `translate()` doesn't work here, I removed the call but didn't change the keys back to their original strings. Good catch!

PR #1461, opened 2023-06-28
Add unit tests to bulk-actions store and remove fiction-store

rossedfort:
## Description & motivation 💭 
- adds unit tests to bulk-actions store, this is important because the logic to determine whether batch actions are enabled has many branches and testing the different possibilities ensures we don't have regressions when updating this util.
- also remove the fiction-store, which was unused.
N/a
### Design Considerations 🎨 
N/a
## Testing 🧪 
N/a
### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

N/a
## Checklists
N/a
### Issue(s) closed 
N/a
## Docs
N/a


src/lib/stores/advanced-visibility.ts
```diff
@@ -1,17 +1,17 @@
...
+import { settings } from './settings';
 
 import {
   advancedVisibilityEnabled,
   advancedVisibilityEnabledWithOrderBy,
 } from '$lib/utilities/advanced-visibility-enabled';
 
 export const isCloud = derived(
-  [page],
-  ([$page]) => $page.data?.settings?.runtimeEnvironment?.isCloud,
```
Alex-Tideman, line 14: I'm thinking we should remove the $settings store. I don't think we need it anymore now that we are setting the settings prop in cloud. Any reference to settings should come from $page.data.settings so we have one source of truth.

Alex-Tideman (approved):

PR #1463, opened 2023-06-29
paginated workflows table

rossedfort:
## Description & motivation 💭 
This PR implements new designs for the workflows summary table. The designs can be found [here](https://www.figma.com/file/u0hxNQYdSWagO8eOlXkgea/Core-UI?type=design&node-id=8073-112933&mode=design&t=UKjntGTgNVV4vXNC-0). 

Notable changes include
* remove the notion of "pinned" columns from the workflows summary table - this allows us to render a single `<table />` element, which facilitates scrolling vertically and horizontally within the table itself. This also removes the ability to resize the "pinned" section, as it no longer exists.
* Move pagination controls to the "footer" of the table
* Style the table such that it only every fills the remaining space in the main content, make the table itself scrollable both vertically and horizontally.

https://github.com/temporalio/ui/assets/11775628/b16bc657-a2ff-4d5b-9998-9fcf008395db

### Design Considerations 🎨 
- Need to confirm that implementation matches designs
  - Should we have a page size selector in the left portion of the footer?
  - Does the border on top of the footer help separate the footer from the table 
### How was this tested 👻 
Manual testing as well as existing tests cover these changes
- [x] Manual testing
- [x] E2E tests added
- [ ] Unit tests added

### Draft Checklist 
- [x] pagination controls should read from/write to URL search params
- [x] fix overscroll issue on workflows list page from new table
- [x] add updating/loading/empty state to table
- [ ] add i18n strings


src/lib/components/workflow/workflows-summary-configurable-table/table-header-row.svelte
```diff
@@ -31,48 +29,39 @@
...
-    <th class="configuration-button-table-cell">
-      <IconButton
-        data-testid="workflows-summary-table-configuration-button"
-        icon="vertical-ellipsis"
-        on:click={onClickConfigure}
-      />
-    </th>
-  {/if}
+  <th class="configuration-button-table-cell">
+    <IconButton
```
laurakwhit, line 56: Looks like this starts floating left a bit with fewer columns

Screenshot 2023-08-02 at 12 04 45 PM

src/lib/pages/workflows-with-new-search.svelte
```diff
@@ -272,22 +271,13 @@
...
-  </svelte:fragment>
-  <svelte:fragment slot="action-top-center">
-    <WorkflowDateTimeFilter />
-  </svelte:fragment>
-  <WorkflowsSummaryConfigurableTable workflows={visibleItems} />
-</Pagination>
+<div class="flex flex-row gap-2">
+  <LabsModeGuard>
+    <svelte:fragment slot="fallback">
+      <WorkflowAdvancedSearch />
```
laurakwhit, line 277: Some wonky things happen with the status filter if the height of the filter is greater than the table
|||
|--|--|
|Screenshot 2023-08-02 at 12 07 05 PM|Screenshot 2023-08-02 at 12 07 47 PM|

And the input filters can get cut off a bit

Screenshot 2023-08-02 at 12 11 46 PM

src/lib/holocene/table/paginated-table.svelte
```diff
@@ -0,0 +1,194 @@
...
+      <slot name="headers" />
+      {#if updating}
+        <ProgressBar />
+      {/if}
+    </thead>
+    <tbody class="paginated-table-body">
+      <slot visibleItems={$store.items} />
+    </tbody>
+  </table>
+  <div class="paginated-table-controls">
```
laurakwhit, line 106: We might want to set a min width for these so things don't start looking like this on smaller screens

Screenshot 2023-08-02 at 12 22 11 PM

src/lib/holocene/table/paginated-table.svelte
```diff
@@ -0,0 +1,194 @@
...
+  };
+
+  $: {
+    if (perPage) store.jumpToPage(currentPage);
+    if (currentPage) store.adjustPageSize(perPage);
+  }
+</script>
+
+<div class="paginated-table-wrapper">
+  <table class="paginated-table">
```
laurakwhit, line 95: Similar to our `Table` component we could go ahead and add a `caption` slot here for accessibility

```suggestion
  <table class="paginated-table">
    <slot name="caption" />
```

And then update instances where `PaginatedTable` is used with a `<caption/>` element.

Alex-Tideman (approved):

src/lib/components/workflow/workflows-summary-configurable-table.svelte
```diff
@@ -35,60 +27,47 @@
...
-      <slot slot="cloud" name="cloud" />
-    </TableEmptyState>
-  {/if}
-</div>
+    {#each columns as column}
+      <TableHeaderCell {column} />
+    {/each}
+  </TableHeaderRow>
+  {#each visibleItems as workflow}
+    <TableRow {workflow}>
```
laurakwhit, line 54: Looks like the multi-select isn't working as expected

Screenshot 2023-08-14 at 10 13 09 AM

laurakwhit: **Update:** Unable to repro. this, but there is an existing bug with multi-select. Will address separately.

laurakwhit (approved):

PR #1476, opened 2023-07-10
Map default columns to the namespace column store

Alex-Tideman:
## Description & motivation 💭 
There are possible edge cases depending on the state of a user's localStorage where $persistedWorkflowTableColumns[namespace] could return an empty array. This plus the check for empty column length for empty prop led to weird situations where it showed 'No Workflows' when there was workflows.

The solution is to check for !columns?.[namespace]?.length, and if false, set columns[namespace] to the default columns.

A side effect of this solution is that if you try to remove all columns, it will reset to the default columns (which I think is an okay behavior)

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/workflow/workflows-summary-configurable-table.svelte
+++ b/src/lib/components/workflow/workflows-summary-configurable-table.svelte
@@ -25,6 +25,7 @@
   $: empty = workflows.length === 0;
   $: pinnedColumns = columns.filter((column) => column.pinned);
   $: otherColumns = columns.filter((column) => !column.pinned);
+  $: totalColumnCount = pinnedColumns.length + otherColumns.length;
 
   const openCustomizationDrawer = () => {
     customizationDrawerOpen = true;
@@ -37,7 +38,7 @@
 
 <div class="workflows-summary-configurable-table">
   <TableWrapper noPinnedColumns={pinnedColumns.length === 0}>
-    <Table pinned {empty} columns={pinnedColumns}>
+    <Table pinned {empty} {totalColumnCount} columns={pinnedColumns}>
       <TableHeaderRow pinned {workflows} {pinnedColumns} {empty} slot="headers">
         {#each pinnedColumns as column}
           <TableHeaderCell {column} />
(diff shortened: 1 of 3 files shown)


src/lib/components/workflow/workflows-summary-configurable-table.svelte
```diff
@@ -19,7 +19,7 @@
 
   $: ({ namespace } = $page.params);
   $: columns = $workflowTableColumns?.[namespace] ?? [];
```
laurakwhit: Should we add this here too?
```suggestion
  $: columns = $workflowTableColumns?.[namespace] ?? getDefaultColumns();
```

src/lib/components/workflow/workflows-summary-configurable-table.svelte
```diff
@@ -18,8 +21,8 @@
   let customizationDrawerOpen: boolean = false;
 
   $: ({ namespace } = $page.params);
-  $: columns = $workflowTableColumns?.[namespace] ?? [];
-  $: empty = workflows.length === 0 || columns.length === 0;
+  $: columns = $workflowTableColumns?.[namespace] ?? getDefaultColumns();
```
laurakwhit, line 24: What do you think about also checking that `JSON.parse(stringifiedOldColumns)` in `getDefaultColumns` has `length` and, if not, setting it to `DEFAULT_COLUMNS`? That way if a user currently has `workflow-table-columns` set to an empty an array (due to a previous edge case) they'll see default columns instead.

src/lib/components/workflow/workflows-summary-configurable-table.svelte
```diff
@@ -18,8 +21,8 @@
...
+  $: empty = workflows.length === 0;
```
laurakwhit, line 25: Couple of edge cases still 😞 

First one is if there are workflows, but no columns are selected (we see the bulk actions)

Screenshot 2023-07-10 at 12 40 53 PM

Second one is if there are only pinned columns

Screenshot 2023-07-10 at 12 41 33 PM

laurakwhit (approved):
Only other thought to perhaps follow-up on is adding a check in the reducer for `WORKFLOW_COLUMN.REMOVE` and preventing a user from ever removing all the columns (but then we run into the question of how to disable the remove button 😅).

Alex-Tideman:
I think we are okay here since it will reset the columns to the default columns if they remove all the columns.

PR #1482, opened 2023-07-12
Remove start and end time sort filter in table header and $workflowSorts

Alex-Tideman:
## Description & motivation 💭 
Since sort is no longer supported, remove the UI filters to sort by Start and End times in the table.

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/components/workflow/dropdown-filter/end-time.svelte
```diff
@@ -1,25 +0,0 @@
-<script lang="ts">
-  import DropdownMenu from '$lib/components/dropdown-menu.svelte';
-  import { translate } from '$lib/i18n/translate';
-  import { workflowSorts } from '$lib/stores/filters';
```
laurakwhit, line 4: > sort is no longer supported

Does this affect all `workflowSort`s or just `StartTime` and `EndTime`?

And can we remove the `workflowSorts` from the `filters` store (or is it needed for backwards compatibility)?

Alex-Tideman: Hmm good question. It's all sorts I believe, but we only allow sorting with StartTime/EndTime in the UI unless you manually enter the query in the input. 

If we are getting rid of our sort UI, it makes sense to gut the workflowSorts store. I really don't want to do another backwards compatible check for this small UI feature that can cause issues with a big number of workflows even when enabled so I'm gonna remove it

laurakwhit (approved):
Just noting that we'll need a follow-up PR for Cloud.

Alex-Tideman:
https://github.com/temporalio/cloud-ui/pull/550

PR #1485, opened 2023-07-18
DT-831 - add i18n to schedules pages

rossedfort:
## Description & motivation 💭 
- Add internationalization to schedules index, view, create, and edit pages.
- n/a
### Design Considerations 🎨 
- n/a
### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/components/schedule/schedule-form-view.svelte
```diff
@@ -31,13 +33,18 @@
   let namespace = $page.params.namespace;
   let scheduleId = $page.params.schedule;
 
-  let title = schedule ? 'Edit Schedule' : 'Create Schedule';
-  let loadingText = schedule ? 'Editing Schedule...' : 'Creating Schedule...';
-  let backTitle = schedule ? 'Back to Schedule' : 'Back to Schedules';
+  let title = translate('schedules', schedule ? 'edit' : 'create');
+  let loadingText = translate('schedules', schedule ? 'editing' : 'creating');
```
Alex-Tideman, line 37: 'edit' and 'create' seem like a common.json term, not specific to schedules

src/lib/components/schedule/schedule-form-view.svelte
```diff
@@ -31,13 +33,18 @@
...
+    'schedules',
+    schedule ? 'back-to-schedule' : 'back-to-schedules',
+  );
   let backHref = schedule
     ? routeForSchedule({ namespace, scheduleId })
     : routeForSchedules({ namespace });
-  let confirmText = schedule ? 'Save' : 'Create Schedule';
+  let confirmText = schedule
+    ? translate('save')
+    : translate('schedules', 'create');
```
Alex-Tideman, line 47: Same here

src/lib/components/schedule/schedule-form-view.svelte
Alex-Tideman: nvm it's Creating Schedules...

rossedfort: The copy here is "Edit Schedule", and "Create Schedule", so we can make them common if we remove "Schedule", just didn't want to change existing copy.

Alex-Tideman (approved):

src/lib/components/schedule/schedule-form-view.svelte
rossedfort: Same response as above, the copy is "Create Schedule". If it were just "Create", def would have put it in common.

src/lib/components/schedule/schedule-form-view.svelte
Alex-Tideman: Yup realized that looking at the json, it's totally fine. We can change next time there is a 'Create text'.

PR #1486, opened 2023-07-18
Download Workflows list

Alex-Tideman:
## Description & motivation 💭 
As a user, you can download a the current page of the workflows list. The downloaded file is named workflows-{timestamp}.json. 

This PR also fixes the logic of show Loading workflows and removes the Namespace name from the header because we show the Namespace name in the top nav. It also fixes showing NaN workflows when there is an error querying workflows.

Screen Shot 2023-07-18 at 3 26 50 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/holocene/button.svelte
```diff
@@ -131,7 +131,7 @@
   }
 
   .link {
-    @apply border-0 bg-none p-0 font-primary text-sm underline shadow-none;
+    @apply border-0 bg-none p-0 font-primary text-sm text-blue-700 leading-5 font-semibold hover:underline shadow-none;
```
laurakwhit, line 134: Wondering if we want another `Button` variant for all the `Cancel` use cases that use `link`.

|Before | After|
|--|--|
|Screenshot 2023-07-18 at 3 02 59 PM|Screenshot 2023-07-18 at 3 04 43 PM|

src/lib/pages/workflows-with-new-search.svelte
```diff
@@ -266,7 +258,11 @@
             />
           {/if}
         </p>
+        <div class="h-1 w-1 rounded-full bg-gray-400" />
       {/if}
+      <Button variant="link" on:click={() => exportWorkflows($workflows)}
+        >Download JSON</Button
```
laurakwhit, line 264: What do you think about hiding or disabling this if there are no Workflows?

Screenshot 2023-07-18 at 3 07 25 PM

src/lib/utilities/export-workflows.ts
```diff
@@ -0,0 +1,16 @@
...
+
+export const exportWorkflows = (workflows: WorkflowExecution[]) => {
+  const content = stringifyWithBigInt({ workflows }, null, 2);
+  download(content, `workflows-${Date.now()}.json`, 'text/plain');
+
+  function download(content: string, fileName: string, contentType: string) {
+    const a = document.createElement('a');
+    const file = new Blob([content], { type: contentType });
+    a.href = URL.createObjectURL(file);
+    a.download = fileName;
```
laurakwhit, line 13: Would including the range (e.g. 1-100) be helpful in the file name?

src/lib/holocene/button.svelte
Alex-Tideman: This is the css for the link variant in the designs, so it's supposed to be new styles for link button. I looked and that data encoder settings UI is the only other spot we have a link button. I say we roll with it there too (I like it more and don't want to add another variant).

src/lib/pages/workflows-with-new-search.svelte
Alex-Tideman: Yeah probably a good idea, it would just download an empty array in this case. Curious though, how did you get this to not show 0 workflows count to the left of Download JSON?

src/lib/utilities/export-workflows.ts
Alex-Tideman: :+1 I included the count

src/lib/pages/workflows-with-new-search.svelte
laurakwhit: Cropped out the important part of that screenshot, that state was when I was on a Namespace that didn't exist. Looks like with your recent changes the `Download JSON` is disabled for it.

Screenshot 2023-07-19 at 9 19 26 AM

src/lib/holocene/button.svelte
laurakwhit: Fair enough. Just wanted to make sure it was a change we wanted in Cloud too since we use the `link` variant for this pattern all over the place ⬇️ 

Screenshot 2023-07-19 at 9 24 42 AM

PR #1487, opened 2023-07-19
DT-1055 - ensure all text in holocene is rendered via slot or prop

rossedfort:
## Description & motivation 💭 
In order to fully internationalize the UI, we need to make sure there are no hard-coded strings in Holocene components. This PR takes care of that with a few exceptions.
- Day of Week Picker - this is hardcoded to use English days of the week for now
- Month Picker - similarly this is hard coded to use English months for now
- Error/Error Boundary - the library we are using to create the Error boundary does seem to let us pass props to the child components. For now the error text is hard coded in English.
 
I also moved some components out of `src/lib/holocene` and into `src/lib/components`, because they dealt with specific Temporal logic. Those include:
- codec-endpoint-settings.svelte
- data-converter-port-settings.svelte
- data-encoder-settings.svelte
- data-encoder-status.svelte
- namespace-selector.svelte

These components can be translated at a later date.
### Design Considerations 🎨 
N/a
### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


Alex-Tideman:
I guess my general question is why pass in a prop with the translated text instead of translating the string directly in the Holocene component?

Alex-Tideman:
Definitely agree with moving that list of components to `$lib/components`

rossedfort:
> I guess my general question is why pass in a prop with the translated text instead of translating the string directly in the Holocene component?

Good question. If we want to publicize Holocene as a UI component library that can be consumed by external users, they may already be using their own internationalization framework other than i18next, or they may not have any i18n requirement at all. If we translated strings directly in our Holocene components, we'd be requiring everyone who uses Holocene to also bundle i18next in their app and host their own json files with the strings in them. Having all the text as props makes the Holocene agnostic to any potential internationalization frameworks.

Alex-Tideman (approved):

PR #1488, opened 2023-07-19
DT-832 - add i18n to top and side nav

rossedfort:
## Description & motivation 💭 
This PR adds internationalization to the Top and Side navs. Also removed some unused components:
- import-nav.svelte
- namespace-selector.svelte

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/i18n/locales/en/common.ts
```diff
@@ -65,4 +65,11 @@ export const Strings = {
   'minutes-abbreviated': 'min.',
   'seconds-abbreviated': 'sec.',
   days: 'days',
+  workflows: 'Workflows',
```
Alex-Tideman, line 68: Something I was thinking about was doing all these in lowercase and we could do capitalize(translate('workflows')). Since there will probably be places where some of these aren't capitalized and I don't want to have to support workflows and Workflows in these translation files.

But then thinking more that's probably not whats going to happen. Workflows should always be capitalized as it's a Temporal term and we want them capitalized.

Alex-Tideman (approved):

src/lib/i18n/locales/en/common.ts
rossedfort: Yeah I think that's a good idea for words/strings we use a lot that aren't Temporal terms. I haven't run into many of those yet. Once we finish this initial pass, we can take a look at our json files and see if they can be cleaned up. One thing I'd like is to support nesting, but currently the script that translates the TS to JSON doesn't handle nested strings.

PR #1490, opened 2023-07-19
DT-841 - data encoder i18n

rossedfort:
## Description & motivation 💭 
Add internationalization for the Data Encoder modal - note this PR depends on #1487 

Screen Shot 2023-07-19 at 3 45 45 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/components/data-converter-port-settings.svelte
```diff
@@ -18,7 +19,7 @@
       on:keydown|stopPropagation
     />
     <p data-testid="data-encoder-info" class="text-sm w-[220px]">
-      *If both are set, the Codec Server endpoint will be used.
+      {translate('data-encoder', 'port-info')}
```
laurakwhit, line 22: Small thing, but wondering if we want to leave `*` out of the translation since it's more of a stylistic aspect.
```suggestion
      *{translate('data-encoder', 'port-info')}
```

laurakwhit (approved):

src/lib/components/data-converter-port-settings.svelte
rossedfort: Good call, I removed it from the i18n string and used `before:content-['*']` to add the asterisk

PR #1494, opened 2023-07-20
DT-932 make Holocene component labels mandatory

rossedfort:
## Description & motivation 💭 
Make `label` props required across all Holocene components that can render a `<label />`. Also add a `labelHidden` prop that defaults to false. If true, add the `sr-only` class to the label

Components Affected
- Checkbox
- ChipInput
- Input
- NumberInput
- RangeInput
- Select
n/a
### Design Considerations 🎨 
n/a
### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/components/workflow/filter-search/list-filter.svelte
```diff
@@ -22,10 +22,12 @@
 
 <div class="flex">
   <ChipInput
+    label={translate('type-or-paste-in')}
```
laurakwhit, line 25: Would it be helpful to include the attribute here as well?

```suggestion
    label="{translate('type-or-paste-in')} {$filter.attribute}"
```

or even just the attribute

```suggestion
    label="{$filter.attribute}"
```

src/lib/components/workflow/filter-search/text-filter.svelte
```diff
@@ -22,6 +22,8 @@
 </script>
 
 <Input
+  label={translate('search')}
```
laurakwhit, line 25: Similarly, how about using the attribute here so it reflects what should go in the input?

```suggestion
  label={$filter.attribute)}
```

laurakwhit (approved):

src/lib/components/workflow/filter-search/list-filter.svelte
rossedfort: I like just using the attribute, good call 👍

src/lib/components/workflow/filter-search/text-filter.svelte
rossedfort: Will use the attribute same as list-filter

PR #1501, opened 2023-07-25
Improved Elapsed Time & Duration

Alex-Tideman:
## Description & motivation 💭 
Our current elapsed time calculation on the event history is super confusing. This PR simplifies the logic and includes the duration of the event group. 

There are now two numbers:
1. Time since first event in the event history to the initial event in the event group or the event (if not an event group).
2. Time since initial event in event group to last event in event group (does not show anything if not an event group).

I've also included support for milliseconds to get more precision on numbers. The calculations should work as negative numbers when in reverse sort.

Screen Shot 2023-07-25 at 10 20 46 AM

Screen Shot 2023-07-25 at 10 20 32 AM

Screen Shot 2023-07-25 at 10 21 21 AM

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/event/event-summary-row.svelte
+++ b/src/lib/components/event/event-summary-row.svelte
@@ -38,6 +38,12 @@
 
   $: expanded = expandAll || active;
 
+  $: descending = $eventFilterSort === 'descending';
+  $: showElapsed = $eventShowElapsed === 'true';
+  $: showElapsedTimeDiff =
+    showElapsed && initialItem && event.id !== initialItem.id;
+  $: attributes = formatAttributes(event, { compact });
+
   $: currentEvent = isEventGroup(event) ? event.events.get(selectedId) : event;
   $: elapsedTime = formatDistanceAbbreviated({
     start: initialItem.eventTime,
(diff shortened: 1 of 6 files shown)


src/lib/components/event/event-summary-row.svelte
```diff
@@ -90,17 +90,27 @@
...
-    <p class="break-word truncate text-sm md:whitespace-normal md:text-base">
+    <div class="flex flex-col gap-0">
       {#if showElapsedTimeDiff}
-        {formatDistanceAbbreviated({
-          start: initialItem.eventTime,
-          end: currentEvent.eventTime,
-        })}
-        {timeDiffChange}
+        <p class="break-word truncate text-sm md:whitespace-normal">
+          {descending ? '-' : '+'}{elapsedTime}
```
laurakwhit, line 96: To avoid a `-` without a time

Screenshot 2023-07-25 at 11 26 36 AM

maybe we can do something like this?
```suggestion
          {#if elapsedTime}
            {descending ? '-' : '+'}{elapsedTime}
          {/if}
```

Alex-Tideman: 👍

Alex-Tideman:
I fixed the Inter font issue for Safari as well

laurakwhit (approved):

PR #1515, opened 2023-07-28
DT-1226 - add labels to icon buttons

rossedfort:
## Description & motivation 💭 
`<IconButton />` now requires a `label` prop, which is set as `aria-label` on the `<button />` element. This ensures screen-readers can provide useful information about the intent of the button for visually impaired users. 

Note - there were many other Holocene components which utilize `<IconButton />` which necessitated exposing a `label` prop in those components as well.

### Design Considerations 🎨 
N/a
### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

## Checklists
N/a
## Docs
N/a


src/lib/components/namespace-selector.svelte
```diff
@@ -10,7 +10,7 @@
   }
 </script>
 
-<Tooltip text="Select a Namespace">
+<Tooltip top text="Select a Namespace">
```
laurakwhit, line 13: Was this change intentional? And is this component currently being used anywhere atm? Wondering if we want to add translations.

laurakwhit (approved):
Two small comments, but lgtm 🚀

src/lib/components/namespace-selector.svelte
rossedfort: Ah yeah good call, I meant to remove this component because it's unused since Tideman implemented the new top-nav.

PR #1520, opened 2023-08-01
DT-1230 - toggle button a11y

rossedfort:
## Description & motivation 💭 
Add proper accessibility behavior to `<ToggleButton />` as well as update to current figma designs
- add keyboard navigation and hover/focus state
- add screen reader support
- use correct colors and remove icons

This PR is based loosely on https://www.w3.org/WAI/ARIA/apg/patterns/switch/examples/switch-checkbox/

default "off" state
Screen Shot 2023-08-01 at 12 02 03 PM

focused "off" state
Screen Shot 2023-08-01 at 12 02 15 PM

hover "off" state
Screen Shot 2023-08-01 at 12 02 19 PM

disabled "off" state
Screen Shot 2023-08-01 at 12 06 43 PM

default "on" state
Screen Shot 2023-08-01 at 12 03 29 PM

focused "on" state
Screen Shot 2023-08-01 at 12 02 10 PM

hover "on" state
Screen Shot 2023-08-03 at 9 04 24 AM

disabled "on" state
Screen Shot 2023-08-01 at 12 07 31 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/holocene/toggle-switch.svelte
+++ b/src/lib/holocene/toggle-switch.svelte
@@ -65,7 +65,7 @@
 
 <style lang="postcss">
   .switch {
-    @apply px-2 relative flex items-center rounded max-w-fit cursor-pointer;
+    @apply px-2 relative flex items-center rounded min-w-fit cursor-pointer;
 
     &.disabled {
       @apply pointer-events-none;
--- a/src/lib/holocene/toggle-switch.svelte
+++ b/src/lib/holocene/toggle-switch.svelte
@@ -91,8 +91,12 @@
   .slider {
     @apply flex relative items-center h-5 w-9 rounded-xl border border-gray-600 bg-gray-50;
 
-    &.hovered {
+    &.hovered:not(.checked) {
       @apply bg-gray-200;
+    }
+
+    &.hovered.checked {
+      @apply bg-indigo-800;
     }
 
     &.focused:not(.checked) {
--- a/src/lib/holocene/toggle-switch.svelte
+++ b/src/lib/holocene/toggle-switch.svelte
@@ -54,7 +54,7 @@
     class:focused
     class:hovered
   >
-    <span class="thumb" class:disabled class:checked />
+    <span class="thumb" class:checked />
   </span>
   {#if labelPosition === 'right'}
     <span class="label right" class:sr-only={labelHidden}>
@@ -68,7 +68,7 @@
     @apply px-2 relative flex items-center rounded max-w-fit cursor-pointer;
 
     &.disabled {
-      @apply cursor-not-allowed pointer-events-none;
+      @apply pointer-events-none;
     }
   }
 
@@ -108,7 +108,7 @@
     }
 
     &.disabled {
-      @apply bg-gray-100;
+      @apply opacity-50;
     }
 
     &.checked {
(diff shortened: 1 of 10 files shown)


src/lib/holocene/toggle-switch.svelte
```diff
@@ -1,53 +1,130 @@
...
+    @apply absolute w-0 opacity-0;
   }
 
-  input.disabled + .slider {
-    @apply bg-gray-200 cursor-not-allowed;
+  .slider {
+    @apply flex relative items-center h-5 w-9 rounded-xl border border-gray-600 bg-gray-50;
+
+    &.hovered {
+      @apply bg-gray-200;
```
laurakwhit, line 95: What do you think about adding a hover state for the toggled on state as well? Maybe a slightly darker indigo?

Screenshot 2023-08-02 at 9 23 32 AM

src/lib/holocene/toggle-switch.svelte
```diff
@@ -1,53 +1,130 @@
...
+    &.focused:not(.checked) {
+      @apply border-indigo-600;
+
+      box-shadow: 0 0 0 4px #2463eb80; /* blue-600 at 50% opacity */
+    }
+
+    &.focused.checked {
+      @apply border-white;
+
+      box-shadow: 0 0 0 4px #2463eb80; /* blue-600 at 50% opacity */
```
laurakwhit, line 107: Should we add `#2463eb80` to the theme? Seems like something we'll likely want to use elsewhere.

src/lib/holocene/toggle-switch.svelte
```diff
@@ -1,53 +1,130 @@
...
+    <span class="label right" class:sr-only={labelHidden}>
+      {label}
+    </span>
+  {/if}
 </label>
 
 <style lang="postcss">
   .switch {
-    @apply relative inline-block h-8 w-[52px] rounded-[8rem];
+    @apply px-2 relative flex items-center rounded max-w-fit cursor-pointer;
```
laurakwhit, line 68: We might want a `min-w-fit` here instead

```suggestion
    @apply px-2 relative flex items-center rounded min-w-fit cursor-pointer;
```

so that it doesn't get messed up on smaller screens

| Before | After |
|--|--|
|Screenshot 2023-08-02 at 9 30 24 AM|Screenshot 2023-08-02 at 9 31 07 AM|

laurakwhit (approved):
So much better 💅

src/lib/holocene/toggle-switch.svelte
rossedfort: I'll check with Ash, but the hover state in Figma is the same as the default checked state.
Screen Shot 2023-08-02 at 3 44 06 PM

src/lib/holocene/toggle-switch.svelte
rossedfort: Yeah good call, I tried that, but couldn't get it to work. Will try again 👍

rossedfort: Turns out you can set the shadow color separately with opacity, like `shadow-blue-600/50` https://tailwindcss.com/docs/box-shadow-color

So I made a `shadow-focus`  for `0 0 0 4px`, which will use the color set by `shadow-blue-600/50` 🎉

laurakwhit (approved):

PR #1528, opened 2023-08-03
Additional i18n translations

laurakwhit:
## Description & motivation 💭 
Updates the following w/ translations
* `/select-namespace` 
* `/namespaces/[namespace]/archival`
* `/import/events`
* `aria-labels` on next/previous buttons in Pagination components 
* Temporal and UI version banners

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
`DT-1324`

### DIFF
--- a/src/routes/(app)/namespaces/[namespace]/archival/_workflow-filters.svelte
+++ b/src/routes/(app)/namespaces/[namespace]/archival/_workflow-filters.svelte
@@ -10,15 +10,31 @@
   import Search from '$lib/components/search.svelte';
 
   import { timeFormat } from '$lib/stores/time-format';
+  import { translate } from '$lib/i18n/translate';
 
   const statuses = {
-    All: null,
-    'Timed Out': 'TimedOut',
-    Completed: 'Completed',
-    Failed: 'Failed',
-    'Continued as New': 'ContinuedAsNew',
-    Canceled: 'Canceled',
-    Terminated: 'Terminated',
+    All: { value: null, label: translate('workflows', 'all-statuses') },
+    'Timed Out': {
+      value: 'TimedOut',
+      label: translate('workflows', 'timed-out'),
+    },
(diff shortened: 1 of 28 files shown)


src/routes/(app)/namespaces/[namespace]/archival/_workflow-filters.svelte
```diff
@@ -81,28 +97,32 @@
...
+        name={translate('workflow-type')}
         value={workflowTypeFilter}
       />
-      <FilterSelect label="Time Range" parameter="time-range" value="24 hours">
+      <FilterSelect
+        label={translate('workflows', 'time-range')}
+        parameter="time-range"
+        value="24 hours"
+      >
         {#each durations as value}
```
Comment (laurakwhit) [inline src/routes/(app)/namespaces/[namespace]/archival/_workflow-filters.svelte:113]: Have we aligned on an approach to durations yet (e.g. hrs/mins/days)?

Comment (rossedfort) [reply src/routes/(app)/namespaces/[namespace]/archival/_workflow-filters.svelte:113]: Not yet, I think we'll need to refactor some components a little, as I don't think we can call `translate()` outside of a svelte component.

rossedfort (approved):

src/routes/(app)/namespaces/[namespace]/archival/_workflow-filters.svelte
Comment (Alex-Tideman) [reply src/routes/(app)/namespaces/[namespace]/archival/_workflow-filters.svelte:113]: like the approach to translate them?

Comment (rossedfort) [reply src/routes/(app)/namespaces/[namespace]/archival/_workflow-filters.svelte:113]: Yeah, we might have to make these i18n keys, https://github.com/temporalio/ui/blob/main/src/lib/utilities/to-duration.ts#L7-L37 but then there are other utils that use those files, so they'd need to be refactored I think. There might be a simpler way to do it though

PR #1535, opened 2023-08-08
DT-1233 and DT-1254 - Namespace switcher and Split Button a11y

rossedfort:
## Description & motivation 💭 
This PR provides better screen reader announcements for the namespace switch dropdown menu by adding an aria label to dropdown menu button. Focusing the dropdown button should now read `Namespace <current namespace>` instead of just `<current namespace>`.

Also adds a labels to the `<Menu />` components in Split Buttons and makes `<MenuDivider />` invisible to the screen reader and not reachable via keyboard navigation.

**NOTE** this changes the `<Menu />` component to use `hidden` instead of `sr-only` when the menu is closed. This prevents the menu and menu-item elements from being reached via keyboard navigation and the screen reader when the menu is closed.

N/a
### Design Considerations 🎨 
N/a
### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

- Open the web UI and navigate the namespace switcher with the screen reader to verify a good experience.
- Open the web UI and navigate a split button with the screen reader to verify a good experience. A good example is the client actions split button for Running workflows.
## Checklists
N/a
### Issue(s) closed 
N/a
## Docs
N/a

### DIFF
--- a/src/lib/components/top-nav.svelte
+++ b/src/lib/components/top-nav.svelte
@@ -41,7 +41,8 @@
     {#if showNamespaceSpecificNav}
       {#key namespace}
         <DropdownMenu
-          id="namespace"
+          label={translate('namespaces', 'namespace-label', { namespace })}
+          id="namespace-switcher"
           position="right"
           menuClass="border-2 bg-purple-200"
           buttonClass="border border-purple-700 rounded-sm"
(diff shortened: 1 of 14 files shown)


laurakwhit (approved):
So much better 🎉

PR #1536, opened 2023-08-08
Versioning Warnings -> No Workers or Incompatible Workers

Alex-Tideman:
## Description & motivation 💭 
We want to alert users if they are using versioning and there are no compatible workers for the version the workflow is running on. A workflow is considered to have no compatible workers if there are no workers with the default version of the version set that the workflow version is on. For example:

Version Sets
[2.0,2.1,**2.2**]
[3.0,3.1,**3.2**]
Workflow is running and using 2.1.
Workers are compatible only if one or more of them is using 2.2

Also included in this PR is an update to the 'No Workers Running' alert to match designs.

I've also had to refactor how we fetch reachability. It used to only fetch on the workers list, but now with this warning we need to fetch it in the workflow run layout.

Screen Shot 2023-08-08 at 12 46 57 PM
Screen Shot 2023-08-08 at 12 48 35 PM

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


static/i18n/locales/en/common.json
```diff
@@ -0,0 +1,94 @@
+{
```
laurakwhit, line 1: Hmm I don't think we want these json files in here 📁

laurakwhit (approved):

PR #1537, opened 2023-08-08
Alternative text for images

laurakwhit:
## Description & motivation 💭 
Adds/verifies all `` and `<svg/>` elements have correct alt text or title.

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
`DT-1257`


src/lib/holocene/icon/svg/svg.svelte
```diff
@@ -11,7 +11,8 @@
   fill="none"
   xmlns="http://www.w3.org/2000/svg"
   class={$$props.class}
+  aria-labelledby="icon-title"
 >
-  <title>{title}</title>
+  <title id="icon-title">{title}</title>
```
rossedfort, line 16: I'm not 100% certain, but I didn't think `aria-labelledby` is necessary with `<title />` as a child of `<svg />`. Also - we should add `role="img"` to `<svg />` here I think.

https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title

check out the Test Cases > Passed Example 1 here https://www.w3.org/WAI/standards-guidelines/act/rules/7d6734/

rossedfort: Given we can have many `<Icon />`'s on the same page, having a static ID here may not be the best choice. If we do need an `id` and `aria-labelledby`, we should pass in a unique ID, or maybe generate one from the icon name.

laurakwhit: R.e. `aria-labelledby`, I've seen both patterns (with and without) so I just went with it to be sure or (as [this resource says](https://www.unimelb.edu.au/accessibility/techniques/accessible-svgs))
> The SVG has an aria-labelledby attribute which references a title element. Note: the title element should be sufficient on it's own, but we added aria-labelledby to be sure.

Good point r.e. unique IDs. I've gone ahead and used the `title` to create the ID, but perhaps we should use something even more unique? I've also added `role="img"`.

src/routes/(login)/signin/+page.svelte
```diff
@@ -17,9 +17,7 @@
 
 <PageTitle title="Login" url={$page.url.href} />
 <header class="flex h-16 w-full items-center justify-between bg-primary px-10">
-  <a href="/" class="inline-block">
-    <img src={Logo} alt="Temporal Logo" class="max-h-10" />
-  </a>
+  <img src={Logo} alt="" class="max-h-10" />
```
rossedfort, line 20: And here

rossedfort (approved):

src/lib/holocene/loading.svelte
```diff
@@ -9,9 +9,12 @@
   <img
     src={Logo}
     style="margin-top: -40px;"
-    alt="Temporal Logo"
+    alt=""
```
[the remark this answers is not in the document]
laurakwhit: Yup! Made them [decorative](https://www.w3.org/WAI/tutorials/images/decorative/) instead.

PR #1538, opened 2023-08-08
Workflow Counts UI (not added until GroupBy API ready in Temporal v1.22)

Alex-Tideman:
## Description & motivation 💭 
Add the count for each ExecutionStatus for workflow list. Clicking on a status will add it as a filter. You can combine multiples of them, and clicking an active status will remove it from the filter.

The <WorkflowCounts /> components isn't added to the Workflow List page since we don't have the GroupBy api ready yet. But when it is, we can add this in. The numbers shown in the count cards are purely random, fake numbers as placeholders.

https://github.com/temporalio/ui/assets/7967403/3e6a1fd3-4814-454f-bbb8-a422680667cf

Screen Shot 2023-08-09 at 10 58 16 AM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/workflow/workflow-counts.svelte
+++ b/src/lib/components/workflow/workflow-counts.svelte
@@ -77,7 +77,7 @@
   };
 </script>
 
-<div class="flex gap-4 flex-wrap">
+<div class="flex gap-2 lg:gap-4 flex-wrap">
   {#each workflowStatuses as status}
     <WorkflowCount
       {status}


src/lib/components/workflow/workflow-count.svelte
```diff
@@ -0,0 +1,30 @@
+<script lang="ts">
+  import type { WorkflowStatus as Status } from '$lib/types/workflows';
+  import WorkflowStatus from '../workflow-status.svelte';
+
+  export let status: Status;
+  export let count: number = Math.round(Math.random() * 10000);
```
rossedfort, line 6: Seems odd to have a default value for the count, is this to avoid an empty box when the counts are loading? Should we use a loading indicator instead?

src/lib/components/workflow/workflow-counts.svelte
```diff
@@ -0,0 +1,88 @@
...
+    });
+  };
+</script>
+
+<div class="flex gap-4 flex-wrap">
+  {#each workflowStatuses as status}
+    <WorkflowCount
+      {status}
+      {onStatusClick}
+      active={Boolean(statusFilters.find((filter) => filter.value === status))}
```
rossedfort, line 85: nit: could use Array.prototype.some() and remove the Boolean coercion.

rossedfort (commented):
Couldn't we just add this to the workflows page under a version guard and then it will only show up when 1.22 is released?

src/lib/components/workflow/workflow-count.svelte
rossedfort: Oh I see this is just for testing purposes. Can remove the default value once we get the API 👍

laurakwhit: Maybe we can add a `TODO` here?

src/lib/components/workflow/workflow-count.svelte
```diff
@@ -0,0 +1,30 @@
...
+  export let active = false;
+  export let onStatusClick: (status: Status) => void = () => {};
+</script>
+
+<div
```
laurakwhit, line 11: There are responsive designs for these, will send you a link! 

Screenshot 2023-08-08 at 4 24 20 PM

src/lib/components/workflow/workflow-counts.svelte
```diff
@@ -0,0 +1,88 @@
...
+    updateQueryParameters({
+      url: $page.url,
+      parameter: 'query',
+      value: searchQuery,
+      allowEmpty: true,
+    });
+  };
+</script>
+
+<div class="flex gap-2 lg:gap-4 flex-wrap">
```
laurakwhit, line 80: Only thought it that maybe we want to keep the `gap-4` here on smaller screens 
```suggestion
<div class="flex gap-4 flex-wrap">
```
so that there's still some space on hover
| gap-2 | gap-4|
|--|--|
|Screenshot 2023-08-09 at 10 13 08 AM|Screenshot 2023-08-09 at 10 13 56 AM|

laurakwhit (approved):

src/lib/components/workflow/workflow-counts.svelte
Alex-Tideman: I'm gonna make that box-shadow not so aggressive on sm screens, I like the smaller gap.

PR #1540, opened 2023-08-09
add new combobox - remove old namespace switcher components

rossedfort:
## Description & motivation 💭 
Adds new `<Combobox />` and `<ComboboxOption />` components, and use them in the namespace switcher.

Combobox is designed to be flexible such that it can take an array of strings and render its own ComboboxOptions, or it can take an array of any object type, and exposes each option in a slot. It also has keyboard navigation and proper aria labels baked in for screen reader and keyboard navigation.

Also refactor`<MenuContainer />`, `<MenuButton />`, `<Menu />` and `<MenuItem />` to have keyboard navigation, hover, and focus states baked in, and use these components anywhere a dropdown menu is utilized. This includes our existing `<Select />` component.

Also remove `src/lib/components/dropdown-menu.svelte`, `src/lib/holocene/dropdown-menu.svelte`, `src/lib/holocene/autocomplete.svelte`, and `src/lib/holocene/dropdown-button/`. 

Move Menu components out of `src/lib/holocene/primitives` to `src/lib/holocene/menu`. 

https://github.com/temporalio/ui/assets/11775628/f0379724-15d0-4cbe-903c-0b12687542f5

### Design Considerations 🎨 
- Need to confirm with Ash that the Component matches Figma
### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

Either in the Web UI or the Holocene preview URL, verify the following
- tabbing into the Combobox displays a focus state and places a cursor in the input
  - hitting "Enter" or "DownArrow" opens the Combobox options and focuses the first option
  - typing opens the Combobox options
  - `Tab` and `Shift+Tab` cycle through the input and each option, without trapping focus
  - `UpArrow` and `DownArrow` cycle through the input an each option _with_ trapping focus
  - `Enter` and `Space` select a Combobox Option while it is focused
- clicking on the Combobox displays a focus state and opens the Combobox options
  - `Tab` and `Shift+Tab` cycle through the input and each option, without trapping focus
  - `UpArrow` and `DownArrow` cycle through the input an each option _with_ trapping focus
  - `Enter` and `Space` select a Combobox Option while it is focused
- A checkmark is displayed for the selected option
- Disabled options cannot be selected


Alex-Tideman:
Some thoughts:

Going to be a challenge to make this select wide enough to show the whole namespace name without making the select obnoxiously wide in the nav.
Screen Shot 2023-08-10 at 2 50 29 PM

We'll want a max height on the dropdown list and better formatting of the list to show the full namespace name
Screen Shot 2023-08-10 at 2 52 36 PM

You can get in a state where the input is cleared and then unfocused and it stays clear. I also think we want the caret-up and caret-down icon clickable to open without needing to focus in the input.
Screen Shot 2023-08-10 at 1 36 30 PM

Alex-Tideman:
Might want break-all here:

Screen Shot 2023-08-17 at 12 27 40 PM

Alex-Tideman:
I think this is going to come up more than we think, where there are multiple namespaces that are fairly long with similar prefixes. When you cut off the end, it's impossible to know the exact namespace you are on (minus looking at the url)

Screen Shot 2023-08-17 at 12 32 52 PM
Screen Shot 2023-08-17 at 12 32 47 PM

rossedfort:
> I think this is going to come up more than we think, where there are multiple namespaces that are fairly long with similar prefixes. When you cut off the end, it's impossible to know the exact namespace you are on (minus looking at the url)

True, I can look at changing some styles in the top-nav to let the combobox grow when there's a long namespace name. And will add break-all to the options.

src/lib/holocene/combobox/combobox.story.svelte
```diff
@@ -0,0 +1,106 @@
...
+      label="Select a Language"
+      noResultsText="No Results"
+      id="combobox-2"
+      options={customOptions}
+      optionLabelKey="label"
+      optionValueKey="value"
+    />
+  </Hst.Variant>
+
+  <Hst.Variant title="a combobox with custom options (controlled)">
```
laurakwhit, line 78: One behavior I noticed is that if you edit an input value and then try and select the same value again, the input value isn't updated (e.g. select Japanese > Delete a few characters > select Japanese again).

Screenshot 2023-08-18 at 11 04 15 AM

src/lib/components/schedule/schedules-interval-view.svelte
```diff
@@ -107,7 +101,7 @@
   <p>
     {translate('schedules', 'offset-description')}
```
laurakwhit, line 102: Are the different rounded borders here expected for the button vs. input? Also, looks like there's an extra border between the input and button now.

| Before | After |
|--|--|
|Screenshot 2023-08-18 at 10 53 40 AM|Screenshot 2023-08-18 at 10 15 06 AM|

src/lib/holocene/select/option-group.svelte
```diff
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import MenuItem from '../primitives/menu/menu-item.svelte';
+  import MenuItem from '$lib/holocene/menu/menu-item.svelte';
 
   export let label: string;
```
laurakwhit, line 4: Wondering if we want a little bit of an indent between the label and the options?

Screenshot 2023-08-18 at 11 01 25 AM

src/lib/components/workflow/filter-search/datetime-filter.svelte
```diff
@@ -141,17 +137,13 @@
...
-      class="flex flex-row items-center p-2 bg-white text-gray-800 hover:border-primary hover:bg-primary hover:text-white border border-gray-800 rounded-r h-10"
     >
       {translate('workflows', 'select-time')}
     </MenuButton>
-    <Menu
-      keepOpen
-      id="time-range-filter-menu"
-      class="flex rounded h-auto w-[400px] flex-col gap-8 bg-white p-2"
-    >
+    <Menu keepOpen id="time-range-filter-menu" class="w-[400px] p-2">
```
laurakwhit, line 146: Not sure how we want to handle dropdowns within dropdowns, but not being able to see the content unless you scroll definitely ins't ideal.

| Before | After |
|--|--|
|Screenshot 2023-08-18 at 11 25 27 AM|Screenshot 2023-08-18 at 11 26 09 AM|

src/lib/holocene/split-button.svelte
```diff
@@ -23,19 +19,17 @@
   export let position: 'left' | 'right' = 'left';
   export let href = '';
   export let primaryActionDisabled = false;
-
-  let show = false;
 </script>
 
 <MenuContainer class={$$props.class}>
-  <div class="split-button" class:disabled>
+  <div class="split-button">
```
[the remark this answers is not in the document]
rossedfort: Hm yeah the new stuff Ash is working on is all `rounded-lg`, i.e. 8px of border-radius, so I think we can just change our Button component accordingly

src/lib/holocene/select/option-group.svelte
rossedfort: We currently have any Selects with grouped options, and Select will be deprecated soon in favor of Combobox, so I'm not too worried about this.

src/lib/components/schedule/schedules-interval-view.svelte
rossedfort: Hm yeah, maybe input should be 8px border-radius now as well

rossedfort: I think I'm going to make a separate component, or maybe extend our `<Input />` component to support an adjacent select in a separate PR.

src/lib/components/workflow/dropdown-filter/workflow-status.svelte
```diff
@@ -75,51 +82,39 @@
...
+    data-testid="execution-status-filter-button"
+    variant="table-header"
+    controls="execution-status-filter"
+  >
+    {translate('status')}
+    <Icon name="filter" slot="trailing" />
+  </MenuButton>
+  <Menu keepOpen id="execution-status-filter">
+    {#each workflowStatusFilters as status}
+      <MenuItem on:click={() => onStatusClick(status)}>
```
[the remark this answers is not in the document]
laurakwhit: @rossedfort looks like the spacing got a little off here

Screenshot 2023-08-21 at 11 08 13 AM

src/lib/components/workflow/dropdown-filter/workflow-datetime-filter.svelte
```diff
@@ -174,18 +176,12 @@
...
         >
           {value}
         </MenuButton>
-        <Menu
-          keepOpen
-          id="time-range-filter-menu"
-          class="flex rounded h-auto w-[400px] flex-col gap-8 bg-white p-2"
-          position="right"
-        >
+        <Menu position="right" keepOpen id="time-range-filter-menu">
```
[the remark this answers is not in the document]
laurakwhit: @rossedfort we might want to add a little padding above the `End` here too

Screenshot 2023-08-21 at 11 22 31 AM

laurakwhit (approved):

Alex-Tideman (approved):

PR #1559, opened 2023-08-17
Fix relative time filter 

laurakwhit:
## Description & motivation 💭 
* Update time range filter values when `Relative` is toggled on/off
* Converts relative time filter to a date to fix Invalid Date error
* Fixes position of datetime filter dropdown

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

- [ ] Verify the datetime filter dropdown is positioned correctly and does not go off the page

In Labs mode
- [ ] Verify a time range value is updated with `Relative` is togged on/off
- [ ] Verify selecting a relative time works as expected (and is converted to the timezone selected)

### DIFF
--- a/src/lib/components/workflow/filter-search/filter-list.svelte
+++ b/src/lib/components/workflow/filter-search/filter-list.svelte
@@ -1,4 +1,5 @@
 <script lang="ts">
+  import { parseJSON } from 'date-fns';
   import { fade } from 'svelte/transition';
   import { getContext } from 'svelte';
   import { type FilterContext, FILTER_CONTEXT } from './index.svelte';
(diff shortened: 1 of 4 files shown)


src/lib/components/workflow/filter-search/filter-list.svelte
```diff
@@ -1,4 +1,5 @@
 <script lang="ts">
+  import { parseJSON } from 'date-fns';
```
rossedfort, line 2: Unused import?

rossedfort (approved):

src/lib/components/workflow/filter-search/filter-list.svelte
laurakwhit: Yup, removed ✅

rossedfort:
> When was it offscreen? When it's small screen, position right is cut off. Position left doesn't.

Yeah agree, I think we want this `position="left"`

laurakwhit:
@Alex-Tideman I just fixed the position for the existing datetime filter dropdown in this PR ⬇️ 
|Before | After |
|--|--|
|Screenshot 2023-08-17 at 10 49 01 AM|Screenshot 2023-08-17 at 10 49 59 AM|

For the timezone dropdown responsiveness I believe @rossedfort has some goodness in the new combox PR.

PR #1560, opened 2023-08-17
Make unique aria-ids

Alex-Tideman:
## Description & motivation 💭 
Make sure all aria-labels are unique.

Also fix version text contrast.

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/components/workflow/batch-operation-confirmation-modal.svelte
```diff
@@ -44,7 +44,7 @@
 </script>
 
 <Modal
-  id="batch-operation-confirmation-modal"
+  id={`batch-operation-confirmation-modal-${action}`}
```
laurakwhit, line 47: Small thing, but we can make these a tiny bit simpler like so

```suggestion
  id="batch-operation-confirmation-modal-{action}"
```

src/lib/components/workflow/batch-operation-confirmation-modal.svelte
```diff
@@ -97,7 +97,7 @@
       {/if}
     </div>
     <Input
-      id="bulk-action-reason"
+      id={`bulk-action-reason-${action}`}
```
laurakwhit, line 100:
```suggestion
      id="bulk-action-reason-{action}"
```

PR #1568, opened 2023-08-22
Descriptive link and various fixes

laurakwhit:
## Description & motivation 💭 
* Removes generic text in links (e.g. `Show all`)
* Adds some missing translations
* Fixes overflow of Namespace name on Alert component

| Before | After |
|--|--|
|Screenshot 2023-08-22 at 11 17 55 AM|Screenshot 2023-08-22 at 11 18 03 AM|
### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
`DT-1236`

### DIFF
--- a/src/lib/components/workflow/workflows-summary-configurable-table/table-empty-state.svelte
+++ b/src/lib/components/workflow/workflows-summary-configurable-table/table-empty-state.svelte
@@ -47,7 +47,7 @@
           intent="caution"
           icon="warning"
           title={translate('workflows', 'workflow-query-error-state')}
-          >{$workflowError}</Alert
+          style="overflow-wrap: anywhere">{$workflowError}</Alert
         >
       {:else if query}
         <p>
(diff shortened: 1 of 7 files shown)


src/lib/components/workflow/workflows-summary-configurable-table/table-empty-state.svelte
```diff
@@ -47,7 +47,7 @@
           intent="caution"
           icon="warning"
           title={translate('workflows', 'workflow-query-error-state')}
-          >{$workflowError}</Alert
+          style="overflow-wrap: anywhere">{$workflowError}</Alert
```
laurakwhit, line 50: Added this here so there wouldn't be any adverse impacts on other `Alert` components, but wondering if we want to do an audit and potentially just add this to the `Alert` component itself since there could be other similar cases.

Alex-Tideman (approved):

PR #1584, opened 2023-08-23
Use Tailwind layers, rename workflow status classes

stevekinney:
- Moves base styles into the appropriate Tailwind layer.
- Makes workflow status classes generally available.
- Renames workflow status classes for semantic meaning.
- Removes reactive statements in favor of view layer reactivity.


src/lib/components/workflow-status.svelte
```diff
@@ -20,58 +19,23 @@
...
-    Terminated: 'gray',
-    Paused: 'yellow',
-  };
-
-  $: color = colors[status];
-  $: label = humanFriendlyNames[status];
-  $: isRunning = label === humanFriendlyNames.Running;
 </script>
 
 <span class="flex text-center text-sm font-medium leading-4">
```
rossedfort, line 24: We could just use `<Badge />` here and get rid of those extra color styles

```
<Badge type={color}>
  <span class="whitespace-nowrap">{label}</span>
  {#if isRunning}
    <HeartBeat {delay} />
  {/if}
</Badge>
```

stevekinney: Let's give it a shot.

rossedfort (approved):

PR #1586, opened 2023-08-24
Fixes based on A11y audit 

laurakwhit:
## Description & motivation 💭 
Various fixes for accessibility violations found in https://github.com/temporalio/ui/pull/1571.

| Before | After |
|--|--|
|Screenshot 2023-08-23 at 2 43 38 PM|Screenshot 2023-08-23 at 2 48 26 PM|
### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

- [ ] Verify `Table` components still look as expected
- [ ] Verify `Copyable` and `Codeblock` components still look/work as expected
- [ ] Verify `Combobox` component still looks/works as expected


src/lib/holocene/tooltip.svelte
```diff
@@ -104,7 +104,6 @@
             {copyIconTitle}
             clickAllToCopy
             content={text}
-            color="white"
```
laurakwhit, line 107: I removed `color` from `Copyable` since it uses an `IconButton` now (it looks like it wasn't working when it was using just an `Icon` either). 

Wondering how we might want to go about passing a different `stroke` color here or if it's even necessary?

src/lib/holocene/pagination.svelte
```diff
@@ -143,7 +143,7 @@
       style={floatStyle}
       bind:clientHeight={height}
       class="flex flex-col items-end gap-4 md:flex-row min-w-fit"
-      aria-label={$$restProps['aria-label']}
+      aria-label="{$$restProps['aria-label']} 1"
```
laurakwhit, line 146: I just added a `1` and a `2` here to delineate pagination navigation, but wondering if there might be a better solution.

src/lib/holocene/code-block.svelte
```diff
@@ -1,15 +1,35 @@
...
 
+  type BaseProps = HTMLAttributes<HTMLDivElement> & {
+    content: Parameters<typeof JSON.stringify>[0];
+    inline?: boolean;
+    language?: string;
+    async?: boolean;
+    testId?: string;
+  };
+
+  type CopyableProps = BaseProps & {
```
laurakwhit, line 19: This typing doesn't appear to be working as expected (I think because we're defaulting `copyable` to `true`). Any ideas how we can ensure `copyIconTitle` and `copySuccessIconTitle` are included if `copyable` is not passed as a prop?

src/lib/holocene/api-pagination.svelte
```diff
@@ -179,7 +179,7 @@
     </div>
     <nav
       class="flex flex-col justify-end gap-4 md:flex-row"
-      aria-label={$$restProps['aria-label']}
+      aria-label="{$$restProps['aria-label']} 1"
```
rossedfort, line 182: Maybe "top" and "bottom" would be better here?

src/lib/holocene/code-block.svelte
rossedfort: Hm ok, I can check this out locally

src/lib/holocene/pagination.svelte
rossedfort: How about "top" and "bottom"?

src/lib/holocene/table/table-header-row.svelte
```diff
@@ -1,5 +1,5 @@
 <tr class={$$props.class}>
-  <th />
+  <td />
   <slot />
-  <th />
+  <td />
```
rossedfort, line 4: I'm going to try to play with getting rid of these entirely 😈

rossedfort (approved):

src/lib/holocene/api-pagination.svelte
laurakwhit: I was trying to keep it something that wouldn't need to be translated (hence the 1 and 2) 😅

src/lib/holocene/table/table-header-row.svelte
laurakwhit: 💯 % behind that endeavor

src/lib/holocene/tooltip.svelte
laurakwhit: **Update**: Fixed in https://github.com/temporalio/ui/pull/1586/commits/e034d84f92064f6961cb181e0dc6c8a4a8a28632

PR #1589, opened 2023-08-24
DT-1248 - consistent link styles

rossedfort:
## Description & motivation 💭 
Use `<Link />` component wherever possible and adjust styles to match new figma designs
### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


Alex-Tideman:
I think this button is a little too tall now, makes wonky spacing in the event history table.

Screen Shot 2023-08-29 at 3 27 04 PM
Screen Shot 2023-08-29 at 3 28 42 PM

Alex-Tideman:
But dang that button is 🔥

src/lib/components/schedule/schedule-form-view.svelte
```diff
@@ -1,13 +1,13 @@
 <script lang="ts">
```
laurakwhit, line 1: Unrelated to this PR, but wondering if we want to make this a bit bigger so `Delete Schedule` fits on one line
Screenshot 2023-08-29 at 3 59 55 PM

src/lib/components/workflow/filter-search/filter-list.svelte
```diff
@@ -134,7 +134,7 @@
     {/if}
   {/each}
   {#if hasMoreFilters}
```
laurakwhit, line 136: Also unrelated to this PR, but it looks like the `BooleanFilter` could use a `borderRadiusModifier`

Screenshot 2023-08-29 at 4 04 03 PM

src/lib/holocene/combobox/combobox.svelte
```diff
@@ -245,13 +244,13 @@
   </label>
 
   <div
-    class="text-sm w-full h-10 flex flex-row items-center rounded-lg bg-white border border-primary"
+    class="combobox-wrapper"
```
laurakwhit, line 247: Wondering if we want some padding on the left here when there isn't an icon

Screenshot 2023-08-29 at 4 17 08 PM

src/lib/holocene/modal.svelte
```diff
@@ -137,15 +137,11 @@
       {/if}
     </div>
     <div class="flex items-center justify-end space-x-2 p-6">
-      <Button
-        thin
-        variant="secondary"
-        disabled={loading}
-        on:click={handleCancel}>{cancelText}</Button
+      <Button variant="secondary" disabled={loading} on:click={handleCancel}
```
laurakwhit, line 140: Do we want a `size="xs"` here to replace `thin`?

src/lib/holocene/scroll-to-top.svelte
```diff
@@ -9,8 +9,7 @@
   <slot />
   <div class="back-to-top" class:hidden>
     <Button
-      class="!py-1.5 !px-1"
-      icon="arrow-up"
+      leadingIcon="arrow-up"
```
laurakwhit, line 12: I don't think this is showing up as intended anymore unfortunately (see [this preview](https://holocene.preview.thundergun.io/story/src-lib-holocene-scroll-to-top-story-svelte?variantId=src-lib-holocene-scroll-to-top-story-svelte-2)) 😞 Perhaps something with other recent changes?

Also wondering if we want it to be an `IconButton` instead? If not, we might want to remove some padding here

Screenshot 2023-08-29 at 4 32 01 PM

src/lib/holocene/split-button.svelte
```diff
@@ -8,26 +8,18 @@
...
-  export let thin = false;
+  export let variant: 'primary' | 'secondary' | 'destructive' | 'ghost' =
+    'primary';
   export let disabled = false;
   export let position: 'left' | 'right' = 'left';
-  export let href = '';
   export let primaryActionDisabled = false;
 </script>
 
 <MenuContainer class={$$props.class}>
```
laurakwhit, line 18: Wondering if we want to do something with `z-index` for the border on hover for these buttons

Screenshot 2023-08-29 at 3 50 14 PM
Screenshot 2023-08-29 at 3 50 34 PM

src/lib/holocene/button.svelte
```diff
@@ -1,222 +1,139 @@
...
-  export interface ButtonProps extends HTMLButtonAttributes {
-    variant?: ButtonVariant;
-    id?: string;
+  import { cva, type VariantProps } from 'class-variance-authority';
+
+  import Badge from '$lib/holocene/badge.svelte';
+  import Icon from '$lib/holocene/icon/icon.svelte';
+  import type { IconName } from '$lib/holocene/icon/paths';
+
+  const buttonStyles = cva(
```
laurakwhit, line 13: 💅

src/lib/components/event/event-summary-table.svelte
```diff
@@ -3,7 +3,7 @@
 
   import EventCategoryFilter from '$lib/components/event/event-category-filter.svelte';
```
laurakwhit, line 4: Looks like the links in the expanded row could use centering
|||
|--|--|
|Screenshot 2023-08-29 at 3 53 44 PM|Screenshot 2023-08-29 at 3 52 47 PM|

src/lib/components/event/event-summary-table.svelte
```diff
@@ -36,15 +36,15 @@
     <th class="w-44"><EventCategoryFilter {compact} /></th>
     <th class="w-auto xl:w-80">
       <div class="flex w-full justify-end">
-        <button
-          class="relative flex w-28 items-center justify-end rounded sm:justify-between"
+        <Button
```
laurakwhit, line 39: Do we want `Button` styling for any of the filters (e.g. `Date & Time`) in the header too?

Screenshot 2023-08-29 at 4 42 39 PM
Screenshot 2023-08-29 at 4 42 49 PM

src/lib/holocene/button.svelte
```diff
@@ -1,222 +1,139 @@
...
-  type ButtonVariant =
-    | 'primary'
-    | 'secondary'
-    | 'search'
-    | 'destructive'
-    | 'login'
-    | 'ghost'
-    | 'link'
-    | 'menu';
+<script lang="ts">
```
laurakwhit, line 1: Also, in no particular order r.e. some other button related components...

Is this the state we want for this icon button? Hover bleeds over header a bit (and it's also on a black background already).
Screenshot 2023-08-29 at 4 11 02 PM



For the copyable icon button,
Screenshot 2023-08-29 at 3 51 53 PM

wondering if we want it to be closer to what we have in the `CodeBlock` component?
Screenshot 2023-08-29 at 3 51 39 PM
And also, if we want to change the fill color of the icon on hover when its a link?
Screenshot 2023-08-29 at 3 57 54 PM

For the `Accordion` component, wondering if this is the style we want for the chevron icon (even when disabled)? Or do we want to make this more similar to the `Button` styling?
Screenshot 2023-08-29 at 4 15 29 PM

src/lib/components/schedule/schedule-form-view.svelte
Alex-Tideman: I'll do that on https://github.com/temporalio/ui/pull/1598

src/lib/components/banner/banner.svelte
```diff
@@ -24,9 +25,9 @@
     class={`relative block text-center leading-10 ${severity}`}
     {...$$restProps}
   >
-    <a href={link} target="_blank" rel="noreferrer">
+    <Link href={link} newTab>
       {message}
```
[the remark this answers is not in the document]
rossedfort: I think this is fine, it's actually an emoji in the string that we're passing in as `message`, so I don't know of an immediate way to do this besides exposing a second slot, which seems overkill.

src/lib/components/import/event-history-file-import.svelte
```diff
@@ -74,6 +74,6 @@
   accept=".json"
   on:change={onFileSelect}
 />
-<Button icon="file-upload" on:click={onConfirm} disabled={!fileLoaded}
```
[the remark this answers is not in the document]
rossedfort: I think we can probably just wait until we have a better FileInput component to address this.

src/lib/components/schedule/schedule-form-view.svelte
rossedfort: I think @Alex-Tideman is on this 💪

src/lib/holocene/combobox/combobox.svelte
rossedfort: Yep good call, fixed that in one of my latest commits.

src/lib/holocene/modal.svelte
rossedfort: Tbh I think the buttons look better here unmodified 🤷

|default|xs|
|---|---|
|Screenshot 2023-08-30 at 12 57 38 PM|Screenshot 2023-08-30 at 12 57 21 PM|

src/lib/holocene/scroll-to-top.svelte
rossedfort: I don't think regression was added in this PR, I'm currently unable to get the scroll-to-top button to show up in cloud. I'll make a new ticket to fix it 👍

src/lib/holocene/split-button.svelte
rossedfort: I'll ask Ash about this, I'm not sure what the expected behavior here is.

src/lib/holocene/button.svelte
rossedfort: Good call, for Icon button I changed the hover/focus state to our blue->purple gradient, so it should be consistent now.

src/lib/components/workflow/filter-search/boolean-filter.svelte
```diff
@@ -23,7 +23,7 @@
 </script>
 
 <MenuContainer>
-  <MenuButton id="boolean-filter" controls="boolean-filter-menu">
+  <MenuButton unroundLeft id="boolean-filter" controls="boolean-filter-menu">
```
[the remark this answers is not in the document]
rossedfort: I think we can make this change for now, but I would eventually like to solve for this in a less one-off fashion 👍

src/lib/holocene/table/table-row.svelte
```diff
@@ -2,26 +2,15 @@
   import type { HTMLAttributes } from 'svelte/elements';
 
   interface $$Props extends HTMLAttributes<HTMLTableRowElement> {
```
laurakwhit, line 4: Looks like something changed with the table row truncation

| Before | After |
|--|--|
|Screenshot 2023-08-31 at 9 20 44 AM|Screenshot 2023-08-31 at 9 22 33 AM|

colors.cjs
```diff
@@ -21,7 +21,7 @@ const colors = {
   },
   indigo: {
     50: '#eef2ff',
-    100: '#e0e7ff',
+    100: '#E0EAFF',
```
stevekinney, line 24: I am not going to block on this, but the inconsistency between capital and lower case makes me sad.

src/lib/components/search.svelte
```diff
@@ -11,27 +11,20 @@
...
-  <label for={id} class="hidden">{label}</label>
-  <input
+<form on:submit|preventDefault role="search" class="flex items-center">
+  <Input
+    class="w-full"
+    unroundRight
+    {label}
     {id}
+    labelHidden
+    icon={icon ? 'search' : null}
```
stevekinney, line 21: This ternary feels weird.

src/lib/components/workflow/filter-search/boolean-filter.svelte
```diff
@@ -23,7 +23,7 @@
 </script>
 
 <MenuContainer>
-  <MenuButton id="boolean-filter" controls="boolean-filter-menu">
+  <MenuButton unroundLeft class="!border-l-0" id="boolean-filter" controls="boolean-filter-menu">
```
stevekinney, line 26: What's going on with the `important`?

rossedfort: By default our inputs, buttons, menu buttons, etc. have borders and border radius, however there are instances where they are combined with other elements, and we want to remove that border and border-radius. I don't like the pattern of overriding default component styles by adding a css class, I would rather it be controlled by a prop so as to maintain visual consistency. That's exactly why I went with [this approach](https://github.com/temporalio/ui/pull/1589#discussion_r1311932741) for `<Button />`, I just haven't done it for `<MenuButton />`.

I'm not sure if important is necessary here though, we may just be able to pass the regular `border-l-0` tailwind class, @laurakwhit did you happen to test that out?

src/lib/components/banner/banner.svelte
```diff
@@ -24,9 +25,9 @@
     class={`relative block text-center leading-10 ${severity}`}
     {...$$restProps}
   >
-    <a href={link} target="_blank" rel="noreferrer">
+    <Link href={link} newTab>
```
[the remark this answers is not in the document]
rossedfort: Sure I agree, although `newTab` was already the name of the prop in `<Link />` so I didn't change it. Can make a follow up PR for that.

colors.cjs
rossedfort: Fair, I just copied the values from Figma, and figured we will be revisiting this file once Design finalizes our color palette.

src/lib/components/workflow/filter-search/list-filter.svelte
```diff
@@ -31,8 +31,8 @@
     placeholder="{translate('type-or-paste-in')} {$filter.attribute}"
   />
   <Button
-    variant="search"
-    unroundLeft
+    variant="secondary"
+    borderRadiusModifier="square-left"
```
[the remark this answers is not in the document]
rossedfort: I think a prop is better so that as much of the css that affects how a component looks is in one place, i.e. the component itself. IMO adding a class to a component should be a last resort and only happen in extreme one-off cases, which we should probably try to avoid.

src/lib/components/search.svelte
rossedfort: `icon` is a boolean prop that controls whether the magnifying glass icon displays in the input. `icon={icon && 'search'}` won't do, because then `false` would be passed to `icon` on `<Input />` which is not valid. Could change `null` to `undefined` or `''` but other than that I'm not sure how to make this better.

src/lib/components/workflow/filter-search/boolean-filter.svelte
stevekinney: I think it would be better to use the `group` or `peer` classes in Tailwind.

src/lib/components/workflow/filter-search/list-filter.svelte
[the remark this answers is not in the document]
stevekinney: To the core problem, _why_ would we need to manually pass this prop _or_ class?

src/lib/components/workflow/filter-search/boolean-filter.svelte
stevekinney: For example, it looks like we _know_ this is the first of its kind the group. Our styling could have affordances for this _without_ us needed to specify it.

laurakwhit: > I'm not sure if important is necessary here though, we may just be able to pass the regular border-l-0 tailwind class, @laurakwhit did you happen to test that out?

I did not. I just pulled that from somewhere similar.. Maybe `ConditonalMenu`?

src/lib/components/workflow/filter-search/list-filter.svelte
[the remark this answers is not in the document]
rossedfort: Because some of our interfaces make use of core components, like dropdown menu, button, input, etc. but need to be tweaked slightly for display with other components. I've yet to come up with a better way to solve this apart from making whole new sets of components that have square and/or borderless left or right edges, but it's on my back burner for sure.


https://github.com/temporalio/ui/assets/11775628/b5b55d9d-8f43-4018-9d79-067a3c8014dd


https://github.com/temporalio/ui/assets/11775628/f16b5985-8b9b-47db-8187-f2132e9e50c7

PR #1595, opened 2023-08-28
ARIA improvements to errors and alerts

laurakwhit:
## Description & motivation 💭 
* Ensures errors are announced with `role="alert"`
* Sets `display: none;` for `Alert` components whenever possible so that they are loaded in the DOM and screen readers are aware of them and can announce changes
* Sets `role="status"` for `Alert` components that are not urgent
* Updates Codec settings to utilize Holocene components with built-in accessibility

| Before | After |
|-- |--|
|Screenshot 2023-08-28 at 3 49 29 PM|Screenshot 2023-08-28 at 3 58 37 PM|

### How was this tested 👻 

- [x] Manual testing
- [x] E2E tests ~added~ updated
- [ ] Unit tests added

### Issue(s) closed 
`DT-1289`


src/lib/components/data-converter-port-settings.svelte
```diff
@@ -1,22 +1,28 @@
 <script lang="ts">
+  import Input from '$lib/holocene/input/input.svelte';
   import { translate } from '$lib/i18n/translate';
   export let port = '';
 </script>
 
 <div class="my-4">
-  <h3 class="text-lg" data-testid="data-encoder-port-title">
+  <label
```
rossedfort, line 8: Can we use `<Input />`'s `label` prop here?

rossedfort (changes requested):

src/lib/components/codec-endpoint-settings.svelte
```diff
@@ -10,26 +11,26 @@
...
+    </div>
+    <Textarea
+      id="data-encoder-endpoint-input"
       rows={3}
       placeholder={translate('data-encoder', 'endpoint-placeholder')}
-      data-testid="data-encoder-endpoint-input"
       bind:value={endpoint}
-      on:keydown|stopPropagation
+      {error}
+      isValid={!error}
```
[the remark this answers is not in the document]
laurakwhit: Ha you read my mind. Was about to ask you if you're cool with me updating the `Textarea` component 😆

src/lib/components/schedule/schedule-form-view.svelte
```diff
@@ -123,11 +122,9 @@
       <h1 class="font-base mt-8 ml-0 text-2xl">{title}</h1>
     </header>
     <form class="mb-4 flex w-full flex-col gap-4 md:w-2/3 xl:w-1/2">
-      {#if $error}
-        <Alert intent="error" title="" bold>
-          {$error}
-        </Alert>
-      {/if}
+      <Alert intent="error" title="" class={$error ? '' : 'hidden'} bold>
```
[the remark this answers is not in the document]
laurakwhit: So I went back and forth about this and experimented with adding a `hidden` prop. Definitely something we can add, but I wasn't sure if it made the most sense since it's technically something that can (and perhaps should) be controlled by it's parent or wherever it's being used.

Happy to have a further discussion on this!

src/lib/holocene/alert.svelte
```diff
@@ -23,7 +23,7 @@
 </script>
 
 <div
-  class="alert {intent} {className}"
+  class="flex alert {intent} {className}"
   class:bold
   role="alert"
```
rossedfort, line 28: I wonder if we should just have this be determined by the component. I.e. `intent="error"` is `role="alert"` and all others are `role="status"` ?

rossedfort: We will undoubtedly add a new info alert and forget to pass `role="status"` in the future

src/lib/holocene/accordion.svelte
```diff
@@ -80,9 +80,9 @@
       {/if}
     </div>
     <p class="flex items-center font-secondary">
-      {#if error}
-        <Badge class="mr-2" type="error">{error}</Badge>
-      {/if}
+      <Badge class="mr-2 {error ? '' : 'hidden'}" type="error" role="alert">
+        {error}
+      </Badge>
```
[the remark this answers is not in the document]
laurakwhit: Yeah perhaps a discussion to have with the design team. I think the `Badge` works best if we want something inline

Screenshot 2023-08-29 at 11 16 36 AM

but if we have a long error it might not be great. I'll try some things out and see if we can get some feedback.

src/lib/components/schedule/schedule-form-view.svelte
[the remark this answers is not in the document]
rossedfort: If it's a prop, the parent can still control it, right?

```
<Alert intent="error" hidden={!$error} />
```

I just don't want to have to think about which specific css class to add to ensure proper accessibility of the component.

src/lib/components/data-converter-port-settings.svelte
rossedfort: This would be my preference 😂 

```svelte
  <Input
    label={translate('data-encoder', 'port-title')}
    placeholder={translate('data-encoder', 'port')}
    id="data-encoder-port-input"
    bind:value={port}
    hintText={translate('data-encoder', 'port-info')}
  />
```

Screen Shot 2023-08-29 at 11 32 41 AM

src/lib/holocene/alert.svelte
rossedfort: That's fair. IMO, being in a box with these kinds of things wouldn't be the worst. Although another way we could do it is to just make the `role` prop required.

laurakwhit: What do you think about defaulting to `status` instead? That seems to be a pattern I see in other design systems.

src/lib/components/schedule/schedule-form-view.svelte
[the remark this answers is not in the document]
laurakwhit: Yeah that's exactly how I had the prop, but something felt not right about it. I'm going to change it back and see if I still have any opinions worth investigating.

src/lib/holocene/accordion.svelte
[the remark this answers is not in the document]
laurakwhit: I'm actually going to revert these changes and leave off a `role` here. This `Badge` is currently being used in instances where the error is there when the component loads so it's not even content that's updating.

Might want to revisit how we display errors on the `Accordion` component in the future though.

src/lib/holocene/alert.svelte
GiantRobots: I think we should jump in the box and infer the role based on the intent. The second we need to break out of it we can update the component to accept a `role` prop to override the inferred ones. I think we'll have generally more accessible components with that calculated and it will automagically add accessibility in areas we already use those components.

I would say it should be like...
```
typescript
'warning' | 'caution' | 'error'  - alert
'success' - status
'info' - no role
```

laurakwhit: Hmm I guess we could go in the box, but still allow it to be overridden? 

There are examples in Cloud (e.g.audit logging) where the error is loaded with the rest of the page content and isn't something that is being updated based on actions or changing data, etc. In those cases we want the `error` `Alert` variant, but it shouldn't be `role="alert"`.

rossedfort: Ahh ok, that's good to know. Sounds good to me 👍

laurakwhit: Updated [here](https://github.com/temporalio/ui/pull/1595/commits/706f9fd64c253d61653f1a3893179a60e9e866bd) ✅ with the following based on our use cases

```
'error' - alert
'success' | 'warning' | 'caution' - status
'info' - no role
```

src/lib/components/schedule/schedule-form-view.svelte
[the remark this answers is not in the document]
laurakwhit: Updated [here](https://github.com/temporalio/ui/pull/1595/commits/005b9f93140f3e975ab71cb478a4eb82f17250b9) ✅

src/lib/holocene/alert.svelte
```diff
@@ -11,21 +11,38 @@
...
+    if (intent === 'error') {
+      return 'alert';
+    }
+
+    if (intent === 'success' || intent === 'warning' || intent === 'caution') {
+      return 'status';
+    }
+
+    return null;
+  }
```
rossedfort, line 38: Will this work even though it's not reactive?

rossedfort (approved):
lgtm, just one question

src/lib/holocene/alert.svelte
laurakwhit: It does... But I think I'm going to refactor this a bit anyways 😅

PR #1599, opened 2023-08-29
Storybook

GiantRobots:
…apper to come

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/stories/Button.stories.ts
rossedfort, line 1: I'm wondering if we want to throw stories in holocene, i.e. `src/lib/holocene/stories` ?

src/stories/ReadMe.mdx
rossedfort, line 1: What is this file, and do we need/want it? If so - can we expand on its contents?

GiantRobots: It's a top level `README` the plan is definitely to fill it out as I get more information on what good ideas are

src/stories/Button.stories.ts
GiantRobots: I think that's a good next step.

 I'm unfamiliar with how storybook is configured these days and figure we can do this temporarily and then move it to holocene/stories. Also probably good to keep them split for a time until we do a proper effort to migrate (and verify the utility of this)

src/lib/holocene/stories/Button.stories.ts
```diff
@@ -0,0 +1,42 @@
...
+  component: ButtonWrapper,
+  argTypes: {
+    leadingIcon: { ...iconControls },
+    trailingIcon: { ...iconControls },
+    variant: {
+      control: {
+        type: 'select',
+      },
+      options: ['primary', 'secondary', 'destructive', 'ghost'],
+    },
```
rossedfort, line 26: Maybe a future project, try to type this such that we can only have controls with valid prop names/values.

rossedfort (approved):
🎨

src/lib/holocene/stories/Button.stories.ts
GiantRobots: 100% They kind of were before but not sure why this wasn't typed correctly. I think it ends up being a weird generic thing

PR #1606, opened 2023-08-31
Remove Parent Workflow ID column

laurakwhit:
## Description & motivation 💭 

This PR remove the `Parent Workflow ID` column, as it is currently always an empty value since
> The `ListWorkflowExecutions` api returns the `parentExecution` field, but it is never set at the server level so it’s always `null`. 
### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests ~added~ updated

### Issue(s) closed 
`DT-1396`


stevekinney (approved):

src/lib/stores/workflow-table-columns.ts
```diff
@@ -89,7 +87,10 @@ export const getDefaultColumns = (): WorkflowHeader[] => {
     const parsedOldColumns = JSON.parse(stringifiedOldColumns);
 
     if (stringifiedOldColumns && parsedOldColumns?.length) {
-      columns = parsedOldColumns;
+      const filteredOldColumns = parsedOldColumns.filter(
```
laurakwhit, line 90: > Make sure Parent Workflow ID column doesn't render if already a selected column

@Alex-Tideman I considered this, but then thought it might be best to leave as is rather than add this extra code. If they remove the column it won't reappear as an option to add, but if it's there it just shows nothing (which is existing behavior).

Alex-Tideman: So my problem with not having this code is that they will leave that column in the table and wait till it's fixed/shows the data. I'd rather proactively filter it out and we can add it back once it's supported.

laurakwhit: Ah ok, I think this makes more sense if we're going to add it back.

rossedfort: @laurakwhit I don't think these changes are actually doing anything, this function gets the old cached columns before we switched to storing them per-namespace. We'll need to add the filter on the `workflowTableColumns` derived store I think.

Side note, this store is becoming a bit convoluted and could probably use some refactoring. Now that it's been a couple months since we started storing columns per namespace, it might be safe to remove that functionality.

Alex-Tideman (approved):

src/lib/stores/workflow-table-columns.ts
laurakwhit: Ah good point, [updated](https://github.com/temporalio/ui/pull/1606/commits/818419c198439a7470049787c7709a3c7b09e878) ✅ 

And yeah, agree that we could probably do some refactoring here. Shall we create a follow-up task for this?

rossedfort: Yes, def can do this in a follow-up task 👍

PR #1607, opened 2023-08-31
Make version checks async, closes #1465

### LINKED ISSUE #1465: Remove Github Dependency for the UI

We are running version 2.14.0 of the Temporal UI.
The UI has a dependency on Github (a GET request to https://api.github.com/repos/temporalio/ui-server/releases), so when Github is down the UI does not load until the request to the Github API times out, causing the page to take 100+ seconds to load.

Remove this Github API request if it is not used for anything in the UI. Alternatively make this request asynchronous so it does not block the page from loading.

## Additional context

This is where the Github API request is happening: 
image

stevekinney:
Stops version checks from blocking the rendering of the page

### DIFF
--- a/src/lib/components/banner/banner-ui-version.svelte
+++ b/src/lib/components/banner/banner-ui-version.svelte
@@ -3,18 +3,19 @@
   import type { BannersState } from '$lib/models/banner-state';
   import type { UiVersionInfo } from '$lib/types/global';
   import { isVersionNewer } from '$lib/utilities/version-check';
-  
+
   import Banner from './banner.svelte';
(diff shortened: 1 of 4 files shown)


rossedfort:
It's not apparent to me how this solves the issue, I think we need to make this function not await the result of the GitHub releases API so that it doesn't block.

Also, something regressed in the logic that displays the banner
Screenshot 2023-08-31 at 9 16 48 AM

rossedfort:
Oh I see, you removed the call to that function, where do we get the recommended (latest) UI version from now?

PR #1610, opened 2023-08-31
focus and select text in combobox when list is opened

rossedfort:
## Description & motivation 💭 
To better indicate that the text in the combobox can be edited, and to allow the text to be deleted in a single keystroke, whenever the list of options is opened (either by clicking the chevron button, or focusing the input), the text in the input will be selected.

https://github.com/temporalio/ui/assets/11775628/13e8fe49-7890-44f6-a7b5-407d64090958

### Design Considerations 🎨 
Approval by @canvanooo 
### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/holocene/combobox/combobox.svelte
+++ b/src/lib/holocene/combobox/combobox.svelte
@@ -113,12 +113,13 @@
       closeList();
     } else {
       openList();
-      inputElement.focus();
     }
   };
 
   const openList = () => {
     $open = true;
+    inputElement.focus();
+    inputElement.select();
   };
 
   const closeList = () => {
@@ -288,7 +289,7 @@
       on:click={toggleList}
     >
       <Icon name={$open ? 'chevron-up' : 'chevron-down'} />
-  </button>
+    </button>
   </div>
 
   <Menu bind:menuElement id="{id}-listbox" role="listbox" class="w-full">


Alex-Tideman (approved):

src/lib/holocene/combobox/combobox.svelte
```diff
@@ -113,12 +113,13 @@
       closeList();
     } else {
       openList();
-      inputElement.focus();
     }
   };
 
   const openList = () => {
     $open = true;
+    inputElement.focus();
```
stevekinney, line 121: QQ: Would `autofocus` work here? Alternatively? I wonder if a Svelte directive that was reusable would be even better?

```ts
import type {Action} from 'svelte/actions';

export const focus: Action = (node: HTMLInputElement, shouldFocus: boolean | undefined) => {
  // Focus on mount if no value is passed or we are told to focus.
  if (shouldFocus === undefined || shouldFocus) {
    node.focus();
    node.select();
  }

  return {
    update: (shouldFocus: boolean | undefined) => {
      if (shouldFocus) {
        node.focus();
        node.select();
      }
    },
    destroy: () => {},
  }
}
```

stevekinney: And then you could just do `use:focus` or `use:focus={$open}`

rossedfort: I don't think autofocus would work, because I think it would focus the input on mount, whereas we only want to focus the input when the button to open the options is clicked. And In a component which such a specific interaction, I don't know that there's any real value in creating/using a more generic directive here. If we find opportunity for re-use later on we can certainly create one.

PR #1611, opened 2023-08-31
Add link to skip to main content 

laurakwhit:
## Description & motivation 💭 
Adds a "Skip to main content" link that is accessible via keyboard.

Screenshot 2023-08-31 at 11 55 15 AM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

- [ ] Verify the "Skip to main content" link places focus in main content
- [ ] Verify the "Skip to main content" link is only visible when focused

### Issue(s) closed 
`DT-1228`

### DIFF
--- a/src/lib/components/skip-nav.svelte
+++ b/src/lib/components/skip-nav.svelte
@@ -1,0 +1,12 @@
+<script lang="ts">
+  import Link from '$lib/holocene/link.svelte';
+  import { translate } from '$lib/i18n/translate';
+</script>
+
+<Link
+  id="skip-nav"
+  href="#content"
+  class="absolute left-1/2 translate-y-[-100%] focus:translate-y-0 z-[100] p-1"
+>
+  {translate('skip-nav')}
+</Link>
--- a/src/lib/components/skip-nav.svelte
+++ b/src/lib/components/skip-nav.svelte
@@ -6,7 +6,7 @@
 <Link
   id="skip-nav"
   href="#content"
-  class="absolute left-1/2 translate-y-[-100%] focus:translate-y-0 z-[100] p-1"
+  class="absolute left-1/2 -translate-y-full focus:translate-y-0 z-100 p-1"
 >
   {translate('skip-nav')}
 </Link>
--- a/src/routes/(app)/+layout.svelte
+++ b/src/routes/(app)/+layout.svelte
@@ -8,6 +8,7 @@
   import Banners from '$lib/components/banner/banners.svelte';
   import DataEncoderSettings from '$lib/components/data-encoder-settings.svelte';
   import SideNavigation from '$lib/components/side-nav.svelte';
+  import SkipNav from '$lib/components/skip-nav.svelte';
   import TopNavigation from '$lib/components/top-nav.svelte';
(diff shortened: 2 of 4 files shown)


Alex-Tideman:
How do you view this? What's the keyboard shortcut?

Alex-Tideman:
Nvm I see it when I tab.

laurakwhit:
@Alex-Tideman it should be the first thing you tab to when you load the page.

Alex-Tideman:
Looks like this doesn't work in Safari or FF. I don't see it when I tab.

src/lib/components/skip-nav.svelte
```diff
@@ -0,0 +1,12 @@
+<script lang="ts">
+  import Link from '$lib/holocene/link.svelte';
+  import { translate } from '$lib/i18n/translate';
+</script>
+
+<Link
+  id="skip-nav"
+  href="#content"
+  class="absolute left-1/2 translate-y-[-100%] focus:translate-y-0 z-[100] p-1"
```
stevekinney, line 9: Why not `translate-y-full`?

Are we okay with the bespoke z-index?

src/routes/(app)/+layout.svelte
```diff
@@ -8,6 +8,7 @@
   import Banners from '$lib/components/banner/banners.svelte';
   import DataEncoderSettings from '$lib/components/data-encoder-settings.svelte';
   import SideNavigation from '$lib/components/side-nav.svelte';
+  import SkipNav from '$lib/components/skip-nav.svelte';
```
stevekinney, line 11: Can I be pendantic and ask that we call this `SkipNavigation`?

laurakwhit:
@Alex-Tideman have you turned on `Full Keyboard Access` in `System Settings` > `Accessibility` > `Keyboard`?

src/lib/components/skip-nav.svelte
laurakwhit: > Why not translate-y-full?

Needs to be a negative (we want it off the page).

> Are we okay with the bespoke z-index?
We can update this to `z-50`, but then we'll need to update `NavContainer` to something else (and potential a few other things). I was 50/50 and so went with bespoke.

src/routes/(app)/+layout.svelte
laurakwhit: [Done](https://github.com/temporalio/ui/pull/1611/commits/741522d878311c87d8d05dbdf699407f1f6e15db) ✅

src/lib/components/skip-nav.svelte
stevekinney: Oh, I can't read. Why not `-translate-y-full`?

stevekinney: Yea, I think maybe we should just update to standardized z-indices.

laurakwhit: > Why not -translate-y-full?

TIL

> update to standardized z-indices.

Opinions on what these might be? Or perhaps I should just add 100 to the config for now...

Alex-Tideman (approved):

PR #1613, opened 2023-09-01
UI updates to WorkflowFilterSearch

laurakwhit:
|  |  |
|--|--|
|Screenshot 2023-09-06 at 3 55 52 PM|Screenshot 2023-09-06 at 3 56 12 PM|

| Before | After |
|--|--|
|Screenshot 2023-09-01 at 3 43 34 PM|Screenshot 2023-09-01 at 3 44 02 PM|
|Screenshot 2023-09-01 at 3 44 42 PM|Screenshot 2023-09-01 at 3 45 02 PM|
|Screenshot 2023-09-01 at 3 45 39 PM|Screenshot 2023-09-01 at 3 46 02 PM|
|Screenshot 2023-09-01 at 3 46 33 PM|Screenshot 2023-09-01 at 3 46 49 PM|

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

- [ ] Verify all types of filters have the close button and it works/looks as expected
- [ ] Verify Relative and Absolute Datetime filters work as expected
- [ ] Verify Start/End time Datetime filter works as expected
### Issue(s) closed 
`DT-1359`


src/lib/components/workflow/filter-search/close-filter-button.svelte
```diff
@@ -0,0 +1,19 @@
...
+
+  const { resetFilter } = getContext<FilterContext>(FILTER_CONTEXT);
+</script>
+
+<Button
+  aria-label={translate('close')}
+  borderRadiusModifier="square-left"
+  leadingIcon="chevron-left"
+  on:click={resetFilter}
+  style="padding: 0"
```
laurakwhit, line 17: Wondering how we feel about overriding css like this? Or do we want to add something to support this in the `Button` component?

src/lib/components/workflow/filter-search/datetime-filter.svelte
```diff
@@ -239,18 +152,117 @@
...
               >
+                {#each TIME_UNIT_OPTIONS as unit}
+                  <Option value={unit}>{unit} {translate('ago')}</Option>
+                {/each}
+              </Select>
+            </div>
+          </div>
+        </MenuItem>
+        <MenuDivider />
+        <MenuItem on:click={() => (type = 'absolute')}>
```
laurakwhit, line 223: We can't make `MenuItem`s full width because of the space reserved for the checkmark

Screenshot 2023-09-01 at 4 03 53 PM


Do we want to update the conditional [here](https://github.com/temporalio/ui/blob/main/src/lib/holocene/menu/menu-item.svelte#L143-L147) to include the container `<div/>`? If we do, selecting something might alter the height of the option selected (e.g. if the text wraps).

rossedfort: Yeah hmm. Maybe we need a different variant of MenuItem that isn't intended to be selected.

src/lib/components/workflow/filter-search/datetime-filter.svelte
```diff
@@ -239,18 +152,117 @@
...
+              <input
+                on:click|stopPropagation={() => (type = 'absolute')}
+                class="w-4 h-4 accent-gray-900"
+                type="radio"
+                checked={type === 'absolute'}
+                id="absolute-time"
+                tabindex="-1"
+              />
+              {translate('absolute')}
+            </label>
```
rossedfort, line 238: I think it would be nice to extend `<Input />` to support `type="radio"`

src/lib/components/workflow/filter-search/close-filter-button.svelte
rossedfort: Oh I see this is for that "collapse" button we were talking about the other day. Personally I could do without it, or maybe just an `<IconButton />` with an `X` icon next to the filters.

rossedfort:
Couple of thoughts

1. Were you going to fix the issue where clicking away from the Workflow Status filter doesn't reset it in the dropdown?
2. Should we also use this PR to remove the labs mode check from the new filters?

src/lib/components/workflow/filter-search/datetime-filter.svelte
laurakwhit: [Added a check](https://github.com/temporalio/ui/pull/1613/commits/9d686f48e41bd5b1a40d2b2a543a0e63b7a54206) for if `selected` is being passed as a prop or not ✅

laurakwhit:
> Were you going to fix the issue where clicking away from the Workflow Status filter doesn't reset it in the dropdown?

@rossedfort I have another task for this (`DT-1360`), but we also may be removing this filter entirely with the addition of Workflow counts.

> Should we also use this PR to remove the labs mode check from the new filters?

I'll create a separate task for this and get it approved by product.

> Looks like there is some slightly inconsistent behavior here. If the user has selected "relative" but not specified a duration, should we disable the "Apply" button? 

I've disabled `Apply` if relative is selected and there is an error or no specified duration (https://github.com/temporalio/ui/pull/1613/commits/b0067201716becadc595e2ac9244932d8979e466).

> Similarly, should we disable it if they select "absolute" and clear the date input?

The existing behavior for the datetime filter is just to use the current date if there is none specified. I think it would be fine to keep that and allow apply even if the date input is empty.

src/lib/components/workflow/filter-search/close-filter-button.svelte
laurakwhit: So the default `<IconButton />` looks a little wonky here

Screenshot 2023-09-06 at 12 39 28 PM

Let's discuss this a bit more and maybe get design's input.

src/lib/components/workflow/filter-search/datetime-filter.svelte
```diff
@@ -239,18 +152,117 @@
...
+                  type = 'relative';
+                }}
+                class="w-4 h-4 accent-gray-900"
+                type="radio"
+                checked={type === 'relative'}
+                id="relative-time"
+                tabindex="-1"
+              />
+              {translate('relative')}
+            </label>
```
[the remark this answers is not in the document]
laurakwhit: I did look into this a bit, but I think it might best if we addressed refactoring these in a separate PR. We have things like the `<div/>` (aka `input-container`), etc. that complicate the styling here.

src/lib/components/workflow/filter-search/datetime-filter.svelte
laurakwhit: See https://github.com/temporalio/ui/pull/1613#discussion_r1317798158.

src/lib/components/workflow/filter-search/close-filter-button.svelte
laurakwhit: [Updated](https://github.com/temporalio/ui/pull/1613/commits/2e3b17f83eba67067ed0dcc924685f8a1c8522c9) to use `<IconButton />`  ✅

rossedfort (approved):
LGTM

Once future enhancement I thought of is disabling the input(s) in the "Relative" option when user has selected "Absolute" and vice versa. It's a bit odd than you can interact with both even when you only have 1 option selected
Screenshot 2023-09-07 at 10 34 50 AM

src/lib/components/workflow/filter-search/datetime-filter.svelte
```diff
@@ -207,13 +207,15 @@
                 error={error(relativeTime)}
                 unroundRight
                 class="h-10"
+                disabled={type !== 'relative'}
```
laurakwhit, line 210: > Once future enhancement I thought of is disabling the input(s) in the "Relative" option when user has selected "Absolute" and vice versa. It's a bit odd than you can interact with both even when you only have 1 option selected

@rossedfort added ✅ 

|||
|--|--|
|Screenshot 2023-09-07 at 10 50 11 AM|Screenshot 2023-09-07 at 10 50 19 AM|

PR #1618, opened 2023-09-06
DT-1255 - toggle button a11y

rossedfort:
## Description & motivation 💭 
- make toggle button either a button or anchor element (previously was a noninteractive div)
- add some hover and focus styles for visual consistency with other interactive elements

https://github.com/temporalio/ui/assets/11775628/02369838-1cfe-431b-ab2a-2278e5ef2160

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/holocene/toggle-button/toggle-button.svelte
+++ b/src/lib/holocene/toggle-button/toggle-button.svelte
@@ -1,10 +1,38 @@
 <script lang="ts">
-  import { page } from '$app/stores';
-  
-  import Icon from '$lib/holocene/icon/icon.svelte';
-import type { IconName } from '$lib/holocene/icon/paths';
-import { getAppContext } from '$lib/utilities/get-context';
+  import type {
+    HTMLAnchorAttributes,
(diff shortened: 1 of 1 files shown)


src/lib/holocene/toggle-button/toggle-button.svelte
```diff
@@ -1,61 +1,74 @@
...
+  href={href ? href + $page.url.search : null}
+  class:rounded-lg={!group}
+  on:click
+  role="button"
+  tabindex="0"
+  {...$$restProps}
+>
+  {#if icon}
+    <div class="flex items-center gap-2">
+      <Icon name={icon} />
```
laurakwhit, line 57: One thing I've noticed with our `Button` like components is that if they have icons **and** text the screen reader reads both. Maybe we consider hiding icons from screen readers that are nested within interactive elements like buttons/toggles?

Screenshot 2023-09-07 at 2 22 40 PM

laurakwhit (approved):

PR #1642, opened 2023-09-20
Fix button so that it can't accept an `href` prop if disabled

rossedfort:
## Description & motivation 💭 
We allow `<Button />` to accept an `href` prop, which will render an Anchor element that is styled to look like a button. Anchor elements cannot be disabled, so this PR fixes `<Button />` to make the `disabled` prop type `never` when `href` is specified.

Also fixes the schedules list page, which is the only place in OSS we were were disabling link buttons.

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/holocene/button.story.svelte
+++ b/src/lib/holocene/button.story.svelte
@@ -18,21 +18,41 @@
   export let Hst: HST;
 </script>
 
-<Hst.Story>
-  <div class="flex items-center justify-center p-4">
-    <Button
-      {variant}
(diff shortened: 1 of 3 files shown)


laurakwhit:
We should look into the `<MenuItem/>` and `<Tab/>` components as well since those can be links and likely also allow `disabled`.

laurakwhit (approved):

PR #1651, opened 2023-09-26
PayloadDecoder

Alex-Tideman:
## Description & motivation 💭 
Use the PayloadDecoder component to decode payloads and pass to the CodeBlock instead of decoding everything upfront.

The PayloadDecoder accepts `PotentiallyDecodable | EventAttribute | WorkflowEvent` as a value. It will iterate over the value and decode each payload. You can also pass a key prop to return a value of the decoded object. This is mostly to not show { payloads: "thing" } and instead show "thing".

There is no caching of decoded values. So when a PayloadDecoder is rerendered, it will send another decode request. It is up to the codec server implementation to handle caching.

⚠️ THIS PR REMOVES SUPPORT FOR WEBSOCKET PORT DECODING.

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added


Alex-Tideman:
Need to figure out the mechanism to tell the PayloadDecoder to not decode something already decoded.

rossedfort (approved):

src/lib/components/workflow/workflow-json-navigator.svelte
```diff
@@ -78,12 +82,27 @@
...
     <slot name="decode" />
   </div>
-  <CodeBlock
-    content={stringifyWithBigInt(events[index - 1])}
-    testId="event-history-json"
-    copyIconTitle={translate('copy-icon-title')}
-    copySuccessIconTitle={translate('copy-success-icon-title')}
-  />
+  {#if decodeEventHistory}
+    {#key [index, decodeEventHistory]}
```
rossedfort, line 86: TIL you can do array keys. is the `decodeEventHistory` necessary so that we re-render encoded/decoded payloads when the switch is toggled on and off?

Alex-Tideman: Yup exactly

PR #1658, opened 2023-10-02
Timeline Improvements

Alex-Tideman:
## Description & motivation 💭 
Improvement to the event history timeline:

- Take out of Labs Mode and add to an Accordion (default open - remembers preference of open/close)
- ToggleButtons for Zoom In/Out/Fit
- Limit the min and max of the timeline to the start of the minute of the workflow execution start time and the end of the minute of the workflow execution end time. This prevents panning way out of the timescale of the timeline.
- Add tooltip with Start/End Time and duration
- Better labels for group names: Name (specific event name). Label: 'Signal received' vs 'Signal'. DisplayName: Combine Label and Name.
- Collapse rows when event isn't visible in the current time frame.
- Replace 'Recent Events' with 'Event History'

Screen Shot 2023-10-02 at 12 16 14 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/components/event/event-history-timeline-helpers.ts
```diff
@@ -0,0 +1,64 @@
...
+
+const groupEventTooltipTemplate = (item): string => {
+  return `<div class="flex flex-col gap-1">
+    <div class="flex gap-2"><div class="font-bold w-16">Start</div>
+      <div>${formatDate(item.start, get(timeFormat))}
+      </div>
+    </div>
+  <div class="flex gap-2">
+    <div class="font-bold w-16">End</div>
+    <div>${formatDate(item.end, get(timeFormat))}</div>
```
laurakwhit, line 53: Do we want to support `relative` time for the start and end times in the tooltips? e.g.

```suggestion
    <div>${formatDate(item.end, get(timeFormat), { relative: get(relativeTime) )}</div>
```

src/lib/components/event/event-history-timeline.svelte
```diff
@@ -112,88 +133,139 @@
...
+        </ToggleButtons>
+      </div>
     </div>
+    <div class="timeline" bind:this={visualizationRef} />
   </div>
-  <div bind:this={visualizationRef} />
-</div>
+</Accordion>
 
 <style lang="postcss">
```
laurakwhit, line 270: Noticed that on some `Completed` workflows you can zoom out and see a red moving line that correlates with the current time, which is a bit confusing since the workflow is `Completed` and there should be no more events 🤔 

Screenshot 2023-10-02 at 2 54 38 PM

src/lib/models/event-groups/get-group-name.ts
```diff
@@ -19,33 +19,80 @@ export const getEventGroupName = (event: CommonHistoryEvent): string => {
...
 
   if (isWorkflowExecutionUpdateAcceptedEvent(event)) {
-    return `${event.workflowExecutionUpdateAcceptedEventAttributes?.acceptedRequest?.input?.name}`;
+    return event.workflowExecutionUpdateAcceptedEventAttributes?.acceptedRequest
+      ?.input?.name;
   }
 };
+
+export const getEventGroupLabel = (event: CommonHistoryEvent): string => {
+  if (!event) return;
```
laurakwhit, line 57: Looks like maybe we want to return an empty string here instead of `undefined` based on the type defined for the return?. Also (nit) it looks like this uses `` instead of `''` in a lot of places (maybe something to have auto-fix on format). 

```suggestion
  if (!event) return '';
```

laurakwhit (commented):
This just keeps getting better and better 🎉

src/lib/components/event/event-history-timeline.svelte
rossedfort, line 1: Not sure the limitations of this `vis-timeline` library, but is there any way we svelte-ify this Component a little more? I don't mind the approach you took for `renderComponentToHTML` can we make the rest of those html strings their own components and use that to render them?

rossedfort: Also this whole file could use some types

src/lib/components/event/event-history-timeline-helpers.ts
rossedfort, line 1: Is there any other way we can do this besides writing html as strings?

Alex-Tideman: Probably my favorite screenshot. Tried doing this by importing a Svelte component... got this message in the tooltip. Gonna keep trying.

Screen Shot 2023-10-03 at 9 42 59 AM

src/lib/components/event/event-history-timeline.svelte
Alex-Tideman: 👍 I'll try, the library is a little tough to use without strings. I'll add types.

src/lib/models/event-groups/get-group-name.ts
Alex-Tideman: 👍

src/lib/components/event/event-history-timeline-helpers.ts
Alex-Tideman: 👍

src/lib/components/event/event-history-timeline.svelte
Alex-Tideman: So this happens because I've set the max time to be the end of the minute of the minute that the workflow execution ends. So you'd only see that red moving line for the <60 seconds after it completes. I do this so you can zoom out a little to see any events that happen right at the end of the workflow.

laurakwhit: Ah, got it. Wanted to make sure I wasn't missing something 😅

src/lib/components/event/event-history-timeline-helpers.ts
Alex-Tideman: Just with how Svelte works, I don't think we can componentize the templates. You could with React/jsx but not svelte

laurakwhit (approved):

PR #1659, opened 2023-10-02
Dt 1142 deep linking activities

GraceGardner:
### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/components/event/event-group-details.svelte
```diff
@@ -7,10 +10,27 @@
...
   } from '$lib/models/event-groups/get-event-in-group';
   import { isLocalActivityMarkerEvent } from '$lib/utilities/is-event-type';
+  import { routeForEventHistory } from '$lib/utilities/route-for';
 
   export let eventGroup: EventGroup;
   export let selectedId: string;
   export let onGroupClick: (id: string) => void;
+
+  const handleGroupClick = (id: string) => {
+    const { namespace, workflow, run } = $page.params;
```
stevekinney, line 20: We can probably just store these variables in the top-level instead of fetching them every time.

src/lib/components/event/event-group-details.svelte
```diff
@@ -7,10 +10,27 @@
...
+    const queryParams: Record<string, string> = {};
+    $page.url.searchParams.forEach((value, key) => {
```
stevekinney, line 22: Why not just use a `reduce`? This can probably be broken out into a utility function and unit tested. Now that I think about it, that's probably true for this entire function.

src/lib/components/event/event-history-timeline.svelte
```diff
@@ -150,7 +152,14 @@
   ): void => {
     timeline.setGroups(groups);
     timeline.setItems(items);
-    timeline.fit();
+
+    if ($page.url.hash) {
+      setTimeout(() => {
```
stevekinney, line 157: Why are we using a `setTimeout` here?

src/lib/components/event/event-summary-row.svelte
```diff
@@ -61,6 +65,19 @@
 
   const onLinkClick = () => {
     expanded = !expanded;
+    // Dont delete query params like ?per-page
+    const { namespace, workflow, run } = $page.params;
+    const queryParams: Record<string, string> = {};
```
stevekinney, line 70: Yea, if we're doing this more than once, then we should definitely break it out.

src/lib/components/event/event-summary.svelte
```diff
@@ -54,6 +74,9 @@
       sort: compact ? 'ascending' : sort,
     });
     loading = false;
+    setTimeout(() => {
```
stevekinney, line 77: The `setTimeout` here feels a little bit like a code smell.

src/lib/components/event/event-summary.svelte
```diff
@@ -31,12 +34,29 @@
...
   const resetFullHistory = () => {
     $fullEventHistory = [];
     loading = true;
   };
 
+  onMount(() => {
+    hash = $page.url.hash;
+  });
+
+  function scrollIntoView(target, eventHistory) {
```
stevekinney, line 48: This should be broken out.

PR #1670, opened 2023-10-05
DT-1407 - add RadioInput component and replace existing native radio inputs

rossedfort:
## Description & motivation 💭 
Adds `RadioInput` and `RadioGroup` components following the new component spec and replaces existing native `<input type="radio" />` with the new Component(s)

Some implementation notes:
`<RadioInput />` can be used with or without a `<RadioGroup />`. To use it standalone, a `name` and `group` prop must be passed to each `<RadioInput />`

```svelte
<script>
  const group = writable<'a' | 'b' | 'c'>('a');
</script>

<RadioInput id="option-a" value="a" label="Option A" name="options" {group} />
<RadioInput id="option-b" value="b" label="Option B" name="options" {group} />
<RadioInput id="option-c" value="c" label="Option C" name="options" {group} />
```

to use with a `<RadioGroup />`, `name` and `group` can be passed as props to `<RadioGroup />` and will be set on child `<RadioInput />`'s via context.

```svelte
<script>
  const group = writable<'a' | 'b' | 'c'>('a');
</script>

<RadioGroup name="options" {group}>
  <RadioInput id="option-a" value="a" label="Option A" />
  <RadioInput id="option-b" value="b" label="Option B" />
  <RadioInput id="option-c" value="c" label="Option C" />
</RadioGroup>
```
Note: binding to `group` is not necessary, as the value is a store. `group` had to be made a store, because context is not reactive by default. For reactivity in svelte context, a reactive value must be used.

Screenshot 2023-10-05 at 10 05 04 AM
Screenshot 2023-10-05 at 10 05 30 AM
Screenshot 2023-10-05 at 10 05 53 AM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

Radio inputs were used in 3 places
- DataEncoderSettings
- WorkflowResetForm
- DatetimeFilter

Test for no regressions in functionality in these 3 areas.


src/lib/components/data-encoder-settings.svelte
```diff
@@ -26,11 +26,14 @@
...
 <script lang="ts">
+  import RadioGroup from '$lib/holocene/radio-input/radio-group.svelte';
+  import RadioInput from '$lib/holocene/radio-input/radio-input.svelte';
+
   let endpoint = $codecEndpoint ?? '';
   let port = $dataConverterPort ?? '';
   let passToken = $passAccessToken ?? false;
   let includeCreds = $includeCredentials ?? false;
-  let override = $overrideRemoteCodecConfiguration ?? false;
+  let override = writable($overrideRemoteCodecConfiguration);
```
laurakwhit, line 36: Curious why you've chosen `writable` over ⬇️ here
```suggestion
  let override = $overrideRemoteCodecConfiguration ?? false;
```
and if we want to change the other variables to stores as well 🤔

laurakwhit (approved):
🎉

src/lib/components/data-encoder-settings.svelte
rossedfort: I did it this way because `override` needs to be a `Writable`, and `let override = $overrideRemoteCodecConfiguration` would make `override` the value of the store, i.e. a `boolean`. Additionally, `let override = overrideRemoteCodecConfiguration` (without the `$`) assign `override` by reference and thus the store value would still be mutated.

laurakwhit: Ah forgot about the context r.e. the group needing to be a store in the note for this PR, thanks!

PR #1671, opened 2023-10-05
[DT-1423] Send signal input payload to data encoder

laurakwhit:
## Description & motivation 💭 
When sending a Signal through the UI, the input payload is sent to the `/encode` endpoint if a codec endpoint is configured. If not, the payload is base64 encoded. Once encoded, the payload is sent as part of the api request. If there is an error when the payload is sent to `/encode`, there is an error message (`Data encoding failed`) and the signal request is not sent.

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

## Checklists
- [ ] Verify a signal input is base64 encoded if there is no codec endpoint configured
- [ ] Verify a signal input is encoded if there is a codec endpoint configured
- [ ] Verify there is an error message if the input can not be encoded
- [ ] Verify the name and input are cleared when the signal modal is closed

### Issue(s) closed 
`DT-1423`

### DIFF
--- a/src/lib/components/event/event-details-row.svelte
+++ b/src/lib/components/event/event-details-row.svelte
@@ -32,7 +32,7 @@
 >
   {#if typeof value === 'object'}
     <div
-      class="flex w-full flex-wrap items-center justify-between pr-1 xl:flex-nowrap xl:gap-4"
+      class="flex w-full flex-wrap items-center justify-between gap-1 pr-1 xl:flex-nowrap xl:gap-4"
     >
       <p class="min-w-fit text-sm">
         {format(key)}
--- a/src/lib/holocene/code-block.svelte
+++ b/src/lib/holocene/code-block.svelte
@@ -128,7 +128,7 @@
   });
 
   const setView = () => {
-    if (view && !editable) {
+    if (view && (!editable || (editable && !value))) {
       const newState = createEditorState(value);
       view.setState(newState);
     }
--- a/src/lib/services/workflow-service.ts
+++ b/src/lib/services/workflow-service.ts
@@ -1,13 +1,19 @@
 import { noop } from 'svelte/internal';
+import { get } from 'svelte/store';
 
 import { v4 } from 'uuid';
 
+import { translate } from '$lib/i18n/translate';
 import type { ResetReapplyType } from '$lib/models/workflow-actions';
 import {
   toWorkflowExecution,
   toWorkflowExecutions,
 } from '$lib/models/workflow-execution';
 import { convertPayloadsWithCodec } from '$lib/services/data-encoder';
+import {
+  lastDataEncoderError,
+  lastDataEncoderStatus,
+} from '$lib/stores/data-encoder-config';
 import type { ResetWorkflowRequest } from '$lib/types';
 import type {
   ValidWorkflowEndpoints,
@@ -31,10 +37,8 @@ import {
 } from '$lib/utilities/handle-error';
 import { stringifyWithBigInt } from '$lib/utilities/parse-with-big-int';
 import { toListWorkflowQuery } from '$lib/utilities/query/list-workflow-query';
-import {
-  type ErrorCallback,
-  requestFromAPI,
-} from '$lib/utilities/request-from-api';
+import type { ErrorCallback } from '$lib/utilities/request-from-api';
+import { requestFromAPI } from '$lib/utilities/request-from-api';
 import { base, pathForApi, routeForApi } from '$lib/utilities/route-for-api';
 
 export type GetWorkflowExecutionRequest = NamespaceScopedRequest & {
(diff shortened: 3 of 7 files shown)


src/lib/components/event/event-details-row.svelte
```diff
@@ -32,7 +32,7 @@
 >
   {#if typeof value === 'object'}
     <div
-      class="flex w-full flex-wrap items-center justify-between pr-1 xl:flex-nowrap xl:gap-4"
+      class="flex w-full flex-wrap items-center justify-between gap-1 pr-1 xl:flex-nowrap xl:gap-4"
```
laurakwhit, line 35: | Before | After |
|--|--|
|Screenshot 2023-10-05 at 10 49 13 AM|Screenshot 2023-10-05 at 10 52 09 AM|

src/lib/services/workflow-service.ts
```diff
@@ -293,30 +304,52 @@ export async function signalWorkflow({
...
+  let payloads = null;
+
+  if (signalInput) {
+    if (settings?.codec?.endpoint) {
+      const awaitData = await convertPayloadsWithCodec({
+        payloads: { payloads: [JSON.parse(signalInput)] },
+        namespace,
+        settings,
+        accessToken,
+        encode: true,
```
rossedfort, line 325: Should `encode` always be true here?

src/lib/services/workflow-service.ts
```diff
@@ -293,30 +304,52 @@ export async function signalWorkflow({
...
+      });
+      if (get(lastDataEncoderStatus) === 'error') {
+        throw new Error(
+          get(lastDataEncoderError) || translate('encode-failed'),
+        );
+      }
```
Alex-Tideman, line 331: So the thing that tripped me up testing this and I'm sure users will run into is knowing the right structure to send payloads when a codec server is configured. 

Options when codec server is configured:
- Add a message to tell users to structure payload as `{ "metadata": { "encoding": "...." }, "data": "...." }`
- Automatically set the initial value of the input to `{ "metadata": { "encoding": "...." }, "data": "...." }` when signal modal is open


Screen Shot 2023-10-10 at 11 03 31 AM

src/lib/services/workflow-service.ts
Alex-Tideman: Yeah should always be true if settings?.codec?.endpoint is set.

src/lib/services/workflow-service.ts
laurakwhit: Hmm yeah let me try both and see how it feels. Wish we could just do a placeholder here, but that would likely require it to be a `<textarea />` 🤔

Alex-Tideman: Could we try JSON.stringify({"metadata": {"encoding": "..."}, "data": "..."}, undefined, 2) to see if it renders it on multiple lines?

Alex-Tideman: Oh its just a placeholder. I would set it as the actual value so they can edit it

laurakwhit: Ah ok, there's a low likelihood of a user just be copy/pasting the input here then?

Alex-Tideman: They certainly could, but it's pretty easy and a common pattern to cmd-a and paste over the defaulted value. I think it would be really helpful to not have users need to enter these required keys.

Screen Shot 2023-10-10 at 11 03 31 AM

laurakwhit: Makes sense, updated ✅

src/lib/holocene/code-block.svelte
```diff
@@ -128,7 +128,7 @@
   });
 
   const setView = () => {
-    if (view && !editable) {
+    if (view && (!editable || (editable && !value))) {
```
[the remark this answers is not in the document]
laurakwhit: Hmm looks like my "solution" here means that the default value isn't used if the signal modal is closed and then opened again. @Alex-Tideman & @rossedfort any ideas on a better fix here?

src/lib/services/workflow-service.ts
rossedfort: I might omit the `. . .` and just use empty strings but not gonna block the PR on it

Alex-Tideman: Yeah I was kinda thinking that too. probably best to do empty strings

src/lib/holocene/code-block.svelte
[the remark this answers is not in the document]
laurakwhit: Updated in https://github.com/temporalio/ui/pull/1671/commits/1dd46db4d5fad3cb5d0ebb13eaea941fd5b0d041 ✅

src/lib/services/workflow-service.ts
laurakwhit: [Done](https://github.com/temporalio/ui/pull/1671/commits/d945dea86fa99152a6c0245ac7ca0b742a648995) ✅

Alex-Tideman (approved):

rossedfort (approved):

PR #1674, opened 2023-10-09
add jobId to batch operation form

rossedfort:
## Description & motivation 💭 
Add an input to the batch operation form so that users can specify a Job ID for their batch operations. Fall back to a uuid if not specified.
Screenshot 2023-10-09 at 3 21 37 PM
Screenshot 2023-10-09 at 3 22 13 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/workflow/batch-operation-confirmation-modal.svelte
+++ b/src/lib/components/workflow/batch-operation-confirmation-modal.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
   import { createEventDispatcher } from 'svelte';
+  import { v4 } from 'uuid';
 
   import Input from '$lib/holocene/input/input.svelte';
   import Modal from '$lib/holocene/modal.svelte';
@@ -17,6 +18,8 @@
   let reason = '';
   let isOpen = false;
   let error = '';
+  let jobId = '';
+  let jobIdPlaceholder = v4();
 
   export const open = () => {
     reason = '';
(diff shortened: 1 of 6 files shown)


src/lib/components/workflow/batch-operation-confirmation-modal.svelte
```diff
@@ -97,17 +101,24 @@
...
+          'batch-operation-confirmation-input-hint',
+          { placeholder },
+        )}
+        {placeholder}
+      />
+      <Input
+        id="batch-operation-job-id"
+        label={translate('job-id')}
+        hintText={translate('batch', 'job-id-input-hint')}
+        bind:value={jobId}
```
rossedfort, line 119: TODO: validate this is URL-safe, as the jobID ends up in the URL

Alex-Tideman (approved):

Alex-Tideman:
👍 Nice

src/lib/components/workflow/batch-operation-confirmation-modal.svelte
laurakwhit: Can we use the utils in `encode-uri.ts` for this? Maybe add something to `routeForBatchOperation` real quick?

rossedfort: Gah sorry I missed this before merging. I suppose we could have done that but I think this is also a totally reasonable approach.

PR #1684, opened 2023-10-12
DT-1150 and DT-1136 - i18n improvements

rossedfort:
## Description & motivation 💭 
- Support nested strings in our locale files - works with strongly typed `translate()` function
- Support creating a `translate` function bound to a namespace to avoid passing in a namespace to every `translate()` call

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/components/batch-operations/details.svelte
```diff
@@ -1,23 +1,24 @@
 <script lang="ts">
-  import { translate } from '$lib/i18n/translate';
+  import { createTranslate, translate } from '$lib/i18n/translate';
   import { relativeTime, timeFormat } from '$lib/stores/time-format';
   import type { BatchOperation } from '$lib/types/batch';
   import { formatDate } from '$lib/utilities/format-date';
 
   export let operation: BatchOperation;
+  const t = createTranslate('batch');
```
stevekinney, line 8: Is this really more ergonomic than what it's replacing? I'd love to see something simple like `translate('batch.details')`.

src/lib/components/batch-operations/results.svelte
```diff
@@ -1,9 +1,9 @@
 <script lang="ts">
-  import { translate } from '$lib/i18n/translate';
+  import { createTranslate } from '$lib/i18n/translate';
   import type { BatchOperation } from '$lib/types/batch';
 
   export let operation: BatchOperation;
-
+  const t = createTranslate('batch');
```
stevekinney, line 6: Alternatively, another possibility would be something like:

```ts
const { results, operationsProgess, operationsSucceeded, operationsFailed } = createTranslate('batch');
```

stevekinney (commented):
Generally speaking, I don't love the ergonomics of this.

src/lib/components/batch-operations/details.svelte
rossedfort: I think it's slightly more ergonomic when translating a bunch of strings in the same namespace, however it's probably not going to be realized until we add new pages. I like the idea of passing in a single string like `"<namespace>.<key>"` though.

src/lib/components/batch-operations/results.svelte
rossedfort: Yeah that would be nice, let me look into this.

Alex-Tideman (approved):

src/lib/components/batch-operations/results.svelte
rossedfort: Team opted not to do this as it will end up creating polluting the `<script />` tag in our svelte components with lots of on-off variable names. Also would likely require a refactor of all of our i18n keys to be `camelCase` instead of `kebab-case`.

PR #1685, opened 2023-10-12
Keyboard navigation improvements

laurakwhit:
## Description & motivation 💭 
Now, when using the keyboard 
* focus goes to the `Drawer` component when it is opened and is trapped there
* all `Link` components are tabbable
* the filter and copy buttons in the Workflows table are tabbable
* the `Menu` component closes when it loses focus

This PR also adds a reusable `focusTrap` and updates z-indices that were causing bugs (see screenshot below) as well as removes unnecessary ones. 
Screenshot 2023-10-11 at 7 38 24 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

- [ ] Verify a `Menu` component closes when tabbed out of it
- [ ] Verify focus is trapped in the `Drawer` component (e.g. `Configure Workflow List`) when tabbing through it
- [ ] Verify focus is trapped in the `Modal` component when tabbing through it
- [ ] Verify the `name` link is tabbable in the `Schedules` table
- [ ] Verify UI appears as expected on all pages (no overlap due to z-indices)

### DIFF
--- a/src/lib/components/schedule/schedules-table-row.svelte
+++ b/src/lib/components/schedule/schedules-table-row.svelte
@@ -1,5 +1,4 @@
 <script lang="ts">
-  import { goto } from '$app/navigation';
   import { page } from '$app/stores';
 
   import WorkflowStatus from '$lib/components/workflow-status.svelte';
@@ -52,12 +51,7 @@
     <WorkflowStatus status={schedule?.info?.paused ? 'Paused' : 'Running'} />
   </td>
   <td class="cell whitespace-pre-line break-words">
-    <Link
-      class="text-base"
-      on:click={() => {
-        goto(route);
-      }}
-    >
+    <Link class="text-base" href={route}>
       {schedule.scheduleId}
     </Link>
   </td>
--- a/src/lib/holocene/drawer.svelte
+++ b/src/lib/holocene/drawer.svelte
@@ -2,6 +2,7 @@
   import { fly } from 'svelte/transition';
 
   import { clickOutside } from '$lib/holocene/outside-click';
+  import { focusTrap } from '$lib/utilities/focus-trap';
 
   import IconButton from './icon-button.svelte';
 
@@ -29,6 +30,7 @@
     {id}
     role="region"
     on:click-outside={onClick}
+    use:focusTrap={true}
   >
     <div class="close-button-wrapper {position}">
       <IconButton
--- a/src/lib/utilities/focus-trap.ts
+++ b/src/lib/utilities/focus-trap.ts
@@ -1,0 +1,70 @@
+export const focusTrap = (node: HTMLElement, enabled: boolean) => {
+  let firstFocusable: HTMLElement;
+  let lastFocusable: HTMLElement;
+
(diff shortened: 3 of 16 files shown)


src/lib/holocene/drawer.svelte
```diff
@@ -29,6 +30,7 @@
     {id}
     role="region"
     on:click-outside={onClick}
+    use:focusTrap={true}
```
laurakwhit, line 33: There is an edge case with the `Configure Workflows List`. When you hit `Enter` on the last item in the list to add it to the headings the drawer loses focus. Not sure on an elegant solution here, since usually that would be expected behavior (e.g. a close button). 

We could consider adding some optional param e.g. `exitOnEnter` and adding something like this to `focusTrap` ⬇️ 
```js
if (!exitOnEnter && event.key === 'Enter' && document.activeElement === lastFocusable) {
    firstFocusable.focus();
    event.preventDefault();
}
```

src/lib/utilities/focus-trap.ts
```diff
@@ -0,0 +1,77 @@
...
+      | HTMLButtonElement
+      | HTMLTextAreaElement
+      | HTMLInputElement
+      | HTMLDivElement
+      | HTMLSelectElement
+    >('[href], button, textarea, input, div[contentEditable="true"], select'),
+  ).filter(
+    (element) =>
+      !element.hasAttribute('disabled') &&
+      !(element.getAttribute('tabindex') === '-1'),
```
laurakwhit, line 14: Tried adding `[tabindex]:not([tabindex="-1"])` to the query selector argument, but it didn't appear to work based on testing.

laurakwhit (commented):
Noting that the `z-index` in `paginated-table` can likely be removed when the table filters are removed (and therefore the z-index in the `Drawer` component can be updated as well).

stevekinney (approved):

src/lib/components/schedule/schedules-table-row.svelte
```diff
@@ -52,12 +51,7 @@
     <WorkflowStatus status={schedule?.info?.paused ? 'Paused' : 'Running'} />
   </td>
   <td class="cell whitespace-pre-line break-words">
-    <Link
-      class="text-base"
-      on:click={() => {
-        goto(route);
-      }}
-    >
+    <Link class="text-base" href={route}>
```
Alex-Tideman, line 54: So why I did this...

With Link href={}, it does a full reload and refetches settings/cluster....

With on:click and goto, it does not

Alex-Tideman: So the question begs.  How do keep a tags accessible (hrefs) but use client-side routing with goto

laurakwhit: Whelp, I'm gonna add that back then 😅 Changed this before I updated the `Link` component, but now it should still be tabbable.

Alex-Tideman:
I'm not sure where the issue stems from, but the Codec Server config doesn't open when you click the glasses icon

PR #1686, opened 2023-10-13
Workflow Counts

Alex-Tideman:
Screen Shot 2023-10-13 at 11 56 57 AM

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/components/workflow/workflow-counts.svelte
```diff
@@ -1,90 +1,61 @@
...
+        const status = decodePayload(
+          group?.groupValues[0],
+        ) as unknown as WorkflowStatus;
+        const count = parseInt(group.count);
+        return {
+          status,
+          count,
+        };
+      })
+      .filter((s) => s.count > 0);
```
rossedfort, line 50: I think this probably fits better in the workflows service

rossedfort (commented):
Looks good except for 1 comment on moving the API logic into a service

src/lib/components/workflow/workflow-counts.svelte
```diff
@@ -1,90 +1,44 @@
...
+      namespace,
+      filters: $workflowFilters,
+    });
+    $workflowCount.totalCount = parseInt(count);
+    statusGroups = groups
+      .map((group) => {
+        const status = decodePayload(
+          group?.groupValues[0],
+        ) as unknown as WorkflowStatus;
+        const count = parseInt(group.count);
```
rossedfort, line 27: nit: `fetchWorkflowCountByExecutionStatus` could probably just do this data massaging

src/lib/services/cluster-service.ts
```diff
@@ -14,7 +13,20 @@ export const fetchCluster = async (
...
+  request = fetch,
+): Promise<GetSystemInfoResponse> => {
+  if (settings.runtimeEnvironment.isCloud) return;
+
+  const route = routeForApi('systemInfo');
+  return await requestFromAPI(route, {
+    request,
+  }).then((systemInformation) => {
+    return systemInformation;
+  });
```
rossedfort, line 31: nit: this is redundant, can just `return requestFromApi(route { request });`

rossedfort (approved):

src/lib/components/workflow/workflow-counts.svelte
Alex-Tideman: I kinda like it in the view layer because we are massaging this data specifically for this component. Down the road we might want the service to do something else, so rather leave it more generic.

rossedfort: Fair point 👍

PR #1688, opened 2023-10-13
Use last WorkflowTaskFailed event for strongly typed error

laurakwhit:
## Description & motivation 💭 
If there are multiple `WorkflowTaskFailed` events in `eventHistory.end` then the first one was being set as the `workflowTaskFailedEvent` in `getWorkflowStartedCompletedAndTaskFailedEvents` since `end` has events in descending order. This change ensures that the last `WorkflowTaskFailed` is from the full event history is passed to the `WorkflowTypedError` component.

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


Alex-Tideman:
Let's add a unit test to get-started-completed-and-task-failed-events.test.ts to cover this case.

Alex-Tideman (approved):

PR #1694, opened 2023-10-16
Individual Event History page

GraceGardner:
## Description & motivation 💭 
This PR creates an individual workflow event page, allowing a user to share specific events via URL to see the details of that event quickly without needing to scroll through event history. If the event id in the URL does not exist in the event history, the page will show a 'No Events Match' result.

This also changes the event summary row opening behavior to allow multiple open rows. It removes auto-opening rows with keyboard navigation.

Screen Shot 2023-10-17 at 10 36 50 AM
Screen Shot 2023-10-17 at 2 55 33 PM
Screen Shot 2023-10-17 at 2 57 22 PM

### Design Considerations 🎨 
Would love to pair with @tigernaut and see their opinions on user flow. 

### How was this tested 👻 

- [X] Manual testing
- [x] E2E tests added
- [ ] Unit tests added

- go to a workflow, select the id number for the workflow, this should take you to that specific page for the selected workflow event. If copy and pasted this should take you to that same event page. 

### Issue(s) closed 
https://temporalio.atlassian.net/browse/DT-1517


src/lib/utilities/route-for.ts
```diff
@@ -40,8 +40,9 @@ export type EventHistoryParameters = Pick<
   RouteParameters,
   'namespace' | 'workflow' | 'run' | 'view' | 'queryParams'
 >;
-export type EventParameters = Required<
-  Pick<RouteParameters, 'namespace' | 'workflow' | 'run' | 'view' | 'eventId'>
+export type EventParameters = Pick<
```
GraceGardner, line 43: @Alex-Tideman this is how I was able to fix the type errors. I'm worried that this may break something in the future. Was their a better way of handling this?

Alex-Tideman: Yup it's okay to remove that. One good way to check is to see where EventParameters is used throughout the app with a global search in your IDE. It was only used in the test file, so I updated the tests to not include view.

Alex-Tideman (approved):

Alex-Tideman (approved):

PR #1699, opened 2023-10-19
Refresh Workflow Counts

Alex-Tideman:
## Description & motivation 💭 
After initial workflowList query and count query completes, set an interval to show changes in status counts in the StatusBadges, as well as count of new workflows in the refresh button.

Interval is set as an exponential backoff interval starting at 5 seconds and up to one hour after 100 attempts to refresh the counts. For example, first update of counts takes 5 seconds. Second update takes 6 seconds, 10th update takes 10 seconds, ..., 100th update takes an hour from 99th attempt. After the 100th attempt, clear the interval and don't update the counts.

This way a user gets a more immediate, short-term feedback whenever they want to watch changes after querying and it will prevent excessive count requests for idle users. They also get to see an indication of changes in status count and new incoming workflows, but still have control over when to view them in the workflow list.

https://github.com/temporalio/ui/assets/7967403/e0cd9dfa-8505-4c40-9f15-c1721ae5dc89

### How was this tested 👻 

- [x] Manual testing
- [x] E2E tests added
- [x] Unit tests added


src/lib/components/workflow/workflow-counts.svelte
```diff
@@ -10,22 +13,48 @@
...
+  const maxAttempts = 100;
+
+  onDestroy(() => {
+    clearNewCounts();
+  });
+
+  const setBackoffInterval = () => {
+    attempt += 1;
+    clearInterval(refreshInterval);
+    if (attempt <= 100) {
```
rossedfort, line 38: Should this be `maxAttempts` ? Or is `100` just an arbitrary choice here

rossedfort (approved):

src/lib/components/workflow/workflow-counts.svelte
Alex-Tideman: Good catch, fixed.

PR #1700, opened 2023-10-19
Update UI to use our official HTTP API

tdeebswihart:
## Description & motivation 💭
In order to update this repo's dependency on gogo/protobuf we need to unify the protos we have here with those in our API repo: the new proto library we're using panics at runtime if it a proto is registered twice, whereas gogo ignored it and giggled (possibly causing bugs).

I updated the UI and our tests to use the routes we agreed upon during our [discussions about an official HTTP API](https://github.com/temporalio/proposals/blob/master/api/http-api.md).

### How was this tested 👻 
No tests were added but all tests were updated.

- [x] Manual testing
- [x] E2E tests updated
- [x] Unit tests updated

Run the tests!

### Draft Checklist
- [x] Wait for temporalio/api#319 to merge and update the submodule
- [x] Wait for temporalio/api-go#126 to merge and remove the `replace` directive in server/go.mod
- [x] Retest

### Any docs updates needed?

I don't believe so


src/lib/utilities/route-for-api.ts
```diff
@@ -99,34 +103,34 @@ export function pathForApi(
...
-    'workflows.count': `/namespaces/${parameters?.namespace}/workflows/count`,
-    'activity.complete': `/namespaces/${parameters.namespace}/workflows/${parameters.workflowId}/runs/${parameters.runId}/activities/${parameters.activityId}/complete`,
-    'activity.fail': `/namespaces/${parameters.namespace}/workflows/${parameters.workflowId}/runs/${parameters.runId}/activities/${parameters.activityId}/fail`,
-    'batch-operations': `/namespaces/${parameters.namespace}/batch-operations`,
-    'batch-operation.describe': `/namespaces/${parameters.namespace}/batch-operations/describe`,
+    'workflows.count': `/namespaces/${parameters?.namespace}/workflow-count`,
+    'activity.complete': `/namespaces/${parameters.namespace}/activities/complete-by-id`,
+    'activity.fail': `/namespaces/${parameters.namespace}/activities/fail-by-id`,
+    'batch-operations': `/namespaces/${parameters.namespace}/batch-operations/${parameters?.batchJobId}`,
+    'batch-operation.describe': `/namespaces/${parameters.namespace}/batch-operations/${parameters?.batchJobId}`,
```
rossedfort, line 133: Since these are now identical, maybe we just keep one? Also just so I understand, to create Batch Operation, a POST is sent to `/namespaces/<namespace>/batch-operations/<jobId>` ? Does the jobId also need to be included in the body of the request?

rossedfort:
@tdeebswihart I can't seem to run this locally, does it require using go v1.20?

```
error when starting dev server:
Error: go: go.mod file indicates go 1.20, but maximum version supported by tidy is 1.18
make: *** [build-server] Error 1
```

tdeebswihart:
> @tdeebswihart I can't seem to run this locally, does it require using go v1.20?
> 
> ```
> error when starting dev server:
> Error: go: go.mod file indicates go 1.20, but maximum version supported by tidy is 1.18
> make: *** [build-server] Error 1
> ```

Yeah. I had to bump it as part of this effort but I can't recall why. I _know_ we need at least go 1.19 for the new GRPC gateway we'll be using as part of the protobuf changes, so we decided to go straight to 1.20

src/lib/utilities/route-for-api.ts
tdeebswihart: Good catch. And I don't believe so: it should be extracted by our GRPC gateway

Alex-Tideman (approved):

PR #1704, opened 2023-10-23
Allow decoding Event History Download through the UI

Alex-Tideman:
## Description & motivation 💭 
Previously we had an env var to allow downloading decoded event histories. This PR removes that env var and replaces it with a toggle from the UI. Decoding is defaulted on.

This also strips the extra UI attributes we add (id, name, classification, timestamp, category, attributes) from downloaded event histories and JSON view events.

Screen Shot 2023-10-23 at 1 52 42 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

### DIFF
--- a/src/lib/components/workflow/workflow-json-navigator.svelte
+++ b/src/lib/components/workflow/workflow-json-navigator.svelte
@@ -3,6 +3,7 @@
   import RangeInput from '$lib/holocene/input/range-input.svelte';
   import Loading from '$lib/holocene/loading.svelte';
   import { translate } from '$lib/i18n/translate';
+  import { fromEventToRawEvent } from '$lib/models/event-history';
   import type { WorkflowEvents } from '$lib/types/events';
   import { stringifyWithBigInt } from '$lib/utilities/parse-with-big-int';
 
@@ -84,7 +85,10 @@
   </div>
   {#if decodeEventHistory}
     {#key [index, decodeEventHistory]}
-      <PayloadDecoder value={events[index - 1]} let:decodedValue>
+      <PayloadDecoder
+        value={fromEventToRawEvent(events[index - 1])}
+        let:decodedValue
+      >
         <CodeBlock
           content={decodedValue}
           testId="event-history-json"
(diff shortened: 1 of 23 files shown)


src/lib/layouts/workflow-history-layout.svelte
```diff
@@ -164,3 +174,26 @@
...
+  on:cancelModal={() => (showDownloadPrompt = false)}
+>
+  <h3 slot="title">
+    {translate('common.download-json')}
+  </h3>
+  <div slot="content" class="flex flex-col gap-4">
+    <ToggleSwitch
+      label={translate('events.decode-event-history')}
+      id="decode-event-history"
+      bind:checked={decodeEventHistory}
```
laurakwhit, line 195: Might be nice to default this to whatever the user has selected for `View decoded event history`

Screenshot 2023-10-23 at 3 07 54 PM

src/lib/components/workflow/workflow-json-navigator.svelte
```diff
@@ -96,7 +100,11 @@
   {:else}
     {#key index}
       <CodeBlock
```
laurakwhit, line 102: Looks like this is existing behavior, but when toggling `View decoded event history` off and then back on again it looks like there is a page "flash" and a full UI re-render. Might be something we can improve...

Screenshot 2023-10-23 at 3 10 13 PM

laurakwhit (approved):

src/lib/components/workflow/workflow-json-navigator.svelte
Alex-Tideman: Good call. We need to rerender the CodeBlock on decode toggle, but I changed the initial decodeValue string from empty string to stringifyWithBigInt(value). So the initial value is the event un-decoded, and then if it decodes successfully, replaces it with the decoded value. This way you don't get that flash which is the change in height from empty string to the decoded value. Much better experience.

Alex-Tideman: I also changed the label to "Decode Event History" to make them consistent

src/lib/layouts/workflow-history-layout.svelte
Alex-Tideman: 👍 Made it a persistStore and used for both

PR #1709, opened 2023-10-26
Schedule Workflows Count and View All link

Alex-Tideman:
## Description & motivation 💭 
It is really helpful to not only show the last 5 runs of a schedule, but the count of all workflows from the schedule and a link to view all runs on the Workflow list page.

This adds the workflow counts on the Schedule view page if that capability is enabled, as well as a link to view all workflows of that schedule with a predefined link.

This PR also fixes the 1 Workflows pluralization to 1 Workflow in both the Workflows list page and here.

Screen Shot 2023-10-26 at 9 15 27 AM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/schedule/schedule-recent-runs.svelte
+++ b/src/lib/components/schedule/schedule-recent-runs.svelte
@@ -51,7 +51,6 @@
       View All Runs
     </Link>
   </div>
-
   {#each sortRecentRuns(recentRuns) as run (run?.startWorkflowResult?.workflowId)}
     {#await fetchWorkflowForSchedule({ namespace, workflowId: decodeURIForSvelte(run.startWorkflowResult.workflowId), runId: run.startWorkflowResult.runId }, fetch) then workflow}
       <div class="row">
--- a/src/lib/components/schedule/schedule-recent-runs.svelte
+++ b/src/lib/components/schedule/schedule-recent-runs.svelte
@@ -48,7 +48,7 @@
         );
       }}
     >
-      View All Runs
+      {translate('common.view-all-runs')}
     </Link>
   </div>
   {#each sortRecentRuns(recentRuns) as run (run?.startWorkflowResult?.workflowId)}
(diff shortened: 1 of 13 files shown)


src/lib/components/schedule/schedule-recent-runs.svelte
```diff
@@ -30,7 +36,21 @@
...
+        goto(
+          routeForWorkflowsWithQuery({
+            namespace,
+            query: workflowQuery,
+          }),
+        );
+      }}
+    >
+      View All Runs
+    </Link>
```
rossedfort, line 52: Why not use `href` here? Also needs an i18n string

src/lib/pages/schedule-view.svelte
```diff
@@ -260,11 +263,40 @@
...
+      {#if $groupByCountEnabled}
+        <div class="flex w-full flex-col gap-2 text-lg">
+          <div class="flex items-center gap-2">
+            <span data-testid="workflow-count"
+              >{$workflowCount.count.toLocaleString()}
+              <Translate
+                key={$workflowCount.count === 1
+                  ? 'common.workflow'
+                  : 'common.workflows'}
+              />
```
rossedfort, line 275: I18next support pluralization out of the box, un `src/lib/i18n/locales/common.ts` you can have 2 versions of the same key:
```
{
  "workflows_one": "Workflow",
  "workflows_other": "Workflows",
}
```
And then just pass a `count` prop to `<Translate />`, i.e. `<Translate key="common.workflows" count={$workflowCount.count} />`

rossedfort (approved):
Looks good just a couple small suggestions

src/lib/components/schedule/schedule-recent-runs.svelte
Alex-Tideman: 👍 Cuz href does a hard reload all the apis instead of client-side routing

src/lib/pages/schedule-view.svelte
Alex-Tideman: Nice nice nice, I was definitely using count wrong when i tried this.

PR #1715, opened 2023-10-27
Check for abandoned child start events

laurakwhit:
## Description & motivation 💭 
* Checks for `ChildWorkflowExecutionStartedEvent`s that don't have a corresponding closed event and displays them under `Relationships` > `Children`.
* Fixes rendering bug if there are children without a `runId`.

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

- [ ] Verify child events show up as expected under `Relationships` > `Children` 
   - when events are in ascending order
   - when events are in descending order
   - when all child events have been completed (no duplicates)
   - when some child started events don't have corresponding closed events

### Issue(s) closed 
`DT-1531`

### DIFF
--- a/src/lib/components/workflow/child-workflows-table.svelte
+++ b/src/lib/components/workflow/child-workflows-table.svelte
@@ -58,7 +58,7 @@
       <th>{translate('workflows.child-id')}</th>
       <th>{translate('workflows.child-run-id')}</th>
     </TableHeaderRow>
-    {#each visibleItems as child (child.runId)}
+    {#each visibleItems as child}
       <TableRow>
         <td class="max-md:hidden">
           <WorkflowStatus status={child.status} />
--- a/src/lib/components/workflow/child-workflows-table.svelte
+++ b/src/lib/components/workflow/child-workflows-table.svelte
@@ -6,12 +6,12 @@
   import Table from '$lib/holocene/table/table.svelte';
   import { translate } from '$lib/i18n/translate';
   import type { WorkflowExecution } from '$lib/types/workflows';
-  import type { ChildWorkflowClosedEvent } from '$lib/utilities/get-workflow-relationships';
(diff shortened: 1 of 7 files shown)


src/lib/components/workflow/child-workflows-table.svelte
```diff
@@ -58,7 +58,7 @@
       <th>{translate('workflows.child-id')}</th>
       <th>{translate('workflows.child-run-id')}</th>
     </TableHeaderRow>
-    {#each visibleItems as child (child.runId)}
+    {#each visibleItems as child}
```
Alex-Tideman, line 61: Can there be duplicate runIds? Or is this to prevent the same workflow shown twice?

Alex-Tideman:
Just a general question: How can you abandon a child workflow? What's the situation where it doesn't complete/fail/cancel/terminate...

src/lib/components/workflow/child-workflows-table.svelte
laurakwhit: > Also when I run samples-typescript / child-workflows, I don't get any errors (not on this branch)

Running that sample was causing this error ⬇️ Which was because multiple children had the same `runId`.

Screenshot 2023-10-26 at 5 06 32 PM

laurakwhit:
> Just a general question: How can you abandon a child workflow? What's the situation where it doesn't complete/fail/cancel/terminate...

Looks like you can abandoned children if the parent needs to close but children are still running. 

> Also when I run samples-typescript / child-workflows, I don't get any errors (not on this branch). Do I need to do something to trigger the error?

See [this comment](https://github.com/temporalio/ui/pull/1715#discussion_r1374972971).

Alex-Tideman (approved):

laurakwhit:
**ON HOLD:** In testing it appears that children that haven't completed once a parent with an abandon child policy closes should continue to show up as pending children (even once the children complete). Awaiting further repro. info.

PR #1726, opened 2023-11-02
Better responsive layout for Workflow Summary to keep times visible

Alex-Tideman:
Screen Shot 2023-11-02 at 12 04 37 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/components/workflow/workflow-summary.svelte
```diff
@@ -31,16 +31,14 @@
       $workflowSummaryViewOpen = !$workflowSummaryViewOpen;
     }}
   >
-    <div
-      class="grid-row-3 grid w-full grid-cols-1 gap-4 md:grid-cols-2 lg:grid-cols-4 xl:w-11/12"
-    >
-      <div class="col-span-1 md:col-span-2">
+    <div class="flex flex-col gap-2 xl:flex-row">
+      <div class="grow">
```
laurakwhit, line 35: Unfortunately, this breaks truncation 😞 

Screenshot 2023-11-03 at 3 58 26 PM

laurakwhit (changes requested):

src/lib/components/workflow/workflow-summary.svelte
Alex-Tideman: 👍

src/lib/components/workflow/workflow-summary.svelte
```diff
@@ -31,16 +31,14 @@
...
-      <div class="col-span-1 md:col-span-2">
+    <div class="flex flex-col gap-2 xl:flex-row">
+      <div class="grow overflow-hidden">
         <h3 class="font-medium">{translate('common.workflow-type')}</h3>
         <div class="h-0.5 rounded-full bg-gray-900" />
         <WorkflowDetail content={workflow?.name} copyable />
         <WorkflowDetail content={workflow?.runId} copyable />
       </div>
-      <div class="col-span-1">
+      <div class="grow overflow-hidden">
```
laurakwhit, line 41: What do you think about using `flex-1` instead of `grow` for the type and task queue sections?

```suggestion
      <div class="flex-1 overflow-hidden">
```

Otherwise the task queue section can get cut off if the workflow name is super long ⬇️ 

| Before | After |
|--|--|
|Screenshot 2023-11-06 at 9 21 05 AM|Screenshot 2023-11-06 at 9 30 25 AM|

Alex-Tideman: It's a balancing act for sure. I think the most likely situation is a workflow type and task queue that are long but not insanely long like that one. In your example, if the Workflow Type was a little shorter in text, the Type would not be truncated and the task queue wouldn't be either (Before with grow). With flex-1, it will still be truncated.

I think I prefer grow because it's a little more flexible on growing with the long text (with the downside of truncating with super long texts, but I think that's less likely).

laurakwhit (approved):

PR #1736, opened 2023-11-08
Apply selected timezone to datetime filters

laurakwhit:
## Description & motivation 💭 
When selecting a timezone, a user should be able to select a datetime filter based on that timezone.

Screenshot 2023-11-08 at 9 39 58 AM
### How was this tested 👻 

- [x] Manual testing
- [x] E2E tests added
- [x] Unit tests added

- [x] Verify when selecting an `Absolute` date with a non-local timezone selected
   - the filter "chip" should show the correct time (based on the timezone)
   - the search input and query param show the time in UTC
   - the filter should work as expected
- [x] Verify when selecting a `Relative` date with a non-local timezone selected
   - the filter "chip" should show the correct time (based on the timezone)
   - the search input and query param show the time in UTC
   - the filter should work as expected
### Issue(s) closed 
`DT-1377`

### DIFF
--- a/src/lib/components/workflow/filter-search/datetime-filter.svelte
+++ b/src/lib/components/workflow/filter-search/datetime-filter.svelte
@@ -8,6 +8,7 @@
     formatISO,
     startOfDay,
   } from 'date-fns';
+  import { zonedTimeToUtc } from 'date-fns-tz';
   import { getContext } from 'svelte';
 
   import Button from '$lib/holocene/button.svelte';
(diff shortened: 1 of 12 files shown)


Alex-Tideman:
This is great! 🎉 

I know it's not the current behavior, but I know at one point we kept Start/End filters when you switched namespaces. Is that something we want to bring back after this fix or was that determined not a good UX? Not something we need to address on this PR regardless.

Alex-Tideman (approved):

laurakwhit:
> I know it's not the current behavior, but I know at one point we kept Start/End filters when you switched namespaces. Is that something we want to bring back after this fix or was that determined not a good UX? Not something we need to address on this PR regardless.

I'm not sure what the right answer is here regarding good UX. Seems like a system of bookmarking/starring filters that could apply across Namespaces might better fit what users are looking for (and that way it's not just the time filters). Will bring to the team for more discussion.

PR #1737, opened 2023-11-09
Fix query param for requests containing a Run ID

rossedfort:
## Description & motivation 💭 
In #1700 we accidentally introduced a query param that is not compatible with the new protos. `runId` should be `execution.runId`
### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


tdeebswihart (approved):
This looks like all of them. It'd be great to have a test to catch this in the future

PR #1749, opened 2023-11-14
Use basic frequency without CodeBlock for schedules list view

Alex-Tideman:
## Description & motivation 💭 
To improve the render performance of the Schedules list page when there are hundreds of schedules, I removed the CodeBlock and used `<code><pre>` with matching css.

### BEFORE
Screen Shot 2023-11-14 at 8 33 02 AM

### AFTER
Screen Shot 2023-11-14 at 8 32 46 AM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/schedule/schedule-basic-frequency.svelte
+++ b/src/lib/components/schedule/schedule-basic-frequency.svelte
@@ -1,0 +1,21 @@
+<script lang="ts">
+  import type { StructuredCalendar } from '$lib/types/schedule';
+  import { stringifyWithBigInt } from '$lib/utilities/parse-with-big-int';
+
+  import type { IntervalSpec } from '$types';
+
+  export let calendar: StructuredCalendar | undefined = undefined;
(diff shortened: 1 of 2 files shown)


src/lib/components/schedule/schedule-basic-frequency.svelte
```diff
@@ -0,0 +1,21 @@
...
+
+  import type { IntervalSpec } from '$types';
+
+  export let calendar: StructuredCalendar | undefined = undefined;
+  export let interval: IntervalSpec | undefined = undefined;
+</script>
+
+{#key [calendar, interval]}
+  <div class="frequency {$$props.class}">
+    <code><pre>{stringifyWithBigInt(calendar || interval)}</pre></code>
```
laurakwhit, line 13: Do we want to add a copy button here?

laurakwhit (approved):

PR #1758, opened 2023-11-17
Add maxLength to Textarea

laurakwhit:
## Description & motivation 💭 
Add `maxLength` prop for Holocene `Textarea` component. 

Screenshot 2023-11-17 at 4 42 52 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

- [x] Verify a `Textarea` with a `maxLength`
   - Shows up as expected (count should be orange within 5 characters and red when it reaches max)
   - Doesn't allow additional characters
   - Still shows an error message as expected (with no overlap of the count)

### DIFF
--- a/src/lib/holocene/textarea.svelte
+++ b/src/lib/holocene/textarea.svelte
@@ -1,4 +1,22 @@
 <script lang="ts">
+  import type { HTMLTextareaAttributes } from 'svelte/elements';
+
+  type $$Props = HTMLTextareaAttributes & {
+    disabled?: boolean;
+    error?: string;
+    isValid?: boolean;
+    placeholder?: string;
+    rows?: number;
+    spellcheck?: boolean;
+    value: string;
+    label: string;
+    labelHidden?: boolean;
+    id: string;
+    required?: boolean;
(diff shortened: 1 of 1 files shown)


src/lib/holocene/textarea.svelte
```diff
@@ -60,6 +95,26 @@
   }
 
   .error-msg {
-    @apply border-danger font-primary text-sm font-normal text-danger;
+    @apply break-words border-danger font-primary text-sm font-normal text-danger;
```
Alex-Tideman, line 98: One suggestion: add `h-auto min-h-[1.5rem]` here to keep a consistent height so there's no jumping when the error message does appear, and on small screens it will grow to give the message and the count more room and not bleed into content below.

Before
Screen Shot 2023-11-20 at 9 17 52 AM

After
Screen Shot 2023-11-20 at 9 18 09 AM

laurakwhit: Done ✅

Alex-Tideman (approved):

PR #1771, opened 2023-12-12
Allow disabled styling on Buttons as links

laurakwhit:
## Description & motivation 💭 
Render `Button` component as `<button/>` instead of `<a/>` if `disabled` (allows for disabled styling on links).

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/holocene/button.svelte
+++ b/src/lib/holocene/button.svelte
@@ -81,7 +81,7 @@
     HTMLAnchorAttributes & {
       href: string;
       target?: HTMLAnchorAttributes['target'];
-      disabled?: never;
+      disabled?: boolean;
     };
 
   type ButtonStyles = VariantProps<typeof buttonStyles>;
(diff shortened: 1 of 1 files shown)


Alex-Tideman (approved):

src/lib/holocene/button.svelte
```diff
@@ -104,10 +104,10 @@
 </script>
 
 <svelte:element
-  this={href ? 'a' : 'button'}
+  this={href && !disabled ? 'a' : 'button'}
   {target}
-  {href}
-  disabled={href ? null : disabled}
+  href={disabled ? null : href}
```
rossedfort, line 109: Might want to retain the check here if href is passed otherwise we may get `<button />`'s with an empty `href` I think?

laurakwhit: This should behave the same since `href` defaults to `null` if not passed, so it would be still be `href={null}` even if `disabled` is `false`.

rossedfort (approved):

PR #1775, opened 2023-12-14
Update ui-server to use the new protos

tdeebswihart:
## Description & motivation 💭

This updates the UI server to use our new google/protobuf-compiled proto code. This introduces a breaking change to our JSON formatting: all proto enums are now in SCREAMING_SNAKE_CASE rather than camelCase.

This PR (and branch) is just to get y'all what you need to update the UI itself and this should not be merged on its own.

N/A

### Design Considerations 🎨

N/A

### How was this tested 👻

No tests were added and all tests have likely broken. This PR is a starting point upon which the UI itself can be updated. I've done some _very_ basic manual testing to ensure the API returns what we expect it to, like the new SCREAMING_SNAKE_ENUMS:

```
$ xh localhost:8081/api/v1/namespaces
HTTP/1.1 200 OK
Content-Length: 1349
Content-Type: *
...
{
  "namespaces": [
    {
      "namespaceInfo": {
        "name": "temporal-system",
        "state": "NAMESPACE_STATE_REGISTERED", # NEW ENUM FORMAT
        "description": "Temporal internal system namespace",
...
```

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

1. Update the UI to handle SCREAMING_SNAKE_CASE enums
2. Run existing tests

### Draft Checklist
- [x] Need to fix query api params


server/go.mod
```diff
@@ -4,24 +4,23 @@ go 1.20
 
 require (
 	github.com/coreos/go-oidc/v3 v3.1.0
-	github.com/gogo/gateway v1.1.0
 	github.com/gorilla/securecookie v1.1.1
-	github.com/grpc-ecosystem/grpc-gateway v1.16.0
+	github.com/grpc-ecosystem/grpc-gateway/v2 v2.18.1
```
Alex-Tideman, line 8: What is the breaking change for this?

src/lib/services/namespaces-service.ts
```diff
@@ -8,11 +8,35 @@ import type { Settings } from '$lib/types/global';
...
+      namespace.config?.historyArchivalState,
+    );
+    namespace.config.visibilityArchivalState = toNamespaceArchivalStateReadable(
+      namespace.config?.visibilityArchivalState,
+    );
+  }
+
+  if (namespace.namespaceInfo) {
+    namespace.namespaceInfo.state = toNamespaceStateReadable(
+      namespace.namespaceInfo?.state,
```
laurakwhit, line 34: Can we remove the `?` here since there is the `namespace.namespaceInfo` check above (same with `namespace.config`)? Or maybe remove the check if `toNamespaceStateReadable` can handle `undefined`?

```suggestion
      namespace.namespaceInfo.state,
```

laurakwhit (approved):
Seeing `undefined` for this column ⬇️ Once that's fixed I say merge it 🚀

Screenshot 2024-01-29 at 11 06 38 AM

src/lib/services/namespaces-service.ts
Alex-Tideman: 👍 Yup good call

Alex-Tideman (approved):

PR #1779, opened 2023-12-19
Schedules with API Pagination

Alex-Tideman:
## Description & motivation 💭 
Use API pagination for schedules so you can view more than 1,000 schedules. Since you cannot search schedules to narrow down what you are looking forward, this is necessary for users to find schedules if they have over 1,000.

Screen Shot 2024-01-04 at 12 12 34 PM
Screen Shot 2024-01-04 at 12 26 53 PM

### Design Considerations 🎨 

- [x] Search only finds current page schedules

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/pages/schedules.svelte
+++ b/src/lib/pages/schedules.svelte
@@ -25,7 +25,9 @@
   const onError: ErrorCallback = (err) =>
     (error =
       err?.body?.message ??
-      `Error fetching schedules: ${err.status}: ${err.statusText}`);
+      `${translate('schedules.error-message-fetching')}: ${err.status}: ${
+        err.statusText
+      }`);
 
   let coreUser = coreUserStore();
   $: createDisabled = $coreUser.namespaceWriteDisabled(namespace);
@@ -46,12 +48,12 @@
     let:visibleItems
     {onFetch}
     {onError}
-    aria-label="schedules"
+    aria-label={translate('common.schedules')}
     pageSizeSelectLabel={translate('common.per-page')}
     nextButtonLabel={translate('common.next')}
     previousButtonLabel={translate('common.previous')}
     emptyStateMessage={translate('schedules.empty-state-title')}
-    fallbackErrorMessage={'Error fetching schedules'}
+    fallbackErrorMessage={translate('schedules.error-message-fetching')}
   >
     <header class="flex flex-row justify-between gap-2" slot="header">
       <div>
--- a/src/lib/pages/schedules.svelte
+++ b/src/lib/pages/schedules.svelte
@@ -6,6 +6,7 @@
   import ApiPagination from '$lib/holocene/api-pagination.svelte';
   import Button from '$lib/holocene/button.svelte';
   import EmptyState from '$lib/holocene/empty-state.svelte';
+  import Link from '$lib/holocene/link.svelte';
   import TableRow from '$lib/holocene/table/table-row.svelte';
   import { translate } from '$lib/i18n/translate';
   import { fetchPaginatedSchedules } from '$lib/services/schedule-service';
@@ -15,9 +16,7 @@
 
   $: namespace = $page.params.namespace;
 
-  let hasSchedules = false;
   let error = '';
-
   const onError: ErrorCallback = (err) =>
     (error =
       err?.body?.message ??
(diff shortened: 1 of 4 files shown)


src/lib/pages/schedules.svelte
```diff
@@ -49,7 +36,7 @@
       {translate('common.schedules')}
     </h1>
   </div>
-  {#if hasSchedules && !createDisabled}
+  {#if !createDisabled}
```
laurakwhit, line 39: Removing the `hastSchedules` check here means that this button shows up in two places on the `No Schedules Found` page. Was there a reason it was removed? 

Screenshot 2024-01-03 at 1 03 19 PM

Alex-Tideman: We don't know the count of schedules anymore outside of the <ApiPagination> component (let:visibleItems are schedules, so only the children of ApiPagination know about the count). And that Create Schedule button is outside of it. So we could refactor a bunch of things to make it work, but it feels like it's not worth it and would add a lot of complexity/slots to ApiPagination.

Alex-Tideman: Or we can remove the button in the EmptyState

laurakwhit: Ah makes sense. I think that was added as part of a design request (to only show one on the page) but I can't find verification of that. We could also wrap that button in an additional `{#await fetchSchedules}` and check  `{#if schedules?.length}`.

src/lib/pages/schedules.svelte
```diff
@@ -41,38 +41,38 @@
...
-      nextButtonLabel={translate('common.next')}
-    >
-      <svelte:fragment slot="action-top-left">
-        <div class="w-full xl:w-1/2">
+{#key namespace}
+  <ApiPagination
+    let:visibleItems
+    {onFetch}
+    {onError}
+    aria-label="schedules"
```
laurakwhit, line 49:
```suggestion
    aria-label={translate('common.schedules')}
```

src/lib/pages/schedules.svelte
```diff
@@ -41,38 +41,38 @@
...
+    pageSizeSelectLabel={translate('common.per-page')}
+    nextButtonLabel={translate('common.next')}
+    previousButtonLabel={translate('common.previous')}
+    emptyStateMessage={translate('schedules.empty-state-title')}
+    fallbackErrorMessage={'Error fetching schedules'}
```
laurakwhit, line 54: Looks like we use "Error fetching schedules" a few places, maybe worth defining in i18n?

src/lib/pages/schedules.svelte
```diff
@@ -1,37 +1,37 @@
...
+      err?.body?.message ??
+      `Error fetching schedules: ${err.status}: ${err.statusText}`);
 
   let coreUser = coreUserStore();
   $: createDisabled = $coreUser.namespaceWriteDisabled(namespace);
 
+  $: onFetch = () => fetchPaginatedSchedules(namespace);
+
   let search = '';
   $: filteredSchedules = (schedules: ScheduleListEntry[]) =>
```
laurakwhit, line 36: Just want to clarify that this is expected behavior. When showing filtered schedules the number of items doesn't change (e.g. only two items are shown, but pagination still says 1-3) ⬇️ 
 
Screenshot 2024-01-04 at 12 53 05 PM

laurakwhit (approved):
One clarification and two small suggestions, but otherwise lgmt 🚀

src/lib/pages/schedules.svelte
Alex-Tideman: 👍

src/lib/pages/schedules.svelte
Alex-Tideman: 👍

src/lib/pages/schedules.svelte
Alex-Tideman: We don't get a total from the api. 

For the total when filtering... totally agree it's not ideal but I think for now we'll live with it. There's a whole lot of edge cases if we update the pagination store with a filter.

laurakwhit: Yeah, I figured it would be a can of worms. I think it's fine for now 👍

PR #1785, opened 2024-01-04
DT-1654 - add scheduled by id and scheduled start time to workflow table columns

rossedfort:
## Description & motivation 💭 
Adds the `TemporalScheduledById` and `TemporalScheduledStartTime` internal search attributes to the list of available columns for the workflows summary table. Note - we should eventually refactor this table to read the internal and custom search attributes from the API and generate a list of available columns.

Screenshot 2024-01-04 at 11 44 42 AM
Screenshot 2024-01-04 at 11 45 00 AM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/workflow/workflows-summary-configurable-table/table-body-cell.svelte
+++ b/src/lib/components/workflow/workflows-summary-configurable-table/table-body-cell.svelte
@@ -102,6 +102,14 @@
       })}
     {:else if label === 'History Length'}
       {parseInt(workflow.historyEvents, 10) > 0 ? workflow.historyEvents : ''}
+    {:else if label === 'Scheduled By ID'}
+      {workflow.searchAttributes?.indexedFields?.TemporalScheduledById ?? ''}
+    {:else if label === 'Scheduled Start Time'}
+      {@const content =
+        workflow.searchAttributes?.indexedFields?.TemporalScheduledStartTime}
+      {content && typeof content === 'string'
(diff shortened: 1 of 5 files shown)


src/lib/components/workflow/workflows-summary-configurable-table/table-body-cell.svelte
```diff
@@ -102,6 +102,14 @@
       })}
     {:else if label === 'History Length'}
       {parseInt(workflow.historyEvents, 10) > 0 ? workflow.historyEvents : ''}
+    {:else if label === 'Scheduled By ID'}
+      {workflow.searchAttributes?.indexedFields?.TemporalScheduledById ?? ''}
+    {:else if label === 'Scheduled Start Time'}
+      {@const content =
+        workflow.searchAttributes?.indexedFields?.TemporalScheduledStartTime}
+      {content && typeof content === 'string'
+        ? formatDate(content, $timeFormat, { relative: $relativeTime })
```
Alex-Tideman, line 111: Are these decoded already? I know they don't need to go to the codec server but they are payloads so need to be base64 decoded

rossedfort: Yep, they are decoded when we get them back from the server: https://github.com/temporalio/ui/blob/main/src/lib/models/workflow-execution.ts#L39

Alex-Tideman (approved):

Issue #1594, opened 2023-08-28
Click Continue to SSO button twice after logout or session timeout


michaeldop:
**Describe the bug**
It appears that when the `returnUrl` is anything but the baseUrl, for example `https://temporal-ui.domain.com/login?returnUrl=https%3A%2F%2Ftemporal-ui.domain.com%2Fnamespaces%2Fdefault%2Fworkflows`, Temporal UI requires you to click on the "Continue to SSO" button twice.

**To Reproduce**

1. Go to temporal login page and click on 'Continue to SSO'
2. See that you are successfully logged in and are on the workflows page
3. Click the top left corner drop down and 'Log out'
4. Click on 'Continue to SSO'
5. See an unauthorized error (see screenshots below)
6. Click on 'Continue to SSO' again 
7. See that you are succesfully logged in

**Expected behavior**
I would expect to only have to click on the "Continue to SSO" button once and after successful authentication go whatever `redirectUrl` was set to

**Screenshots**
First time clicking "Continue to SSO" redirects back to this page with "unauthorized"
![Screenshot 2023-08-25 at 1 52 01 PM](https://github.com/temporalio/ui/assets/64210583/fcac252b-c429-4f83-a4ab-c7697cb59319)

There is a 401 error being thrown here after clicking on 'Continue to SSO'
![Screenshot 2023-08-28 at 9 52 01 AM](https://github.com/temporalio/ui/assets/64210583/268d5642-7016-4e7c-9e02-21645d0942bc)

**Desktop (please complete the following information):**

- OS: Mac
- Browser: Chrome
- Version: 116

**Additional context**
 - Temporal UI 2.18.0

joshbranham:
We are also noticing this, as well as the `returnUrl` being perpetually appended to the URL when you click the Temporal logo in the top left of the UI on the login screen.

MB175:
This issue is caused by the search-attributes endpoint, which is called soon before the auth token is set as a header parameter.

`/api/v1/namespaces/default/search-attributes?`

Since the UI is receiving a 401 Statuscode it falls back to the login page. To prove that the search-attributes endpoint is the issue, it can simply be blocked out in Chrome, and the login process works flawlessly.

![Bildschirmfoto 2024-01-04 um 11 06 18](https://github.com/temporalio/ui/assets/57674787/71197d47-d67b-4403-809d-c8f5c5fe86cf)

MB175:
[...]

PR #1787, opened 2024-01-08
Move search-attributes call to prevent SSO loop

MB175:
## Description & motivation 💭 

In this fix, I tackled an issue where a component was triggering a request for search attributes before authentication, causing a 401 error and looping back to the login mask. To address this, I moved the request for loading search attributes into the component responsible for loading workflows.

Now, the request for search attributes is part of the workflow loading process, ensuring it happens after authentication. This adjustment prevents the 401 error and eliminates the loop in the user interface. It's a straightforward modification.

Further description of the problem can be found in the issue #1594 

### How was this tested 👻 

(no new test have been added)

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

- Enable SSO authentication and notice the login loop is gone

### Issue(s) closed 

#1594 

### Any docs updates needed? 

No


Alex-Tideman:
Thanks @MB175! We don't want to move to workflows store so that call isn't included in our NPM package, but I've solved this with https://github.com/temporalio/ui/pull/1794. Going to close this PR.

PR #1788, opened 2024-01-09
Checkboxiis are now laid out with flex instead of only using absolute…

GiantRobots:
…. Hover element is a pseudo element now and seems to be generally centered. Removed some of the extra fluff/padding that hover-able introduced around elements. Made it so text can wrap on multiple lines and checkbox stays correct. Used sr-only for input.

Updates checkbox styles so we can properly do the side modal for global namespace design (having multiple lines for a checkbox label).

https://www.figma.com/file/8dSoDU2sWeQf7CgZzTqzeW/Cloud-Account-Management?type=design&node-id=6203-8475&mode=design&t=ika7oA0QQw4FkYMO-0

https://holocene-git-fixcheckboxlayout.preview.thundergun.io/?path=%2Fstory%2Fcheckbox--so-many

Added story to see edge cases for many checkboxii

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/holocene/checkbox.svelte
+++ b/src/lib/holocene/checkbox.svelte
@@ -84,15 +84,17 @@
       {...omit($$restProps, 'data-testid')}
     />
 
-    <div class="checkbox-hover-wrapper">
-      <span class="checkmark" class:hoverable class:on-dark={onDark}>
-        {#if indeterminate}
-          <Icon class="h-full w-full" name="hyphen" />
-        {:else if checked}
-          <Icon class="h-full w-full" name="checkmark" strokeWidth={1} />
-        {/if}
-      </span>
-    </div>
+    <span class="checkmark" class:hoverable class:on-dark={onDark}>
+      {#if indeterminate}
+        <Icon class="absolute top-0 left-0 h-4 w-4" name="hyphen" />
+      {:else if checked}
+        <Icon
+          class="absolute top-0 left-0 h-4 w-4"
+          name="checkmark"
+          strokeWidth={3}
+        />
+      {/if}
+    </span>
 
     <span class="label" class:hoverable class:sr-only={labelHidden}>
       {label}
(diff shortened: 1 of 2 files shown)


src/lib/holocene/checkbox.svelte
```diff
@@ -86,51 +99,34 @@
...
 <style lang="postcss">
   .checkbox {
-    @apply block h-[18px] w-[18px] cursor-pointer select-none text-sm leading-[18px] text-primary;
+    @apply flex cursor-pointer select-none items-center gap-3 text-sm leading-[18px] text-primary;
   }
 
-  .checkbox.hoverable {
-    @apply h-9 w-9 rounded-full hover:bg-purple-200;
+  .checkbox.hoverable:hover .checkmark::before {
+    @apply absolute -left-[.65rem] -z-10 h-9 w-9 self-center rounded-full bg-purple-200;
```
rossedfort, line 111: `-left-[.65rem]` is heinous, is there a better way we can position it? Also, the hover state doesn't seem to be working for all checkboxes in the story

https://github.com/temporalio/ui/assets/11775628/158e575f-28fc-46f7-bca4-b85c1b9cf794

src/lib/holocene/checkbox.svelte
```diff
@@ -86,51 +99,34 @@
...
-        />
-      {/if}
-    </span>
   </label>
 </div>
 
 <style lang="postcss">
   .checkbox {
-    @apply block h-[18px] w-[18px] cursor-pointer select-none text-sm leading-[18px] text-primary;
+    @apply flex cursor-pointer select-none items-center gap-3 text-sm leading-[18px] text-primary;
```
rossedfort, line 107: On the designs, it looks like this should be `items-start` ?
Screenshot 2024-01-09 at 2 14 09 PM

src/lib/holocene/checkbox.svelte
GiantRobots: > -left-[.65rem] is heinous, is there a better way we can position it?

It's so heinous.

But not that I can think of we were doing that previously by using the transform to move it to the center of the object which also changed the size of the checkbox which is just differently heinous. There's probably something we can do with `calc` but I'm too dumb at css to figure that one out. Also since it uses REM it is appropriately sized when the page increases or decreases in size through `CMD + +` (mostly sometimes it's a pixel or so off center). 

Hover state is applied to any of the elements that have hoverable (every other one) or if you update the story component to use hoverable true/false it will update all of them. I had hoverable and non hoverable next to eachother to make sure they held the same dimensions.

rossedfort (approved):

Alex-Tideman (approved):

Issue #1786, opened 2024-01-05
Feature Request: Timeline view allows navigation to child workflows


edclement:
## Is your feature request related to a problem? Please describe.

The timeline view is great. However, clicking on a child workflow from the timeline view does nothing. When there are lots of child workflows named the same, it's difficult to determine which child workflow id lines up with the visual representation in the timeline view.

## Describe the solution you'd like

Clicking on a child workflow in the timeline view opens that workflow in a new tab.

## Describe alternatives you've considered

Using the "Relationships" > "Children" table to select a child workflow. But you loose the context of when that workflow started/stopped and how long it took in relation to other child workflows.

## Additional context

None.

Alex-Tideman:
Added in new Timeline view (can also view Timeline of child directly without navigating)

PR #1790, opened 2024-01-11
Timeline UI improvements

laurakwhit:
## Description & motivation 💭 
*  Adds a `history/event-groups/:id` route to show that specific event group in the event history table with the timeline above it. A user can navigate to it by clicking on an event group on the timeline.
   * **Note**: A `text` prop has been added to our `Link` Holocene component so it could be used with `renderComponentToHTM`.
* `[WIP]`: Updates colors to reflect current colors for workflow statuses and adds icons based on event classification for Running, Completed, Canceled, Terminated, Failed, and TimedOut. Also removes rounded border on right for Running events.
* Fixes the `Cannot read properties of null (reading 'left')` error thrown by vis-timeline when auto-refresh was on and a user was scrolling in the timeline.

| Event group page |
|-|
|Screenshot 2024-01-10 at 4 16 54 PM|

| Running | Completed | Canceled | 
|--|--|--|
|Screenshot 2024-01-10 at 4 15 46 PM|Screenshot 2024-01-10 at 4 16 13 PM| Screenshot 2024-01-10 at 4 44 45 PM|

|Terminated | Failed | TimedOut |
|--|--|--|
|Screenshot 2024-01-10 at 4 43 43 PM |Screenshot 2024-01-10 at 4 51 52 PM |Screenshot 2024-01-10 at 4 33 51 PM|
### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 

`DT-1545`

### DIFF
--- a/src/lib/components/event/event-history-timeline.svelte
+++ b/src/lib/components/event/event-history-timeline.svelte
@@ -4,7 +4,9 @@
 
   import { page } from '$app/stores';
 
+  import HeartBeat from '$lib/components/heart-beat-indicator.svelte';
   import Icon from '$lib/holocene/icon/icon.svelte';
+  import type { IconName } from '$lib/holocene/icon/paths';
   import Link from '$lib/holocene/link.svelte';
   import Loading from '$lib/holocene/loading.svelte';
   import ToggleButton from '$lib/holocene/toggle-button/toggle-button.svelte';
@@ -20,6 +22,7 @@
   } from '$lib/stores/workflow-run';
   import type {
     CommonHistoryEvent,
+    EventClassification,
     EventTypeCategory,
   } from '$lib/types/events';
   import { capitalize } from '$lib/utilities/format-camel-case';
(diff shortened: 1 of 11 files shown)


Alex-Tideman:
Should we add the heartbeat icon for Running?

Alex-Tideman:
The one thing that I noticed is if it's a Running workflow and you have auto-refresh on, it keeps shifting the spans to the right and the 'Fit' button doesn't work as expected.

Alex-Tideman:
What do you think about making all the rows this same shorter height? (the height of the row when there's no color span)

Screen Shot 2024-01-11 at 8 43 22 AM

Alex-Tideman:
What do you think of linking to the child workflow in the title here?

Screen Shot 2024-01-11 at 9 17 23 AM

This would help with this feature request: https://github.com/temporalio/ui/issues/1786

src/lib/components/event/event-history-timeline.svelte
```diff
@@ -207,9 +262,15 @@
     visualizationRef;
 
   const drawTimeline = () => {
-    if (timeline) {
-      timeline.destroy();
```
laurakwhit, line 211: > The one thing that I noticed is if it's a Running workflow and you have auto-refresh on, it keeps shifting the spans to the right and the 'Fit' button doesn't work as expected.

@Alex-Tideman we may need that `timeline.destroy()` after all... There's some wonkiness with sizing going on with auto-refresh and `timeline.redraw()` doesn't seem to fix it 😞 

> What do you think about making all the rows this same shorter height? (the height of the row when there's no color span)

There is a "fixed" `groupHeightMode` option, unfortunately, it appears not to be working (looks like there is already an open issue for it). Another reason to maybe make our own timeline 😅 

> What do you think of linking to the child workflow in the title here? 

On it! Might need to fix some edge cases with navigation though (e.g. hitting the back button after navigating from an event group to child).

Alex-Tideman (approved):

PR #1795, opened 2024-01-12
Show Activity Type for local activities in all SDKs

Alex-Tideman:
## Description & motivation 💭 
When we moved to decoding payloads on demand, we lost the ability to show LocalActivity activityType on the event summary row. This adds back that functionality by decoding the LocalActivity payloads and finding the correct value for ActivityType based on SDK.

**AFTER**
Screen Shot 2024-01-22 at 1 42 40 PM

**BEFORE**
Screen Shot 2024-01-12 at 9 25 05 AM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added


src/lib/utilities/get-single-attribute-for-event.ts
```diff
@@ -208,20 +215,20 @@ const getSummaryAttribute = (event: WorkflowEvent): SummaryAttribute => {
   return first;
 };
 
-export const getSummaryForEventGroup = ({
+export const getSummaryForEventGroup = async ({
   lastEvent,
-}: EventGroup): SummaryAttribute => {
-  return getSummaryAttribute(lastEvent);
+}: EventGroup): Promise<SummaryAttribute> => {
+  return await getSummaryAttribute(lastEvent);
```
laurakwhit, line 221: Is the addition of async/await here needed? `getSummaryAttribute` already returns `Promise<SummaryAttribute>`.

src/lib/utilities/get-single-attribute-for-event.ts
```diff
@@ -208,20 +215,20 @@ const getSummaryAttribute = (event: WorkflowEvent): SummaryAttribute => {
...
-): SummaryAttribute => {
+): Promise<SummaryAttribute> => {
   if (!event) return emptyAttribute;
 
   if (isEventGroup(event)) {
     return getSummaryForEventGroup(event);
   }
 
-  return getSummaryAttribute(event);
+  return await getSummaryAttribute(event);
```
laurakwhit, line 233: Similar comment as above ⬆️

src/lib/utilities/get-single-attribute-for-event.ts
Alex-Tideman: You are absolutely correct

laurakwhit: Think we can remove the async [here](https://github.com/temporalio/ui/pull/1795/commits/910dc77d274b840979e8e393cc19c2f65c54ca2d#diff-42a5d42a3ea1e209202949d938fa2ca7271443c0279d04c2991e2cf844f8e0c5R218) too.

src/lib/holocene/select/multi-select.svelte
```diff
@@ -38,7 +38,12 @@
     } else {
       selectedOptions = [...selectedOptions, option];
     }
-    onChange(selectedOptions);
+
+    if (selectedOptions.length === options.length) {
+      onChange([]);
```
laurakwhit, line 43: Wondering if we want this conditional here or in the `onChange`. Might there be an instance were we want all the options to be passed to the `onChange` when they are all selected?

Alex-Tideman: You are absolutely right, I'll move it to the onChange. Good catch

laurakwhit (approved):

PR #1803, opened 2024-01-18
Add back single banner

Alex-Tideman:
## Description & motivation 💭 
Per request of @antmendoza, add back the banner but simplified. It does not check for versions, it only checks for the TEMPORAL_BANNER_TEXT env var and if it exists, show it. No links. Once closed, the banner will not show up again (until localStorage is cleared).

Screen Shot 2024-01-18 at 9 35 55 AM

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/server/docker/config-template.yaml
+++ b/server/docker/config-template.yaml
@@ -2,6 +2,7 @@ temporalGrpcAddress: {{ default .Env.TEMPORAL_ADDRESS "127.0.0.1:7233" }}
 port: {{ default .Env.TEMPORAL_UI_PORT "8080" }}
 publicPath: {{ default .Env.TEMPORAL_UI_PUBLIC_PATH "" }}
 enableUi: {{ default .Env.TEMPORAL_UI_ENABLED "true" }}
+bannerText: {{ default .Env.TEMPORAL_BANNER_TEXT "" }}
 cloudUi: {{ default .Env.TEMPORAL_CLOUD_UI "false" }}
 defaultNamespace: {{ default .Env.TEMPORAL_DEFAULT_NAMESPACE "default" }}
 feedbackUrl: {{ default .Env.TEMPORAL_FEEDBACK_URL "" }}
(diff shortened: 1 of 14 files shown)


src/lib/holocene/banner/banner.svelte
```diff
@@ -0,0 +1,28 @@
...
+
+{#if show}
+  <section class="banner" {...$$restProps}>
+    {message}
+    <IconButton
+      on:click={() => ($bannerClosed = true)}
+      data-testid="close-banner"
+      class="text-black-600 absolute top-0 right-0 mr-5 h-10 w-10"
+      icon="close"
+      label={translate('common.close')}
```
rossedfort, line 19: Can we pass this label as a prop? I don't want Holocene to depend on i18next

src/lib/stores/banner.ts
rossedfort, line 1: This can probably just live alongside the banner component in holocene

server/docker/config-template.yaml
```diff
@@ -2,6 +2,7 @@ temporalGrpcAddress: {{ default .Env.TEMPORAL_ADDRESS "127.0.0.1:7233" }}
 port: {{ default .Env.TEMPORAL_UI_PORT "8080" }}
 publicPath: {{ default .Env.TEMPORAL_UI_PUBLIC_PATH "" }}
 enableUi: {{ default .Env.TEMPORAL_UI_ENABLED "true" }}
+bannerText: {{ default .Env.TEMPORAL_BANNER_TEXT "" }}
```
rossedfort, line 5: Should we make this name more specific?

Alex-Tideman: Got any suggestions?

src/lib/holocene/banner/banner.svelte
Alex-Tideman: 👍

server/docker/config-template.yaml
rossedfort: Meh it's probably fine

rossedfort (approved):

PR #1806, opened 2024-01-18
MultiSelect component used in Event Type filter

Alex-Tideman:
## Description & motivation 💭 
To support multi-select filtering of event types in the event history, such as filtering out signals and showing the rest.

You can also Select All or Clear All.

This changes the $eventCategoryFilter from a string to string[].

https://github.com/temporalio/ui/assets/7967403/9c17cf7e-fe84-4803-898b-33032573317a

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/holocene/select/multi-select.svelte
```diff
@@ -0,0 +1,103 @@
...
+          onOptionClick(option);
+        }}
+      >
+        <Checkbox
+          on:click={() => onOptionClick(option)}
+          slot="leading"
+          {checked}
+          label={option.label}
+          labelHidden
+        />
```
rossedfort, line 85: check out `bind:group` 👀

src/lib/holocene/select/multi-select.svelte
```diff
@@ -0,0 +1,103 @@
...
+  export let clearAllLabel: string;
+
+  let selectedOptions = initialSelected.length ? initialSelected : options;
+
+  const open = writable(false);
+
+  const resetFilter = () => {};
+
+  const onOptionClick = (option: Option) => {
+    if (selectedOptions.find((s) => s.value === option.value)) {
```
rossedfort, line 37: `.some()` would be slightly more correct here I think

src/lib/holocene/select/multi-select.svelte
```diff
@@ -0,0 +1,103 @@
...
+  export let variant: MenuButtonVariant = 'secondary';
+  export let icon: IconName | undefined = undefined;
+  export let selectAllLabel: string;
+  export let clearAllLabel: string;
+
+  let selectedOptions = initialSelected.length ? initialSelected : options;
+
+  const open = writable(false);
+
+  const resetFilter = () => {};
```
rossedfort, line 34: is this used? If so, could use `noop`

rossedfort (commented):
Looks pretty good so far, just some small suggestions

src/lib/holocene/select/multi-select.svelte
```diff
@@ -0,0 +1,103 @@
...
+      selectedOptions = selectedOptions.filter((s) => s.value !== option.value);
+    } else {
+      selectedOptions = [...selectedOptions, option];
+    }
+    onChange(selectedOptions);
+  };
+
+  const onSelectAllOptionClick = () => {
+    selectedOptions = options;
+    onChange([]);
```
rossedfort, line 47: Don't we want `onChange(selectedOptions)` ?

src/lib/holocene/select/multi-select.svelte
stevekinney: Good catch.

src/lib/holocene/select/multi-select.svelte
Alex-Tideman: We could but then it would be ?category=marker,command,timer....

By using an empty array, doesn't have any category which means it removes that query param altogether and so it's just /history which includes all types

rossedfort: Ahh ok makes sense

src/lib/holocene/select/multi-select.svelte
Alex-Tideman: Normally would agree, but we need to call onChange as well

PR #1819, opened 2024-01-23
Use goto in Link

Alex-Tideman:
## Description & motivation 💭 
For some mystical, unexplained reason Sveltekit (or us) does a full page reload on certain links when using `<a href>`. To fix this, I use goto on the on:click in Link. Now all Links should be simpler and not accept an on:click, just an href.

Other things of note:

- Split up Button component to use Link or button so we get the same behavior for button links.
- Remove all newTab props for app links. I think this should be the default behavior. If they want to open in new tab, they right click. All external tabs open in new tab.
- Remove the old workflows page. It's been over 3 major server versions, we can deprecate. This cleans up some old uses of links.

### How was this tested 👻 

- [ ] Manual testing
- [x] E2E tests added
- [x] Unit tests added


rossedfort (approved):
Haven't tested anything but code looks good

PR #1821, opened 2024-01-24
Workflow Error Page

Alex-Tideman:
## Description & motivation 💭 
Previously all we did was show a spinner and throw a toast notification for 404/other errors. This was very difficult to read and didn't last long. It also led to confusion on if the workflow was still loading.

This is also better than our default Error component because it allows the breadcrumb Back to Workflows and gives more humanized errors.

I added custom onError handling for fetchWorkflow and a new WorkflowError component to handle the error.

Screen Shot 2024-01-24 at 1 07 17 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


rossedfort (approved):
Nice improvement, I'll let design weigh in but my only suggestion would maybe be make the error text smaller and omit the logo, or move the logo above the text

src/lib/components/workflow/workflow-error.svelte
```diff
@@ -0,0 +1,46 @@
...
+      icon="chevron-left"
+    >
+      {translate('workflows.back-to-workflows')}
+    </Link>
+  </div>
+  <h1
+    data-testid="workflow-error-heading"
+    class="overflow-hidden text-3xl font-medium"
+  >
+    {error.statusText}
```
laurakwhit, line 37: What do you think about making this a little more similar to the existing `Error` page and have the status text centered with the other info. ⬇️ 

Screenshot 2024-01-24 at 11 32 11 AM

src/lib/services/workflow-service.ts
```diff
@@ -206,17 +207,28 @@ export const fetchAllArchivedWorkflows = async (
 export async function fetchWorkflow(
   parameters: GetWorkflowExecutionRequest,
   request = fetch,
-): Promise<WorkflowExecution> {
+): Promise<{
+  workflow: WorkflowExecution | undefined;
```
laurakwhit, line 211: Noticed that workflow or error doesn't actually show up as possibly `undefined` ⬇️  Wondering if there's a better way to type this so it does...

Screenshot 2024-01-24 at 11 34 35 AM

Also wondering if we shouldn't just return the object without the `undefined` property (e.g. `{ error: e }`.

laurakwhit (approved):
Some small comments r.e. types and design, but otherwise lgtm

src/lib/components/workflow/workflow-error.svelte
Alex-Tideman: 👍 updated

src/lib/services/workflow-service.ts
Alex-Tideman: Good call

PR #1833, opened 2024-01-29
Drawer fix

GiantRobots:
### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/event/event-shortcut-keys.svelte
+++ b/src/lib/components/event/event-shortcut-keys.svelte
@@ -1,5 +1,6 @@
 <script lang="ts">
   import Button from '$lib/holocene/button.svelte';
+  import DrawerContent from '$lib/holocene/drawer-content.svelte';
   import Drawer from '$lib/holocene/drawer.svelte';
   import Shortcut from '$lib/holocene/keyboard-shortcut/shortcut.svelte';
   import { translate } from '$lib/i18n/translate';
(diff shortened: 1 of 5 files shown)


src/lib/components/event/event-shortcut-keys.svelte
```diff
@@ -20,49 +21,51 @@
...
-      {translate('common.next-page')}
-      <div class="shortcut">
-        <Shortcut
-          arrow="right"
-          title={translate('common.arrow-key-right')}
-        />|<Shortcut title="">L</Shortcut>
+  <DrawerContent title={translate('common.keyboard-shortcuts')}>
+    <div
+      class="grid grid-cols-1 gap-8 text-gray-500 dark:text-gray-400 md:grid-cols-2 xl:grid-cols-4"
+    >
```
rossedfort, line 30: Maybe `<DrawerContent />` should take a class and we can remove this extra `<div />` ?

rossedfort (approved):

src/lib/holocene/drawer.svelte
```diff
@@ -1,57 +1,55 @@
...
   import IconButton from './icon-button.svelte';
 
   export let open = false;
-  export let title: string;
   export let position: 'bottom' | 'right' = 'bottom';
   export let dark = true;
   export let onClick: () => void;
   export let id = 'navigation-drawer';
   export let closeButtonLabel: string;
+  export let closePadding: boolean = true;
```
rossedfort, line 17: Is there a way to apply the padding based on an existing prop, like `position`?

src/lib/components/event/event-shortcut-keys.svelte
GiantRobots: Hmm that would kind of be confusing I think. 

Because it could target the title or the content, this just targets the content and essentially maintains the previous API. 

Could also switch title to be it's own slot and then it would get the same semantics but I was lazy and didn't want to make this super flexible/correct because it would induce change in the calling components and also we don't use it all over and didn't want to establish a different paradigm.

Happy to switch it to `<DrawerContent /> and <DrawerTitle />`

src/lib/holocene/drawer.svelte
GiantRobots: I couldn't think of a way to do that well without losing intent.

The issue is we need to be able to address the entire area of the drawer, and the padding we give content is the opposite of that.

Screenshot 2024-01-30 at 9 55 03 AM

The other ways to accomplish that is to give more access to the aside which felt worse as well so I settled on that prop to control it and default it because most times you won't need it.

PR #1843, opened 2024-01-31
Add trailing-slash utility and use on codec server endpoint

Alex-Tideman:
## Description & motivation 💭 
If a user enters `https://something-of-a-url.com/` for their codec, the decode endpoint would incorrectly be `https://something-of-a-url.com//decode`. This trims it for that case.

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

### DIFF
--- a/src/lib/utilities/trim-trailing-slash.ts
+++ b/src/lib/utilities/trim-trailing-slash.ts
@@ -1,0 +1,3 @@
+export const trimTrailingSlash = (x: string): string => {
+  return x.endsWith('/') ? x.slice(0, -1) : x;
+};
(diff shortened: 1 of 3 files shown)


laurakwhit (approved):

src/lib/utilities/trim-trailing-slash.ts
```diff
@@ -0,0 +1,3 @@
+export const trimTrailingSlash = (x: string): string => {
+  return x.endsWith('/') ? x.slice(0, -1) : x;
```
laurakwhit, line 2: We could also do something like this to trim all trailing slashes (e.g. for something like `http://cats.meow//`)
```suggestion
  return x.replace(/\/+$/, "")

PR #1845, opened 2024-01-31
Refactor select-namespace page

stevekinney:
- Use CSS selectors to round first and last elements
- Reduce DOM elements and logic

### DIFF
--- a/src/routes/(app)/select-namespace/+page.svelte
+++ b/src/routes/(app)/select-namespace/+page.svelte
@@ -16,32 +16,27 @@
 
   import type { DescribeNamespaceResponse as Namespace } from '$types';
 
-  $: namespaceNames = $namespaces.map(
-    (namespace: Namespace) => namespace?.namespaceInfo?.name,
-  );
-
-  $: namespaceList = namespaceNames.map((namespace: string) => {
-    return {
-      namespace,
-      onClick: async (namespace: string) => {
-        const { authorized } = await fetchWorkflowForAuthorization(namespace);
-        if (authorized) {
-          $lastUsedNamespace = namespace;
-          goto(routeForWorkflows({ namespace }));
-        } else {
-          toaster.push({
-            variant: 'error',
-            message: translate('namespaces.unauthorized-namespace-error'),
-          });
-        }
-      },
-    };
-  });
-
   let searchValue = '';
-  $: filteredList = namespaceList.filter(({ namespace }) =>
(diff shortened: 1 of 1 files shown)


src/routes/(app)/select-namespace/+page.svelte
```diff
@@ -64,42 +59,28 @@
...
-            >
-              {item.namespace}
-            </button>
-          </div>
-        </VirtualList>
-      {:else}
-        <EmptyState title={translate('common.no-results')} />
-      {/if}
+  <div
+    class="group h-screen w-full"
```
laurakwhit, line 63: This list actually goes off screen with `h-screen`. Wondering if we want to add a `calc()` here to set the max height so it does appear all on screen.

src/routes/(app)/select-namespace/+page.svelte
```diff
@@ -16,32 +16,27 @@
...
-  let searchValue = '';
-  $: filteredList = namespaceList.filter(({ namespace }) =>
-    namespace.includes(searchValue),
-  );
+  $: filteredList = $namespaces
+    .map((namespace: Namespace, index: number) => ({
+      namespace: namespace?.namespaceInfo?.name,
+      index,
+    }))
+    .filter(({ namespace }) => namespace.includes(searchValue));
```
laurakwhit, line 39: Is separating these two potentially a little more performant since we wouldn't be iterating over the list twice on search?

```suggestion
  $: namespaceList = $namespaces.map((namespace: Namespace, index: number) => ({
    namespace: namespace?.namespaceInfo?.name,
    index,
  }));
  $: filteredList = namespaceList.filter(({ namespace }) =>
    namespace.includes(searchValue),
  );
```

laurakwhit (approved):
Nice little improvement. Left some thoughts, but it's fine as is 🚀

src/routes/(app)/select-namespace/+page.svelte
stevekinney: I really wish the `VirtualList` would give me the index and then I wouldn't have to even map over the namespaces at all.

src/routes/(app)/select-namespace/+page.svelte
stevekinney: This is why we never used the virtual list on the event history. I struggled with this for a long time in 2021. 🤣 I might play around with it tomorrow to see what I can do.

PR #1846, opened 2024-02-01
Simplify side navigation; add dark mode support

stevekinney:
- Remove `isCloud` prop drilling and rely on CSS instead.
- Move properties into navigation items
- Remove commented-out Labs code (for now)
- Add dark mode support
- Adjust borders for side and top navigation


src/lib/holocene/navigation/navigation-container.svelte
```diff
@@ -0,0 +1,50 @@
...
+    <Icon name="chevron-right" />
+  </button>
+  <div role="list">
+    <slot name="top" />
+    <hr class="my-8 border-subtle" />
+    <slot name="middle" />
+  </div>
+  <div class="space-y-2 self-center">
+    <slot name="bottom" />
+  </div>
```
Alex-Tideman, line 45: Just my personal preference, but I'd but the bottom slot down in the div below, above the version. It's a little weird floating where it is, I think at the bottom is probably better?

src/lib/holocene/navigation/navigation-item.svelte
```diff
@@ -0,0 +1,34 @@
...
+  export let label: string;
+  export let icon: IconName;
+  export let tooltip = label;
+
+  const attributes = link.startsWith('http')
+    ? {
+        rel: 'noreferrer',
+        target: '_blank',
+      }
+    : {};
```
Alex-Tideman, line 15: This is going to give us some trouble in Cloud. All the AppRoutes are full urls so this will always open cloud nav items in a new tab

stevekinney: Ugh. Good catch. Do you think we should maybe be more specific? Maybe it's not worth it since we _only_ use this for the feedback link anyway. Also: I hate this code anyway.

Alex-Tideman: Might just need an external prop to set this. Feedback/support and potentially status + any future nav item item that links externally

src/lib/holocene/navigation/navigation-container.svelte
stevekinney: It _really_ doesn't like that.

Alex-Tideman (approved):

PR #1851, opened 2024-02-02
Dynamically import icons

stevekinney:
Adding each icon to a file is tedious. Let's just fetch them dynamically!

This change also adds `pnpm generate:icons` to generate a union type of all available icons.


src/lib/holocene/icon/icon-names.ts
rossedfort, line 1: This is very cool

PR #1855, opened 2024-02-02
combobx/input/menu updates for self serve signup

rossedfort:
## Description & motivation 💭 
- add invalid state to combobox
- fix some colors for dark mode components
### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/holocene/combobox/combobox.svelte
```diff
@@ -77,7 +78,8 @@
   export let minSize = 0;
   export let maxSize = 120;
   export let theme: 'light' | 'dark' = 'light';
-  export let filterable = true;
+  export let error = '';
+  export let valid = true;
```
Alex-Tideman, line 82: Do we need both? Whats the use case for having valid false and no error or the other way around?

rossedfort: We definitely don't need both but I think I prefer the ergonomics of this approach to just passing `error`. The idea being when the value isn't valid, we may or may not want to show an error message.

`<Combobox valid={!value} error="This is the error" />`

`<Combobox error={value ? '' : 'This is the error'} />`

It keeps the props single responsibility so that `error` isn't controlling style _and_ rendering in the DOM.

Alex-Tideman: 👍 Then I would just use {#if error} and not {#if error && !valid} for the error message. Let error be the single responsibility of the error message

rossedfort: Tbh I think I like it as it is. If we only check `#if error` that means the parent needs to toggle the error to a falsey value when valid, like this:

```
<Combobox valid={!value} error={valid ? '' : 'This is the error'} />
```

Right now, it's much cleaner. `error` will only display when `valid` is false.
```
<Combobox valid={!value} error="This is the error" />
```

PR #1858, opened 2024-02-09
Stop UI server from using stdout/stderr inadvertently

cretz:
## Description & motivation 💭

Stop UI server from using stdout/stderr inadvertently


lorensr (approved):
Chad and I tested CLI server with this change

cretz:
Will wait for approval from frontend before merging...

Alex-Tideman (approved):

PR #1863, opened 2024-02-13
Move tailwind theme to lib to import in theme

Alex-Tideman:
## Description & motivation 💭 
Moved the theme directory into lib so it's a part of npm package. Made a new preset file that can be used in Cloud.

Made top-nav bottom border subtle.

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/theme/colors.ts
```diff
@@ -157,6 +157,10 @@ const colors = {
   offWhite: '#F8FAFC',
   black: '#141414',
   offBlack: '#0F172A',
+  spaceGray: '#141414',
```
stevekinney, line 160: What's the difference between `spaceGray` and `black`?

src/lib/theme/colors.ts
```diff
@@ -157,6 +157,10 @@ const colors = {
...
+  asteroidBeige: '#f8f8f7',
+  lightBlue: '#DFFBFF',
```
stevekinney, line 162: Should we just find the nearest `blue` and `purple` for these two? `purple-100` and `blue-100` seem dangerously close.

src/lib/theme/colors.ts
```diff
@@ -157,6 +157,10 @@ const colors = {
   offWhite: '#F8FAFC',
   black: '#141414',
   offBlack: '#0F172A',
+  spaceGray: '#141414',
+  asteroidBeige: '#f8f8f7',
```
stevekinney, line 161: Should we just make an entire range here like we did for the other colors?

stevekinney (approved):

PR #1866, opened 2024-02-14
Small tweaks to Holocene components for self serve signup

rossedfort:
## Description & motivation 💭 
- Support label slot in checkbox. This is so that links (or any other markup) can be rendered in the checkbox label
Screenshot 2024-02-14 at 10 32 15 AM

- Add close event to combobox. This is so that we have an event to catch when the menu is closed _without_ the user having selected a value (i.e. they open the list of options by focusing the combobox, but then click away). This is useful to add inline errors to the combobox when a value is required.
- Only dispatch close even from `<MenuContainer />` if menu is open

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/holocene/checkbox.svelte
+++ b/src/lib/holocene/checkbox.svelte
@@ -10,7 +10,7 @@
 
   interface $$Props extends HTMLInputAttributes {
     checked?: boolean;
-    label: string;
+    label?: string;
     labelHidden?: boolean;
     theme?: 'dark' | 'light';
     indeterminate?: boolean;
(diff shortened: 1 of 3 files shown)


Alex-Tideman (approved):

src/lib/holocene/checkbox.svelte
```diff
@@ -95,9 +95,11 @@
       {/if}
     </span>
 
-    <span class="label" class:sr-only={labelHidden}>
-      {label}
-    </span>
+    <slot name="label">
+      <span class="label" class:sr-only={labelHidden}>
```
laurakwhit, line 99: Could a checkbox potentially have both a label slot and a label prop? Wondering if we want to make this a conditional so that there's no potential for both.

PR #1871, opened 2024-02-15
Fix backdrop for Modal component

laurakwhit:
## Description & motivation 💭 

Due to [::backdrop](https://developer.mozilla.org/en-US/docs/Web/CSS/::backdrop) not being inherited
> ::backdrop neither inherits from nor is inherited by any other elements. No restrictions are made on what properties apply to this pseudo-element.

CSS variables (e.g. `bg-inverse`) won't work. This PR removes the unused background color and unnecessary opacity.

 | Before | After |
|---|---|
 |Screenshot 2024-02-15 at 3 12 14 PM|Screenshot 2024-02-15 at 4 15 51 PM|
 |Screenshot 2024-02-15 at 3 12 25 PM|Screenshot 2024-02-15 at 4 16 00 PM|

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/holocene/modal.stories.svelte
+++ b/src/lib/holocene/modal.stories.svelte
@@ -28,6 +28,7 @@
     confirmText: 'Confirm',
     cancelText: 'Cancel',
     confirmType: 'destructive',
+    open: false,
   }}
   let:args
 >
--- a/src/lib/holocene/modal.svelte
+++ b/src/lib/holocene/modal.svelte
@@ -131,11 +131,11 @@
 
 <style lang="postcss">
   .body {
-    @apply surface-primary  z-50 w-full max-w-lg overflow-y-auto rounded-lg p-0 text-primary shadow-xl md:h-max;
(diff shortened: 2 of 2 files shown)


src/lib/holocene/modal.stories.svelte
```diff
@@ -28,6 +28,7 @@
     confirmText: 'Confirm',
     cancelText: 'Cancel',
     confirmType: 'destructive',
+    open: true,
```
rossedfort, line 31: I figured out why this isn't working. In the modal component, we have `$: open, toggleModal();`, and in toggleModal, we call `modalElement?.show()` if `open` is true, however `modalElement` is not defined yet. We could change that reactive statement to `$: (open, modalElement), toggleModal();`, however it's not really a true use case to mount the modal already opened, so I don't think we really need to.

rossedfort (approved):

src/lib/holocene/modal.stories.svelte
laurakwhit: In that case maybe the better default is `false`. Will update!

PR #1884, opened 2024-02-27
Add support for WorkflowExecutionUpdateRequested event

Alex-Tideman:
### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/utilities/is-event-type.ts
```diff
@@ -116,6 +116,8 @@ export const eventTypes = [
   'UpsertWorkflowSearchAttributes',
   'WorkflowExecutionUpdateAccepted',
   'WorkflowExecutionUpdateCompleted',
+  'WorkflowExecutionUpdateRejected',
+  'WorkflowExecutionUpdateRequested',
```
laurakwhit, line 120: Do we want to add these to `all-event-types.json` and add tests to `get-group-id.test.ts` as well?

Alex-Tideman: Yup good call. Going to add those just as soon as I can figure out how to actually run a workflow with this new type (it's on a forked branch using a in-progress temporal server version).

dandavison:
Thanks very much for doing this work. I've tried testing this locally and the UI page loading the new event is showing errors, but there are a lot of moving parts and this may have been an error on my part. Sorry for holding this up but I think it's best if we wait until I get some feature branches merged first so that we can test this more sanely.

dandavison:
I've tested this (briefly, so far) against my server feature branch and can confirm that it is correctly displaying `WorkflowExecutionUpdateRequested` in the history list view and that there are no network fetch errors or JS errors in the console.

image

stevekinney (approved):

PR #1887, opened 2024-02-27
remove enabled from buttons

GraceGardner:
## Description & motivation 💭 
Removed enabled: because it doesn't work on <a>

## Testing 🧪 
Used Story book and checked in the UI but would like someone to double check my work please. 

### How was this tested 👻 

** Should we create some tests to make sure this doesn't happen again? **

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/holocene/button.svelte
+++ b/src/lib/holocene/button.svelte
@@ -33,13 +33,13 @@
       variants: {
         variant: {
           primary:
-            'bg-interactive border-interactive text-inverse hover:enabled:bg-interactive-hover hover:enabled:border-interactive-hover focus-visible:enabled:bg-interactive-hover focus-visible:enabled:border-white dark:focus-visible:enabled:border-black focus-visible:shadow-primary ',
+            'bg-interactive border-interactive text-inverse [&:not(:disabled):hover]:bg-interactive-hover hover:border-interactive-hover focus-visible:bg-interactive-hover focus-visible:border-white dark:focus-visible:border-black focus-visible:shadow-primary ',
           secondary:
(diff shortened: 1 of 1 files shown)


src/lib/holocene/button.svelte
```diff
@@ -100,24 +100,30 @@
...
+  export let icon: IconName = null;
   export let href: string = null;
   export let target: string = null;
   export let active = false;
 </script>
 
 {#if href && !disabled}
-  <Link
-    {target}
+  <a
```
Alex-Tideman, line 110: I think we'll want to copy the behavior of Link on the on:click instead of using native <a> behavior. An <a> does a hard refresh navigation (triggers all load functions) instead of doing client-side navigation where it pushes to history.

  const onLinkClick = (e: MouseEvent) => {
    // Skip if middle mouse click or new tab
    if (e.button === 1 || target || e.metaKey) return;
    e.preventDefault();
    e.stopPropagation();
    goto(href);
  };
</script>

<a
  {href}
  target={newTab ? '_blank' : null}
  rel={newTab ? 'noreferrer' : null}
  class={merge('link', icon ? 'inline-flex' : 'inline', className)}
  class:active
  on:click={onLinkClick}
  tabindex={href ? null : 0}
  {...$$restProps}
>

GraceGardner: Okay, cool! I added it in and then removed it when I was switching it over. I honestly wasn't sure what the difference was. Thanks for the explanation Alex! I'll add this in now :)

PR #1889, opened 2024-02-28
add tooltip to menu items

GraceGardner:
## Description & motivation 💭 
Adds Tooltip to disabled menu items 

## Testing 🧪 
Localhost and storybook

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

Check that tooltip appears for disabled menu items. 

either bottom positioning or right positioning seem to look the least bad, let me know if there's a better option. When I do bottomRight it looks really funky and I struggled to fix it.

### DIFF
--- a/src/lib/holocene/menu/menu-item.svelte
+++ b/src/lib/holocene/menu/menu-item.svelte
@@ -107,7 +107,7 @@
 </script>
 
 {#if href}
-  <Tooltip bottom text={disabledText} hide={!disabled}>
+  <Tooltip bottomRight text={disabledText} hide={!disabled}>
     <a
       {href}
       role="menuitem"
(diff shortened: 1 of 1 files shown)


Alex-Tideman (approved):

src/lib/holocene/menu/menu-item.svelte
```diff
@@ -37,6 +40,7 @@
   export let selected = undefined;
   export let destructive = false;
   export let disabled = false;
+  export let disabledText: string = 'unavailable';
```
laurakwhit, line 43: I would leave out the default text here (or set it to an empty string). I think with i18n we want to avoid having any text in Holocene components themselves.

laurakwhit: Also, maybe we want to make the `Tooltip` optional? If `disabledText` isn't assigned we could just not have the `Tooltip`.

PR #1893, opened 2024-02-28
Move reachability out of workflow run and to workflow workers page

Alex-Tideman:
## Description & motivation 💭 
Don't await the reachability api in workflow-run-layout since it's used only in the workers page. Await it there.

Tested with and without versioning.

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/pages/workflow-workers.svelte
+++ b/src/lib/pages/workflow-workers.svelte
@@ -1,13 +1,28 @@
 <script lang="ts">
-  import WorkersList from '$lib/components/workers-list.svelte';
-  import { workflowRun } from '$lib/stores/workflow-run';
+  import { page } from '$app/stores';
 
-  $: ({ workers, workflow, compatibility, reachability } = $workflowRun);
+  import WorkersList from '$lib/components/workers-list.svelte';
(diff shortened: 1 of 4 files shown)


src/lib/pages/workflow-workers.svelte
```diff
@@ -1,13 +1,28 @@
...
 
-<WorkersList
-  taskQueue={workflow?.taskQueue}
-  {workers}
-  {compatibility}
-  {reachability}
-/>
+{#if versioningEnabled}
+  {@const buildIds = getUniqueBuildIdsFromPollers(workers.pollers)}
+  {#await getWorkerTaskReachability( { namespace, buildIds, taskQueue }, ) then reachability}
```
laurakwhit, line 21: Do we want some kind of loading state here? Maybe a skeleton table?

laurakwhit (approved):
One small comment, but otherwise lgtm!

src/lib/pages/workflow-workers.svelte
Alex-Tideman: 👍 good call, adding it

PR #1904, opened 2024-03-12
Small Holocene fixes

laurakwhit:
## Description & motivation 💭 
* Adds a loading slot for the ApiPagination component so that a custom loading element may be used in cases where it might not be a table
* Adds `stopPropagation` to the on:click on the IconButton component (we use an IconButton on the Modal component and this prevents a button from being triggered if it is behind the Modal component)
* Adds `text-nowrap` to the Badge component so that text does not wrap on small screens
### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


rossedfort (approved):
Looks good, although I'm confused why the button snaps changed when we didn't update that component?

laurakwhit:
> why the button snaps changed when we didn't update that component


@rossedfort it's the `Badge` component we use for the count on the `Button` component (it now has the `text-nowrap`).

PR #1925, opened 2024-03-21
Update Skeleton Table and Loader components for dark mode support

KATIETOLER:
## Description & motivation 💭 
Adds design changes to Skeleton Table and Skeleton Loader components for dark mode support. 

[DT-1767](https://temporalio.atlassian.net/browse/DT-1767)
[DT-1768](https://temporalio.atlassian.net/browse/DT-1768)

Screenshot from storybook:
skeleton loader

### Design Considerations 🎨 
[Design screenshot](https://www.figma.com/file/LNN3Cu8Akqb240WmJmbtsS/Holocene?type=design&node-id=2607-30655&mode=design&t=SbNeVJoU7Clgyn1N-0):
Screenshot 2024-03-21 at 12 39 13 PM

## Testing 🧪 
Storybook


vitest.config.ts
```diff
@@ -24,6 +24,7 @@ export default defineConfig({
         'src/lib/svelte-mocks/**/*',
         'src/lib/utilities/get-environment.ts',
         '**/*.test.ts',
+        '**/.svelte-kit',
```
rossedfort, line 27: I think this needs to go in the `exclude` directly under `test`, not `test.coverage`

src/lib/theme/plugin.ts
```diff
@@ -13,6 +13,7 @@ export const variables = {
   '--color-interactive-hover': rgb(getColor('indigo', 700)),
 
   '--color-text-black': rgb(colors.black),
+  '--color-text-white': rgb(colors.white),
```
rossedfort, line 16: I think we can just use off-white instead, which already exists as `--color-text-inverse`

src/lib/theme/plugin.ts
```diff
@@ -32,6 +33,7 @@ export const variables = {
   '--color-surface-disabled': rgb(getColor('slate', 50)),
   '--color-surface-inverse': rgb(colors.primary),
   '--color-surface-subtle': rgb(getColor('slate', 100)),
+  '--color-surface-table': rgb(colors.black),
```
rossedfort, line 36: I think we can use `--color-surface-inverse` instead of a new var here

src/lib/theme/plugin.ts
```diff
@@ -42,6 +44,7 @@ export const variables = {
   '--color-border-primary': rgb(colors.black),
   '--color-border-secondary': rgb(getColor('slate', 200)),
   '--color-border-subtle': rgb(getColor('slate', 300)),
+  '--color-border-table': rgb(colors.black),
```
rossedfort, line 47: Same here, can just use `--color-border-primary`

src/lib/theme/plugin.ts
```diff
@@ -78,6 +81,7 @@ const dark: Partial<Variables<keyof typeof variables>> = {
   '--color-surface-interactive-secondary': rgb(getColor('slate', 800)),
   '--color-surface-secondary-active': rgb(getColor('slate', 700)),
   '--color-surface-subtle': rgb(getColor('slate', 900)),
+  '--color-surface-table': rgb(getColor('slate', 900)),
```
rossedfort, line 84: And here we can use `--color-surface-subtle`

src/lib/theme/plugin.ts
```diff
@@ -88,6 +92,7 @@ const dark: Partial<Variables<keyof typeof variables>> = {
   '--color-border-primary': rgb(colors.offWhite),
   '--color-border-secondary': rgb(getColor('slate', 600)),
   '--color-border-subtle': rgb(getColor('slate', 800)),
+  '--color-border-table': rgb(getColor('slate', 900)),
```
rossedfort, line 95: can use `--color-border-subtle` here as well

src/lib/theme/plugin.ts
```diff
@@ -137,6 +142,10 @@ const temporal = plugin(
         backgroundColor: css('--color-surface-subtle'),
         color: css('--color-text-primary'),
       },
+      '.surface-table': {
+        backgroundColor: css('--color-surface-table'),
+        color: css('--color-text-white'),
+      },
```
rossedfort, line 148: I think we can use surface-inverse here instead?

rossedfort (changes requested):
I think I would rather try to use existing css vars if possible. You added some new vars that use the same colors as existing vars, and I think we should try to avoid that.

Alex-Tideman:
@rossedfort So I would argue we want this. Even though we are duplicating variables in a single theme, it's the combination of light and dark themes that matter. 

Light
  '--color-border-table': rgb(colors.black)
Dark
   '--color-border-table': rgb(getColor('slate', 900))

And if we wanted to change the border of the table in a theme, we would have to do this same thing.

rossedfort:
Chatted with @Alex-Tideman and we agree that little bit of duplication is fine 👍

Alex-Tideman (approved):

rossedfort (approved):

PR #1956, opened 2024-04-01
Update Date Picker component for dark mode support

KATIETOLER:
## Description & motivation 💭 

This will add initial styles to the date-picker calendar to support dark and light mode. 

This change has not yet been approved by design but I wanted to get some feedback from the team first as far as how I should approach a few challenges in terms of the design organization in the code. 
DT-1755


src/lib/holocene/calendar.svelte
```diff
@@ -66,11 +66,17 @@
   }
 
   .cell {
-    @apply m-1 inline-flex h-[24px] w-[24px] items-center justify-center rounded p-1 text-sm;
+    @apply m-1 inline-flex h-[24px] w-[24px] items-center justify-center rounded p-1 text-sm 
+    hover:surface-interactive
+    active:surface-interactive  
+    hover:rounded-2xl 
+    hover:text-inverse 
+    active:text-inverse;
```
Alex-Tideman, line 74: I think we want to move these to the .highlight class. Otherwise the days of the week ("Su","Mo"...) and the days that are not of the current month get this background too. We just want the cells you can select to have the background. I'd also probably use bg-interactive and text-offWhite since you are overriding the surface text color anyways.

KATIETOLER: Thank you so much, this is very helpful.

src/lib/holocene/date-picker.svelte
```diff
@@ -89,7 +89,7 @@
   />
   {#if showDatePicker}
     <div
-      class="surface-primary absolute z-30 inline-block rounded border shadow"
+      class="surface-primary absolute z-30 inline-block rounded border border-subtle shadow dark:bg-[#101828]"
```
laurakwhit, line 92: Is `#101828` a new color? If so, maybe something we want to add to `/theme/plugins.ts` instead of having as a one-off here. 

Otherwise maybe there's an existing black we can use here instead?

src/lib/holocene/calendar.svelte
```diff
@@ -66,11 +66,11 @@
   }
 
   .cell {
-    @apply m-1 inline-flex h-[24px] w-[24px] items-center justify-center rounded p-1 text-sm;
+    @apply m-1 inline-flex h-[24px] w-[24px] items-center justify-center rounded-2xl p-1 text-sm text-primary;
   }
 
   .selected {
-    @apply bg-gradient-to-br from-blue-100  to-purple-100 text-primary;
+    @apply surface-interactive rounded-2xl  p-2 text-offWhite dark:bg-indigo-300 dark:text-black;
```
laurakwhit, line 73: Looks like the hover and selected style is the same in light mode ⬇️ 

Screenshot 2024-04-05 at 11 40 45 AM

Might we want to make them slightly different similar to how they are in dark mode?

Screenshot 2024-04-05 at 11 40 13 AM

src/lib/holocene/date-picker.svelte
KATIETOLER: This is the 'main' background color for dark mode -- it only exists in one other place right now. This is the color from the designs. 

Do you have an opinion here for either switching to our black and checking with Ash or adding to plugin file? I truly don't know what the right move here would be (thus the hard coding) and would love guidance.

src/lib/holocene/calendar.svelte
KATIETOLER: Yes, this is from the designs. I will ask Ash if they want to change this.

src/lib/holocene/date-picker.svelte
laurakwhit: I think we don't want any one-off colors and should add this to the plugin file. If there's not a noticeable difference between this and an existing black than I think it would be fine to just use the existing.

laurakwhit:
| Design | Before | After |
|--|--|--|
|Screenshot 2024-04-11 at 10 59 32 AM|Screenshot 2024-04-11 at 11 14 07 AM|Screenshot 2024-04-11 at 10 57 21 AM|
|Screenshot 2024-04-11 at 10 59 42 AM|Screenshot 2024-04-11 at 11 14 26 AM|Screenshot 2024-04-11 at 10 57 43 AM|

| Before | After |
|--|--|
|Screenshot 2024-04-11 at 11 15 08 AM|Screenshot 2024-04-11 at 11 18 45 AM|
|Screenshot 2024-04-11 at 11 14 51 AM|Screenshot 2024-04-11 at 11 18 29 AM|

laurakwhit (approved):

PR #1963, opened 2024-04-04
update copyable button

GraceGardner:
## Description & motivation 💭 
Updates copyable button, I just used the designs for the Icon button on darkmode being it's always going to be on a dark background.  

without hover state
Screenshot 2024-04-04 at 1 14 23 PM
With hover state
Screenshot 2024-04-04 at 1 14 33 PM

### Design Considerations 🎨 
Needs approval from Ash 

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
DT-1987


src/lib/holocene/copyable/button.svelte
```diff
@@ -29,6 +29,6 @@
 
 <style lang="postcss">
   .copy-button {
-    @apply m-1 rounded-md border-2 border-[transparent] hover:border-indigo-600 hover:bg-gradient-to-br hover:from-blue-100 hover:to-purple-100 hover:shadow-focus hover:shadow-blue-600/50 focus-visible:border-indigo-600 focus-visible:bg-gradient-to-br focus-visible:from-blue-100 focus-visible:to-purple-100 focus-visible:shadow-focus focus-visible:shadow-blue-600/50 focus-visible:outline-none;
+    @apply m-1 rounded-md border-2 border-[transparent] bg-transparent text-white hover:bg-slate-800 focus-visible:border-indigo-600 focus-visible:bg-slate-800 focus-visible:outline-none;
```
laurakwhit, line 32: We'll need styling for light mode too. This isn't only used within a `CodeBlock` component, it's also used in the `Copyable` component which can be on any background (e.g. with the `Workflow ID` ⬇️ ).

Screenshot 2024-04-04 at 6 04 29 PM

laurakwhit (changes requested):

src/lib/holocene/copyable/button.svelte
GraceGardner: @laurakwhit would it make sense to have a conditional for when it's used in the `CodeBlock`? That way we can have it show the darkmode styles in the `CodeBlock` component?

laurakwhit: It looks like [we're currently passing](https://github.com/temporalio/ui/blob/main/src/lib/holocene/code-block.svelte#L168) a `class` that overrides the styles. We might just want to do something similar, but I think a prop wouldn't be a bad idea either.

GraceGardner: @laurakwhit So I tried to use the class, and the other styles are overriding the things I'm passing in. Now sure why? What do you think?

src/lib/holocene/copyable/button.svelte
```diff
@@ -29,6 +29,6 @@
 
 <style lang="postcss">
   .copy-button {
-    @apply m-1 rounded-md border-2 border-[transparent] hover:border-indigo-600 hover:bg-gradient-to-br hover:from-blue-100 hover:to-purple-100 hover:shadow-focus hover:shadow-blue-600/50 focus-visible:border-indigo-600 focus-visible:bg-gradient-to-br focus-visible:from-blue-100 focus-visible:to-purple-100 focus-visible:shadow-focus focus-visible:shadow-blue-600/50 focus-visible:outline-none;
+    @apply m-1 rounded-md border-2 border-[transparent] bg-transparent text-primary hover:surface-interactive-secondary focus-visible:surface-interactive-secondary focus-visible:border-indigo-600 focus-visible:outline-none;
```
laurakwhit, line 32: Looks like `text-primary` is causing the icon to "disappear" in light mode

```suggestion
    @apply m-1 rounded-md border-2 border-[transparent] bg-transparent hover:surface-interactive-secondary focus-visible:surface-interactive-secondary focus-visible:border-indigo-600 focus-visible:outline-none;
```

| Before | After |
|--|--|
|Screenshot 2024-04-11 at 11 49 18 AM|Screenshot 2024-04-11 at 11 51 25 AM|

laurakwhit (changes requested):

laurakwhit (approved):

PR #1983, opened 2024-04-09
update panel colors

GraceGardner:
## Description & motivation 💭 
Update border color for the panel component 

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
DT-1992

### DIFF
--- a/src/lib/components/panel.svelte
+++ b/src/lib/components/panel.svelte
@@ -8,10 +8,10 @@
 
 <style lang="postcss">
   .panel {
-    @apply rounded-xl border-2  p-8;
+    @apply rounded-xl border-2 p-8 dark:border-subtle;
   }
 
   .error {
-    @apply border-red-700 bg-red-50;
+    @apply border-error bg-error;
   }
 </style>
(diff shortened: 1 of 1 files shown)


src/lib/components/panel.svelte
```diff
@@ -8,10 +8,10 @@
...
 <style lang="postcss">
   .panel {
-    @apply rounded-xl border-2  p-8;
+    @apply rounded-xl border-2 p-8 dark:border-subtle;
   }
 
   .error {
-    @apply border-red-700 bg-red-50;
+    @apply border-error bg-error;
   }
```
laurakwhit, line 16: Looks like `border-error` here ⬇️  is being overridden by `border-subtle` in dark mode

Screenshot 2024-04-09 at 8 28 19 PM


Moving the `error` class should fix it

```js
  .panel {
    @apply rounded-xl border-2 p-8 dark:border-subtle;
    &.error {
      @apply border-error bg-error;
    }
  }
```

laurakwhit (changes requested):

src/lib/components/panel.svelte
GraceGardner: Thanks Laura! I appreciate you!

laurakwhit (approved):

PR #1984, opened 2024-04-09
Fix table-header class in menu-button component

GraceGardner:
Before: 
Screenshot 2024-04-09 at 2 26 40 PM

After: 
Screenshot 2024-04-11 at 10 49 32 AM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/lines-and-dots/event-sort-filter.svelte
+++ b/src/lib/components/lines-and-dots/event-sort-filter.svelte
@@ -57,7 +57,7 @@
   <MenuButton
     active={!!value}
     hasIndicator
-    variant="primary"
+    variant="table-header"
     controls="event-date-filter-menu"
     data-testid="event-date-filter-button"
   >
--- a/src/lib/components/lines-and-dots/event-sort-filter.svelte
+++ b/src/lib/components/lines-and-dots/event-sort-filter.svelte
@@ -57,7 +57,7 @@
   <MenuButton
     active={!!value}
     hasIndicator
-    variant="table-header"
+    variant="primary"
     controls="event-date-filter-menu"
     data-testid="event-date-filter-button"
   >
--- a/src/lib/holocene/menu/menu-button.svelte
+++ b/src/lib/holocene/menu/menu-button.svelte
@@ -174,7 +174,7 @@
   }
 
   .table-header {
-    @apply bg-inverse text-white focus-visible:border-white focus-visible:shadow-focus focus-visible:shadow-primary/50;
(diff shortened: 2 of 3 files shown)


src/lib/components/lines-and-dots/event-sort-filter.svelte
```diff
@@ -57,7 +57,7 @@
   <MenuButton
     active={!!value}
     hasIndicator
-    variant="primary"
+    variant="table-header"
```
laurakwhit, line 60: We don't need to update this component since it isn't currently being used anywhere. It also looks like it's part of the lines and dots and so it likely wouldn't be in a table anyways.

```suggestion
```

src/lib/holocene/menu/menu-button.svelte
```diff
@@ -174,7 +174,7 @@
   }
 
   .table-header {
-    @apply h-auto max-w-fit border-2 border-transparent bg-primary p-0 text-sm text-white focus-visible:border-white;
+    @apply bg-inverse text-white focus-visible:border-white focus-visible:shadow-focus focus-visible:shadow-primary/50;
```
laurakwhit, line 177: Do we want to match the existing style or are there updated designs? 

If we want to just make the existing style compatible with dark mode (e.g. ⬇️)

| Light | Dark |
|--|--|
|Screenshot 2024-04-09 at 8 00 23 PM| Screenshot 2024-04-09 at 8 11 46 PM|

 all we need to do here is to remove the `bg-primary` 

```suggestion
    @apply h-auto max-w-fit border-2 border-transparent p-0 text-sm text-white focus-visible:border-white;
```

and update the `table-header` variant in the `Button` component by replacing `bg-inverse` with `border-transparent` [here](https://github.com/temporalio/ui/blob/main/src/lib/holocene/button.svelte#L45).

laurakwhit (changes requested):

src/lib/holocene/menu/menu-button.svelte
laurakwhit: > and update the table-header variant in the Button component by replacing bg-inverse with border-transparent [here](https://github.com/temporalio/ui/blob/main/src/lib/holocene/button.svelte#L45).

@GraceGardner we'll need this part too in order fix this button ⬇️ 

Screenshot 2024-04-11 at 9 40 25 AM

GraceGardner: Thanks for catching this <3

laurakwhit (approved):

PR #1986, opened 2024-04-10
Show tooltip of relative/local time on workflow start/end times

Alex-Tideman:
## Description & motivation 💭 
For Workflow Start/End Times in Summary, add tooltip to show relative time if timezone selected is local, and show local time if timezone selected is relative.

Small UI improvements to switch order in new UI to Start/End/Duration and only show icon for duration.

Screen Shot 2024-04-10 at 9 43 41 AM
Screen Shot 2024-04-10 at 9 43 52 AM
Screen Shot 2024-04-10 at 9 51 46 AM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/lines-and-dots/workflow-details.svelte
+++ b/src/lib/components/lines-and-dots/workflow-details.svelte
@@ -24,26 +24,39 @@
     class="surface-primary grid w-full grid-flow-row grid-cols-1 gap-2 rounded-xl border-2 border-primary p-2 md:grid-cols-2 xl:grid-cols-3"
   >
     <WorkflowDetail
-      title={translate('common.duration')}
-      content={elapsedTime}
-      class="order-1 text-sm"
-      icon="clock"
-    />
-    <WorkflowDetail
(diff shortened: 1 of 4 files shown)


src/lib/components/lines-and-dots/workflow-details.svelte
```diff
@@ -23,27 +23,40 @@
...
   >
-    <WorkflowDetail
-      title={translate('common.duration')}
-      content={elapsedTime}
-      class="order-1 text-sm"
-      icon="clock"
-    />
     <WorkflowDetail
       title={translate('common.start')}
+      tooltip={$relativeTime
```
GiantRobots, line 28: Probably my intense hatred for ternaries is clouding my judgement here, but maybe the tooltip should be an object with both time stamps or implemented in workflow detail so it always hits this codepath?

GiantRobots (approved):

PR #1995, opened 2024-04-11
Darkmode/pills

GraceGardner:
Before: 
Screenshot 2024-04-11 at 12 40 45 PM
Screenshot 2024-04-11 at 3 51 28 PM

After: 

Screenshot 2024-04-11 at 4 18 54 PM
Screenshot 2024-04-11 at 4 19 21 PM

### Design Considerations 🎨 
Designs weren't completed for this, so we might need to do a fast follow on this one 🏃‍♀️‍➡️

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
DT-1960


src/lib/components/pill.svelte
```diff
@@ -22,15 +22,15 @@
 
 <style lang="postcss">
   .pill {
-    @apply flex items-center justify-center gap-1 rounded-full border-2 border-slate-100 px-3 py-1 text-sm;
+    @apply flex items-center justify-center gap-1 rounded-full border-2 border-slate-50 px-3 py-1  text-sm text-black text-inverse;
```
laurakwhit, line 25: Looks like there's an unnecessary `text-black` and we can probably just make the `border-transparent` here

```suggestion
        @apply flex items-center justify-center gap-1 rounded-full border-2 border-transparent px-3 py-1 text-sm text-inverse;
```

src/lib/components/pill.svelte
```diff
@@ -22,15 +22,15 @@
...
   }
 
   .gray {
-    @apply bg-badge text-slate-600;
+    @apply border-none bg-transparent text-primary;
```
laurakwhit, line 29: Small suggestion for the text color here

```suggestion
    @apply bg-badge text-disabled;
```
Screenshot 2024-04-11 at 5 09 47 PM

src/lib/components/pill.svelte
```diff
@@ -22,15 +22,15 @@
...
   }
 
   .gray.active {
-    @apply surface-primary border-2  text-primary;
+    @apply surface-interactive;
```
laurakwhit, line 33: We're actually never using this class currently, so we can leave it as is for now

```suggestion
    @apply surface-primary border-2 text-primary;
```

src/lib/components/pill.svelte
```diff
@@ -50,7 +50,7 @@
   }
 
   .lightBlue.active {
-    @apply border-2  bg-blue-50 text-primary;
+    @apply border-2 border-indigo-500 bg-indigo-500;
```
laurakwhit, line 53: Let's just use the existing `interactive` color here

```suggestion
    @apply border-2 border-interactive bg-interactive;
```

laurakwhit (commented):
Just noting that we'll definitely want to audit this component again in the future as there are a bunch of possible colors other than `gray` and `lightBlue` that we aren't accounting for here. 

Maybe we can add a follow-up task?

src/lib/components/pill.svelte
GraceGardner: @laurakwhit The interactive color  was what I thought should be used too, but it shows the indigo on light mode and then it goes to bg-transparent with a white border on dark mode. I also tried brand, but the color was too dark.

GraceGardner: Looks like you were able to use interactive in your PR and it worked fine.... Not sure if this was a me problem?

KATIETOLER (approved):
Looks good to me!

PR #2013, opened 2024-04-17
Fetch child workflows with ParentWorkflowId if Temporal server v1.23+

Alex-Tideman:
## Description & motivation 💭 
In the Relationships tab, child workflows can show stale statuses due to the parent completing. The completed parent won't update it's pending children, it's a snapshot at the time of completion.

With v1.23, we can now query workflows with ParentWorkflowId. This allows us to show current status of children regardless of the parent status.

This shows the difference in the two tables after terminating the workflow. The live children table correctly shows the Terminated child workflow status. The old table shows it still running.

Screen Shot 2024-04-17 at 12 45 13 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/workflow/live-child-workflows-table.svelte
+++ b/src/lib/components/workflow/live-child-workflows-table.svelte
@@ -1,0 +1,72 @@
+<script lang="ts">
+  import { page } from '$app/stores';
+
+  import Link from '$lib/holocene/link.svelte';
+  import Pagination from '$lib/holocene/pagination.svelte';
+  import TableHeaderRow from '$lib/holocene/table/table-header-row.svelte';
+  import TableRow from '$lib/holocene/table/table-row.svelte';
+  import Table from '$lib/holocene/table/table.svelte';
+  import { translate } from '$lib/i18n/translate';
+  import type { WorkflowExecution } from '$lib/types/workflows';
+  import { routeForEventHistory } from '$lib/utilities/route-for';
+
+  import WorkflowStatus from '../workflow-status.svelte';
+
+  export let children: WorkflowExecution[] = [];
+
+  $: ({ namespace } = $page.params);
+</script>
+
+<Pagination
+  items={children}
+  let:visibleItems
+  aria-label={translate('workflows.child-workflows')}
+  pageSizeSelectLabel={translate('common.per-page')}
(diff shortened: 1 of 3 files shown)


src/lib/components/workflow/live-child-workflows-table.svelte
```diff
@@ -0,0 +1,72 @@
...
+    <caption class="sr-only" slot="caption"
+      >{translate('workflows.child-workflows')}</caption
+    >
+    <TableHeaderRow slot="headers">
+      <th class="max-md:hidden">{translate('common.status')}</th>
+      <th class="max-lg:hidden">{translate('common.type')}</th>
+      <th>{translate('workflows.child-id')}</th>
+      <th>{translate('workflows.child-run-id')}</th>
+    </TableHeaderRow>
+    {#each visibleItems as child}
```
laurakwhit, line 39: What do you think about just using `ChildWorkflowsTable` here and instead refactoring it to only take `children` and `namespace` props. 

The children could be formatted in the parent component to something like this ⬇️ 

```suggestion
    {#each visibleItems as {status, name, namespace, workflow, run}}
```

That way if the any future UI changes for this table only need to happen in one place.

src/lib/components/workflow/workflow-relationships.svelte
```diff
@@ -52,7 +63,9 @@
         />
       {/if}
     </div>
-    {#if hasChildren}
+    {#if liveChildren.length}
+      <LiveChildWorkflowsTable children={liveChildren} />
```
laurakwhit, line 67: Do we want any kind of loading state on this table?

Alex-Tideman: I did it this way so that the "non"-live table shows up immediately (i.e. that's the loading state), and then the "live" table replaces it.

src/lib/components/workflow/live-child-workflows-table.svelte
laurakwhit: Definitely not a blocker! Especially if we think this table might diverge from the other.

laurakwhit (approved):

PR #2014, opened 2024-04-17
DT-1740 - update workflow reset UI for new reapply types in 1.24

rossedfort:
## Description & motivation 💭 
in temporal server version 1.24, a new API will be available for resetting workflows that supports reapplying Updates after the reset point, in addition to Signals. This PR makes the changes necessary to the UI support this new option. Note - the behavior is inverse, i.e. the API expects opting out of reapplies as opposed to opting in.

- For UI running against temporal server version <1.24 - no change. Show a checkbox whether to re-apply signals after reset point, if checked send `RESET_REAPPLY_TYPE_SIGNAL`, if unchecked send `RESET_REAPPLY_TYPE_NONE` in `resetReapplyType` [field](https://github.com/temporalio/api/blob/master/temporal/api/workflowservice/v1/request_response.proto#L699)
- For UI running against temporal server version >=1.24 - keep checkbox for Signal, add a checkbox for Update - inverse behavior of <1.24. if only “Exclude signals” is checked - send `RESET_REAPPLY_EXCLUDE_TYPE_SIGNAL`, if only “Exclude updates” is checked, send `RESET_REAPPLY_EXCLUDE_TYPE_UPDATE`. If both are checked send both, if none are checked, send `RESET_REAPPLY_EXCLUDE_TYPE_UNSPECIFIED` in `resetReapplyExcludeTypes` [field](https://github.com/temporalio/api/blob/master/temporal/api/workflowservice/v1/request_response.proto#L701)
|<1.24|>=1.24|
|---|---|
|Screenshot 2024-04-17 at 2 12 52 PM|Screenshot 2024-04-17 at 2 19 33 PM|
### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/components/workflow/client-actions/reset-confirmation-modal.svelte
```diff
@@ -81,31 +80,39 @@
...
-        checked={reapplyType === ResetReapplyType.Signal}
-        on:change={handleReapplyTypeChange}
-        label={translate('workflows.reset-reapply-type-label')}
-      />
+      </Select>
+      {#if minimumVersionRequired('1.24', $temporalVersion)}
+        <Checkbox
+          id="reset-exclude-signals-checkbox"
+          bind:checked={excludeSignals}
+          label={translate('workflows.reset-exclude-signals')}
```
Alex-Tideman, line 102: Tiny nit, capitalize Signals as well as Updates below in the translate

src/lib/i18n/locales/en/workflows.ts
```diff
@@ -90,6 +90,8 @@ export const Strings = {
   'reset-event-radio-group-description': 'Choose an Event to reset to',
   'reset-reapply-type-label':
     'Reapply Signals that happened after the Reset point',
+  'reset-exclude-signals': 'Exclude signals that occur after the reset point.',
+  'reset-exclude-updates': 'Exclude updates that occur after the reset point.',
```
Alex-Tideman, line 94: Let's capitalize Signals and Updates here

dandavison:
Screenshots LGTM!

src/lib/i18n/locales/en/workflows.ts
alexshtin: and Reset too.

alexshtin (approved):
Approving just screenshots :-)

src/lib/components/workflow/client-actions/reset-confirmation-modal.svelte
```diff
@@ -63,6 +68,8 @@
     }
     hideResetModal();
   };
+
+  $: console.log($temporalVersion);
```
Alex-Tideman, line 72: 🗑️

Alex-Tideman (approved):

src/lib/services/workflow-service.ts
```diff
@@ -363,18 +379,53 @@ export async function resetWorkflow({
...
+    }
+
+    if (excludeUpdates) {
+      resetReapplyExcludeTypes.push(
+        ResetReapplyExcludeType.RESET_REAPPLY_EXCLUDE_TYPE_UPDATE,
+      );
+    }
+
+    body.resetReapplyExcludeTypes = resetReapplyExcludeTypes;
+    body.resetReapplyType = ResetReapplyType.RESET_REAPPLY_TYPE_ALL_ELIGIBLE;
```
dandavison, line 416: It might be worth leaving a comment in the code saying that in the future we won't _need_ to explicitly send `body.resetReapplyType`, since `ResetReapplyType.RESET_REAPPLY_TYPE_ALL_ELIGIBLE` will be the server-side default. AIUI that will become true for server >= 1.25.1 (see https://github.com/temporalio/temporal/pull/5956)

But to be clear, this code is correct as-is.

dandavison (approved):
LGTM!

PR #2021, opened 2024-04-18
Update loading component to mp4 

laurakwhit:
## Description & motivation 💭 
Updates the `Loading` component to use a `.mp4` or `.webm` file instead of a `.gif`.

| Before | After |
|-|-|
|Screenshot 2024-04-18 at 2 28 37 PM| Screenshot 2024-04-19 at 12 59 25 PM|
|Screenshot 2024-04-18 at 2 28 26 PM|Screenshot 2024-04-19 at 12 57 48 PM|
|Screenshot 2024-04-18 at 3 45 31 PM|Screenshot 2024-04-19 at 12 58 58 PM|

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
`DT-1952`


src/lib/holocene/loading.svelte
```diff
@@ -1,19 +1,25 @@
...
+  class="my-12 flex flex-col items-center justify-start gap-4 {$$props.class}"
+>
+  <video
+    autoplay
+    loop
+    muted
+    playsinline
+    width="100"
+    height="100"
+    class="dark:invert-[.92]"
```
laurakwhit, line 18: I tried adjusting the filter here to account for the background it is most likely to be on, but not sure if that's really what we want to do here... 

There are still other backgrounds that won't match the background of the mp4 ⬇️ 
|||
|-|-|
|Screenshot 2024-04-18 at 4 50 53 PM|Screenshot 2024-04-18 at 4 51 00 PM|

Alex-Tideman: Any way to get the mp4 with a transparent bg?

laurakwhit: Unfortunately, no 😞 Need something with an alpha channel.

laurakwhit: **Update:** Apparently an mp4 **can** support an alpha channel with the H.265 codec. But looks like only in Safari, so added a link to a `.webm` source as well.

rossedfort (approved):

rossedfort:
Did you happen to test this if you have autoplay disabled?

rossedfort:
> Did you happen to test this if you have autoplay disabled?

Hm maybe it's not possible to disable autoplay, although I thought it was 🤔

src/lib/holocene/loading.svelte
```diff
@@ -1,19 +1,28 @@
...
-    height="200px"
-  />
-  <h2 class="text-xl font-medium" style="margin-top: -40px;">
+<div
+  class="my-12 flex flex-col items-center justify-start gap-4 {$$props.class}"
+>
+  <video
+    autoplay
+    loop
+    muted
```
laurakwhit: > Autoplay blocking is not applied to [<video>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video) elements when the source media does not have an audio track, or if the audio track is muted.

@rossedfort I had assumed [this](https://developer.mozilla.org/en-US/docs/Web/Media/Autoplay_guide) ⬆️  meant we could get around autoplay blocking.

But after further testing it definitely does not work in certain browsers with `autoplay` blocked (e.g. Firefox). I've gone ahead and [added a check](https://github.com/temporalio/ui/pull/2021/commits/8a1cc2789937a3d5ba8bc311186d57c35e3fb4ec) to see if it is allowed and a fallback to a logo ⬇️ if it is not.

Screenshot 2024-04-25 at 12 28 40 PM

PR #2045, opened 2024-04-24
Fix alignment on checkbox; address accessibility concerns

stevekinney:
Addresses the following accessibility warnings:

- A11y: `<div>` with click, keypress handlers must have an ARIA role
- A11y: Non-interactive element `<label>` should not be assigned mouse or keyboard event listeners.

### Before

checkbox-before@2x

### After

checkbox-after@2x

### DIFF
--- a/src/lib/holocene/checkbox.svelte
+++ b/src/lib/holocene/checkbox.svelte
@@ -119,7 +119,7 @@
 
 <style lang="postcss">
   .checkbox {
-    @apply flex cursor-pointer select-none items-start items-center gap-3 text-sm leading-[18px] text-primary;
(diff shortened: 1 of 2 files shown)


src/lib/holocene/checkbox.svelte
```diff
@@ -118,7 +119,7 @@
 
 <style lang="postcss">
   .checkbox {
-    @apply flex cursor-pointer select-none items-start gap-3 text-sm leading-[18px] text-primary;
+    @apply flex cursor-pointer select-none items-start items-center gap-3 text-sm leading-[18px] text-primary;
```
Alex-Tideman, line 122: Remove that items-start here eh

rossedfort (approved):

PR #2049, opened 2024-04-25
Get rid of split button variants, support passing classes to buttons, use a Button in Split Buttons

stevekinney:
- Adds support for stories anywhere
- Uses only primary variants of split buttons
- Uses a `Button` component for the menu button and removes duplicate classes


src/lib/holocene/button.stories.svelte
```diff
@@ -73,22 +73,6 @@
...
-        },
-      },
-      borderRadiusModifier: {
-        name: 'Border Radius Modifier',
-        control: 'select',
-        options: ['square', 'square-left', 'square-right'],
-        table: {
-          category: 'Modifiers',
-        },
-      },
```
rossedfort, line 91: There's a few places we're using this where the button is next to an input, will need to account for that

stevekinney: This is only removing the control from Storybook, not the actual functionality. I figure the stories where it's used should capture the functionality. A random knob that lets you make the left-side of a button square isn't _super useful_. But, I can be persuaded to put it back.

rossedfort (approved):

PR #2055, opened 2024-04-29
Don't return WorkflowTaskFailed error if a WorkflowTaskCompleted event occurs after it

Alex-Tideman:
## Description & motivation 💭 
We don't want to show WorkflowTaskFailed errors if a WorkflowTaskCompleted event occurs after it since the Workflow has made progress.

Screen Shot 2024-04-29 at 9 52 16 AM
Screen Shot 2024-04-29 at 9 52 29 AM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

### DIFF
--- a/src/lib/utilities/get-workflow-task-failed-event.ts
+++ b/src/lib/utilities/get-workflow-task-failed-event.ts
@@ -22,10 +22,11 @@ const getFailedWorkflowTask = (
   fullEventHistory: WorkflowEvents,
 ): WorkflowTaskFailedEvent | undefined => {
   const failedWorkflowTaskIndex = fullEventHistory.findIndex(isFailedTaskEvent);
-  const completedWorkflowTaskIndex =
-    fullEventHistory.findIndex(isCompletedTaskEvent);
 
   if (failedWorkflowTaskIndex < 0) return;
+
+  const completedWorkflowTaskIndex =
+    fullEventHistory.findIndex(isCompletedTaskEvent);
 
   const failedWorkflowTask = fullEventHistory.find((event) =>
     isFailedTaskEvent(event),
--- a/src/lib/utilities/get-workflow-task-failed-event.ts
+++ b/src/lib/utilities/get-workflow-task-failed-event.ts
@@ -2,6 +2,7 @@ import type { EventSortOrder } from '$lib/stores/event-view';
 import type {
   WorkflowEvent,
   WorkflowEvents,
+  WorkflowTaskCompletedEvent,
   WorkflowTaskFailedEvent,
 } from '$lib/types/events';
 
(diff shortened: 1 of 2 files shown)


src/lib/utilities/get-workflow-task-failed-event.ts
```diff
@@ -11,18 +12,42 @@ const isFailedTaskEvent = (
...
+  ) as WorkflowTaskFailedEvent;
+
+  if (completedWorkflowTaskIndex < 0 && failedWorkflowTaskIndex >= 0)
+    return failedWorkflowTask;
+
+  // History is sorted in descending order, so index of failed task should be less than index of completed task
+  if (failedWorkflowTaskIndex < completedWorkflowTaskIndex) {
+    return failedWorkflowTask;
+  }
+};
```
rossedfort, line 41: Not a biggie but I feel like this function could be optimized slightly. Could return early if there is no failed wf task, and remove some iteration. Also this may need some tweaking, I wrote it in GitHub without running/testing.

```suggestion
const getFailedWorkflowTask = (
  fullEventHistory: WorkflowEvents,
): WorkflowTaskFailedEvent | undefined => {
  
  let failedWorkflowTaskIndex = 0;
  let failedWorkflowTask: WorkflowTaskFailedEvent | undefined = undefined;
  let completedWorkflowTaskIndex = 0;

  for (let i = 0; i < fullEventHistory.length; i++) {
    const event = fullEventHistory[i];
    if (isFailedTaskEvent(event) && !failedWorkflowTask) {
      failedWorkflowTaskIndex = i;
      failedWorkflowTask = event;
    }
    
    if (isCompletedTaskEvent(event)) {
      completedWorkflowTaskIndex = i;
    }
  }

  if (!failedWorkflowTaskIndex) return;

  if (completedWorkflowTaskIndex < 0 && failedWorkflowTaskIndex >= 0)
    return failedWorkflowTask;

  // History is sorted in descending order, so index of failed task should be less than index of completed task
  if (failedWorkflowTaskIndex < completedWorkflowTaskIndex) {
    return failedWorkflowTask;
  }
};
```

rossedfort (approved):

src/lib/utilities/get-workflow-task-failed-event.ts
Alex-Tideman: I like findIndex more cuz it's about finding the first occurance (really the latest but since it's reversed history the first) and so you don't need to iterate over the whole event history. But Ill move  if (failedWorkflowTaskIndex < 0) return; above the completedWorkflowTaskIndex so it returns before doing that check if no failed workflow tasks

rossedfort: Sounds good!

PR #1926, opened 2024-03-22
Audit Split Button

GraceGardner:
## Description & motivation 💭 
fixes split button padding for the dark mode update. 

### Design Considerations 🎨 
Needs approval from Ash

## Testing 🧪 
Testing through Story book using the split button component. 

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
https://temporalio.atlassian.net/browse/DT-1769

PR #2062, opened 2024-04-30
Padding, border-radius fixes

Alex-Tideman:
## Description & motivation 💭 
Remove unused unround props
Add gap to Schedule Offset
Add more x-padding on split button to make it easier to click
Add padding to RadioGroup so on:hover effect doesn't get cut off
Fix rounding on Workflow Search filters
Fix gap and max height on DateTime filter inputs
Add top padding on Loading for Labs Mode

Screen Shot 2024-04-30 at 10 23 30 AM
Screen Shot 2024-04-30 at 10 24 46 AM
Screen Shot 2024-04-30 at 10 27 25 AM
Screen Shot 2024-04-30 at 12 11 23 PM
Screen Shot 2024-04-30 at 12 27 16 PM

**Loading After**
Screen Shot 2024-04-30 at 12 13 16 PM

**Loading Before**
Screen Shot 2024-04-30 at 12 13 27 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/holocene/split-button.svelte
+++ b/src/lib/holocene/split-button.svelte
@@ -26,7 +26,7 @@
       {label}
     </Button>
     <MenuButton
-      class="max-w-fit grow-0 px-1"
+      class="max-w-fit grow-0 px-3"
       id="{id}-menu-button"
       label={menuLabel}
       controls="{id}-menu"
--- a/src/lib/components/schedule/schedules-interval-view.svelte
+++ b/src/lib/components/schedule/schedules-interval-view.svelte
@@ -100,7 +100,7 @@
   <p>
     {translate('schedules.offset-description')}
   </p>
-  <div class="flex w-48 gap-0">
+  <div class="flex w-48 gap-2">
     <Input
       id="phase"
       class="w-28"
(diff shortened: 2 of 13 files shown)


src/lib/holocene/split-button.svelte
```diff
@@ -26,7 +26,7 @@
       {label}
     </Button>
     <MenuButton
-      class="max-w-fit grow-0 px-1"
```
rossedfort, line 29: I'm pretty sure this was intentional based on feedback from Ash in https://github.com/temporalio/ui/pull/1926

I'd double check with them that this is the padding we want. Looks like the designs specify p-2
Screenshot 2024-04-30 at 9 37 01 AM

rossedfort (approved):

src/lib/holocene/split-button.svelte
Alex-Tideman: That was before the icons were changed to 16x16, I think px-3 is the way to go now.

Screen Shot 2024-04-30 at 10 49 13 AM

PR #2088, opened 2024-05-06
Fix accessibility violations for the date picker

stevekinney:
- Addresses contrast issues in dark mode.
- Swaps out label-less buttons for empty elements for days outside of the month.
- Adds labels to previous and next month buttons.
- Adds previous and next text to `i18n`.

### DIFF
--- a/src/lib/holocene/calendar.svelte
+++ b/src/lib/holocene/calendar.svelte
@@ -38,20 +38,24 @@
 
   <div class="row">
     {#each cells as { allowed, value }, index (index)}
-      <button
-        type="button"
-        on:click={allowed && value ? () => onChange(value) : noop}
-        class="cell"
-        class:highlight={allowed && value}
-        class:disabled={!allowed}
-        class:selected={new Date(
-          date.getFullYear(),
-          date.getMonth(),
-          date.getDate(),
-        ).getTime() === new Date(year, month, value).getTime()}
-      >
-        {value || ''}
-      </button>
+      {#if value}
+        <button
+          type="button"
+          on:click={allowed && value ? () => onChange(value) : noop}
+          class="cell"
+          class:highlight={allowed && value}
+          class:disabled={!allowed}
+          class:selected={new Date(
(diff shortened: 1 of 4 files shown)


src/lib/i18n/locales/en/date-picker.ts
rossedfort, line 1: These could probably just go in the `common.ts` strings file, but this is fine too

src/lib/holocene/date-picker.svelte
```diff
@@ -93,18 +97,21 @@
     >
       <div class="mx-3 my-2 flex items-center justify-around">
         <div class="flex items-center justify-center">
-          <button type="button" on:click={prev}
-            ><Icon name="chevron-left" /></button
+          <button type="button" on:click={prev} title={previousMonth}>
+            <span class="sr-only">{previousMonth}</span>
```
rossedfort, line 101: Is this preferred over just specifying `aria-label` on the button? I imagine it accomplishing the same thing.

src/lib/holocene/calendar.svelte
```diff
@@ -38,20 +38,24 @@
...
-        class:disabled={!allowed}
-        class:selected={new Date(
-          date.getFullYear(),
-          date.getMonth(),
-          date.getDate(),
-        ).getTime() === new Date(year, month, value).getTime()}
-      >
-        {value || ''}
-      </button>
+      {#if value}
```
rossedfort, line 41: Why are we only showing the button if it has a value? So it doesn't have an empty label? If so - maybe we should just change the empty state label down on line 54.

stevekinney: Well, it's not _really_ a button at that point either. So, we really just need something to take up space.

src/lib/i18n/locales/en/date-picker.ts
stevekinney: I wasn't sure if it might grow with months and whatnot.

PR #2096, opened 2024-05-08
Remove double encoding for api route

Alex-Tideman:
## Description & motivation 💭 
I'm guessing this is an ancient relic of pre-v1.0 Sveltekit where we needed to double encode the URI. This can cause issues with WorkflowIds like `workflow ()` and I'm sure other combinations. It also adds extra %25 characters which I'm shocked haven't been an issue on the server. I've removed the double encoding and added more unit tests for special characters.

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

### DIFF
--- a/src/lib/utilities/route-for-api.test.ts
+++ b/src/lib/utilities/route-for-api.test.ts
@@ -181,7 +181,7 @@ describe('API Request Encoding', () => {
       workflowId: 'workflow#with#hashes',
     });
     expect(route).toBe(
-      'http://localhost:8233/api/v1/namespaces/namespace/workflows/workflow%2523with%2523hashes',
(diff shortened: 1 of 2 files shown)


laurakwhit (approved):

tsurdilo:
hi, seems this was reverted in pr https://github.com/temporalio/ui/pull/2107
can we get some info if using parenthesis in workflow id is going to be supported at some point / will be worked on in a different issue/commit?

Alex-Tideman:
This PR reverted a fix for this specific issue but it broke slashes in Workflow ids. You can use parenthesis, just not a space before parenthesis. We will continue to work on this specific issue but I would also recommend not using this as a Workflow Id.

PR #2097, opened 2024-05-08
Refactor Versioning apis and UIs

Alex-Tideman:
## Description & motivation 💭 
This simplifies how we fetch and display versioning info. Now only the /workers and /task-queues/{queue} pages will fetch the versioning rules and compatibility. If either or both fail, show WorkerTable. 

`/workers` will check if the workflow is using versioning before fetching the rules and compatibilities. 
`/task-queues/{queue}` will fetch both always.

Both the rules and compatibility apis will swallow any errors and return undefined if they fail.

The Workflow Header will only show current BuildId of the workflow if using versioning.

NOTE: Rules api will always fail right now in OSS.

Screen Shot 2024-05-08 at 2 13 44 PM
Screen Shot 2024-05-08 at 2 13 54 PM
Screen Shot 2024-05-08 at 2 14 06 PM
Screen Shot 2024-05-08 at 2 14 18 PM
Screen Shot 2024-05-08 at 2 14 29 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/pages/task-queue.svelte
```diff
@@ -0,0 +1,13 @@
...
+  import { page } from '$app/stores';
+
+  import { getPollers } from '$lib/services/pollers-service';
+
+  import TaskQueueWorkers from './task-queue-workers.svelte';
+
+  $: ({ queue: taskQueue, namespace } = $page.params);
+</script>
+
+{#await getPollers({ queue: taskQueue, namespace }) then workers}
```
laurakwhit, line 11: Only thought is if we want some kind of loading state here?

laurakwhit (approved):

PR #2106, opened 2024-05-09
Simplify CSS variables for light and dark mode

stevekinney:
Keep two different objects of colors for light and dark mode can be a lot to hold in your head. Instead, let's just have one object that stores values for both light and dark modes.


src/lib/theme/plugin.ts
```diff
@@ -1,122 +1,13 @@
...
-} as const;
+import { colors } from './colors';
+import { css } from './utilities';
+import { dark, light } from './variables';
 
 const temporal = plugin(
   ({ addComponents, addBase }) => {
     addBase({
-      ':root': variables,
+      ':root': light,
```
laurakwhit, line 10: Something is off... Noticing alerts, the table header, side nav, and various other colors aren't showing as expected in day/night modes ⬇️ 

| Before | After |
|--|--|
|Screenshot 2024-05-13 at 12 10 40 PM|Screenshot 2024-05-13 at 12 10 48 PM|

Screenshot 2024-05-13 at 12 14 43 PM

laurakwhit (changes requested):

src/lib/theme/utilities.ts
```diff
@@ -11,5 +12,26 @@ export const rgb = (hexColor: `#${string}`): RGB => {
...
+export const isHexColor = (color: unknown): color is HexColor => {
+  if (typeof color !== 'string') return false;
+  return /^#[0-9A-F]{6}$/i.test(color);
+};
+
+export const isPaletteColor = (color: string): color is PaletteColor =>
+  color in palette;
+
+export const isShade = (shade: unknown): shade is Shade =>
+  typeof shade === 'number' || shade === 'DEFAULT';
```
laurakwhit, line 37: Since we're using `name.split('.')` on a string in `toColor` we likely want to account for a `string` here
```suggestion
  typeof shade === 'number' || typeof Number(shade) === 'number' || shade === 'DEFAULT';
```

src/lib/theme/variables.ts
```diff
@@ -0,0 +1,231 @@
...
+    light: 'primary',
+    dark: 'offWhite',
+  },
+  '--color-text-secondary': {
+    light: 'secondary',
+    dark: 'slate.200',
+  },
+  '--color-text-inverse': {
+    light: 'offWhite',
+    dark: 'primary',
```
laurakwhit, line 63: Looks like we didn't have a dark option for `--color-text-inverse` before, so it was always defaulting to `offWhite` ⬇️ 

```suggestion
    dark: 'offWhite',
```

laurakwhit (approved):

PR #2109, opened 2024-05-10
Add HTTP client

stevekinney:
Adds the new HTTP API client


Alex-Tideman (approved):

src/api/index.ts
```diff
@@ -0,0 +1,7 @@
+import { Client } from './client.js';
+
+const client = new Client();
+
+export const api = {
```
stevekinney, line 5: I don't feel strongly about this.

PR #2130, opened 2024-05-30
add filtering to api pagination

rossedfort:
## Description & motivation 💭 
- Adds a `filterable` prop to `<ApiPagination />` if present, will render an Input. The input is debounced and will refetch items after 1 second using the existing `onFetch` prop. Extends the `PaginatedRequest` type to take an optional `query` field, which can be used by callers

https://github.com/temporalio/ui/assets/11775628/456d9460-4105-4290-a600-3bf026ed28f0

## Testing 🧪 
- I pnpm linked ui to cloud-ui and tested against Namespaces and Users
### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/holocene/api-pagination.svelte
+++ b/src/lib/holocene/api-pagination.svelte
@@ -1,9 +1,11 @@
 <script lang="ts">
   import type { HTMLAttributes } from 'svelte/elements';
 
+  import debounce from 'just-debounce';
   import { onMount } from 'svelte';
 
   import Alert from '$lib/holocene/alert.svelte';
+  import Input from '$lib/holocene/input/input.svelte';
   import FilterSelect from '$lib/holocene/select/filter-select.svelte';
   import SkeletonTable from '$lib/holocene/skeleton/table.svelte';
   import {
@@ -29,11 +31,14 @@
     itemsKeyname?: string;
     previousButtonLabel: string;
     nextButtonLabel: string;
+    filterable?: boolean | undefined;
+    filterInputLabel?: string | undefined;
   }
 
   type PaginatedRequest<T> = (
     size: number,
     token: string,
+    query?: string,
   ) => Promise<{ items: T[]; nextPageToken: string }>;
 
   export let onError: (error: Error) => void | undefined = undefined;
@@ -52,6 +57,10 @@
   export let itemsKeyname = 'items';
   export let previousButtonLabel: string;
   export let nextButtonLabel: string;
+  export let filterable = false;
+  export let filterInputLabel: string = undefined;
+
+  let query = '';
 
   let store: PaginationStore<T> = createPaginationStore(
     pageSizeOptions,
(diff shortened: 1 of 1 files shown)


Alex-Tideman (approved):

src/lib/holocene/api-pagination.svelte
```diff
@@ -52,6 +57,10 @@
   export let itemsKeyname = 'items';
   export let previousButtonLabel: string;
   export let nextButtonLabel: string;
+  export let filterable = false;
+  export let filterInputLabel: string = undefined;
```
laurakwhit, line 61: What do you think about making this ⬇️ instead?
```suggestion
  export let filterInputPlaceholder: string = undefined;
```

src/lib/holocene/api-pagination.svelte
```diff
@@ -29,11 +31,14 @@
     itemsKeyname?: string;
     previousButtonLabel: string;
     nextButtonLabel: string;
+    filterable?: boolean | undefined;
+    filterInputLabel?: string | undefined;
```
laurakwhit, line 35: Since the input only shows up if `filterable && filterInputLabel`, can we make `filterInputLabel` required if `filterable` is `true`?

src/lib/holocene/api-pagination.svelte
```diff
@@ -29,11 +31,14 @@
     itemsKeyname?: string;
     previousButtonLabel: string;
     nextButtonLabel: string;
+    filterable?: boolean | undefined;
```
laurakwhit, line 34: We should replace the [custom input](https://github.com/temporalio/ui/blob/main/src/lib/pages/schedules.svelte#L84) on the Schedules page ⬇️  with this prop 🎉 

Screenshot 2024-06-03 at 11 21 50 AM

src/lib/holocene/api-pagination.svelte
```diff
@@ -175,11 +200,27 @@
...
-      <slot name="action-top-left" visibleItems={$store.visibleItems} />
-    </div>
+    {#if $$slots['action-top-left']}
+      <div class="flex shrink-0 items-center gap-1 lg:gap-2 xl:gap-3">
+        <slot name="action-top-left" visibleItems={$store.visibleItems} />
+      </div>
+    {/if}
+    {#if filterable && filterInputLabel}
+      <Input
+        id="api-pagination-search-input"
```
laurakwhit, line 210: Do we want a search icon here, similar to our other "search" inputs?
```suggestion
        icon="search"
        id="api-pagination-search-input"
```

src/lib/holocene/api-pagination.svelte
stevekinney: I could be talked into just using `placeholder` to keep it inline with HTML.

rossedfort: `<ApiPagination placeholder="whatever" />` doesn't make a ton of sense, cause it's not clear what the placeholder is for. I'll go with Laura's suggesiton.

src/lib/holocene/api-pagination.svelte
rossedfort: We should, I can make a follow up task for that

PR #2135, opened 2024-06-03
Audit a11y issues

stevekinney:
**For the `CodeBlock` component**: I could not find a single use of `CodeBlock` that had an `on:keydown` event, so it doesn't _appear_ that we're using it?

**For `filter-or-copy-buttons`**: These both appear to be `noop` functions?

### DIFF
--- a/src/lib/holocene/filter-or-copy-buttons.svelte
+++ b/src/lib/holocene/filter-or-copy-buttons.svelte
@@ -23,11 +23,7 @@
 </script>
 
 {#if show}
-  <div
-    class={merge('copy-or-filter', className)}
-    on:click|preventDefault|stopPropagation={noop}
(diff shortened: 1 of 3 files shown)


Alex-Tideman (approved):

src/lib/holocene/filter-or-copy-buttons.svelte
```diff
@@ -23,11 +23,7 @@
 </script>
 
 {#if show}
-  <div
-    class={merge('copy-or-filter', className)}
-    on:click|preventDefault|stopPropagation={noop}
-    on:keyup|preventDefault|stopPropagation={noop}
```
rossedfort, line 29: we probably had these for when the table rows used to be links I'm guessing

rossedfort (approved):

PR #2138, opened 2024-06-04
Use globalThis.accessToken() for codec server calls

Alex-Tideman:
## Description & motivation 💭 
Now with 15 minute token refreshes, we want to grab the access token for every request to codec (similar to request-from-api) instead of using the token from authUser after login since it can expire.

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


GiantRobots (approved):
Glorious

PR #2143, opened 2024-06-06
Update timeline workflow status colors

laurakwhit:
## Description & motivation 💭 
Updates `event-history-timeline` component to use the same colors that the `workflow-status` component uses with Tailwind's `theme()` function instead of hex codes.

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
`DT-2160`

### DIFF
--- a/src/lib/components/event/event-history-timeline.svelte
+++ b/src/lib/components/event/event-history-timeline.svelte
@@ -346,15 +346,15 @@
 
   :global(.vis-time-axis .vis-text) {
     font-size: 10px;
-    color: white !important;
+    color: theme(colors.white) !important;
   }
 
   :global(.vis-background, .vis-timeline) {
-    background-color: #000;
-    color: white;
+    background-color: theme(colors.black);
+    color: theme(colors.white);
     border-radius: 0.75rem;
     border-width: 2px;
-    border: 2px solid #000;
+    border: 2px solid theme(colors.black);
   }
 
   :global(.vis-content, .vis-group) {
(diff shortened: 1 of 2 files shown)


src/lib/components/event/event-history-timeline.svelte
```diff
@@ -478,72 +504,72 @@
...
   :global(.vis-item.vis-point.ContinuedAsNew) {
-    color: #18181b;
+    color: theme(colors.purple.300);
   }
 
   :global(.vis-item.vis-range.Completed, .vis-item.vis-range.Open) {
-    background-color: #dcfce7;
-    border-color: #15803d;
+    background-color: #71d375;
+    border-color: #71d375;
```
rossedfort, line 520: Could we use a theme color here? cc @Alex-Tideman

src/lib/components/event/event-history-timeline.svelte
```diff
@@ -478,72 +504,72 @@
...
     border-radius: 9999px;
     border-width: 2px;
-    color: #15803d;
+    color: theme(colors.black);
   }
 
   :global(.vis-item.vis-point.Completed, .vis-item.vis-point.Open) {
-    color: #15803d;
+    color: #71d375;
```
rossedfort, line 527: And here

rossedfort (approved):

src/lib/components/event/event-history-timeline.svelte
laurakwhit: That `#71d375` is actually a special green (from our brand colors I believe) and not from Tailwind. We're using it in `workflow-status` as well. Looks like we'll be changing it in the near future though, otherwise I would just add it in our tailwind config.

PR #2152, opened 2024-06-10
Dark mode audit style updates

laurakwhit:
## Description & motivation 💭 
* Update `Menu` component border color
* Add border and make "Cancel" a ghost button by default in the `Modal` component
* Update border color in `DayOfMonthPicker` component
* Audit and update `Badge` component types and their corresponding colors
* Add `interactive-table-hover` background color and update table rows to use it on hover
* Revert green status color back to `green-300` from `#71D375`
* Keep `off-white` background and `space-black` text color for Timeline in dark mode

| Before | After |
|--|--|
|Screenshot 2024-06-10 at 4 50 24 PM|Screenshot 2024-06-10 at 4 50 42 PM|
|Screenshot 2024-06-10 at 4 50 02 PM|Screenshot 2024-06-10 at 4 49 35 PM|
|Screenshot 2024-06-10 at 4 24 15 PM|Screenshot 2024-06-10 at 4 48 40 PM|

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
`DT-2200`
`DT-2202`
`DT-2207`
`DT-2208`

### DIFF
--- a/src/lib/holocene/link.svelte
+++ b/src/lib/holocene/link.svelte
@@ -16,7 +16,7 @@
     class?: string;
     icon?: IconName;
     text?: string;
-    inverse?: boolean;
+    light?: boolean;
     'data-testid'?: string;
   };
 
@@ -27,7 +27,7 @@
   export let newTab = false;
   export let icon: IconName = null;
   export let text: string = '';
-  export let inverse = false;
+  export let light = false;
 
   const onLinkClick = (e: MouseEvent) => {
     // Skip if middle mouse click or new tab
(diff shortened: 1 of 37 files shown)


Alex-Tideman:
One thing we could probably throw in here is the <Link inverse /> property. We should make that maybe <Link white instead so it's always white text instead of inverse since the background is always the same color regardless of the dark/light mode.


Screenshot 2024-06-11 at 3 53 48 PM

Alex-Tideman (approved):

src/lib/holocene/link.svelte
```diff
@@ -66,8 +66,8 @@
       @apply text-brand;
     }
 
-    &.inverse {
-      @apply text-inverse hover:text-brand focus-visible:text-brand;
+    &.light {
+      @apply text-off-white;
```
laurakwhit, line 70: @Alex-Tideman changed the `inverse` prop to `light` here.

PR #2155, opened 2024-06-11
Schedules Count

Alex-Tideman:
## Description & motivation 💭 
Query schedule workflows to give a count of Schedules

Screenshot 2024-06-11 at 2 03 15 PM

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/schedule/schedules-count.svelte
+++ b/src/lib/components/schedule/schedules-count.svelte
@@ -1,0 +1,88 @@
+<script lang="ts">
+  import { onDestroy } from 'svelte';
+
+  import { page } from '$app/stores';
+
+  import Skeleton from '$lib/holocene/skeleton/index.svelte';
+  import { translate } from '$lib/i18n/translate';
+  import { fetchScheduleCount } from '$lib/services/workflow-counts';
+  import { getExponentialBackoffSeconds } from '$lib/utilities/refresh-rate';
+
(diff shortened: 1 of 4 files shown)


src/lib/components/schedule/schedules-count.svelte
```diff
@@ -0,0 +1,36 @@
...
+  $: namespace, fetchCounts();
+</script>
+
+<div class="flex flex-wrap items-center gap-2">
+  {#if !loading}
+    {$schedulesCount}
+  {:else}
+    <Skeleton class="h-6 w-6 rounded" />
+  {/if}
+  {translate('common.schedules')}
```
laurakwhit, line 35: What do you think about doing something similar to what we have for `Workflows` here and adding a plural translation? E.g. updating this ⬇️ 

```suggestion
  <Translate key="common.schedules-plural" count={Number($schedulesCount)} />
```
and adding the following to `common.ts`
```js
  'schedules-plural_one': 'Schedule',
  'schedules-plural_other': 'Schedules',
```

so that we have `1 Schedule` instead of `1 Schedules`

Screenshot 2024-06-11 at 3 01 50 PM

Alex-Tideman: 👍

laurakwhit (approved):

PR #2171, opened 2024-06-27
Print raw json

Alex-Tideman:
## Description & motivation 💭 
When `?json` searchParam is added to a Workflow Execution URL, print the Workflow, Pollers, and Event History data as plain text. 

Helpful for debugging large events. Payloads will not be decoded.

Screenshot 2024-06-27 at 4 42 22 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/layouts/workflow-run-layout.svelte
+++ b/src/lib/layouts/workflow-run-layout.svelte
@@ -24,6 +24,7 @@
   import type { NetworkError } from '$lib/types/global';
 
   $: ({ namespace, workflow: workflowId, run: runId } = $page.params);
+  $: raw = $page.url.searchParams.has('json');
 
   let workflowError: NetworkError;
   let eventHistoryController: AbortController;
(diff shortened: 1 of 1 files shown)


src/lib/layouts/workflow-run-layout.svelte
```diff
@@ -133,32 +134,42 @@
...
 </script>
 
-<LabsModeGuard>
-  <div
-    class="absolute bottom-0 left-0 right-0 {$viewDataEncoderSettings
-      ? 'top-[540px]'
-      : 'top-0'}
+{#if raw}
+  <div class="h-auto whitespace-break-spaces rounded bg-primary p-4">
+    {JSON.stringify(
```
laurakwhit, line 139: Would it be useful to add a copy button here so users can copy the JSON and inspect it elsewhere?

Alex-Tideman: Yeah that's a good idea. These can be pretty huge (~50mb) but there's no limit to copyboard copy size.

src/lib/layouts/workflow-run-layout.svelte
```diff
@@ -133,32 +144,45 @@
...
-<LabsModeGuard>
-  <div
-    class="absolute bottom-0 left-0 right-0 {$viewDataEncoderSettings
-      ? 'top-[540px]'
-      : 'top-0'}
+{#if raw}
+  <div class="relative h-auto whitespace-break-spaces rounded bg-primary p-4">
+    <CopyButton
+      copyIconTitle="Copy JSON to clipboard"
+      copySuccessIconTitle="JSON copied to clipboard"
```
laurakwhit, line 151: We could also just use the existing copy for copy icons here if you wanted

```suggestion
      copyIconTitle={translate('common.copy-icon-title')}
      copySuccessIconTitle={translate('common.copy-success-icon-title')}
```

src/lib/layouts/workflow-run-layout.svelte
```diff
@@ -22,12 +23,22 @@
     workflowRun,
   } from '$lib/stores/workflow-run';
   import type { NetworkError } from '$lib/types/global';
+  import { copyToClipboard } from '$lib/utilities/copy-to-clipboard';
+  import { stringifyWithBigInt } from '$lib/utilities/parse-with-big-int';
 
   $: ({ namespace, workflow: workflowId, run: runId } = $page.params);
+  $: raw = $page.url.searchParams.has('json');
```
laurakwhit, line 30: Wondering if we want something slightly more descriptive here for the variable name, e.g. ⬇️ 
```suggestion
  $: showJSON = $page.url.searchParams.has('json');
```

src/lib/layouts/workflow-run-layout.svelte
```diff
@@ -133,32 +144,45 @@
   });
 </script>
 
-<LabsModeGuard>
-  <div
-    class="absolute bottom-0 left-0 right-0 {$viewDataEncoderSettings
-      ? 'top-[540px]'
-      : 'top-0'}
+{#if raw}
+  <div class="relative h-auto whitespace-break-spaces rounded bg-primary p-4">
```
laurakwhit, line 148: Might want to add `overflow-x` 
```suggestion
  <div
    class="relative h-auto overflow-x-auto whitespace-break-spaces rounded bg-primary p-4"
  >
```
or `break-words`
```suggestion
  <div
    class="relative h-auto whitespace-break-spaces break-words rounded bg-primary p-4"
  >
```
for small screens ⬇️ 

Screenshot 2024-06-27 at 4 16 36 PM

laurakwhit (approved):
Some small comments below, but otherwise lgtm!

Also thought I'd note it might be a little confusing when there are other search params (e.g. `sort=descending&category=workflow&json`) since none of them are applied to the JSON. Might we want to consider applying filters to the JSON in the future?

src/lib/layouts/workflow-run-layout.svelte
Alex-Tideman: 👍

src/lib/layouts/workflow-run-layout.svelte
Alex-Tideman: Good call

src/lib/layouts/workflow-run-layout.svelte
Alex-Tideman: 👍

Alex-Tideman:
This is really made for support, and def can think about adding filters in future. But for now the use case is just the ability to see the full JSON from a url. I don't see support doing filters and such and then adding ?json.

PR #2197, opened 2024-07-10
Mobile Nav

Alex-Tideman:
## Description & motivation 💭 
To improve our mobile experience and give more horizontal real estate when on small screens.

Besides adding a Bottom Nav, there's been several other changes:

1. Pass in the linkList as a prop with all the necessary information to render it - `href, icon, label, tooltip, divider, external, hidden, animate`. There are no more slots, each global layout is responsible for providing the linkList and the Navs (side/bottom) will render them for you in the order providered.
2. The Namespace switcher is available on all pages now. It will show the current namespace or the last used namespace. This allows you to jump to a namespace whenever in the app.
3. There is also some smaller general UI updates to improve responsiveness and dark mode.

Screenshot 2024-07-12 at 9 32 13 AM
Screenshot 2024-07-12 at 9 32 23 AM
Screenshot 2024-07-12 at 9 32 44 AM
Screenshot 2024-07-12 at 9 32 52 AM
Screenshot 2024-07-12 at 9 32 59 AM
Screenshot 2024-07-12 at 9 33 48 AM

### How was this tested 👻 

- [x] Manual testing
- [x] E2E tests added
- [ ] Unit tests added

### Merge Checklist 

- [ ] Better slide transition?


src/lib/components/bottom-nav-links.svelte
```diff
@@ -0,0 +1,25 @@
...
+  import NavigationItem from '$lib/holocene/navigation/navigation-item.svelte';
+  import type { NavLinkListItem } from '$lib/types/global';
+
+  export let open = false;
+  export let linkList: NavLinkListItem[];
+</script>
+
+{#if open}
+  <div class="flex h-full flex-col justify-start gap-4">
+    {#each [...linkList].reverse() as item}
```
rossedfort, line 11: I think we could just do this with css with `flex-col-reverse`

src/lib/components/bottom-nav-settings.svelte
```diff
@@ -0,0 +1,88 @@
...
+      tooltip={labsHoverText}
+      label={labsText}
+      icon="labs"
+      active={$labsMode}
+      data-testid="labs-mode-button"
+    />
+    {#if $authUser.accessToken}
+      <div class="border-b-2 border-subtle" />
+      <NavigationButton
+        onClick={() => userEmaiLink && goto(userEmaiLink)}
```
rossedfort, line 73: `NavigationButton` should probably accept an `href` so that we can render an anchor tag here, or can we use `NavigationItem` ?

rossedfort (commented):
Looks good to me, just a few small suggestions

src/lib/components/bottom-nav-links.svelte
Alex-Tideman: Good call.

src/lib/components/bottom-nav-settings.svelte
Alex-Tideman: 👍 I should use NavigationItem

rossedfort (approved):

PR #2215, opened 2024-07-17
Display Local Activity, move Close button to right

Alex-Tideman:
## Description & motivation 💭 
Previously in Lines/Dots we displayed Local Activity as Marker Recorded. This updates the labels to Local Activity.

Also move Close button in Full History view to far right 

Screenshot 2024-07-17 at 3 54 41 PM
Screenshot 2024-07-17 at 3 54 48 PM
Screenshot 2024-07-17 at 3 54 58 PM

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/lines-and-dots/constants.ts
+++ b/src/lib/components/lines-and-dots/constants.ts
@@ -300,13 +300,13 @@ export const getCategoryColor = (type: EventTypeCategory): string => {
   switch (type) {
     case 'command':
     case 'marker':
-    case 'local-activity':
       return '#ebebeb';
     case 'timer':
       return '#fbbf24';
     case 'signal':
       return '#ec4899';
     case 'activity':
+    case 'local-activity':
       return '#a78bfa';
     case 'workflow':
       return '#059669';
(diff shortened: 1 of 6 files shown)


src/lib/components/lines-and-dots/constants.ts
```diff
@@ -300,13 +300,13 @@ export const getCategoryColor = (type: EventTypeCategory): string => {
...
     case 'marker':
-    case 'local-activity':
       return '#ebebeb';
     case 'timer':
       return '#fbbf24';
     case 'signal':
       return '#ec4899';
     case 'activity':
+    case 'local-activity':
       return '#a78bfa';
```
laurakwhit, line 310: Could we add an `import { colors } from '$lib/theme/colors'` to this component and use `colors` here instead?

src/lib/models/event-groups/get-group-name.ts
```diff
@@ -78,6 +78,10 @@ export const getEventGroupLabel = (event: CommonHistoryEvent): string => {
     return 'Signal received';
   }
 
+  if (isLocalActivityMarkerEvent(event)) {
+    return 'Local Activity';
```
laurakwhit, line 82:
```suggestion
    return translate('events.category.local-activity');
```

laurakwhit (approved):
A few small comments, but otherwise lgtm 🚀 I am curious why that holocene build failed though 🤔

PR #2225, opened 2024-07-22
Small CodeBlock in Modal fixes

laurakwhit:
## Description & motivation 💭 
We're seeing the occasional `RangeError: Invalid change range [num] to [num] (in doc of length [num])` error coming from `@codemirror` and it looks like it's happening for `Codeblock`s in `Modal` components. Haven't been able to reproduce, but did notice that this ⬇️  was being called twice on close in the signal confirmation modal. 

https://github.com/temporalio/ui/blob/50a5e10e55c3ea7b0771e58360b2bb994c0615aa/src/lib/components/workflow/client-actions/signal-confirmation-modal.svelte#L28

This PR calls `closeModal` instead of `handleCancel` on close buttons in the `Modal` component, so that `open` is updated to `false` > `triggerModal` is called > then `modal.close()` is called and finally `handleCancel` is only called/dispatched once. This PR also adds a `destroy()` for the `@codemirror` editor view to the `CodeBlock` component unmount.

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/holocene/modal.svelte
+++ b/src/lib/holocene/modal.svelte
@@ -66,8 +66,12 @@
     dispatch('confirmModal');
   };
 
+  const closeModal = () => {
+    open = false;
+  };
+
   const handleClick = (event: MouseEvent) => {
-    if (event.target === modalElement) handleCancel();
+    if (event.target === modalElement) closeModal();
   };
 
   $: {
@@ -97,7 +101,7 @@
       label={cancelText}
       icon="close"
       class="float-right m-4"
-      on:click={handleCancel}
+      on:click={closeModal}
     />
   {/if}
   <div id="modal-title-{id}" class="title">
@@ -115,7 +119,7 @@
       </p>
     </div>
     <div class="flex items-center justify-end space-x-2 p-6">
-      <Button variant="ghost" disabled={loading} on:click={handleCancel}
+      <Button variant="ghost" disabled={loading} on:click={closeModal}
         >{cancelText}</Button
       >
       {#if !hideConfirm}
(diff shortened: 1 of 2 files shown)


src/lib/holocene/modal.svelte
```diff
@@ -66,8 +66,12 @@
     dispatch('confirmModal');
   };
 
+  const closeModal = () => {
+    open = false;
```
Alex-Tideman, line 70: What did the dispatch('cancelModal') do?

laurakwhit: In the case of the signal confirmation modal it calls `hideSignalModal` ⬇️ 
https://github.com/temporalio/ui/blob/50a5e10e55c3ea7b0771e58360b2bb994c0615aa/src/lib/components/workflow/client-actions/signal-confirmation-modal.svelte#L70

which calls `resetView()` on the code block
https://github.com/temporalio/ui/blob/50a5e10e55c3ea7b0771e58360b2bb994c0615aa/src/lib/components/workflow/client-actions/signal-confirmation-modal.svelte#L28

Alex-Tideman: But like for all modals won't that not be called now?

laurakwhit: It will still be called because we are updating `open` to `false` which triggers `toggleModal` which then calls `modal?.close()`
https://github.com/temporalio/ui/blob/50a5e10e55c3ea7b0771e58360b2bb994c0615aa/src/lib/holocene/modal.svelte#L44-L52

which then calls `handleCancel`
https://github.com/temporalio/ui/blob/50a5e10e55c3ea7b0771e58360b2bb994c0615aa/src/lib/holocene/modal.svelte#L84

Alex-Tideman: 🫠 ohhh Modals

Alex-Tideman (approved):

Issue #2227, opened 2024-07-23
Forward Authorization-Extras header to Codec Server


omm-mishra:
## Is your feature request related to a problem? Please describe.

In our Temporal setup we use OIDC with Azure as provider for authentication, the roles and metadata of the users for the application are present in the "Authorization-Extras" header which the web forwards to the frontend for authorization. But the web does not forward this header to the codec server, only the "Authorization" header.

We use the roles to restrict access to users to only their registered namespaces.

## Describe the solution you'd like

We would like the "Authorization-Extras" header to be forwarded as well, to the codec server when "Pass the user access token with your endpoint." option is selected.

ivanmartinvalle:
We can close this right? #2250 looks to have implemented the feature.

omm-mishra:
Yes

PR #2228, opened 2024-07-23
added Authorization-Extras header for codec server

omm-mishra:
## Description & motivation 💭 
The "Authorization-Extras" header is not forwarded to the codec-server when pass access token option is selected. In our Temporal set up we use this header to extract roles and restrict namespace access, we would like to implement this with our codec server as well, this change enables the "Authorization-Extras" header to be forwarded. Have opened an issue as well for this: https://github.com/temporalio/ui/issues/2227

### How was this tested 👻 
- [X] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Any docs updates needed?


Alex-Tideman:
Closing for https://github.com/temporalio/ui/pull/2250

PR #2230, opened 2024-07-24
Fix WorkflowTaskFailed Alert when Cause is not provided

Alex-Tideman:
## Description & motivation 💭 
An API update caused the 'Cause' field on WorkflowTaskFailed to not return at all when previously it returned "cause": null. This then prevented the UI from rendering WorkflowTaskFailed error when there was no cause. This PR fixes that to default to 'Unspecified' if no cause. It also updates the WorkflowTaskFailed Alert to be more informative.

**Before**
Screenshot 2024-07-24 at 12 50 08 PM

**After**
Screenshot 2024-07-24 at 12 47 35 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/components/lines-and-dots/workflow-error.svelte
```diff
@@ -29,7 +29,7 @@
       return 'WorkflowTaskHeartbeatError';
     }
 
-    return toWorkflowTaskFailureReadable(cause);
+    return toWorkflowTaskFailureReadable(cause || 'Unspecified');
```
rossedfort, line 32: Why not just have `toWorkflowTaskFailureReadable` do this if `cause` is `undefined`?

Alex-Tideman: Good call. Added unit test for it

rossedfort (approved):

PR #2236, opened 2024-07-24
Text styles audit

laurakwhit:
## Description & motivation 💭 
* Adds `textStyles` plugin with
   * default font sizing and weight for heading elements
   * body text style classes
* Audits existing heading elements for font sizing and weight
* Sets `sans` font family (instead of `primary` vs. `secondary`)
* Removes set icon size in `Link` component
### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
`DT-2000`


src/app.css
```diff
@@ -15,7 +15,9 @@
 
   body {
     @apply surface-background;
-    font-family: theme('fontFamily.primary');
+    font-family: theme('fontFamily.sans');
+    font-size: theme('fontSize.sm');
```
rossedfort, line 19: One thing we could consider is adjusting our tailwind theme so this is size is `base` and remove anything smaller that we don't need 🤔

src/app.css
```diff
@@ -25,10 +27,56 @@
...
+    }
+  }
+
+  h6 {
+    font-size: theme('fontSize.xs');
+
+    @media (min-width: theme('screens.md')) {
+      font-size: theme('fontSize.sm');
+    }
   }
```
rossedfort, line 81: Might be nice to add some helper classes here for emphasized or other types of frequently used text styles, but not a blocker for this PR.

rossedfort (approved):

src/app.css
laurakwhit: We still have a bunch of places that use/need `text-xs`. Not quite sure how we want to handle that... I've left them as is for now.

rossedfort (approved):

PR #2238, opened 2024-07-26
use order names

bgerrity:
These are more conventional than `Sort`.


Alex-Tideman:
Fixed in https://github.com/temporalio/ui/pull/2269

PR #2239, opened 2024-07-26
Fix editable CodeBlock

laurakwhit:
## Description & motivation 💭 
Sets CodeMirror `EditorView.editable` based on `editable` prop in `CodeBlock` component in addition to `EditorState.readOnly`.

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "@codemirror/language": "^6.10.2",
     "@codemirror/legacy-modes": "^6.4.0",
     "@codemirror/state": "^6.4.1",
-    "@codemirror/view": "^6.28.5",
+    "@codemirror/view": "^6.29.0",
     "@crownframework/svelte-error-boundary": "^1.0.3",
     "@fontsource-variable/inter": "^5.0.8",
     "@fontsource/noto-sans-mono": "^5.0.9",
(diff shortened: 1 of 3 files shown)


rossedfort (approved):

Alex-Tideman:
Kinda confused by this, is this a bug in CodeMirror? We shouldn't need two configurations to make it readOnly.

Also not seeing this same behavior on OSS. Do we just need to set this on Cloud?

rossedfort:
> Kinda confused by this, is this a bug in CodeMirror? We shouldn't need two configurations to make it readOnly.
> 
> Also not seeing this same behavior on OSS. Do we just need to set this on Cloud?

I think it was a bug in CodeMirror https://github.com/codemirror/dev/issues/1411

Also, it appears `readonly` and `editable` are actually 2 different things according to CodeMirror. If you look at a [Holocene preview link before the bug was introduced](https://holocene-h4qiwzr2c.preview.thundergun.io/?path=/story/code-block--default), notice you can click inside the CodeBlock to drop a cursor, but you can't type/delete anything. If you look at the [Holocene preview link for this PR](https://holocene-git-fix-editable-code-block.preview.thundergun.io/?path=/story/code-block--default), notice you can't even drop a cursor in the CodeBlock.

We probably could just use `EditorView.editable.of(editable)` instead of `EditorState.readOnly.of(!editable)`, but I think this fine.

PR #2240, opened 2024-07-29
Dark slider

GraceGardner:
## Description & motivation 💭 
Adjust Range input styling to match new designs. 

Screenshot 2024-07-29 at 9 39 10 AM
Screenshot 2024-07-29 at 9 39 18 AM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
DT-2061

### DIFF
--- a/src/lib/services/data-encoder.test.ts
+++ b/src/lib/services/data-encoder.test.ts
@@ -1,0 +1,81 @@
+import { describe, expect, it, vi } from 'vitest';
+
+import { codeServerRequest } from './data-encoder';
+
+const settings = {
+  codec: {
+    endpoint: 'http://localcodecserver.com',
+    passAccessToken: false,
+    includeCredentials: false,
+  },
+};
+
+describe('Codec Server Requests for Decode and Encode', () => {
+  const payloads = { payloads: [{}] };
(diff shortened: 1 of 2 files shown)


KATIETOLER (approved):
I think this looks great!

src/lib/theme/plugin.ts
```diff
@@ -76,6 +76,9 @@ const temporal = plugin(
           backgroundColor: css('--color-interactive-ghost-active'),
         },
       },
+      '.surface-information': {
+        backgroundColor: css('--color-surface-information'),
```
laurakwhit, line 80: Just wanted to follow-up here. Was there a reason `surface-information` was added here? Looks like we're just using `backgroundColor` and that already exists ⬇️ 

https://github.com/temporalio/ui/blob/5f41480f10ace77f9ee0e31102402786583ca606/src/lib/theme/plugin.ts#L131

laurakwhit: I've gone ahead and removed it [here](https://github.com/temporalio/ui/pull/2248/commits/354df4f9b8f002c78797c53e83e9301929a4a558).

PR #2241, opened 2024-07-29
Schedule search attributes

laurakwhit:
## Description & motivation 💭 
Adds ability to set custom search attribute values on a schedule via the UI.

| Create | View | Edit |
|--|--|--|
|Screenshot 2024-07-29 at 5 25 05 PM|Screenshot 2024-08-05 at 4 10 39 PM|Screenshot 2024-08-02 at 6 40 43 PM|

| Before | After |
|--|--|
|Screenshot 2024-08-05 at 3 56 21 PM|Screenshot 2024-08-05 at 4 08 06 PM|

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

When creating or editing a schedule 
   - [ ] Verify a `Bool` attribute can be added
   - [ ] Verify a `Datetime` attribute can be added (and is saved in UTC)
   - [ ] Verify a `Double` attribute can be added
   - [ ] Verify an `Int` attribute can be added
   - [ ] Verify a `Keyword` attribute can be added
   - [ ] Verify a `Text` attribute can be added
   - [ ] Verify no `KeywordList` attributes show up as options in the attribute select
   - [ ] Verify `+ Add a search attribute` is disabled if
      - there are no more search attributes available
      - an attribute does not have a value set

When editing a search attribute on a schedule
   - [ ] Verify changing the attribute clears the value
   - [ ] Verify `x` removes the attribute
   - [ ] Verify changing the value is correctly reflected on `Save`
   - [ ] Verify there is an error if `Save` is attempted and an attribute does not have a value

--- 
When starting a workflow 
   - [ ] Verify search attributes follow same criteria as above ⬆️ 
### Issue(s) closed 
`DT-2327`


src/lib/pages/schedule-view.svelte
```diff
@@ -393,6 +394,7 @@
             policies={schedule?.schedule?.policies}
             notes={schedule?.schedule?.state?.notes}
           />
+          <ScheduleSearchAttributes schedule={schedule?.schedule} />
```
laurakwhit, line 397: Do we want to list search attributes in the schedule view? Or should it just be on create/edit?

src/lib/components/workflow/search-attribute-input/index.svelte
```diff
@@ -0,0 +1,96 @@
...
+      bind:value={attribute.value}
+    >
+      <Option value={true}>{translate('common.true')}</Option>
+      <Option value={false}>{translate('common.false')}</Option>
+    </Select>
+  {:else if type === 'Datetime'}
+    <Input
+      label={translate('common.value')}
+      id="attribute-value"
+      type="datetime-local"
```
laurakwhit, line 81: For type `Datetime` attributes, this is currently being set based on local time. Should this be based on UTC instead? Maybe we want to make use of our `DatePicker` and `TimePicker` components here instead of using `type="datetime-local"`.

src/lib/components/workflow/search-attribute-input/number-input.svelte
```diff
@@ -0,0 +1,18 @@
+<script lang="ts">
+  import NumberInput from '$lib/holocene/input/number-input.svelte';
+  import { translate } from '$lib/i18n/translate';
+  import type { SearchAttributeInputValue } from '$lib/stores/search-attributes';
+
+  export let value: SearchAttributeInputValue;
+  let _value = value ? Number(value) : null;
```
laurakwhit, line 7: Made separate components here for `NumberInput` and `TextInput` to make types happy, but definitely open to ideas on alternatives.

src/lib/components/workflow/search-attribute-input/index.svelte
```diff
@@ -0,0 +1,96 @@
...
+
+  export let attributesToAdd: SearchAttributeInput[] = [];
+  export let searchAttributes: SearchAttributes = $customSearchAttributes;
+  export let attribute: SearchAttributeInput;
+  export let onRemove: (attribute: string) => void;
+
+  $: type = searchAttributes[attribute.attribute];
+  $: searchAttributesOptions = [...Object.entries(searchAttributes)]
+    .map(([key, value]) => ({ label: key, value: key, type: value }))
+    .filter(({ type }) => type !== 'KeywordList');
```
laurakwhit, line 27: Do we want to support `KeywordList` attributes here? Not sure what that would look like, maybe a [ChipInput](https://holocene.preview.thundergun.io/?path=/story/chip-input--default)?

Alex-Tideman: I think we'll skip KeywordList for now and do a full update of that everywhere when ready including the Workflow Search UI

src/lib/components/workflow/search-attribute-input/index.svelte
Alex-Tideman: This is a great question. Ya I think the Date/TimePicker would definitely be better here

src/lib/components/schedule/schedule-form-view.svelte
```diff
@@ -195,6 +207,10 @@
         payloads={schedule?.action?.startWorkflow?.input}
         error={errors['input']}
       />
+      <AddSearchAttributes
```
laurakwhit, line 210: Do we want to add this UI yet? Or should we wait until filtering by search attribute is supported?

Alex-Tideman (approved):

PR #2242, opened 2024-07-30
remove expand when slot is empty

GraceGardner:
## Description & motivation 💭 
Removes hover states, chevron and ability to expand accordion when slot is empty. 

## Testing 🧪 
Storybook 

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

Please checkout the accordion component in storybook 

### Issue(s) closed 
DT-2312


server/go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/temporalio/ui-server/v2
 
-go 1.20
+go 1.21
+
+toolchain go1.22.4
```
laurakwhit, line 5: Definitely don't want this change in here. Same with changes in `server/go.sum`.

```suggestion
go 1.20
```

laurakwhit (changes requested):

src/lib/holocene/accordion.svelte
```diff
@@ -59,14 +65,16 @@
...
-        <slot name="action" />
-      </div>
-      <Icon class="m-2" name={open ? 'chevron-up' : 'chevron-down'} />
+      {#if $$slots.default}
+        <div
+          class="flex flex-row items-center gap-2 pr-2"
+          on:click|stopPropagation
+          on:keyup|stopPropagation
+        >
+          <slot name="action" />
```
laurakwhit, line 74: We may still want the action slot even if there is no default slot (there can be links, etc. in it). 

Maybe we move the `#if $$slots.default` conditional to just be for the `Icon`?

GraceGardner: That's a good point! Thanks Laura!

server/go.mod
GraceGardner: I don't know how I did that. Thanks for catching this.

laurakwhit: @GraceGardner looks like changes to `server/go.sum` still need to be removed.

laurakwhit (approved):

PR #2251, opened 2024-08-01
Extra Test Id's for easy testing

GiantRobots:
### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/holocene/select/option.svelte
+++ b/src/lib/holocene/select/option.svelte
@@ -5,6 +5,7 @@
     description?: string;
     disabled?: boolean;
     class?: string;
+    ['data-testid']?: string;
   }
 
   export const EMPTY_OPTION: OptionType<string> = {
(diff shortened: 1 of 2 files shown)


src/lib/holocene/select/option.svelte
```diff
@@ -5,6 +5,7 @@
     description?: string;
     disabled?: boolean;
     class?: string;
+    ['data-testid']?: string;
```
Alex-Tideman, line 8: Wouldn't this just be 'data-testid'?: string;

Alex-Tideman: Or just not have it here if it's used on $$restProps

Alex-Tideman (approved):

src/lib/holocene/select/option.svelte
GiantRobots: I think I wanna leave it because it's something I want to be explicitly in the definition, didn't know how to access a variable with a - in int without restProps tho

Alex-Tideman: 🪃 🪃 🪃 🪃 🪃

PR #2252, opened 2024-08-02
Revert "Update base images"

Alex-Tideman:
Reverts temporalio/ui#2226


Alex-Tideman:
Reverting due to missing Dockerize dependency in base image

PR #2256, opened 2024-08-07
Nexus Callback Information

Alex-Tideman:
## Description & motivation 💭 
When a workflow has a Nexus callback (only one for now, could be more in the future), show an Alert that provides the Nexus URL, the state, attempt, timestamps and failure if it exists.

Updates protos and APIs for new Nexus protos.

Also includes correctly showing pending status for Pending Nexus Operations. I will do a follow-up PR to include the Pending Nexus Operations information when the group/event is clicked.

Screenshot 2024-08-07 at 12 48 39 PM
Screenshot 2024-08-07 at 12 49 31 PM
Screenshot 2024-08-07 at 1 00 00 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added


src/lib/components/lines-and-dots/workflow-callback.svelte
```diff
@@ -0,0 +1,73 @@
...
+  import { translate } from '$lib/i18n/translate';
+  import { timeFormat } from '$lib/stores/time-format';
+  import type { CallbackInfo } from '$lib/types';
+  import { formatDate } from '$lib/utilities/format-date';
+
+  export let callback: CallbackInfo;
+
+  $: completedTime = formatDate(callback.lastAttemptCompleteTime, $timeFormat);
+  $: nextTime = formatDate(callback.nextAttemptScheduleTime, $timeFormat);
+  $: failure = callback?.lastAttemptFailure?.message;
```
stevekinney, line 13: Does this need to be reactive?

src/lib/components/lines-and-dots/workflow-callback.svelte
```diff
@@ -0,0 +1,73 @@
...
+
+  const titles = {
+    Standby: translate('nexus.callback.standby'),
+    Scheduled: translate('nexus.callback.scheduled'),
+    'Backing Off': translate('nexus.callback.backing-off'),
+    Failed: translate('nexus.callback.failed'),
+    Succeeded: translate('nexus.callback.succeeded'),
+  };
+
+  $: title = titles[callback.state] || translate('nexus.nexus-callback');
```
stevekinney, line 23: Same comment here—I'd think that a referentially different `callback` property would trigger a re-render anyway?

stevekinney (approved):

src/lib/components/lines-and-dots/workflow-callback.svelte
Alex-Tideman: It might go from no message to message or the message might change

PR #2260, opened 2024-08-09
Upgrade api to v1.37.0

yycptt:
### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


server/go.mod
```diff
@@ -1,17 +1,19 @@
 module github.com/temporalio/ui-server/v2
 
-go 1.20
+go 1.21
+
+toolchain go1.22.1
```
yycptt, line 5: hmm looks like I need to use golang 1.20 on my local to update the api version?

yycptt:
Looks like there's no need to eagerly update the dependency

PR #2273, opened 2024-08-20
Fix the wrong nano conversion in timestampToDate

vshy108:
the date timestamp in unit of milliseconds, hence nano seconds should divide by 10^6 to provide milliseconds value instead of the previous 10^3


antlai-temporal (approved):
Thanks a lot!

Alex-Tideman (approved):

PR #2295, opened 2024-08-29
Make split card have nicer width styles and support hrefs

GiantRobots:
Need href to support certain layouts. Prefer `min-w-fit` over `w-full` to give more control to sub elements in the menu.


Alex-Tideman:
Should we add the external-link icon if there's an href? Not sure what the text is when you have an href but it'd be nice to know you are linking out

Screenshot 2024-08-29 at 3 07 02 PM

GiantRobots:
> Should we add the external-link icon if there's an href? Not sure what the text is when you have an href but it'd be nice to know you are linking out
> 
> ```
> {#if icon}
>    <Icon name={icon} />
>    {:else if href}
>    <Icon name="external-link" />
>  {/if}
> ```
> 
> Screenshot 2024-08-29 at 3 07 02 PM

Hmm we could try detecting if it's an external link. The href I'm using it for is internal and designs don't have that. Also it makes sense because it's the "edit" me button that should just navigate you away. 

Could detect a non base url or protocol to add the external link icon

Alex-Tideman:
Ahhhh more links as buttons. then no icon for sure

Alex-Tideman (approved):

PR #2299, opened 2024-09-03
Bidirectional Links

Alex-Tideman:
## Description & motivation 💭 
Add support to map links to event history and show as link in the summary row. For now there will only every be a single link, but when expanded support multiple links in the future. Include nexusEnabled check to add Nexus filter to event type filter.

Also updates the temporal api version to 1.28.0, which changes the Nexus endpoint regex to use dashes instead of underscores.

Screenshot 2024-09-03 at 8 35 14 AM
Screenshot 2024-09-03 at 8 35 42 AM
Screenshot 2024-09-03 at 9 26 11 AM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


package.json
```diff
@@ -137,13 +137,13 @@
...
-    "@temporalio/testing": "^1.10.3",
-    "@temporalio/worker": "^1.10.3",
-    "@temporalio/workflow": "^1.10.3",
+    "@temporalio/activity": "1.10.3",
+    "@temporalio/client": "1.10.3",
+    "@temporalio/common": "1.10.3",
+    "@temporalio/proto": "1.10.3",
+    "@temporalio/testing": "1.10.3",
+    "@temporalio/worker": "1.10.3",
+    "@temporalio/workflow": "1.10.3",
```
Alex-Tideman, line 146: Locking this dependency because the newest (v1.11.1) has incorrect Nexus description types

src/lib/components/event/event-links-expanded.svelte
```diff
@@ -0,0 +1,42 @@
+<script lang="ts">
+  import Link from '$lib/holocene/link.svelte';
+  import type { EventLink } from '$lib/types/events';
+  import { routeForEventHistory } from '$lib/utilities/route-for';
+
+  export let links: EventLink[] = [];
+</script>
+
+{#if links?.length}
```
rossedfort, line 9: I think this is redundant because `{#each}` won't iterate/render anything for an empty array, but nbd.

src/lib/components/event/event-link.svelte
```diff
@@ -0,0 +1,37 @@
...
+</div>
+
+<style lang="postcss">
+  .badge {
+    @apply surface-subtle overflow-hidden rounded-sm p-1;
+  }
+
+  .payload {
+    @apply overflow-hidden rounded bg-space-black px-1 py-0.5 font-mono text-xs text-white;
+  }
```
rossedfort, line 36: I don't see this class used in the component

src/lib/components/event/event-link.svelte
```diff
@@ -0,0 +1,37 @@
...
+        })}
+      >
+        {link.workflowEvent.workflowId}
+      </Link>
+    </div>
+  </div>
+</div>
+
+<style lang="postcss">
+  .badge {
```
rossedfort, line 30: Would it make sense to extend our badge component to support a link here instead?

src/lib/components/event/event-links-expanded.svelte
```diff
@@ -0,0 +1,42 @@
...
+  {#each links as link}
+    {#if link?.workflowEvent}
+      <div class="content detail-row">
```
rossedfort, line 12: Why 2 classes here if they are only used in tandem on this one element?

src/lib/components/event/event-links-expanded.svelte
```diff
@@ -0,0 +1,42 @@
...
+<style lang="postcss">
+  .content {
+    @apply block w-full px-2 py-1 text-left;
+  }
+
+  .detail-row {
+    @apply flex w-full items-center gap-4 py-1 text-left xl:flex;
+  }
+
+  .badge {
```
rossedfort, line 39: 2nd new `badge` class, def could be worth using `<Badge />` here

rossedfort (commented):
Looks good just a few suggestions/nits

src/lib/components/event/event-links-expanded.svelte
Alex-Tideman: 👍

src/lib/components/event/event-link.svelte
Alex-Tideman: 👍

src/lib/components/event/event-link.svelte
Alex-Tideman: Yeah good idea

rossedfort (approved):

PR #2300, opened 2024-09-03
update paginated table to support custom page size options

rossedfort:
## Description & motivation 💭 
add a `pageSizeOptions` prop which defaults to `['100', '250', '500']` but allows for specifying different page size options
### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/holocene/table/paginated-table/paginated.svelte
+++ b/src/lib/holocene/table/paginated-table/paginated.svelte
@@ -26,10 +26,10 @@
   export let nextPageButtonLabel: string;
   export let previousPageButtonLabel: string;
   export let maxHeight = '';
+  export let pageSizeOptions: string[] = options;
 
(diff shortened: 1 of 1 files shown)


src/lib/holocene/table/paginated-table/paginated.svelte
```diff
@@ -26,10 +26,10 @@
   export let nextPageButtonLabel: string;
   export let previousPageButtonLabel: string;
   export let maxHeight = '';
+  export let pageSizeOptions: string[] = options;
```
GiantRobots, line 29: How do you feel about updating the options here to have a 50 page size default? It makes things load pretty noticeably faster in my testing

GiantRobots (approved):

PR #2319, opened 2024-09-12
Fix workflow counts and reactive selectedId

Alex-Tideman:
## Description & motivation 💭 
Fix workflow count filter click to show the filter tag on click.

Fix compact view expanded row click on a single event on large histories.

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/workflow/workflow-counts.svelte
+++ b/src/lib/components/workflow/workflow-counts.svelte
@@ -3,18 +3,25 @@
 
   import { page } from '$app/stores';
 
-  import Link from '$lib/holocene/link.svelte';
   import Skeleton from '$lib/holocene/skeleton/index.svelte';
   import { workflowStatuses } from '$lib/models/workflow-status';
   import { fetchWorkflowCountByExecutionStatus } from '$lib/services/workflow-counts';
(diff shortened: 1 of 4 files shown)


src/lib/components/workflow/workflow-counts.svelte
```diff
@@ -116,24 +123,43 @@
...
 
+  const onStatusClick = (status) => {
+    const filter = {
+      attribute: 'ExecutionStatus',
+      type: SEARCH_ATTRIBUTE_TYPE.KEYWORD,
+      value: status,
+      operator: '',
+      conditional: '=',
+      parenthesis: '',
+    };
```
rossedfort, line 134: Do we have a type or a function to generate this filter?

rossedfort (approved):

src/lib/components/workflow/workflow-counts.svelte
Alex-Tideman: It's all tangled up in the Context for the workflow filter search stuff. But it's typed and would blow up if in the $workflowFilters = [...$workflowFilters, filter] logic

PR #2328, opened 2024-09-13
Show all possible failures for WorkflowTaskError

Alex-Tideman:
## Description & motivation 💭 
Currently we only show the first failure message and stack trace. It's possible to have deeply nested failures for some SDKS (Java, .NET). This updates our WorkflowTaskError alert to show all failures with message and stack trace to allow for better debugging.

Screenshot 2024-09-13 at 1 25 58 PM
Screenshot 2024-09-13 at 1 26 09 PM
Screenshot 2024-09-13 at 1 26 23 PM

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/server/go.mod
+++ b/server/go.mod
@@ -1,8 +1,6 @@
 module github.com/temporalio/ui-server/v2
 
-go 1.21
-
-toolchain go1.23.0
+go 1.20
 
 require (
 	github.com/coreos/go-oidc/v3 v3.1.0
--- a/server/go.mod
+++ b/server/go.mod
@@ -1,6 +1,8 @@
 module github.com/temporalio/ui-server/v2
 
-go 1.20
+go 1.21
+
+toolchain go1.23.0
 
 require (
 	github.com/coreos/go-oidc/v3 v3.1.0
--- a/src/lib/components/lines-and-dots/workflow-error.svelte
+++ b/src/lib/components/lines-and-dots/workflow-error.svelte
@@ -1,4 +1,5 @@
 <script lang="ts">
+  import AccordionGroup from '$lib/holocene/accordion/accordion-group.svelte';
   import Alert from '$lib/holocene/alert.svelte';
   import CodeBlock from '$lib/holocene/code-block.svelte';
   import Icon from '$lib/holocene/icon/icon.svelte';
@@ -16,6 +17,8 @@
   import { toWorkflowTaskFailureReadable } from '$lib/utilities/screaming-enums';
 
   import { CategoryIcon } from './constants';
+
+  import WorkflowErrorStackTrace from './workflow-error-stack-trace.svelte';
 
   export let error: WorkflowTaskFailedEvent;
   export let pendingTask: PendingWorkflowTaskInfo | undefined = undefined;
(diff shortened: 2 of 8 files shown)


src/lib/components/lines-and-dots/workflow-error.svelte
```diff
@@ -78,28 +80,12 @@
...
-          content={error.attributes?.failure?.source || ''}
-          language="text"
-          copyIconTitle={translate('common.copy-icon-title')}
-          copySuccessIconTitle={translate('common.copy-success-icon-title')}
-        />
+      <div class="flex flex-col gap-2 bg-primary p-4">
+        {#if error.attributes?.failure}
+          <AccordionGroup>
+            <WorkflowErrorStackTrace failure={error.attributes.failure} />
+          </AccordionGroup>
```
laurakwhit, line 87: Is it possible to have more than one failure here? Looking at the screenshots there are multiple, but looking at the code it looks like we'd only every have one `Accordion` in which case we probably don't need an `AccordionGroup` here 

```suggestion
          <WorkflowErrorStackTrace failure={error.attributes.failure} />
```

src/lib/holocene/accordion/accordion.svelte
```diff
@@ -56,11 +56,12 @@
       on:click={toggleAccordion}
     >
       <div class="space-between flex w-full flex-row items-center">
-        <h3 class="flex w-full items-center gap-2">
+        <h3 class="flex min-w-fit items-center gap-2">
```
laurakwhit, line 59: The change to `min-w-fit` here from `w-full` means that the chevron icon on the `Accordion` isn't positioned correctly when there isn't a "description" slot 

| Before | After |
|-|-|
|Screenshot 2024-09-16 at 3 21 59 PM|Screenshot 2024-09-16 at 3 20 20 PM|

src/lib/components/lines-and-dots/workflow-error.svelte
Alex-Tideman: If you look in the WorkflowErrorStackTrace, it recursively calls itself. These failures can be deeply nested (apparently how some SDKs work like .NET and Java). So it can have one or many of the accordions.

src/lib/holocene/accordion/accordion.svelte
Alex-Tideman: Good call, changed it back.

src/lib/holocene/accordion/accordion.svelte
```diff
@@ -61,6 +61,7 @@
           {title}
```
laurakwhit, line 61: | Before | After |
|-|-|
|Screenshot 2024-09-17 at 11 03 36 AM|Screenshot 2024-09-17 at 11 03 53 AM|

Might want to update the following ⬇️  in order to fix the icon and word wrapping ⬆️ 

```js
      <div class="justify-between flex w-full flex-row items-center gap-2">
        <h3 class="flex items-center gap-2">
```
and

```js
        <Icon class="m-2 shrink-0" name={open ? 'chevron-up' : 'chevron-down'} />
```

server/go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/temporalio/ui-server/v2
 
-go 1.20
+go 1.21
+
+toolchain go1.23.0
```
laurakwhit, line 5: Do we want to add `toolchain go1.23.0` here?

src/lib/holocene/accordion/accordion.svelte
Alex-Tideman: 👍

server/go.mod
Alex-Tideman: 👍

laurakwhit (approved):

PR #2333, opened 2024-09-18
Add new responsive styles to input on api pagination

KATIETOLER:
Adjust responsive breakpoints for namespace filtering

This adds a change to the input and nav styles on the API pagination component in holocene to fix breakpoints for namespace filtering. 

MD
---
Screenshot 2024-09-17 at 6 39 07 PM

XS
---
Screenshot 2024-09-17 at 6 39 29 PM

LG
---
Screenshot 2024-09-17 at 6 39 46 PM

XL
---
Screenshot 2024-09-17 at 6 40 00 PM

### DIFF
--- a/src/lib/holocene/api-pagination.svelte
+++ b/src/lib/holocene/api-pagination.svelte
@@ -226,7 +226,7 @@
       <Input
         icon="search"
         id="api-pagination-search-input"
-        class="w-full grow sm:w-1/2"
+        class="grow"
         bind:value={query}
         label={filterInputPlaceholder}
         labelHidden
--- a/src/lib/holocene/api-pagination.svelte
+++ b/src/lib/holocene/api-pagination.svelte
@@ -215,7 +215,7 @@
 <slot name="header" visibleItems={$store.visibleItems} />
 <div class="relative mb-8 flex flex-col gap-4">
   <div
-    class="flex flex-row items-center gap-4 flex-wrap {$$slots[
+    class="flex flex-row flex-wrap items-center gap-4 {$$slots[
       'action-top-left'
     ]
       ? 'justify-between'
@@ -226,7 +226,7 @@
       <Input
         icon="search"
         id="api-pagination-search-input"
-        class="w-full sm:w-1/2"
+        class="w-full grow sm:w-1/2"
         bind:value={query}
         label={filterInputPlaceholder}
         labelHidden
@@ -237,7 +237,7 @@
       />
     {/if}
     <nav
-      class="flex gap-4 flex-row flex-wrap justify-center"
+      class="flex flex-row flex-wrap justify-center gap-4"
       aria-label="{$$restProps['aria-label']} 1"
     >
       <slot name="action-top-center" />
--- a/src/lib/holocene/api-pagination.svelte
+++ b/src/lib/holocene/api-pagination.svelte
@@ -237,7 +237,7 @@
       />
     {/if}
     <nav
-      class="flex shrink-0 flex-col gap-4 sm:flex-row"
+      class="flex gap-4 flex-row flex-wrap justify-center"
       aria-label="{$$restProps['aria-label']} 1"
     >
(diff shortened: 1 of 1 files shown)


src/lib/holocene/api-pagination.svelte
```diff
@@ -226,7 +226,7 @@
       <Input
         icon="search"
         id="api-pagination-search-input"
-        class="grow"
+        class="w-1/2 grow"
```
Alex-Tideman, line 229: I would make this `w-full sm:w-1/2` so on xs screens the input takes up the full width

Alex-Tideman:
Let's point this at `codefreeze-09.12.24` branch instead of `main`

src/lib/holocene/api-pagination.svelte
KATIETOLER: Will do, thank you!

src/lib/holocene/api-pagination.svelte
```diff
@@ -226,7 +226,7 @@
       <Input
         icon="search"
         id="api-pagination-search-input"
-        class="grow"
+        class="w-full grow sm:w-1/2"
```
laurakwhit, line 229: Looks like on big screens this goes back to columns instead of staying in a row ⬇️ 

Screenshot 2024-09-18 at 9 00 10 AM

Screenshot 2024-09-18 at 9 01 08 AM

src/lib/holocene/api-pagination.svelte
```diff
@@ -215,7 +215,7 @@
 <slot name="header" visibleItems={$store.visibleItems} />
 <div class="relative mb-8 flex flex-col gap-4">
   <div
-    class="flex flex-col items-center gap-4 lg:flex-row {$$slots[
+    class="flex flex-col items-center gap-4 max-xl:flex max-xl:flex-col sm:flex-row {$$slots[
```
laurakwhit, line 218: I think we can simplify this a bit with `flex-wrap` ⬇️ 
```suggestion
    class="flex flex-row items-center gap-4 flex-wrap {$$slots[
```

src/lib/holocene/api-pagination.svelte
```diff
@@ -226,7 +226,7 @@
       <Input
         icon="search"
         id="api-pagination-search-input"
-        class="grow"
+        class="w-full sm:w-1/2"
```
laurakwhit, line 229: I think we want to keep this `grow` 
```suggestion
        class="grow"
```

| Before | After |
|-|-|
|Screenshot 2024-10-07 at 9 53 05 AM|Screenshot 2024-10-07 at 10 02 31 AM|

src/lib/holocene/api-pagination.svelte
```diff
@@ -237,7 +237,7 @@
       />
     {/if}
     <nav
-      class="flex shrink-0 flex-col gap-4 md:flex-row"
+      class="flex shrink-0 flex-col gap-4 sm:flex-row"
```
laurakwhit, line 240: This seems to work a bit better with responsiveness. Give it a try and let me know what you think! 
```suggestion
      class="flex gap-4 flex-row flex-wrap justify-center"
```

laurakwhit (changes requested):
Let's point this PR at `main` again since `codefreeze-09.12.24` has been merged

src/lib/holocene/api-pagination.svelte
KATIETOLER: Let me know if this seems better! Thank you Laura! <3

src/lib/holocene/api-pagination.svelte
laurakwhit, line 229: Other width classes aren't needed here with the addition of `grow`

```suggestion
        class="grow"
```

laurakwhit (approved):
One more change (see comment below) and then should be good to go 🚀

PR #2335, opened 2024-09-19
Workflow Electron

Alex-Tideman:
## Description & motivation 💭 
We now have the ability to fetch the full relationship tree of a workflow. This visual will not only show the child workflows of a workflow, but all child workflows of the root workflow of a workflow. The maximum depth of generations shown is 4, but we can add more if we want. Each node is colored to match it's current status, and the active node (the current page user is on) has an additional black/white background circle to easily discover where you are in the electron.

The visual will replace the table of children.

If there are no child workflows, show the Schedule or First/Next/Previous table.

This PR includes a generic zoom-svg component that let's you zoom in/out and pan.

Screenshot 2024-09-19 at 11 20 24 AM
Screenshot 2024-09-19 at 11 21 28 AM
Screenshot 2024-09-19 at 11 21 43 AM
Screenshot 2024-09-19 at 11 21 57 AM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/layouts/workflow-header.svelte
```diff
@@ -136,6 +129,15 @@
...
       </Tab>
+      <Tab
+        label={translate('workflows.relationships')}
+        id="relationships-tab"
+        href={routeForRelationships(routeParameters)}
+        active={pathMatches(
+          $page.url.pathname,
+          routeForRelationships(routeParameters),
+        )}
+      />
```
Alex-Tideman, line 140: I removed the count since we make the query when you go to the tab. And also... would we want to show the full count of relationship (parent/children/grandchildre/cousins) or just children??

src/lib/components/workflow/workflow-atom.svelte
```diff
@@ -0,0 +1,204 @@
...
+    children: RootNode[];
+    workflow: WorkflowExecution;
+  };
+</script>
+
+<script lang="ts">
+  import { page } from '$app/stores';
+
+  import Button from '$lib/holocene/button.svelte';
+  import Tooltip from '$lib/holocene/tooltip.svelte';
```
GiantRobots, line 14: Now that we can have more complex tooltips what if we put the rest of the electrons "tree" into a tooltip 🤔

Alex-Tideman: Hehe yes. But electrons all have the same tree. Maybe on node click we zoom in on node and show just it's subtree and hide/fade the rest

PR #2339, opened 2024-09-20
Event Type multiselect dropdown with All/None options

Alex-Tideman:
## Description & motivation 💭 
Better UX with type filter as a dropdown and includes All / None options to quickly clear or add all filters.

Screenshot 2024-09-20 at 12 32 08 PM
Screenshot 2024-09-20 at 12 32 26 PM
Screenshot 2024-09-20 at 12 39 46 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/holocene/menu/menu.svelte
+++ b/src/lib/holocene/menu/menu.svelte
@@ -12,7 +12,7 @@
   interface $$Props extends HTMLAttributes<HTMLUListElement> {
     id: string;
     keepOpen?: boolean;
-    position?: 'left' | 'right' | 'top-light' | 'top-right';
+    position?: 'left' | 'right' | 'top-left' | 'top-right';
     menuElement?: HTMLUListElement;
   }
 
--- a/src/lib/holocene/menu/menu.svelte
+++ b/src/lib/holocene/menu/menu.svelte
@@ -12,7 +12,7 @@
   interface $$Props extends HTMLAttributes<HTMLUListElement> {
     id: string;
     keepOpen?: boolean;
-    position?: 'left' | 'right' | 'topLeft' | 'topRight';
+    position?: 'left' | 'right' | 'top-light' | 'top-right';
     menuElement?: HTMLUListElement;
   }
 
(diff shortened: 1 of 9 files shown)


src/lib/holocene/menu/menu.svelte
```diff
@@ -47,8 +48,12 @@
   class:hidden={!$open}
   aria-labelledby={id}
   tabindex={-1}
+  style={position === 'topRight' || position === 'topLeft'
+    ? `top: -${height + 16}px;`
+    : ''}
```
rossedfort, line 53: Could we do this in the css declarations below

```css
.topRight,
.topLeft {
  @apply -top-4;
}
```

src/lib/holocene/menu/menu.svelte
```diff
@@ -12,15 +12,16 @@
...
     menuElement?: HTMLUListElement;
   }
 
   let className = '';
+  let height = 0;
   export { className as class };
   export let id: string;
   export let keepOpen = false;
-  export let position: 'left' | 'right' = 'left';
+  export let position: 'left' | 'right' | 'topLeft' | 'topRight' = 'left';
```
rossedfort, line 24: can you make this `kebab-case`? I hate `camelCase` string props, when they're kebab-case, they read more like regular html attributes

src/lib/holocene/menu/menu.svelte
Alex-Tideman: Need to keep the style because it's dynamic based on the height of the menu

src/lib/holocene/menu/menu.svelte
Alex-Tideman: 👍

src/lib/holocene/menu/menu.svelte
rossedfort: Ahh right

rossedfort (approved):

PR #2341, opened 2024-09-24
Codefreeze 09.12.24 -> Main

Alex-Tideman:
### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/components/lines-and-dots/event-type-filter.svelte
```diff
@@ -46,32 +57,79 @@
...
+        checked={!$eventTypeFilter.length}
+        label={translate('common.none')}
+        labelHidden
+      />
+      {translate('common.none')}
+    </MenuItem>
+    <MenuDivider />
+    {#each options as option}
+      <MenuItem
+        data-testid={option.label}
```
laurakwhit, line 118: Wondering if we want to align items to start instead of center here (or in the `MenuItem` component itself) if there is a description ⬇️ 
```suggestion
        class="!items-start"
        data-testid={option.label}
```

| Before | After |
|-|-|
|Screenshot 2024-09-24 at 11 02 53 AM|Screenshot 2024-09-24 at 11 03 05 AM|

laurakwhit (approved):

src/lib/components/lines-and-dots/event-type-filter.svelte
Alex-Tideman: Hmm yeah I could be convinced either way

Alex-Tideman: But should do it MenuItem if we want to do it

PR #2353, opened 2024-09-26
Fix differences in duration 

laurakwhit:
## Description & motivation 💭 
The `toSeconds` util uses [intervalToDuration](https://date-fns.org/v4.1.0/docs/intervalToDuration#) which is expecting an interval object (consisting of two `Dates`) to be passed as the parameter and therefore it utilizes [the Date constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date). `new Date()` creates a datetime in the client timezone which was causing differences in duration (e.g. timer duration ⬇️ ) 
Screenshot 2024-09-26 at 3 31 50 PM
for different users based on what their system timezone was set to.

This PR passes two `Dates` based on UTC instead of relying on the client timezone in instances where `intervalToDuration` is being determined based on seconds.

## Testing 🧪 
If system settings are set to `PDT`
```js 
intervalToDuration({ start: 0, end: (2652361 * 1000) })
``` 
will return `{ months: 1, hours: 7, minutes: 853, seconds: 102419 }` and
```js 
intervalToDuration({ start: new Date(Date.UTC(0, 0, 0, 0, 0, 0)), end: new Date(Date.UTC(0, 0, 0, 0, 0, 2652361))})
```
will return `{ days: 30, hours: 16, minutes: 46, seconds: 1 }`

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
`DT-2484`

### DIFF
--- a/src/lib/utilities/format-time.ts
+++ b/src/lib/utilities/format-time.ts
@@ -193,9 +193,11 @@ export function getMilliseconds(date: ValidTime | undefined | null): number {
 
 export function fromSecondsToMinutesAndSeconds(seconds: number): string {
   if (!seconds) return '';
+  const start = new Date(Date.UTC(0, 0, 0, 0, 0, 0));
+  const end = new Date(Date.UTC(0, 0, 0, 0, 0, Math.floor(seconds)));
   const duration = intervalToDuration({
-    start: 0,
-    end: Math.floor(seconds) * 1000,
+    start,
+    end,
   });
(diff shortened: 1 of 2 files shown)


Alex-Tideman (approved):

Alex-Tideman:
Incredible detective work!

PR #2359, opened 2024-09-30
DT-2499 - wf reset ux improvement

rossedfort:
## Description & motivation 💭 
Apparently a `WorkflowTaskFailed` event in the even history of a workflow execution that is the result of a reset is expected. This PR adjusts the styling/content of the workflow details page for this case so that it doesn't look like things went wrong.

- Remove the Typed Error when cause == ResetWorkflow
- Remove red background from timeline when WorkflowTaskFailed event is a ResetWorkflow task failed event
- Remove red background from event history row when WorkflowTaskFailed event is a ResetWorkflow task failed event

|before|after|
|---|---|
|Screenshot 2024-09-30 at 2 07 27 PM|Screenshot 2024-09-30 at 2 27 21 PM|
### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

- With the UI running, run any sample workflow from a samples repo
- Kill the worker process for that sample
- Reset the workflow that was created via the sample
- View the workflow execution that was the result of the workflow reset


src/lib/components/lines-and-dots/workflow-error.svelte
```diff
@@ -25,30 +22,12 @@
...
-    return toWorkflowTaskFailureReadable(cause);
-  }
-
   $: {
     cause = getErrorCause(error);
   }
 </script>
 
-{#if cause}
+{#if cause && cause !== 'ResetWorkflow'}
```
Alex-Tideman, line 30: I'm sure there will be more of these cases but not gonna overthink trying to abstract that way yet. This is fine.

Alex-Tideman (approved):

PR #2364, opened 2024-10-02
DT-2502 - handle WorkflowExecutionUpdateAdmitted events

rossedfort:
## Description & motivation 💭 
this event type is only used on post-reset workflows when updates are reapplied, but we still want to show their attributes in the event history

|before|after|
|---|---|
|Screenshot 2024-10-02 at 10 50 59 AM|Screenshot 2024-10-02 at 12 01 18 PM|

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/models/event-groups/get-group-name.ts
+++ b/src/lib/models/event-groups/get-group-name.ts
@@ -17,7 +17,7 @@ import {
 
 export const getEventGroupName = (
   event: CommonHistoryEvent,
-  initialEvent: CommonHistoryEvent,
+  initialEvent?: CommonHistoryEvent,
 ): string => {
   if (!event) return '';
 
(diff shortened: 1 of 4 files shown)


src/lib/models/event-groups/get-group-name.ts
```diff
@@ -12,9 +12,13 @@ import {
   isTimerStartedEvent,
   isWorkflowExecutionSignaledEvent,
   isWorkflowExecutionUpdateAcceptedEvent,
+  isWorkflowExecutionUpdateAdmittedEvent,
 } from '$lib/utilities/is-event-type';
 
-export const getEventGroupName = (event: CommonHistoryEvent): string => {
+export const getEventGroupName = (
+  event: CommonHistoryEvent,
+  initialEvent: CommonHistoryEvent,
```
Alex-Tideman, line 20: Probably make this an optional prop to appease the tests

Alex-Tideman (approved):

PR #2373, opened 2024-10-08
DT-1495 - use description instead of tooltip on reset menu item

rossedfort:
## Description & motivation 💭 
Instead of changing Menu and/or MenuItem to support tooltips, this PR removes the only occurrence of using a Tooltip on a MenuItem (workflow reset) and uses the MenuItem's description prop instead

Screenshot 2024-10-08 at 11 26 32 AM
Screenshot 2024-10-08 at 11 27 18 AM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/workflow-actions.svelte
+++ b/src/lib/components/workflow-actions.svelte
@@ -119,6 +119,7 @@
 {#if isRunning}
   <SplitButton
     id="workflow-actions"
+    menuClass="w-[24rem]"
     position="right"
     disabled={actionsDisabled}
     primaryActionDisabled={!cancelEnabled || cancelInProgress}
@@ -161,6 +162,7 @@
 {:else if !workflowCreateDisabled($page)}
   <SplitButton
     id="workflow-actions"
+    menuClass="w-[16rem]"
     position="right"
     primaryActionDisabled={!resetEnabled}
     on:click={() => (resetConfirmationModalOpen = true)}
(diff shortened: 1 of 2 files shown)


Alex-Tideman:
What are your thoughts on making the width wider on the menu item? It's a lot of text in a narrow width, makes it hard to read.

rossedfort:
> What are your thoughts on making the width wider on the menu item? It's a lot of text in a narrow width, makes it hard to read.

I agree, although this could be tricky to solve generically. The width of the MenuItem is inherited from the width of the Menu, which is inherited from the width of the trigger element (in this case the split button). We can hardcode a width on this specific menu I suppose, but may be tricky if we want to solve across the board.

Alex-Tideman:
I think it's probably fine to do it just for this Menu and not generically. I do the same on the Event Type Filter because of the long descriptions.

rossedfort:
Part of the problem here is the workflow task names don't have any whitespace, so they don't break.

Screenshot 2024-10-08 at 11 14 06 AM

Breaking on all isn't great either

Screenshot 2024-10-08 at 11 15 47 AM

Alex-Tideman:
I'd probably just make the width wide enough to fit the first full event type. Also should we change the direction to go left?

rossedfort:
> I'd probably just make the width wide enough to fit the first full event type. Also should we change the direction to go left?

@Alex-Tideman Yep, good call, updated screenshot in the PR description

Alex-Tideman (approved):

PR #2382, opened 2024-10-10
Use encodeUriComponent and decodeUriComponent and add tests

Alex-Tideman:
## Description & motivation 💭 
Remove custom Svelte encode/decode logic and use encodeUriComponent and decodeUriCompoent. Trim the end of schedule names.

Tested with schedules with space at end and correctly fetches schedule.

Screenshot 2024-10-10 at 8 50 04 AM
Screenshot 2024-10-10 at 8 54 32 AM
Screenshot 2024-10-10 at 8 56 37 AM
Screenshot 2024-10-10 at 9 38 22 AM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

### DIFF
--- a/src/lib/components/schedule/schedule-form-view.svelte
+++ b/src/lib/components/schedule/schedule-form-view.svelte
@@ -89,7 +89,7 @@
 
   const handleConfirm = (preset: SchedulePreset, schedule?: Schedule) => {
     const args: Partial<ScheduleParameters> = {
-      name,
+      name: name.trimEnd(),
       workflowType,
       workflowId,
       taskQueue,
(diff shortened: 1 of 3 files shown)


rossedfort (approved):

rossedfort:
Maybe eventually we replace all the `encodeURIForSvelte` and `decodeURIForSvelte` calls with `encodeURIComponent` and `decodeURIComponent` but this is good for now

Alex-Tideman:
Yeah I kept it in that function in case there's some other tweaks we need to make.

PR #2385, opened 2024-10-10
Show correct encoding value in PayloadInput

laurakwhit:
# Description & motivation 💭 
This PR prevents the remount of `PayloadInput` after loading the initial values which was triggering `clearValues()` in the `PayloadInput` component and ensures the correct `Encoding` value is shown.

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

- Create a schedule with a `Schedule Input` and select `json/protobuf` for `Encoding` and Save 
   - [ ] Verify `json/protobuf` is set with the create and/or update schedule request 
- Select "Edit" for that schedule
   - [ ] Verify `json/protobuf` is selected
- Create a schedule with a `Schedule Input` and select `json/plain` for `Encoding` and Save 
   - [ ] Verify `json/plain` is set with the create and/or update schedule request 
- Select "Edit" for that schedule
   - [ ] Verify `json/plain` is selected

### DIFF
--- a/src/lib/components/payload-input.svelte
+++ b/src/lib/components/payload-input.svelte
@@ -1,5 +1,9 @@
 <script context="module" lang="ts">
-  export type PayloadInputEncoding = 'json/plain' | 'json/protobuf';
+  const encoding = ['json/plain', 'json/protobuf'] as const;
+  export type PayloadInputEncoding = (typeof encoding)[number];
+  export const isPayloadInputEncodingType = (
+    x: any, // eslint-disable-line @typescript-eslint/no-explicit-any
+  ): x is PayloadInputEncoding => encoding.includes(x);
 </script>
 
 <script lang="ts">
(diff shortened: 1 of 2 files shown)


src/lib/components/payload-input.svelte
```diff
@@ -1,5 +1,9 @@
 <script context="module" lang="ts">
-  export type PayloadInputEncoding = 'json/plain' | 'json/protobuf';
+  const encoding = ['json/plain', 'json/protobuf'] as const;
+  export type PayloadInputEncoding = (typeof encoding)[number];
+  export const isPayloadInputEncodingType = (
+    x: any, // eslint-disable-line @typescript-eslint/no-explicit-any
```
rossedfort, line 5: One way to get around the `//eslint-disable` comment is to use `unknown` as a type.

rossedfort (approved):

PR #2392, opened 2024-10-15
Fix pending activity text in timeline

laurakwhit:
| Day mode | Night mode |
|-|-|
|Screenshot 2024-10-15 at 9 23 31 AM|Screenshot 2024-10-15 at 9 23 55 AM|
|Screenshot 2024-10-15 at 8 55 01 AM|Screenshot 2024-10-15 at 9 24 30 AM|

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/lines-and-dots/svg/text.svelte
+++ b/src/lib/components/lines-and-dots/svg/text.svelte
@@ -17,7 +17,6 @@
   export let icon: IconName | undefined = undefined;
   export let config: GraphConfig | undefined = undefined;
   export let label = false;
-  export let dark = false;
 
   $: [x, y] = point;
 
@@ -37,12 +36,12 @@
     strokeWidth={backdropHeight}
   />
 {/if}
-{#if showIcon && textAnchor === 'start'}
+{#if showIcon}
   <Icon
     name={icon}
-    {x}
+    x={textAnchor === 'end' ? x - textWidth - backdropHeight : x}
     y={y - 8}
-    class={dark ? 'text-black' : !backdrop ? 'text-primary' : 'text-white'}
+    class={!backdrop ? 'text-primary' : 'text-white'}
   />
 {/if}
 <text
@@ -50,7 +49,6 @@
   class="cursor-pointer select-none outline-none {category} text-primary"
   class:label
   class:backdrop
-  class:dark
   x={textX}
   {y}
   font-size={fontSize}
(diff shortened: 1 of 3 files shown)


src/lib/components/lines-and-dots/svg/text.svelte
```diff
@@ -48,36 +36,27 @@
     strokeWidth={backdropHeight}
   />
 {/if}
-{#if showIcon && textAnchor === 'start'}
+{#if showIcon}
   <Icon
     name={icon}
-    x={x - offset}
+    x={textAnchor === 'end' ? x - textWidth - backdropHeight : x}
     y={y - 8}
```
Alex-Tideman, line 43: One thing to try is instead of this terrible math I did for y, use text-anchor="middle"

src/lib/components/lines-and-dots/svg/text.svelte
```diff
@@ -48,36 +36,27 @@
...
-  x={textX - offset}
+  x={textX}
   {y}
   font-size={fontSize}
   font-weight={fontWeight}
   text-anchor={textAnchor}
 >
   <slot />
 </text>
-{#if showIcon && textAnchor === 'end'}
```
Alex-Tideman, line 73: What icon was this? I guess what was removed?

src/lib/components/lines-and-dots/svg/text.svelte
laurakwhit: We want the icon to show up at the start of the text though, so we need to do a bit of math to make that happen when `textAnchor === 'end'`.

src/lib/components/lines-and-dots/svg/text.svelte
laurakwhit: Not sure why this code existed actually... It was redundant since the icon would show up in the same place (x and y values were the same for both blocks of code). Only difference between this `Icon` and the other one was the ` && textAnchor === 'end'` vs. ` && textAnchor === 'start'` check.

src/lib/components/lines-and-dots/svg/text.svelte
Alex-Tideman: Like centering the y position without needing to do the y - 8

src/lib/components/lines-and-dots/svg/text.svelte
Alex-Tideman: 👍

src/lib/components/lines-and-dots/svg/text.svelte
laurakwhit: Ah I see what you're saying, but the `text-anchor` attribute is for `<text/>` elements and that icon is not in one (it's just an `<svg/>` next to one).

Alex-Tideman: 🤦 My bad. I thought it was text not the icon

Alex-Tideman (approved):

PR #2402, opened 2024-10-22
Add toast instead of redirect on Start Workflow with link to workflow

Alex-Tideman:
## Description & motivation 💭 
Instead of guessing how long it will take to update visibility store, add a toast with link to started workflow and stay on Start Workflow page.

Screenshot 2024-10-22 at 10 10 56 AM
Screenshot 2024-10-22 at 10 21 58 AM

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/holocene/toaster.svelte
+++ b/src/lib/holocene/toaster.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
   import type { Toaster } from '../stores/toaster';
 
+  import Link from './link.svelte';
   import ToastComponent from './toast.svelte';
 
   export let pop: Toaster['pop'];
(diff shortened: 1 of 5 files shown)


src/lib/holocene/toaster.svelte
```diff
@@ -16,9 +17,15 @@
   class="fixed bottom-5 right-5 z-[99999] flex flex-col items-end gap-2"
   role="log"
 >
-  {#each $toasts as { message, variant, id } (id)}
+  {#each $toasts as { message, variant, id, link } (id)}
     <ToastComponent {closeButtonLabel} {variant} {id} on:dismiss={dismissToast}>
-      {message}
+      {#if link}
```
stevekinney, line 22: I could be convinced that should be it's own component or something someday. But, today doesn't have to be that day.

src/lib/pages/start-workflow.svelte
```diff
@@ -82,11 +79,22 @@
         encoding: $encoding,
         searchAttributes,
       });
-      await new Promise((resolve) => setTimeout(resolve, 2000));
-      goto(routeForWorkflows({ namespace }));
+      toaster.push({
+        variant: 'success',
+        duration: 5000,
+        message: translate('workflows.start-workflow-success'),
```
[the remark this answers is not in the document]
Alex-Tideman: Yeah that's probably a good idea, don't let any willy nilly string in here.

src/lib/holocene/toaster.svelte
Alex-Tideman: Ehhh this is so simple that i don't think the juice is worth the squeeze, but if it gets any more complicated for sure.

PR #2405, opened 2024-10-23
Encode header fields if they exist on schedule edit

Alex-Tideman:
## Description & motivation 💭 
If a schedule startWorkflow has been created or edited to include header fields, and then edited in the UI, upon save it would error due to the header fields not being encoded. This ensures the fields get encoded upon save if they exist.

Screenshot 2024-10-23 at 10 07 52 AM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/stores/schedules.ts
+++ b/src/lib/stores/schedules.ts
@@ -14,6 +14,7 @@ import type {
   ScheduleSpecParameters,
 } from '$lib/types/schedule';
 import { encodePayloads } from '$lib/utilities/encode-payload';
+import { stringifyWithBigInt } from '$lib/utilities/parse-with-big-int';
 import { routeForSchedule, routeForSchedules } from '$lib/utilities/route-for';
 import {
   convertDaysAndMonths,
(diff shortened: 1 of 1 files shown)


src/lib/stores/schedules.ts
```diff
@@ -204,6 +205,18 @@ export const submitEditSchedule = async (
     },
   };
 
+  const fields = body.schedule.action.startWorkflow?.header?.fields;
+  if (fields && Object.keys(fields).length > 0) {
+    const entries = Object.entries(fields);
+    for (const [key, value] of entries) {
+      const encodedValue = await encodePayloads(
+        stringifyWithBigInt(value),
+        'json/plain',
```
laurakwhit, line 214: Do we want to use the `encoding` value here as well? 

```suggestion
        encoding
```

Alex-Tideman: I think we just use plain/json for now. There's no UI for editing headers for schedules and not sure we want to add that yet. If we need to we can, but headers are almost always json/plain.

laurakwhit: Yeah only concern here is that if a user has Codec Server set up with users who don't have `/encode` permissions then it will error on save unless we do a similar approach to https://github.com/temporalio/ui/pull/2384.

laurakwhit (approved):

src/lib/stores/schedules.ts
Alex-Tideman: 👍 I added a try/catch for now and return if encoding errors for header fields

PR #2407, opened 2024-10-24
Add query argument support

laurakwhit:
## Description & motivation 💭 
Allows users to include an argument when running a query for a Workflow in the UI.

| Before | After |
|-|-|
|Screenshot 2024-10-23 at 8 55 45 PM|Screenshot 2024-10-23 at 8 32 55 PM|

| | Day mode | Night mode|
|-|-|-|
| Invalid/Error |Screenshot 2024-10-25 at 4 31 23 PM|Screenshot 2024-10-25 at 4 31 37 PM|
| Query Result |Screenshot 2024-10-25 at 4 36 18 PM|Screenshot 2024-10-25 at 4 36 28 PM|
| New Query |Screenshot 2024-10-25 at 4 32 58 PM|Screenshot 2024-10-25 at 4 33 10 PM|
| Loading |Screenshot 2024-10-25 at 4 33 47 PM|Screenshot 2024-10-25 at 4 33 21 PM|

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

* On a running workflow > go to `Queries`
   - [ ] Verify query is run and there is a query result for first `Query Type` selected on page load
   - Select a different `Query Type` and click `Query`
      - [ ] Verify query result appears as expected
   - Add `Query Args`
      - [ ] Verify there is an error if input is not JSON
      - [ ] Verify clicking `Query` returns query result as expected

---

* On a running workflow > go to `Send a Signal`
   - Enter in `Signal Name`, `Data` and select `Encoding`
   - Select `Cancel` > go to `Send a Signal` again
      - [ ] Verify all inputs have been cleared and default `Encoding` selected

### Issue(s) closed 
`DT-2513`


src/lib/components/payload-input-with-encoding.svelte
```diff
@@ -0,0 +1,61 @@
+<script context="module" lang="ts">
+  const encoding = ['json/plain', 'json/protobuf'] as const;
+  export type PayloadInputEncoding = (typeof encoding)[number];
+  export const isPayloadInputEncodingType = (
+    x: any, // eslint-disable-line @typescript-eslint/no-explicit-any
```
stevekinney, line 5: I almost _always_ go with `unknown` over `any`. It doesn't really matter in this case, but it will at least make `eslint` happier.

src/lib/components/payload-input-with-encoding.svelte
```diff
@@ -0,0 +1,61 @@
...
+
+  import PayloadInput from './payload-input.svelte';
+
+  export let input: string;
+  export let encoding: Writable<PayloadInputEncoding>;
+  export let error = false;
+  export let resetValues = false;
+  export let loading = false;
+
+  $: {
```
stevekinney, line 27: This kind of feels a bit weird. It feels a bit off that we're using a state change to trigger an action. Like, I assume `resetValues` is being triggered from outside this component and that why when we're looking at the reactive statement, we're like if it's now `true` then do this action. But, I wonder if that means that the value should be stored outside of this component and be passed in?

src/lib/components/payload-input-with-encoding.svelte
```diff
@@ -0,0 +1,61 @@
...
+    if (resetValues) {
+      clearValues();
+    }
+  }
+
+  const clearValues = () => {
+    $encoding = 'json/plain';
+  };
+
+  onDestroy(() => {
```
stevekinney, line 37: This is obviously _not_ a change you need to make, but this could probably just be `onDestroy(clearValues)`.

src/lib/components/payload-input.svelte
```diff
@@ -1,31 +1,20 @@
...
-  export const isPayloadInputEncodingType = (
-    x: any, // eslint-disable-line @typescript-eslint/no-explicit-any
-  ): x is PayloadInputEncoding => encoding.includes(x);
-</script>
-
 <script lang="ts">
-  import { type Writable } from 'svelte/store';
-
   import { onDestroy } from 'svelte';
+  import { v4 } from 'uuid';
```
stevekinney, line 3: I sometimes like to do `import { v4 as id } from 'uuid'` or `import { v4 as uuid } from 'uuid'` because I always forget what `v4` is supposed to be. 😆

src/lib/components/payload-input-with-encoding.svelte
laurakwhit: I tried `unknown` first but was getting the error `Argument of type 'unknown' is not assignable to parameter of type '"json/plain" | "json/protobuf"'.`

Maybe if I just need to add an `as` (e.g. `encoding.includes(x as PayloadInputEncoding)`) 🤔

src/lib/components/payload-input-with-encoding.svelte
laurakwhit: I think this was added so that if this is in a `Modal` the value can be reset when it's closed (cc @Alex-Tideman to confirm). I'll look into it a bit more though and see if there's a better solution!

laurakwhit: **Update**: I think[ this](https://github.com/temporalio/ui/pull/2407/commits/b276e06b04d556b5dc9c7c06916fea55aaa9f441) will work instead of having to trigger `resetView` outside of the `CodeBlock` component.

src/lib/pages/workflow-query.svelte
```diff
@@ -63,26 +82,31 @@
...
-          leadingIcon="retry"
-          loading={isLoading}
-        >
-          {translate('common.refresh')}
-        </Button>
-      </div>
+    <div class="flex items-end justify-between gap-2 max-sm:flex-wrap">
+      <Card class="flex w-full flex-col gap-2 xl:w-2/3">
+        <div class="flex flex-col gap-1">
+          <PayloadInput bind:input label={translate('workflows.query-args')} />
```
Alex-Tideman, line 88: I'd probably make this 'Query Arg' since we only allow a single argument and not multiple

Alex-Tideman (approved):

src/lib/pages/workflow-query.svelte
```diff
@@ -25,86 +29,127 @@
...
+  let jsonFormatting = true;
+
+  $: edited = initialQueryType !== queryType || input !== initialInput;
 
   let queryTypes = getQueryTypes({
     namespace,
     workflow: params,
   }).then((queryTypes) => {
     queryType = queryType || queryTypes[0];
+    query(queryType);
```
laurakwhit, line 45: > The only thing I might change is not fetching the initial query by default on page load and instead you need to click "Run Query" always

@Alex-Tideman we can definitely remove this initial query. Wondering if we want a different empty/initial state though ⬇️ 

Screenshot 2024-10-29 at 11 26 02 AM

PR #2408, opened 2024-10-24
Use attempt versus retry

Alex-Tideman:
## Description & motivation 💭 
Raphael suggested replacing 'Retry' with 'Attempt' since we are really showing the current attempt number. It's confusing when it's the first attempt and we show Retry 1 / ...

Screenshot 2024-10-24 at 1 04 11 PM

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/lines-and-dots/svg/timeline-graph-row.svelte
+++ b/src/lib/components/lines-and-dots/svg/timeline-graph-row.svelte
@@ -121,7 +121,7 @@
             config={TimelineConfig}
             icon="retry"
           >
-            {group?.displayName} • {translate('workflows.retry')}
+            {group?.displayName} • {translate('workflows.attempt')}
             {pendingActivity.attempt} / {pendingActivity.maximumAttempts || '∞'}
           </Text>
         {/key}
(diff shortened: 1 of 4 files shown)


src/lib/components/lines-and-dots/svg/timeline-graph-row.svelte
```diff
@@ -121,7 +121,7 @@
             config={TimelineConfig}
             icon="retry"
           >
-            {group?.displayName} • {translate('workflows.retry')}
+            {group?.displayName} • {translate('workflows.attempt')}
```
laurakwhit, line 124: Do we want to update `Retry Expiration` or `Next Retry` as well?

laurakwhit (approved):

src/lib/components/lines-and-dots/svg/timeline-graph-row.svelte
Alex-Tideman: I think those still make sense. Like, you are on Attempt 4, but your next retry is in X or will expire in X. Retry is like future attempt.

PR #2412, opened 2024-10-30
Add mobile option for playwright tests

laurakwhit:
## Description & motivation 💭 
Adds mobile vs. desktop projects to run playwright tests against different viewports. 

### How was this tested 👻 

- [ ] Manual testing
- [x] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
`DT-1935`


playwright.config.ts
```diff
@@ -31,8 +31,22 @@ export default defineConfig({
...
+        ...devices['Desktop Chrome'],
+        viewport: { width: 1920, height: 1080 },
+        isMobile: false,
+      },
+    },
+    {
+      name: 'chromium mobile',
+      testIgnore: /.*desktop.spec.ts/,
+      use: {
+        ...devices['Desktop Chrome'],
```
rossedfort, line 46: Should this be `...devices['Mobile Chrome'].` ?

src/lib/components/bottom-nav-settings.svelte
```diff
@@ -24,12 +24,17 @@
...
+  <div class="flex h-full flex-col justify-start gap-6 overflow-auto px-4 py-8">
     <TimezoneSelect position="left" />
     <NavigationButton
       onClick={onCodecServerClick}
       tooltip={translate('data-encoder.codec-server')}
       label={translate('data-encoder.codec-server')}
+      data-testid="data-encoder-status{hasCodecServer
+        ? $dataEncoder.hasError
+          ? '-error'
+          : '-configured'
```
rossedfort, line 36: could have the `-` in the the static part of the string, but nbd

rossedfort (approved):

playwright.config.ts
laurakwhit: We could select [a specific mobile device](https://playwright.dev/docs/api/class-playwright#playwright-devices), but I opted to just stick with `Desktop Chrome` and set the viewport based on what we use in designs instead.

src/lib/components/bottom-nav-settings.svelte
laurakwhit: We use the testid `data-encoder-status` without the `-` if `hasCodecServer` is false.

PR #2431, opened 2024-11-18
Update combobox to support integration with async code

GiantRobots:
Update combobox so we can update options async and respond to events in the correct order.


src/routes/(app)/+layout.svelte
```diff
@@ -188,7 +189,8 @@
   </div>
   <MainContentContainer>
     <DataEncoderSettings class="hidden md:flex" />
-    <TopNavigation {namespaceList}>
+    <TopNavigation>
+      <NamespacePicker {namespaceList} slot="left" />
```
Alex-Tideman, line 193: Will this actually update with a new namespaceList (say when new ones are created and we bust the cache after a minute)? Since it's in a slot I wonder...

Alex-Tideman (approved):

rossedfort (approved):
This looks good to me but wouldn't mind one other reviewer

PR #2447, opened 2024-11-27
Add additional conditional support to KeywordList filter

laurakwhit:
## Description & motivation 💭 
Adds support for `Equal to`, `Not equal to`, `Is null` and `Is not null` to `KeywordList` attributes in the filter search.

| Before | After |
|--|--|
| Screenshot 2024-11-27 at 2 21 37 PM|Screenshot 2024-11-27 at 2 18 57 PM|

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

* Add a `KeywordList` custom search attribute (e.g. `temporal operator search-attribute create --namespace default --name CustomKeywordListField --type KeywordList`)
* Run a workflow with the custom search attribute set (e.g. [this one](https://github.com/temporalio/samples-typescript/blob/main/search-attributes/src/client.ts#L13)). 
 ---
* Go to the Workflows list view
   * Select Filter > `CustomKeywordListField` > select `In` > enter multiple values
      - [ ] Verify searching works as expected
   * Select the filter that was created
       - [ ] Verify editing the filter works as expected (including changing the conditional)
    * Toggle `View Search Input`
        - [ ] Verify filters show up as expected
    * Refresh the page
       - [ ] Verify filters still show up and work as expected

### Issue(s) closed 
`DT-2635`

### DIFF
--- a/src/lib/components/search-attribute-filter/list-filter.svelte
+++ b/src/lib/components/search-attribute-filter/list-filter.svelte
@@ -3,51 +3,79 @@
 
   import Button from '$lib/holocene/button.svelte';
   import ChipInput from '$lib/holocene/input/chip-input.svelte';
+  import Input from '$lib/holocene/input/input.svelte';
   import { translate } from '$lib/i18n/translate';
+  import { isInConditional } from '$lib/utilities/is';
+  import { formatListFilterValue } from '$lib/utilities/query/search-attribute-filter';
 
+  import ConditionalMenu from './conditional-menu.svelte';
   import { FILTER_CONTEXT, type FilterContext } from './index.svelte';
 
   const { filter, handleSubmit } = getContext<FilterContext>(FILTER_CONTEXT);
 
-  $: ({ value } = $filter);
-  $: list =
-    value.length > 0
-      ? value
-          .slice(1, -1)
-          .split(', ')
-          .map((v) => v.slice(1, -1))
-      : [];
+  $: ({ value, conditional } = $filter);
+  $: _value = value;
+  $: chips = formatListFilterValue(_value);
+  $: options = [
+    { value: 'in', label: 'In' },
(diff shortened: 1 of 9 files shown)


Alex-Tideman:
What are the uses of the conditionals that aren't "In"? Like is "Equal to" checking the exact list? "Dog, Cat, Bird" Equal to "Dog, Cat, Bird"?

src/lib/components/search-attribute-filter/list-filter.svelte
```diff
@@ -3,51 +3,79 @@
...
-          .slice(1, -1)
-          .split(', ')
-          .map((v) => v.slice(1, -1))
-      : [];
+  $: ({ value, conditional } = $filter);
+  $: _value = value;
+  $: chips = formatListFilterValue(_value);
+  $: options = [
+    { value: 'in', label: 'In' },
+    { value: '=', label: translate('common.equal-to') },
```
laurakwhit, line 21: > What are the uses of the conditionals that aren't "In"? Like is "Equal to" checking the exact list? "Dog, Cat, Bird" Equal to "Dog, Cat, Bird"?

@Alex-Tideman say a workflow has a `CustomKeywordListField` with the values `keywordA` and `keywordB`. Checking for `CustomKeywordListField = keywordA` or `CustomKeywordListField = keywordB` would return the workflow, checking for`CustomKeywordListField = keywordC` would not. Checking for `CustomKeywordListField IN ("keywordA", "keywordB", "keywordC")` would also return the workflow even though `keywordC` is not a defined value for the workflow. 

Not sure on usefulness of having both options (`CustomKeywordListField = keywordA`  and `CustomKeywordListField In ("keywordA")` are essentially the same), but I figure if it's possible to have in the search input we should support it in the filter search.

Alex-Tideman (approved):

PR #2448, opened 2024-12-02
Border width update

laurakwhit:
## Description & motivation 💭 
Changes border width from `2px` to `1px`.

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/pages/nexus-endpoint.svelte
```diff
@@ -33,9 +33,7 @@
     </div>
     <p>UUID: {endpoint.id}</p>
   </div>
-  <div
-    class="surface-primary max-w-fit rounded-lg border-2 border-secondary p-4"
-  >
+  <div class="surface-primary max-w-fit rounded-lg border border-secondary p-4">
```
laurakwhit, line 36: > I'm wondering if we want to override our tailwind theme so that there's only 1 border class and it applies a 1px border.

@rossedfort would we still need to add a `border` class to every component that needs it though? Just wondering what this would save us, besides maybe "enforcing" border width.

rossedfort: Yeah, it'd be more about enforcing that we only use a single border width. We can always use arbitrary values if necessary, like `border-[3px]`. I don't think it necessary, was just a thought.

PR #2454, opened 2024-12-04
Better styles and flex for Workflow Details

Alex-Tideman:
## Description & motivation 💭 
Mono font style for titles and improved ordering of details. Added Link to parent workflow in header if it exists. Remove bottom padding of tabs.

I've also included a grow prop on CodeBlock that defaults to true. When set to false, it stacks multiple CodeBlocks on top each other which looks nicer. The only place I set this false is in the Input/Results

Screenshot 2024-12-04 at 9 11 17 AM
Screenshot 2024-12-04 at 9 30 37 AM
Screenshot 2024-12-04 at 9 12 59 AM

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/holocene/code-block.svelte
+++ b/src/lib/holocene/code-block.svelte
@@ -57,7 +57,6 @@
   export { className as class };
   export let editable = false;
   export let inline = false;
-  export let grow = true;
   export let language = 'json';
   export let copyable = true;
   export let copyIconTitle = '';
@@ -190,7 +189,7 @@
   $: content, language, setView();
 </script>
 
-<div class:grow class="relative min-w-[80px]">
+<div class="relative min-w-[80px] grow-0">
   <div
     bind:this={editor}
     class={className}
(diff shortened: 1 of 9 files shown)


src/lib/holocene/code-block.svelte
```diff
@@ -56,6 +57,7 @@
   export { className as class };
   export let editable = false;
   export let inline = false;
+  export let grow = true;
```
rossedfort, line 60: I don't love adding a prop for a single css class. Why don't we want the code block to grow in some instances? We are already exposing a `class` prop, can't we just do `class="grow-0"` where we don't want this to grow?

Alex-Tideman: If it always grows, then the input looks like this:
Screenshot 2024-12-04 at 10 38 34 AM

The class prop is on the codeblock div itself, not on the container that needs to grow.

rossedfort: Ohh gotcha. I think we can probably just add `grow-0` to the top level div of the code block component. I don't see a reason why we'd ever want the code block wrapper to be bigger than the codeblock itself, and this probably only occurs if the codeblock is in a flex element 🤷

Alex-Tideman: 👍 good call

rossedfort (approved):

PR #2466, opened 2024-12-06
Current Details

Alex-Tideman:
## Description & motivation 💭 
We currently fetch the workflow metadata query on page load to get the list of signals/queries. This also includes the current details. Add a "Current Details" accordion to show the current details if they exist. So a Workflow can have:

1. Summary & Details
2. Current Details
3. Both 1 & 2
4. Neither 1 & 2

Current details can only be set in code with an SDK, not through the UI.

Screenshot 2024-12-06 at 9 11 50 AM
Screenshot 2024-12-06 at 12 38 44 PM
Screenshot 2024-12-06 at 12 39 14 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/layouts/workflow-header.svelte
+++ b/src/lib/layouts/workflow-header.svelte
@@ -123,13 +123,10 @@
   </div>
   {#if hasUserMetadata}
     <AccordionLight let:open>
-      <div
-        slot="title"
-        class="flex w-full items-center gap-2 rounded p-2 text-xl"
-      >
+      <div slot="title" class="flex w-full items-center gap-2 p-2 text-xl">
         <Icon
           name="info"
-          class="text-indigo-600/80"
+          class="text-brand"
           width={32}
           height={32}
         />{translate('workflows.summary-and-details')}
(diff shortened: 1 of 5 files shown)


src/lib/layouts/workflow-header.svelte
```diff
@@ -145,6 +144,24 @@
       {/if}
     </AccordionLight>
   {/if}
+  {#if currentDetails}
+    <AccordionLight let:open>
+      <div
+        slot="title"
+        class="flex w-full items-center gap-2 rounded p-2 text-xl"
```
laurakwhit, line 151: Looks like this isn't doing anything style wise

```suggestion
        class="flex w-full items-center gap-2 p-2 text-xl"
```

src/lib/layouts/workflow-header.svelte
```diff
@@ -145,6 +144,24 @@
...
+      >
+        <Icon
+          name="flag"
+          class="text-indigo-600/80"
```
laurakwhit, line 155: Can we use `text-brand` here instead?

```suggestion
          class="text-brand"
```

laurakwhit (approved):

PR #2467, opened 2024-12-06
Include WorkflowUpdateAdmitted in Event Type filter

Alex-Tideman:
### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

### DIFF
--- a/src/lib/models/event-history/get-event-categorization.test.ts
+++ b/src/lib/models/event-history/get-event-categorization.test.ts
@@ -64,6 +64,7 @@ describe('Event Category Data Structures', () => {
         "WorkflowExecutionTerminated": "workflow",
         "WorkflowExecutionTimedOut": "workflow",
         "WorkflowExecutionUpdateAccepted": "update",
+        "WorkflowExecutionUpdateAdmitted": "update",
         "WorkflowExecutionUpdateCompleted": "update",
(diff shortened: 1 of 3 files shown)


src/lib/models/event-history/get-event-categorization.test.ts
```diff
@@ -64,6 +64,7 @@ describe('Event Category Data Structures', () => {
         "WorkflowExecutionTerminated": "workflow",
         "WorkflowExecutionTimedOut": "workflow",
         "WorkflowExecutionUpdateAccepted": "update",
+        "WorkflowExecutionUpdateAdmitted": "update",
```
laurakwhit, line 67: We might want to add this to `const categories` here ⬇️ in this test file as well

```js
  update: [
    'WorkflowExecutionUpdateAccepted',
    'WorkflowExecutionUpdateAdmitted',
    'WorkflowExecutionUpdateCompleted',
  ],
```

laurakwhit (approved):

src/lib/models/event-history/get-event-categorization.test.ts
Alex-Tideman: 👍

PR #2469, opened 2024-12-10
Bottom Nav fixes for external control

GiantRobots:
…allow the escape button to close the nav.

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/bottom-nav.svelte
+++ b/src/lib/components/bottom-nav.svelte
@@ -2,6 +2,7 @@
   import { writable } from 'svelte/store';
   import { slide } from 'svelte/transition';
 
+  import { onDestroy, onMount } from 'svelte';
   import { twMerge as merge } from 'tailwind-merge';
 
   import { beforeNavigate } from '$app/navigation';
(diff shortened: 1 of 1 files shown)


Alex-Tideman (approved):

src/lib/components/bottom-nav.svelte
```diff
@@ -27,6 +28,33 @@
...
 
+  const escHandler = new AbortController();
+
+  onMount(() => {
+    document.addEventListener(
+      'keydown',
+      (e) => {
+        if (
+          e.key === 'Escape' &&
+          [viewSettings, viewNamespaces, viewSettings].some((isOpen) => isOpen)
```
rossedfort, line 39: `viewSettings` is specified here twice

src/lib/components/bottom-nav.svelte
```diff
@@ -27,6 +28,33 @@
   let viewNamespaces = writable(false);
   let viewSettings = false;
 
+  const escHandler = new AbortController();
```
rossedfort, line 31: What is the purpose of this? I see it added in the event listener, but where is it referenced, in a `fetch` somewhere?

GiantRobots: You can use abortControllers to remove event listeners now. https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#signal

src/lib/components/bottom-nav.svelte
GiantRobots: Wisdom

rossedfort (approved):

PR #2485, opened 2025-01-07
Add support for config templating.

robholland:
## Description & motivation 💭 

This is to replace our use of dockerize in our docker containers. Using built-in templating will avoid the need to write to the filesystem, improving the security of our images by allowing read-only mounting.

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


rossnelson (approved):
Chatted offline and all looks good functionally. @robholland is going to update the template patterns to use [sprig functions](https://github.com/temporalio/temporal/pull/6251/files#diff-484cfe7544c4b077e30381131adafea68235e6b0f5ca0d6188f20db1d40a3e6e). Then we will be ready to merge.

PR #2508, opened 2025-01-15
Workflow Update

Alex-Tideman:
## Description & motivation 💭 
Send a Workflow Update from the UI. Allow users to pick an update name from the list of registered updates or to send a custom update name. 

When a validation error occurs in the update, the modal will remain open and show the validation error. If the update is in progress, the modal will remain open and an "In Progress" alert will show. If the update is successful, the modal will close and show a "Update successful" toast notification.

Screenshot 2025-01-15 at 3 03 44 PM
Screenshot 2025-01-15 at 3 03 53 PM
Screenshot 2025-01-15 at 3 04 26 PM
Screenshot 2025-01-15 at 3 04 50 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Merge Checklist 

- [x] Do we need some kind of server version check to enable (disable if < 1.25 or whatever is the right version)? I'm thinking we don't but will double check. I've opted to add an env var workflowUpdateDisabled like the rest of our actions so users can disable the feature if they wish.


rossedfort (approved):
Code looks good, haven't done any testing yet

src/lib/i18n/locales/en/workflows.ts
```diff
@@ -105,6 +110,7 @@ export const Strings = {
   'signal-name-label': 'Signal name',
   'signal-payload-input-label': 'Data',
   'signal-payload-input-label-hint': '(only single JSON payload supported)',
+  'update-modal-title': 'Send an Update',
```
laurakwhit, line 113: Do we need this and `update` on line 56?

src/lib/components/workflow/client-actions/update-confirmation-modal.svelte
```diff
@@ -0,0 +1,182 @@
...
+      success = result?.outcome?.success;
+      updateId = uuid();
+
+      if (success) {
+        toaster.push({
+          message: translate('workflows.update-success'),
+          variant: 'success',
+        });
+        hideModal();
+      }
```
laurakwhit, line 82: Looks like there's an edge case here where it doesn't error but isn't successful and the modal remains open with no error after update ⬇️ 

| During Update | After Update |
|-|-|
|Screenshot 2025-01-21 at 3 05 46 PM|Screenshot 2025-01-21 at 3 06 05 PM|

This is what the result looks like, it appears not to have an `outcome` ⬇️ 

Screenshot 2025-01-21 at 3 16 25 PM

Alex-Tideman: 👍 added check for success if also not a failure

src/lib/i18n/locales/en/workflows.ts
Alex-Tideman: Probably not, but I'd rather error on flexibility of separate translations in case the modal designs change.

PR #2511, opened 2025-01-16
Make WorkflowId the link in relationships, add interactive class to Link

Alex-Tideman:
## Description & motivation 💭 
Replace the Link icon with making the WorkflowId the link so it's easier to jump to the workflow. 

Screenshot 2025-01-16 at 10 48 49 AM

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/holocene/link.svelte
+++ b/src/lib/holocene/link.svelte
@@ -12,6 +12,7 @@
   type $$Props = HTMLAnchorAttributes & {
     href: string;
     active?: boolean;
+    interactive?: boolean;
     newTab?: boolean;
     class?: string;
     icon?: IconName;
@@ -24,6 +25,7 @@
   export { className as class };
   export let href: string;
   export let active = false;
+  export let interactive = false;
   export let newTab = false;
   export let icon: IconName = null;
   export let text: string = '';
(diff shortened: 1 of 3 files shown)


src/lib/holocene/link.svelte
```diff
@@ -12,6 +12,7 @@
   type $$Props = HTMLAnchorAttributes & {
     href: string;
     active?: boolean;
+    interactive?: boolean;
```
Alex-Tideman, line 15: Do I love this? No. But it's the least worst option

rossedfort: Looks like link takes a `class` prop, can you just specify the styles from the parent without using a prop?

```
<Link
      href={routeForEventHistory({
        namespace,
        workflow: workflow.id,
        run: workflow.runId,
      })}
      class={isActive ? 'text-white hover:text-indigo-200 focus-visible:text-indigo-200' : ''}
    >
      {workflow.id}
  </Link>
  ```

Alex-Tideman: That doesn't work because it gets merged with text styles that are the defaults and the defaults override

rossedfort: Gotcha, you would just need to inline the text styles then and use `twMerge`.

PR #2529, opened 2025-01-30
User Metadata for Activity and Timers

Alex-Tideman:
## Description & motivation 💭 
Users are now able to set User Metadata at the event level (Timers and Activities only for now). This PR will decode and show the User Metadata summary field as the primary label on the Timeline and in the Event History row. If it doesn't exist, fall back to previous primary label (Activity Type, Start to Fire Timeout).

Also includes a lot of cleaning up of unused code, simplifying the Timeline Expanded view so that everything is HTML inside a single SVG ForeignObject instead of multiple. The Payload fields will be stacked vertically and not have large gaps between them.

Screenshot 2025-01-31 at 1 39 54 PM
Screenshot 2025-01-31 at 1 40 03 PM
Screenshot 2025-01-31 at 1 41 09 PM
Screenshot 2025-01-31 at 1 41 20 PM
Screenshot 2025-01-31 at 1 41 35 PM
Screenshot 2025-01-31 at 2 33 20 PM

## Updated SS for Status Badge
Screenshot 2025-02-03 at 9 18 59 AM
Screenshot 2025-02-03 at 9 48 41 AM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/components/lines-and-dots/svg/group-details-row.svelte
```diff
@@ -77,96 +47,117 @@
...
+      if (group.pendingActivity.attempt > 1) {
+        status = translate('events.event-classification.retrying');
+      } else {
+        status = translate('events.event-classification.pending');
+      }
+    }
+  }
+
+  const onDecode = async () => {
+    await tick();
```
Alex-Tideman, line 68: Let it render first, then calc the height

src/lib/components/event/event-metadata-expanded.svelte
```diff
@@ -0,0 +1,37 @@
...
+  import MetadataDecoder from './metadata-decoder.svelte';
+
+  export let value: Payload;
+</script>
+
+<div class="flex {$$props.class}">
+  <div class="content code-block-row">
+    <div class="flex flex-col">
+      <p class="text-sm">Summary</p>
+      <MetadataDecoder {value} fallback="" let:decodedValue>
```
rossedfort, line 14: I think `fallback=""` is unnecessary here because the default value for that prop is already an empty string.

src/lib/components/event/event-summary-row.svelte
```diff
@@ -198,16 +200,31 @@
         {/if}
       </div>
     {/if}
+    {#if currentEvent?.userMetadata?.summary}
+      <MetadataDecoder
+        value={currentEvent.userMetadata.summary}
+        fallback=""
```
rossedfort, line 206: Same here

src/lib/components/event/metadata-decoder.svelte
```diff
@@ -0,0 +1,57 @@
...
+      }
+    } catch (e) {
+      console.error('Could not decode payloads');
+    }
+  };
+
+  $: decodePayload(value);
+</script>
+
+<slot {decodedValue} />
```
rossedfort, line 57: I think it makes more sense to have `decodePayload` return the decoded value instead of mutating local state. Then this would be

```
{#await decodeValue(value) then decodedValue}
  <slot {decodedValue} />
{/await}
```

Could then also add a slot for an error or something if desired.

src/lib/components/lines-and-dots/svg/group-details-row.svelte
```diff
@@ -77,96 +47,117 @@
...
-    <div class="flex h-full items-center justify-between text-sm text-white">
-      <div class="flex h-full items-center gap-2">
-        <div
-          class="px-4 py-1 text-black"
-          style="background-color: {getStatusColor(status)};"
-        >
-          {status ? spaceBetweenCapitalLetters(status) : group.label}
+<g role="button" tabindex="0" class="relative z-50">
+  <foreignObject {x} {y} {width} height={contentHeight}>
+    <div bind:this={innerContent} class="flex flex-col border-b border-subtle">
```
rossedfort, line 75: Couldn't you use `bind:offsetHeight` instead here? https://svelte.dev/docs/svelte/bind#Dimensions

src/lib/components/event/metadata-decoder.svelte
Alex-Tideman: Good call, for now I'm going to use the fallback for the :catch block value

src/lib/components/lines-and-dots/svg/group-details-row.svelte
Alex-Tideman: 👍

rossedfort (approved):

PR #2541, opened 2025-02-06
Add scroll to Event ID for links to Events

Alex-Tideman:
## Description & motivation 💭 
Allow users to navigate to the history page with a specific event `.../history#138` - and scroll to it. It will not open the row by default, just scroll to it.

Also add full event group on specific event page /history/event/138.

https://github.com/user-attachments/assets/31f28248-eb80-43a7-a560-841dddb90c9c

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

### Merge Checklist 

- [x] Need to figure out how to wait on page change before scrolling if event is not on current page


src/lib/stores/pagination.ts
```diff
@@ -151,21 +152,23 @@ export const outOfBounds = (
   return false;
 };
 
+export const hasId = (item: unknown): item is { id: string } => {
+  return (
+    typeof item === 'object' && Object.prototype.hasOwnProperty.call(item, 'id')
+  );
+};
+
```
rossedfort, line 160: I think we already have a `has()` helper function somewhere

src/lib/holocene/table/paginated-table/paginated.svelte
```diff
@@ -83,8 +86,21 @@
...
+    await tick();
+    let id = hash.slice(1);
+    const row = document?.querySelector(`[data-eventid="${id}"]`);
+    if (row) {
+      setTimeout(() => {
+        row?.scrollIntoView({ behavior: 'smooth' });
+      }, 500);
+    }
+  };
+
```
rossedfort, line 100: I'm not sure how I feel about this logic being in a holocene component and enabled by default. Was there a reason you didn't add this to the parent wherever it was needed?

src/lib/stores/pagination.ts
Alex-Tideman: 👍

src/lib/holocene/table/paginated-table/paginated.svelte
Alex-Tideman: Yeah I was on the fence about that too. I could expose jumpToHash to the slot but then you would need to run the scrollToHashEvent function on every row, which doesn't work if the row isn't rendered (i.e. on the same page you are looking at).

I do think we'll want this eventually for other paginated tables but agreed that this isn't ideal in that it's for events only. I added a hashField prop so that it only runs if that's provided and then you can pass whatever data-{hashField} attribute you want to scroll

rossedfort (approved):

PR #2559, opened 2025-02-14
Fix error in trimTrailingSlash

laurakwhit:
## Description & motivation 💭 

Seeing Sentry errors for `trimTrailingSlash` due to the argument being `null`. This PR checks for potentially `null` endpoint before calling `trimTrailingSlash`.

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/stores/data-encoder-config.ts
+++ b/src/lib/stores/data-encoder-config.ts
@@ -4,7 +4,7 @@ import { persistStore } from '$lib/stores/persist-store';
 import type { DataEncoderStatus } from '$lib/types/global';
 import { has } from '$lib/utilities/has';
 
-export const codecEndpoint = persistStore('endpoint', null, true);
+export const codecEndpoint = persistStore<string>('endpoint', null, true);
 
 export const passAccessToken = persistStore<boolean>(
   'passAccessToken',
--- a/src/lib/components/data-encoder-settings.svelte
+++ b/src/lib/components/data-encoder-settings.svelte
@@ -66,7 +66,7 @@
 
   const onConfirm = () => {
     error = '';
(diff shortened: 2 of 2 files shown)


src/lib/stores/data-encoder-config.ts
```diff
@@ -4,7 +4,7 @@ import { persistStore } from '$lib/stores/persist-store';
 import type { DataEncoderStatus } from '$lib/types/global';
 import { has } from '$lib/utilities/has';
 
-export const codecEndpoint = persistStore('endpoint', null, true);
+export const codecEndpoint = persistStore<string>('endpoint', null, true);
```
laurakwhit, line 7: Wondering if we should consider enabling `strictNullChecks`?

rossedfort: We probably should just not use `null` here, and use `undefined instead

`export const codecEndpoint = persistStore<string | undefined>('endpoint', undefined, true);`

Although I see `persistStore` is explicitly typed to accept `null` as a default value, I'm not sure why that is.

rossedfort (approved):

src/lib/stores/data-encoder-config.ts
rossedfort: (this is probably fine for now though)

laurakwhit: Yeah left it `null` for now as I didn't track down if there were any conditionals that relied on it being `null` or something like that. I turned on `strictNullChecks` just to see if we do this elsewhere and there were a bunch of places that popped up.

PR #2562, opened 2025-02-19
use maxlength instead

GraceGardner:
## Description & motivation 💭 
Add maxlength to workflowid. I went with 36 as the max length because that's what I guess is UUID standard. Original ticket was suggesting a max for an INT, but it looks like since then we changed from an int to a UUID. 

## Testing 🧪 
Go to start a work flow, try to put in an id longer than 36 characters. It now maxes out at 36 and does not allow the user to put in a larger input. 

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

Go to start a work flow, try to put in an id longer than 36 characters. It now maxes out at 36 and does not allow the user to put in a larger input. 

### Issue(s) closed 
DT-2701

### DIFF
--- a/src/lib/pages/start-workflow.svelte
+++ b/src/lib/pages/start-workflow.svelte
@@ -216,6 +216,7 @@
         bind:value={workflowId}
         label="Workflow ID"
         class="w-full grow"
+        maxlength={36}
         on:blur={(e) => onInputChange(e, 'workflowId')}
       />
       <Button


Alex-Tideman:
Let's check out the ticket again, it's talking about a Search Attribute with Int type not the Workflow Type field which is a string.

Alex-Tideman (changes requested):
I'm a little surprised the type check didn't catch this. The prop is maxLength, not maxlength.

src/lib/pages/start-workflow.svelte
```diff
@@ -216,6 +216,7 @@
         bind:value={workflowId}
         label="Workflow ID"
         class="w-full grow"
+        maxLength={36}
```
laurakwhit, line 219: Just noting that whatever we do here we'll likely also need to do for the Workflow ID input in the schedule form.

laurakwhit: Actually, looking at the original issue here and I think it lies with the custom search attribute input not with the inputs on these forms.

Alex-Tideman: ^ Yup this is for the Int type input for a Search Attribute

GraceGardner: Thank you both!

PR #2569, opened 2025-02-20
fix/remove imports from svelte/internal

rossedfort:
## Description & motivation 💭 
As of Svelte 5, svelte/internal has been removed
Screenshot 2025-02-20 at 10 51 34 AM

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/event/event-summary-row.svelte
+++ b/src/lib/components/event/event-summary-row.svelte
@@ -1,5 +1,4 @@
 <script lang="ts">
-  import { noop } from 'svelte/internal';
   import { fade, slide } from 'svelte/transition';
 
   import { page } from '$app/stores';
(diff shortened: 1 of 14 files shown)


src/lib/components/event/event-summary-row.svelte
```diff
@@ -46,7 +45,7 @@
   export let expandAll = false;
   export let typedError = false;
   export let active = false;
-  export let onRowClick: () => void = noop;
+  export let onRowClick: () => void = () => {};
```
GiantRobots, line 48: These look goo to me but this function could also be `() => void` no need to change just commenting.

GiantRobots (approved):

src/lib/components/event/event-summary-row.svelte
rossedfort: I thought `void` was only valid as a type declaration, not the function itself 🤔

PR #2571, opened 2025-02-20
Add lineBreakReplacer extension for CodeBlock

Alex-Tideman:
## Description & motivation 💭 
Replace \n in CodeBlock values with actual line breaks to make it easier to read. Will not replace if editing CodeBlock values

**Before**
Screenshot 2025-02-20 at 3 22 06 PM

**After**
Screenshot 2025-02-20 at 3 21 54 PM

**Before**
Screenshot 2025-02-20 at 3 28 57 PM

**After**
Screenshot 2025-02-20 at 3 29 10 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/holocene/code-block.svelte
+++ b/src/lib/holocene/code-block.svelte
@@ -91,6 +91,16 @@
 
   $: value = formatValue({ value: content, language });
 
+  const lineBreakReplacer = EditorView.updateListener.of((update) => {
+    if (editable) return;
+    const newText = update.state.doc.toString().replace(/\\n/g, '\n');
+    if (newText !== update.state.doc.toString()) {
+      update.view.dispatch({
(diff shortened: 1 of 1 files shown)


rossedfort (commented):
What about text vs. json? We might not want to add line breaks if there's a `\n` in the middle of a JSON payload or something, like

```
{
  "key": "my super long value\nwith a new line in it"
}
```

Alex-Tideman:
That's kinda the whole point, when it's json with several fields, and users want to line break long strings in a json payload. Like that second pair of screenshots, I think it's always easier to read with a linebreak.

rossedfort (approved):

PR #2581, opened 2025-02-27
Support messageType metadata on protobuf encoded Payloads

Alex-Tideman:
## Description & motivation 💭 
To use the `json/protobuf` encoding, you also need to pass the messageType metadata in the Payload. This PR adds support for including that field if `json/protobuf` encoding is used. If editing a schedule or starting a workflow from an existing workflow, if the encoding is `json/protobuf` and a messageType exists, set it as the initial value.

As part of this PR, I've also added base64 decoding all metadata fields instead of just the data field. This should only decode the metadata when 'Human readable' setting is set. For example, if you download the event history and select 'Decoded', the data and metadata fields should still be base64 encoded.

**To test:**

Start a new Workflow:

1. With json/plain
2. With json/protobuf + messageType and make sure initial encoding and messageType is set

Start a new Workflow from existing Workflow:

1. With json/plain
2. With json/protobuf + messageType

Create a new Schedule:

1. With json/plain
2. With json/protobuf + messageType

Edit a Schedule:

1. With json/plain
2. With json/protobuf + messageType and make sure initial encoding and messageType is set

Send a Signal:

1. With json/plain
2. With json/protobuf + messageType

Screenshot 2025-02-27 at 10 09 36 AM
Screenshot 2025-02-27 at 10 08 10 AM
Screenshot 2025-02-27 at 10 08 26 AM
Screenshot 2025-02-27 at 10 08 51 AM
Screenshot 2025-02-27 at 10 12 17 AM
Screenshot 2025-02-27 at 10 09 28 AM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

### DIFF
--- a/src/lib/components/payload-input-with-encoding.svelte
+++ b/src/lib/components/payload-input-with-encoding.svelte
@@ -39,7 +39,9 @@
   <h5 class="pb-1 text-sm font-medium">{label}</h5>
   <Card class="flex flex-col gap-2">
     <PayloadInput bind:input bind:loading {error} {id} {editing} />
-    <div class="flex items-end {editing ? 'justify-between' : 'justify-end'}">
+    <div
+      class="flex items-end gap-2 {editing ? 'justify-between' : 'justify-end'}"
+    >
       {#if editing}
         <div class="flex w-full flex-col gap-2">
           <RadioGroup
--- a/src/lib/components/payload-input-with-encoding.svelte
+++ b/src/lib/components/payload-input-with-encoding.svelte
@@ -9,7 +9,6 @@
 <script lang="ts">
   import { type Writable } from 'svelte/store';
 
-  import { onDestroy } from 'svelte';
   import { v4 as uuid } from 'uuid';
 
   import Card from '$lib/holocene/card.svelte';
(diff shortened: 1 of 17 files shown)


src/lib/components/payload-input-with-encoding.svelte
```diff
@@ -22,36 +22,49 @@
...
+              value="json/protobuf"
+              label="json/protobuf"
+            />
+          </RadioGroup>
+          {#if $encoding === 'json/protobuf'}
+            <Input
+              label={translate('workflows.message-type')}
+              bind:value={messageType}
+              {error}
+              id="messageType"
```
laurakwhit, line 64: Since there isn't a max character limit wondering if we want a `<Textarea />` here instead 🤔

laurakwhit (approved):

src/lib/components/payload-input-with-encoding.svelte
Alex-Tideman: Good call out. Generally speaking this is used as a single Proto definition name like `temporal.api.v1.ThingWithType` so it shouldn't need a textarea

PR #2589, opened 2025-03-07
Full Height Timeline and Event History

Alex-Tideman:
## Description & motivation 💭 
Allow users to expand the Timeline and the Event History table to full height so there is no inner scrolling. Move Sort/Expand/Filter/Pause/Download toggle buttons to above Timeline since they can apply to both Timeline and Event History. I've moved the Failed & Pending filter from a checkbox to inside the Event Type status filter and added a description as well. When Expanded, the Event History toggle buttons are sticky to quickly change. The Expand/Minimize option is a sticky preference.

Remove Expand All event history rows option since it can cause severe performance issues and you can open multiple rows which makes it's value limited.

Also included is an Expand/Collapse caret for CodeBlocks. If the height of the inner scrollable contents of the CodeBlock is greater than the maxHeight, show caret to expand to view the full CodeBlock.

Screenshot 2025-03-07 at 1 56 39 PM
Screenshot 2025-03-07 at 1 57 42 PM
Screenshot 2025-03-07 at 1 57 55 PM
Screenshot 2025-03-10 at 3 30 19 PM
Screenshot 2025-03-14 at 3 00 43 PM

Updates with new Tab Button

Screenshot 2025-03-17 at 11 40 04 AM
Screenshot 2025-03-17 at 11 40 17 AM
Screenshot 2025-03-17 at 11 42 17 AM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Draft Checklist 

- [ ] Icon/Verbiage for Minimize/Expand


Alex-Tideman:
One thing I'm not sure if we want to handle or do anything about: if the height is very short, do I disable the Expand button? I opted to not since it's sticky and they can land on workflows where it's already been toggled and it would be weird to have the current state disabled. However I do think it's not ideal to have a button that doesn't do anything in certain circumstances.

src/lib/layouts/workflow-history-layout.svelte
```diff
@@ -71,18 +88,73 @@
...
+            on:click={onSort}
+            >{reverseSort ? 'Descending' : 'Ascending'}</ToggleButton
+          >
+        {/if}
+        <ToggleButton
+          icon={$minimizeEventView ? 'minimize' : 'expand'}
+          data-testid="expandAll"
+          on:click={() => ($minimizeEventView = !$minimizeEventView)}
+          >{$minimizeEventView ? 'Minimized' : 'Expanded'}</ToggleButton
+        >
```
rossedfort, line 116: If it's easy, it might be worth adding a tooltip here explaining what "minimize" and "expand" do.

src/lib/layouts/workflow-history-layout.svelte
```diff
@@ -71,18 +88,73 @@
...
+<DownloadEventHistoryModal
+  bind:open={showDownloadPrompt}
+  {namespace}
+  workflowId={workflow.id}
+  runId={workflow.runId}
+/>
+
+<style lang="postcss">
+  .sticky-header {
+    @apply sticky top-12 z-50 border-b border-subtle backdrop-blur-sm;
```
rossedfort, line 158: This is an interesting styling choice 😂 
Screenshot 2025-03-17 at 9 50 03 AM

Why not just a solid bg?

rossedfort (approved):

src/lib/layouts/workflow-history-layout.svelte
Alex-Tideman: bg-off-white/50 👍

src/lib/layouts/workflow-history-layout.svelte
Alex-Tideman: It's easy to add, now for what the words should be...

PR #2604, opened 2025-03-18
Copy query

GraceGardner:
## Description & motivation 💭 
Adds a new copy button to the workflow page in two places. See in photos 

Screenshot 2025-03-18 at 3 49 10 PM
Screenshot 2025-03-18 at 3 49 22 PM

### Design Considerations 🎨 
Placement approved by Ash

## Testing 🧪 
Test by going to workflow page, add search params, press copy button to copy query. Check that the expected query is copied to clipboard. 

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
DT-2313


src/lib/components/workflow/workflow-advanced-search.svelte
```diff
@@ -87,6 +95,12 @@
       >
         {translate('common.search')}
       </Button>
+      <CopyButton
```
laurakwhit, line 98: We should only need this `CopyButton` in `src/lib/components/search-attribute-filter/index.svelte` since `SearchAttributeFilter` is the parent component of `WorkflowAdvancedSearch`, but let me know if you run into issues with that approach.
We should just need to remove the `showFilter` conditional so it also shows when the raw query is shown and maybe fix some styling. 

We might also want to consider adding a tooltip (e.g. "Copy Raw Query") so users know what the button does.

src/lib/components/search-attribute-filter/index.svelte
```diff
@@ -209,6 +217,14 @@
         {/if}
       </div>
     {/if}
+    {#if showFilter}
+      <CopyButton
```
laurakwhit, line 221: We'll also need to take into account that `SearchAttributeFilter` is used on the `/schedules` page as well ⬇️ 

Screenshot 2025-03-18 at 5 08 41 PM

GraceGardner: Thanks for noticing this Laura. I should have searched the repo for where this was imported. I appreciate you seeing this.

src/lib/components/search-attribute-filter/index.svelte
```diff
@@ -209,15 +218,22 @@
...
-      class="flex flex-col sm:flex-row {showClearAllButton
-        ? 'w-full justify-between'
-        : 'justify-end'}"
-    >
-      {#if showClearAllButton}
-        <Button variant="ghost" on:click={handleClearInput}
-          >{translate('common.clear-all')}</Button
-        >
+
+    <div class="flex flex-col items-center justify-end gap-1 sm:flex-row">
```
laurakwhit, line 222: | | |
|-|-|
|Screenshot 2025-03-20 at 9 34 30 AM|Screenshot 2025-03-20 at 9 48 40 AM|

PR #2621, opened 2025-03-26
Revert "Revert "ChipInput Updates""

laurakwhit:
* Reverts temporalio/ui#2619 
* Fixes `scrollIntoView()` so screen no longer jumps
* Accounts for length of `placeholder` r.e. the size of the input

| Before | After |
|-|-|
|Screenshot 2025-03-26 at 4 45 22 PM|Screenshot 2025-03-26 at 4 45 35 PM|

### DIFF
--- a/src/lib/holocene/combobox/combobox.svelte
+++ b/src/lib/holocene/combobox/combobox.svelte
@@ -387,9 +387,12 @@
       <Icon class="ml-2 shrink-0" name={leadingIcon} />
     {/if}
     <div
-      class="input-wrapper"
-      class:gap-1={multiselect}
-      class:p-1={multiselect && displayChips}
+      class={merge(
+        'input-wrapper',
+        multiselect && 'gap-1',
+        multiselect && 'm-1',
+        leadingIcon && multiselect && 'ml-2',
+      )}
     >
       {#if multiselect && isArrayValue(value) && value.length > 0}
         {#if displayChips}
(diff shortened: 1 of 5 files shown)


src/lib/holocene/input/chip-input.svelte
```diff
@@ -144,15 +132,38 @@
       on:blur={handleBlur}
       on:keydown|stopPropagation={handleKeydown}
       on:paste={handlePaste}
+      maxlength={maxLength && $values.length >= maxLength ? 0 : undefined}
+      size={placeholder.length || undefined}
```
laurakwhit, line 136: **Note:** Because `grow` is set in the css for the input, once the user starts typing the actual size isn't restricted to the length of the placeholder.

Alex-Tideman:
Screenshot 2025-03-27 at 8 49 43 AM

Could we give a wee bit more space after the leading icon?

src/lib/holocene/combobox/combobox.svelte
```diff
@@ -415,7 +418,15 @@
         type="text"
         value={displayValue}
         class:disabled
-        class={merge('combobox-input', className)}
+        class={merge(
+          'combobox-input',
+          multiselect
```
laurakwhit, line 423: > Could we give a wee bit more space after the leading icon?

@Alex-Tideman made some updates to our `Combobox` component (separate from the `ChipInput`) ⬇️ 

Screenshot 2025-03-27 at 9 23 43 AM

Alex-Tideman: 🤦 I didn't realize that was Combobox. What's the different between the combobox with chips and chip input?

Alex-Tideman: Ok just input and tab to create Chip, where Combobox is pick from select to create Chip

Alex-Tideman (approved):

src/lib/holocene/combobox/combobox.svelte
laurakwhit: Yeah we might want to revisit some of our inputs... I think we can narrow them down.

PR #2630, opened 2025-04-01
feat: integrate Temporal HTTP client for search attributes

rossnelson:
## Description & motivation 💭 

This PR refactors the search attributes service by implmenting the API requests
using the new `temporal-http-client` package.

The client provides a more strucuted approach to interacting with the temporal API.
By abstracting the request client, we are not only able to iterate on the client
functionality in isolation but also, the types used within the client can be
generated from the temporal OpenAPI specification.

This means that we can easily update the client to match the latest API changes
without having to worry about the types in the codebase.

For more insight into how this is type generation is accomplished, peep the
source here: https://github.com/stevekinney/temporal-http-client

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

Navigate to any page that loads namespace details. 

```
http://localhost:3000/namespaces/{namespace}
```

This will call the `getSearchAttributes` API and load the search attributes. It
also maps the response to a more human readable format.

### Draft Checklist 

- [ ] Fix up unit tests
- [ ] Resolve fetch warning

Fetch Warning
```
client.ts:1559 Loading http://localhost:8081/api/v1/namespaces/default/search-attributes using `window.fetch`. For best results, use the `fetch` that is passed to your `load` function: https://kit.svelte.dev/docs/load#making-fetch-requests
```

### DIFF
--- a/.env.http-api
+++ b/.env.http-api
@@ -1,0 +1,4 @@
+VITE_TEMPORAL_PORT="7134"
+VITE_API="http://localhost:8233"
+VITE_MODE="development"
+VITE_TEMPORAL_UI_BUILD_TARGET="local"
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "eslint:fix": "eslint --ignore-path .gitignore --fix .",
     "dev": "pnpm dev:ui-server -- --open",
     "dev:ui-server": ". ./.env.ui-server && vite dev --mode ui-server",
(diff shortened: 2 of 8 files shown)


rossnelson:
foregoing the client for now, migrating the schemas to workspaces https://temporalio.atlassian.net/browse/DT-2864

PR #2661, opened 2025-04-07
Progressive Relationships for large Root Workflows

Alex-Tideman:
## Description & motivation 💭 
To prevent over-fetching root workflow descendants, we put a upper limit (5,000) on how many it can fetch to show the relationship tree. This PR first fetches total rootWorkflow count, if over 5000 fetch only the parent + all child workflows of the current workflow. That way we can show a partial tree, and when a user clicks a node, it fetches that workflow's parent + all child workflows. Users can jump up and down the tree as they need to without fetching all descendants of root workflow.

If under 5000, fetch the full tree of workflows.

I've also removed the separate views for Schedule / CaN relationships and added them so it's a single visualization.

Screenshot 2025-04-16 at 1 55 14 PM
Screenshot 2025-04-16 at 1 55 24 PM
Screenshot 2025-04-16 at 1 56 01 PM
Screenshot 2025-04-16 at 1 56 15 PM
Screenshot 2025-04-16 at 1 56 33 PM
Screenshot 2025-04-16 at 1 56 43 PM
Screenshot 2025-04-16 at 1 57 35 PM
Screenshot 2025-04-16 at 1 57 48 PM
Screenshot 2025-04-16 at 1 58 03 PM

## Testing 🧪 

To test, create a tree of workflows under and over 5000 from the root workflow.

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Merge Checklist 

- [x] Fix refresh issue


src/lib/components/workflow/workflow-relationships.svelte
```diff
@@ -4,115 +4,60 @@
...
-                {first}
-                {previous}
-                {next}
-                current={runId}
-                {workflowId}
-                {namespace}
-              />
-            {/if}
-          </div>
+          {#await fetchAllDirectWorkflows( { namespace, parentWorkflowId, parentRunId, workflow: initialWorkflow }, )}
```
laurakwhit, line 40: NIT: Wondering if there's a solution here we put the `fullTree` value in a store elsewhere
```js
  let showFullTree = writable(false);
```
and set a generalized `fetchWorkflows` variable, something like this ⬇️ 

```js
    fetchAllRootWorkflowsCountPromise = fetchAllRootWorkflowsCount(
      namespace,
      rootWorkflowId,
      rootRunId,
    );
    const { count } = await fetchAllRootWorkflowsCountPromise;
    const intCount = parseInt(count);
    const overMaxLimit = intCount > MAX_UPPER_LIMIT;
    if (overMaxLimit) $showFullTree = true;

    fetchWorkflowPromise = overMaxLimit
      ? fetchAllDirectWorkflows({
          namespace,
          parentWorkflowId,
          parentRunId,
          workflow: initialWorkflow,
        })
      : fetchAllRootWorkflows(namespace, rootWorkflowId, rootRunId);
```

src/lib/components/workflow/relationships/workflow-family-node-tree.svelte
```diff
@@ -207,6 +227,115 @@
...
+        class="stroke-slate-100 stroke-2 transition-all duration-300 ease-in-out dark:stroke-slate-800"
+      />
+      <a href={routeForSchedule({ namespace, scheduleId: root.scheduleId })}>
+        <text
+          x={x - 5.25 * radius}
+          y={y - 2.5 * radius}
+          fill="currentColor"
+          text-decoration="underline"
+          text-anchor="end"
+          font-weight="500">{root.scheduleId}</text
```
laurakwhit, line 323: Looks like the schedule Id can go off the page ⬇️ 

Screenshot 2025-04-18 at 5 19 32 PM

src/lib/components/workflow/workflow-relationships.svelte
Alex-Tideman: 👍

src/lib/components/workflow/relationships/workflow-family-node-tree.svelte
Alex-Tideman: SVG text doesn't have widths/text-wrap. However you can zoom out or pan to see the full schedule name, so I think this is okay.

PR #2676, opened 2025-04-15
Let browser handle ctrl/shift+click on Link

Alex-Tideman:
## Description & motivation 💭 
To allow Windows users to open links in new tab, allow ctrl+click or shift+click to be handled by browser.

Also add noopener for newTab links

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/holocene/link.svelte
+++ b/src/lib/holocene/link.svelte
@@ -32,8 +32,9 @@
   export let light = false;
 
   const onLinkClick = (e: MouseEvent) => {
-    // Skip if middle mouse click or new tab
-    if (e.button === 1 || newTab || e.metaKey) return;
+    if (e.button === 1 || newTab || e.metaKey || e.ctrlKey || e.shiftKey)
+      return;
+
     e.preventDefault();
     goto(href);
   };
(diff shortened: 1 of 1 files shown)


src/lib/holocene/link.svelte
```diff
@@ -32,8 +32,9 @@
...
 
   const onLinkClick = (e: MouseEvent) => {
-    // Skip if middle mouse click or new tab
-    if (e.button === 1 || newTab || e.metaKey) return;
+    if (e.button === 1 || newTab || e.metaKey || e.ctrlKey || e.shiftKey)
+      return;
+
     e.preventDefault();
     goto(href);
   };
```
rossedfort, line 40: Remind me why we need to do any of this? Is it a remnant of some old url encoding stuff? Seems to me we should just remove the on:click handler from the anchor tag entirely.

Alex-Tideman: So we need this because otherwise the <a href...> would do a hard navigation and reload everything from the top level layout every time we navigate. So we need to use goto instead. But in the case of middle button click/newTab/meta/ctrl/shift, we don't want to use goto, but instead use the default broswer behavior of open in new tab/window

Alex-Tideman: But now that you say that, I just tested this with Svelte5 and i think its fixed

Alex-Tideman: Nevermind, we still need it

rossedfort: Ah ok, makes sense. Maybe there's something [here](https://svelte.dev/docs/kit/load#Rerunning-load-functions) we can do, but this is fine for now

rossedfort (approved):

PR #2691, opened 2025-04-22
Add optional scrollTo action to ChipInput 

laurakwhit:
## Description & motivation 💭 
Adds optional `scrollTo` prop to the `ChipInput` component so that the input can be scrolled into view after adding a value.

Screenshot 2025-04-22 at 4 21 49 PM

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
`DT-2912`

### DIFF
--- a/src/lib/holocene/input/chip-input.stories.svelte
+++ b/src/lib/holocene/input/chip-input.stories.svelte
@@ -20,6 +20,7 @@
       validator: isEmail,
       maxLength: undefined,
       chips: ['tobias@temporal.io'],
+      scrollTo: false,
     },
     argTypes: {
       label: { name: 'Label', control: 'text' },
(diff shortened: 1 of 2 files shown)


rossedfort:
I think #2689 might make this unnecessary, but I'll let @Alex-Tideman weigh in

src/lib/holocene/input/chip-input.stories.svelte
```diff
@@ -100,3 +102,14 @@
     await userEvent.click(input);
   }}
 />
+
+<Story
+  name="Scroll Input Into View"
+  args={{ class: 'max-h-20 w-96', scrollTo: true }}
```
laurakwhit, line 108: > I think https://github.com/temporalio/ui/pull/2689 might make this unnecessary,

We can still set the max height on the component via the class. Might be nice to have this option if it comes up in the future.

Alex-Tideman (approved):

PR #2692, opened 2025-04-23
Set position of Toast Notification container

Alex-Tideman:
## Description & motivation 💭 
Ability to set position of toaster container to:

'top-left' | 'top-center' | 'top-right' | 'bottom-left' | 'bottom-center' | 'bottom-right' with default as current 'bottom-right' position.

Set with toaster.setPosition(position). Once all toasts are dismissed, reset position to default position.

Screenshot 2025-04-23 at 9 28 24 AM
Screenshot 2025-04-23 at 9 28 07 AM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/holocene/toaster.svelte
+++ b/src/lib/holocene/toaster.svelte
@@ -23,10 +23,10 @@
     variants: {
       position: {
         'top-left': 'top-16 left-5',
-        'top-center': 'top-16 left-1/2 -translate-x-1/2',
+        'top-center': 'top-16 left-1/2 -translate-x-1/2 items-center',
         'top-right': 'top-16 right-5',
         'bottom-left': 'bottom-5 left-5',
-        'bottom-center': 'bottom-5 left-1/2 -translate-x-1/2',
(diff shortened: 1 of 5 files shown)


rossedfort (approved):

src/lib/holocene/toaster.svelte
```diff
@@ -7,16 +13,30 @@
   export let pop: Toaster['pop'];
   export let toasts: Toaster['toasts'];
   export let closeButtonLabel: string;
+  export let position: Writable<ToastPosition>;
```
laurakwhit, line 16: Should we be using the `toaster.position` from `stores/toaster` here instead?
```suggestion
    import { toaster } from '$lib/stores/toaster';
    $: ({position} = toaster)
```

Alex-Tideman: Will do

Alex-Tideman: ✅

PR #2693, opened 2025-04-23
DT-2884 - updates to Select and Input for quickstart guide

rossedfort:
## Description & motivation 💭 
- Add a `prefix` prop to `<Input />` 
- Add a `loading` prop to `<Select />`

#### input w/ prefix
Screenshot 2025-04-23 at 1 18 24 PM

#### select loading state

https://github.com/user-attachments/assets/d9de7edc-1cdc-47a3-9c85-ca6067273d11

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

### DIFF
--- a/src/lib/holocene/select/select.svelte
+++ b/src/lib/holocene/select/select.svelte
@@ -41,6 +41,7 @@
     value?: T;
     placeholder?: string;
     disabled?: boolean;
+    loading?: boolean;
     leadingIcon?: IconName;
     onChange?: (value: T) => void;
     'data-testid'?: string;
@@ -57,6 +58,7 @@
   export let value: T = undefined;
   export let placeholder = '';
   export let disabled = false;
+  export let loading = false;
   export let leadingIcon: IconName = null;
   export let onChange: (value: T) => void = () => {};
   export let menuClass: string | undefined = undefined;
@@ -113,7 +115,7 @@
     <MenuButton
       class={twMerge('w-full', !valid ? 'border-danger' : undefined)}
       hasIndicator={!disabled}
-      {disabled}
+      disabled={disabled || loading}
       controls="{id}-select"
       {variant}
       data-testid={`${$$restProps['data-testid'] ?? ''}-button`}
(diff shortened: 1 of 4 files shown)


src/lib/holocene/select/select.svelte
```diff
@@ -113,14 +115,16 @@
...
       controls="{id}-select"
       {variant}
       data-testid={`${$$restProps['data-testid'] ?? ''}-button`}
     >
       <slot name="leading" slot="leading">
         {#if leadingIcon}
           <Icon name={leadingIcon} />
+        {:else if loading}
+          <Icon name="spinner" class="animate-spin" />
         {/if}
```
Alex-Tideman, line 128: Should we try to keep the input the same width even when it's loading so it doesn't have that jump when switching between loading/not loading?

laurakwhit: Also wondering if we could just pass `loading` to the `MenuButton` since the underlining `Button` component takes it as a prop.

rossedfort: I'm not sure how we would do that within the component itself, other than specifying a fixed on the MenuContainer (which we probably shouldn't do). Currently you can add a fixed with to the parent element and it won't grow.


https://github.com/user-attachments/assets/918fb235-391a-48a9-bef1-e299f602dced

rossedfort: Although I don't like how the text still jumps. I thought about replacing the chevron icon with the spinner, but Button has the spinner leading the text, so I wanted to keep it consistent

rossedfort: How about this?

https://github.com/user-attachments/assets/b6ff66a4-ee37-4a17-bccd-165e1779ab74

Alex-Tideman: I like that

rossedfort: @laurakwhit 

> Also wondering if we could just pass loading to the MenuButton since the underlying Button component takes it as a prop.

Good ideal, although currently `<Button />` doesn't disable itself when `loading` is true, and I want that behavior for `<Select />`. I thought `<Button />` used to do that, but maybe not.

Alex-Tideman (approved):

PR #2699, opened 2025-04-28
Don't use all namespaces store for namespace columns

Alex-Tideman:
## Description & motivation 💭 
Remove the need to iterate over all the namespaces in the namespace store.

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


rossedfort (approved):
Lgtm but am curious why we don't need to be doing this? What happens if someone visits a workflow list page for a namespace that doesn't already have the columns configured?

PR #2705, opened 2025-05-02
Add date clear funtionality to datepicker

KATIETOLER:
This adds functionality to fix a bug on the auditlog filter where a user cannot clear dates. 

[cloud PR](https://github.com/temporalio/cloud-ui/pull/1408)

### DIFF
--- a/src/lib/holocene/date-picker.svelte
+++ b/src/lib/holocene/date-picker.svelte
@@ -46,6 +46,8 @@
       const date = parseInt(inputDateSplit[1]);
       const newDate = new Date(year, month, date);
       dispatch('datechange', newDate);
+    } else if (inputDate.length === 0) {
+      dispatch('dateclear', {});
     }
   };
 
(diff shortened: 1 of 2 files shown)


src/lib/holocene/date-picker.svelte
```diff
@@ -46,6 +46,8 @@
       const date = parseInt(inputDateSplit[1]);
       const newDate = new Date(year, month, date);
       dispatch('datechange', newDate);
+    } else if (inputDate.length === 0) {
+      dispatch('dateclear', {});
```
Alex-Tideman, line 50: I think we'll want to do this when you click the x and clear the whole input as well. That's more likely than backspacing the full date.

laurakwhit: The `Input` has a [clear](https://github.com/temporalio/ui/blob/main/src/lib/holocene/input/input.svelte#L83) event we could utilize here.

KATIETOLER:
Not currently needed.

PR #2708, opened 2025-05-06
API Pagination for Workflows List

Alex-Tideman:
## Description & motivation 💭 
Let users view all Workflows. This should also increase initial page load since by default we will load only 100 results instead of 1,000.

When a user pages forward, they will receive the next page of workflows at the time of the start of the search (page load, query/filter). Results won't shift to next page when you go forward or backwards in pages.

The table is reset to the first page with new results when a query changes (either with filters, raw query string or page size) or when they click the Refresh button or manually refresh the page. I've also added a default debounce of 350ms for nextPage and better checks to prevent over-fetching when you hold down the right arrow or click the right arrow button.

Batch operations/filtering should work exactly as before.

Screenshot 2025-05-06 at 10 27 18 AM

https://github.com/user-attachments/assets/3a9d969f-e9aa-41d7-99a1-b2b1ae0e2180

### How was this tested 👻 

- [x] Manual testing
- [x] E2E tests added
- [ ] Unit tests added


src/lib/components/workflow/workflows-summary-configurable-table.svelte
```diff
@@ -66,78 +66,81 @@
...
-    {empty}
-    slot="headers"
+{#key [namespace, $workflowsQuery, $refresh]}
+  <PaginatedTable
+    total={$workflowCount.count}
+    {onFetch}
     let:visibleItems
-    workflows={visibleItems}
+    aria-label={translate('common.workflows')}
+    pageSizeSelectLabel={translate('common.per-page')}
```
laurakwhit, line 79: Noticed when changing the pageSize (and there are new Workflows) the total count gets a bit off ⬇️ 

Screenshot 2025-05-07 at 3 59 06 PM

Alex-Tideman: Good catch. A couple of things here:

1. I wasn't refretching the workflow counts when the perPage size changed. Fixed that.
2. There can be a slight difference between the workflow counts total and the actual number returned in ListWorkflows due to timing. So the last page (which can be the first page) might have an endIndex (total number) different than the total count. So I adjusted it to use endIndex instead of total in that situation so it always used the stores total instead of the total prop if it's different.

rossnelson (approved):
I also tried this out in a cloud test [pr](https://cloud-ui-git-test-ui-8759b2b.preview.thundergun.io/namespaces/cross-provider-migration-test.temporal-dev/workflows). The problem is that I don't see a cloud namespace with over 100 workflows. So I spun up Canary and tested it locally. It's looking good! You got to the issue @laurakwhit mentioned above before I could reproduce it. 

nice work

PR #2721, opened 2025-05-12
Migrate accordion

GraceGardner:
## Description & motivation 💭 
Migrate accordion, accordion group, and accordion light to runes mode. 

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
DT-2938

### DIFF
--- a/src/lib/holocene/accordion/accordion-group.svelte
+++ b/src/lib/holocene/accordion/accordion-group.svelte
@@ -1,3 +1,7 @@
+<script lang="ts">
+  let { children } = $props();
+</script>
+
 <div class="*:border-t-0 [&>*:first-child]:border-t">
-  <slot />
+  {@render children?.()}
 </div>
--- a/src/lib/holocene/accordion/accordion.stories.svelte
+++ b/src/lib/holocene/accordion/accordion.stories.svelte
@@ -5,7 +5,17 @@
 
   import Accordion from './accordion.svelte';
 
-  export const meta = {
+  interface AccordionProps {
+    title: string;
+    subtitle?: string;
+    open?: boolean;
+    expandable?: boolean;
+    error?: string;
+    icon?: string;
+    onToggle?: () => void;
+  }
+
+  export const meta: Meta<AccordionProps> = {
     title: 'Accordion',
     component: Accordion,
     args: {
@@ -16,18 +26,14 @@
       error: '',
     },
     argTypes: {
-      title: { name: 'Title', control: 'text' },
-      subtitle: { name: 'Subtitle', control: 'text' },
-      open: { name: 'Open', control: 'boolean' },
-      expandable: { name: 'Expandable', control: 'boolean' },
-      error: { name: 'Error', control: 'text' },
-      icon: {
-        name: 'Icon',
-        control: 'select',
-        options: iconNames,
-      },
+      title: { control: 'text' },
+      subtitle: { control: 'text' },
+      open: { control: 'boolean' },
+      expandable: { control: 'boolean' },
+      error: { control: 'text' },
+      icon: { control: 'select', options: iconNames },
     },
-  } satisfies Meta<Accordion>;
+  };
 </script>
 
 <script lang="ts">
--- a/src/lib/holocene/accordion/accordion.svelte
+++ b/src/lib/holocene/accordion/accordion.svelte
@@ -1,6 +1,7 @@
 <script lang="ts">
   import type { HTMLAttributes } from 'svelte/elements';
 
+  import type { Snippet } from 'svelte';
   import { twMerge as merge } from 'tailwind-merge';
   import { v4 } from 'uuid';
 
@@ -8,7 +9,7 @@
   import type { IconName } from '$lib/holocene/icon';
   import Icon from '$lib/holocene/icon/icon.svelte';
 
-  interface $$Props extends HTMLAttributes<HTMLDivElement> {
+  type Props = HTMLAttributes<HTMLDivElement> & {
     title: string;
     id?: string;
     subtitle?: string;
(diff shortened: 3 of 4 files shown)


src/lib/holocene/accordion/accordion-group.svelte
```diff
@@ -1,3 +1,7 @@
+<script lang="ts">
+  let { children } = $props();
```
Alex-Tideman, line 2: Let's type this:

` let { children }: { children?: Snippet }`

GraceGardner: Thanks for catching this one!

src/lib/holocene/accordion/accordion.svelte
```diff
@@ -59,21 +67,21 @@
...
-            <slot name="summary" />
+            {@render summary?.()}
           </div>
         </div>
         <div
           class="flex flex-row items-center gap-2 pr-2"
-          on:click|stopPropagation
-          on:keyup|stopPropagation
+          onclick={(e) => e.stopPropagation()}
+          onkeyup={(e) => e.stopPropagation()}
```
rossedfort, line 78: I think this will break anywhere we're using click or keyup handlers like `<Accordion on:click={doSomething} />`. The old syntax was forwarding the listener down to this div element, this new syntax is just calling `stopPropagation()`, but doesn't account for if an `onclick` or `onkeyup` callback function is passed to Accordion.

andrewzamojc: Not sure about this spot, but rather than stopping propagation, I tend to `preventDefault`. It sort of marks it as handled, so up the chain, a handler can check `if (e.defaultPrevented)`.

src/lib/holocene/accordion/accordion-group.svelte
```diff
@@ -1,3 +1,9 @@
+<script lang="ts">
+  import type { Snippet } from 'svelte';
+
+  const { children }: { children?: Snippet } = $props();
```
andrewzamojc: Comment: I'm not sure about convention, but I don't mind required children. The component is pointless without them.

src/lib/holocene/accordion/accordion-group.svelte
```diff
@@ -1,3 +1,9 @@
...
+</script>
+
 <div class="*:border-t-0 [&>*:first-child]:border-t">
```
andrewzamojc, line 7: Question: Could the border-t go on the accordion directly using `first:border-t` rather than here on the parent? I see its existing so not critical.

src/lib/holocene/accordion/accordion.stories.svelte
```diff
@@ -5,7 +5,17 @@
 
   import Accordion from './accordion.svelte';
 
-  export const meta = {
+  interface AccordionProps {
```
andrewzamojc: Question: Could we import these from the Accordion component instead of redefining them?

src/lib/holocene/accordion/accordion.stories.svelte
```diff
@@ -16,18 +26,14 @@
...
-        name: 'Icon',
-        control: 'select',
-        options: iconNames,
-      },
+      title: { control: 'text' },
+      subtitle: { control: 'text' },
+      open: { control: 'boolean' },
+      expandable: { control: 'boolean' },
+      error: { control: 'text' },
+      icon: { control: 'select', options: iconNames },
```
andrewzamojc, line 24: Comment: Storybook might set up most of these automatically now with `Meta<AccordionProps>`.

src/lib/holocene/accordion/accordion.svelte
```diff
@@ -1,14 +1,15 @@
...
+  import type { Snippet } from 'svelte';
   import { twMerge as merge } from 'tailwind-merge';
   import { v4 } from 'uuid';
 
   import Badge from '$lib/holocene/badge.svelte';
   import type { IconName } from '$lib/holocene/icon';
   import Icon from '$lib/holocene/icon/icon.svelte';
 
-  interface $$Props extends HTMLAttributes<HTMLDivElement> {
+  type Props = HTMLAttributes<HTMLDivElement> & {
```
andrewzamojc: Comment: possibly could use `interface ... extends` so its still an interface.

src/lib/holocene/accordion/accordion.svelte
```diff
@@ -19,19 +20,26 @@
...
-  export let error = '';
-  export let onToggle = () => {};
+    summary?: Snippet;
+    action?: Snippet;
+    children?: Snippet;
+  };
 
-  let className = '';
-  export { className as class };
+  let {
```
andrewzamojc, line 30: Comment: maybe `const`

src/lib/holocene/accordion/accordion.svelte
rossedfort: Oh ya know what, I read `stopPropagation` as `preventDefault` hah, so I think this is totally fine. I think we were getting double click events or something when a custom action was passed in, hence why the `stopPropagation` directives were added.

src/lib/holocene/accordion/accordion.svelte
andrewzamojc: oh...actually most of the docs destructure with let. Maybe to do with reactivity. I'm still adjusting from React 😅

src/lib/holocene/accordion/accordion.stories.svelte
GraceGardner: I'm a bit less sure about this one... Let me know what you think we should do here or if you have any docs that are helpful pointing me in the right direction.

src/lib/holocene/accordion/accordion.svelte
GraceGardner: @rossedfort Tested this by adding an alert in an onclick event and it seems to be working fine.  I left the alert on the first accordion in storybook if you want to check?

rossedfort (approved):

src/lib/holocene/accordion/accordion.svelte
rossedfort: @GraceGardner I tested and it seems to be working. Let's remove the alert before merging though. Instead, we can add `action('onclick')`

GraceGardner: Will do, thanks Ross

src/lib/holocene/accordion/accordion.stories.svelte
andrewzamojc: No change needed:

No worries. This is a sidetrack and doesn't need to be considered in this PR. For interest sake or future work, I found this bit on storybook choosing controls automatically https://storybook.js.org/docs/essentials/controls#choosing-the-control-type

PR #2724, opened 2025-05-13
Allow Terminate for closed workflows with newExecutionRunId

Alex-Tideman:
## Description & motivation 💭 
If a workflow has a newExecutionRunId - a newer workflow in the chain of workflows - allow it to terminate the latest run and include the firstExecutionRunId in the terminate call.

This is already the behavior on batch if a single workflow with newExecutionRunId is selected and terminated.

Note: We will not check if latest run is open/closed. If it is closed, it will just error with the usual `workflow execution already completed` message.

Screenshot 2025-05-14 at 2 03 03 PM
Screenshot 2025-05-14 at 2 03 14 PM

https://github.com/user-attachments/assets/25f71c0a-ead1-49d8-b0ae-66cf6eba6b84

### How was this tested 👻 

- [x] Manual testing
- [x] E2E tests added
- [ ] Unit tests added


src/lib/components/workflow-actions.svelte
```diff
@@ -206,23 +206,22 @@
...
           }),
         )}
+      disabled={workflowCreateDisabled($page)}
       data-testid="start-workflow-button"
     >
       {translate('workflows.start-workflow-like-this-one')}
     </MenuItem>
+    {#if terminateEnabled && next}
+      <MenuDivider />
+      <MenuItem
```
laurakwhit, line 216: If we're no longer showing the tooltip on disabled can we remove `resetDescription`?

src/lib/components/workflow-actions.svelte
```diff
@@ -206,23 +206,22 @@
             workflowType: workflow.name,
           }),
         )}
+      disabled={workflowCreateDisabled($page)}
```
laurakwhit, line 209: Wondering if we want some tooltip text here for disabled now that we're not just hiding the menu item.

src/lib/components/workflow-actions.svelte
Alex-Tideman: We still use resetDescription on running workflows for the reset menu item

src/lib/components/workflow-actions.svelte
Alex-Tideman: Probably? We're going to refactor these to a new UI instead of the split button, so I think this is okay for now.

Alex-Tideman: We have a bunch of spots where we just disable if read-only so I'm okay with this

PR #2741, opened 2025-05-28
[DT-2881] info alignment

andrewzamojc:
## Description & motivation 💭 

This PR aligns the details in the workflow header, so the information is visually associated. To accomplish this, the PR:
- Uses CSS grid to align detail values in a stack next to the labels.
- Adds a new component DetailList. Semantically it is a definition list and offers a similar structure.
  - This component captures the required CSS grid.
  - Includes a DetailListColumn to position details into columns on larger screens. It is a `div` under the `dl` which is allowed in HTML. On small screens the columns stack.
  - Moves link and text functionality from `WorkflowDetail` component to the more generic `DetailsList`, including tooltips and truncation.
  - Updates `WorkflowDetails` to use `DetailList` instead of `WorkflowDetail`. Tested turning on `deployments` column and everything seemed ok.
  - Does NOT delete `WorkflowDetail` since it is used in a couple other places that can be updated in the future: `version-details.svelte` and `workflow-summary.svelte`.
- Adds a little extra vertical space between major elements in the workflow header, to get closer to the design and feel less cramped.
- Changes clock icon to the text "Duration" as requested in the design.
- Move SDK to last column, requested by Design
- Removed `mono` spaced font style from labels, requested by Design
- The design mentions heading sizes, but they seem like they're already updated and consistent.

#### BEFORE
Screenshot 2025-06-04 at 11 15 14 AM

#### AFTER
Screenshot 2025-06-06 at 11 12 32 AM

#### TRUNCATION

Screenshot 2025-06-05 at 3 49 04 PM

#### STACKED
Screenshot 2025-06-06 at 11 12 50 AM

### Design Considerations 🎨 
- spacing is easy to tweak if needed
- due to existing truncation and now the css grid, if a future design requires multi-line values, this component would need a code change

## Testing 🧪 

I don't see the `<SDKLogo>` so it's hard to know if that'll work :S How can I test this?

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added
- [x] screen reader

- Run the code locally or in preview
- Make a workflow and go to its details page
- Notice the details near the top are aligned nicely
- squish the page and see it truncate and then stack


src/lib/components/lines-and-dots/workflow-details.svelte
```diff
@@ -55,106 +62,104 @@
...
       {/if}
-    </div>
+    </DetailListColumn>
   {/if}
-  <div
-    class="flex w-full flex-col gap-2 {deployment ? '2xl:w-1/4' : 'xl:w-1/3'}"
-  >
-    <SdkLogo />
+
+  <SdkLogo />
```
andrewzamojc: Annotation: I don't see this when I run locally. How can I test this?

src/lib/components/detail-list/detail-list.svelte
```diff
@@ -0,0 +1,18 @@
...
+  let { children, 'aria-label': ariaLabel, rowCount }: Props = $props();
+</script>
+
+<dl
+  class="grid auto-cols-[max-content_minmax(0,max-content)] grid-flow-col gap-x-4 gap-y-1"
+  style={`--row-count: ${rowCount}`}
+  aria-label={ariaLabel}
+>
+  {@render children()}
+</dl>
```
andrewzamojc, line 18: Annotation: In general, this grid works like so:
- implicit grid, always adding two columns, one for label one for value
- the label and value components are always set to column 1 and column 2
- the column component is a subgrid in both directions, making the above work in columns
- the `rowCount` prop is required due to constraints of the CSS subgrid, which does not tend to create implicit rows. In the future we could use new CSS selectors like "sibling index" to count the elements in CSS.

Alex-Tideman:
Screenshot 2025-06-04 at 11 05 17 AM

Think we may need some responsive love. We might need to make all the columns equal widths instead of growing since some of these values can be very long

src/lib/components/lines-and-dots/workflow-details.svelte
andrewzamojc: aha, local canary data!

andrewzamojc:
@Alex-Tideman Thanks for the canary data to test out real values and see that sdk logo!

src/lib/components/lines-and-dots/workflow-details.svelte
```diff
@@ -55,106 +63,102 @@
...
+
       {#if deploymentVersion}
-        <WorkflowDetail
-          title={translate('deployments.deployment-version')}
-          content={workflow.searchAttributes.indexedFields[
+        <DetailListLabel>
+          {translate('deployments.deployment-version')}
+        </DetailListLabel>
+        <DetailListTextValue
+          text={workflow.searchAttributes.indexedFields[
```
Alex-Tideman: Let's use the deploymentVersion var in script

src/lib/components/lines-and-dots/workflow-details.svelte
```diff
@@ -55,106 +63,102 @@
   };
```
Alex-Tideman, line 65: I'd say we just go ahead and update the above $: to the new Svelte5 hotness while we're in here

src/lib/components/lines-and-dots/workflow-details.svelte
andrewzamojc: Done

src/lib/components/lines-and-dots/workflow-details.svelte
andrewzamojc: Done. Let me know if it looks correct. Values still appearing in the UI so that's a good sign.

andrewzamojc: Oops, I'll need to fix the latestRunId, based on failing integration tests.

andrewzamojc: All set!

Issue #1266, opened 2022-11-03
[Feature Request] Support ui-server running over https


tengfmu:

adammy123:
Suggested small code tweak which worked for me (required re-building of binary):
```
// Start UI server.
func (s *Server) Start() error {
	fmt.Println("Starting UI server...")
	cfg, err := s.cfgProvider.GetConfig()
	if err != nil {
		return err
	}

	address := fmt.Sprintf("%s:%d", cfg.Host, cfg.Port)
	if cfg.TLS.CertFile != "" && cfg.TLS.KeyFile != "" {
		fmt.Println("Starting UI server with TLS...")
		s.httpServer.Logger.Fatal(s.httpServer.StartTLS(address, cfg.TLS.CertFile, cfg.TLS.KeyFile)) //<-- start serving over https
	} else {
		fmt.Println("Starting UI server without TLS...")
		s.httpServer.Logger.Fatal(s.httpServer.Start(address)) //<-- original line
	}
	return nil
}
```

PR #2745, opened 2025-05-28
Add TLS support for ui-server

karelinoleg:
## Description & motivation 💭 
Currently when start ui-server through command line or docker image, it always running over http. To make it running from more production env, it would be better to support running the server over https.

As a workaround, we could setup reverse proxy to provide https access to ui server, but still it would be great if we could support https natively.

There are an existing issues:
- https://github.com/temporalio/ui/issues/1262
- https://github.com/temporalio/ui/issues/1266

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/server/config/development.yaml
+++ b/server/config/development.yaml
@@ -48,6 +48,9 @@ tls:
   keyData:
   enableHostVerification: false
   serverName:
+uiServerTLS:
+  certFile:
+  keyFile:
 codec:
   endpoint:
   passAccessToken: false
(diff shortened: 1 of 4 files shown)


karelinoleg:
@GiantRobots Hi! May I ask you to check my pull request?

pvf4HQM6dn:
@GiantRobots this PR adds a very needed feature

PR #2752, opened 2025-05-29
add an integration test to make sure dark mode toggling works through localstorage

andrewzamojc:
### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/dark-mode-icon-button.svelte
+++ b/src/lib/components/dark-mode-icon-button.svelte
@@ -35,5 +35,6 @@
     label={buttonText}
     icon={buttonIcon}
     on:click={cycleDarkModePreference}
+    data-testid="dark-mode-icon-button"
   />
 </Tooltip>
--- a/tests/integration/dark-mode-toggle.spec.ts
+++ b/tests/integration/dark-mode-toggle.spec.ts
@@ -1,0 +1,53 @@
+import { expect, test } from '@playwright/test';
+
+import {
+  mockWorkflowsApis,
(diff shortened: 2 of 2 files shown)


src/lib/components/dark-mode-icon-button.svelte
```diff
@@ -35,5 +35,6 @@
     label={buttonText}
     icon={buttonIcon}
     on:click={cycleDarkModePreference}
+    data-testid="dark-mode-icon-button"
```
andrewzamojc, line 38: Comment: Since the aria-label changes when you press the button, getByRole doesn't work well without re-selecting over and over. By using testid, I can get the element once and use it throughout the test.

src/lib/holocene/navigation/navigation-button.svelte
```diff
@@ -25,6 +25,7 @@
     <div
       class="flex h-6 w-6 items-center after:absolute after:left-[calc(100%_+_1.5rem)] after:top-0 after:hidden after:h-8 after:items-center after:bg-slate-800 after:p-1 after:px-2 after:text-xs after:text-white after:content-[attr(data-tooltip)] group-data-[nav=closed]:hover:after:flex"
       data-tooltip={tooltip}
+      aria-hidden="true"
```
andrewzamojc, line 28: Annotation: This icon is pure decoration / redundant information to a non-visual user. By aria-hiding it, the navigation button now has a clean accessible name.

PR #2755, opened 2025-06-03
Add activity type to row

GraceGardner:
## Description & motivation 💭 
Local activities should show activity type instead of the market name. 

- Changes Event Summary Row to Svelte 5 
- Adds new decodeLocalActivity function that returns the key and value of the activity type using the codec server
- Uses the new function to assign the activityType key and value to the PrimaryLocalAttribute 
- Displays the PrimaryLocalAttribute if it is a PrimaryLocalAttribute 
- Needed to make the decodeLocalActivity function async, so I did move it back to the event-summary-row. I was getting errors about async needing to be in the top layer. Hope that's okay. 

Local Activity now shows as expected when linking ui to cloud ui on the path from ticket. 

Screenshot 2025-06-06 at 9 18 38 AM

## Testing 🧪 
Pull down this pr, link ui to cloud ui and check that all local activities no longer show "Marker Name" and now show "Activity Type" with correct value 

http://localhost:3000/namespaces/prod.cp/workflows/export-status-backfill-cron/d337b16f-92ea-4e40-b2c5-906d6ea9fa52/history
### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
DT-2962

### DIFF
--- a/src/lib/components/event/event-summary-row.svelte
+++ b/src/lib/components/event/event-summary-row.svelte
@@ -20,7 +20,11 @@
   } from '$lib/models/event-groups/get-event-in-group';
   import { authUser } from '$lib/stores/auth-user';
   import { relativeTime, timeFormat } from '$lib/stores/time-format';
-  import type { Payload, WorkflowEvent } from '$lib/types/events';
+  import type {
+    IterableEvent,
+    Payload,
+    WorkflowEvent,
+  } from '$lib/types/events';
   import {
     cloneAllPotentialPayloadsWithCodec,
     decodePayloadAttributes,
@@ -54,8 +58,8 @@
   import MetadataDecoder from './metadata-decoder.svelte';
 
   interface Props {
-    event: WorkflowEvent;
-    group?: EventGroup;
+    event: IterableEvent;
+    group: EventGroup | undefined;
     initialItem: WorkflowEvent;
     index: number;
     compact?: boolean;
@@ -88,7 +92,7 @@
     };
   };
 
-  export const decodeLocalActivity = async (
+  const decodeLocalActivity = async (
     event,
   ): Promise<SummaryAttribute | undefined> => {
     const settings = {
@@ -102,7 +106,6 @@
     };
     const accessToken = $authUser.accessToken;
     const namespace = page.params.namespace;
-    if (!isLocalActivityMarkerEvent(event)) return;
     try {
       const convertedAttributes = await cloneAllPotentialPayloadsWithCodec(
         event.attributes,


src/lib/components/event/event-summary-row.svelte
```diff
@@ -37,104 +39,148 @@
...
-  export let event: IterableEvent;
-  export let group: EventGroup | undefined = undefined;
-  export let initialItem: IterableEvent | undefined;
-  export let index = 0;
-  export let compact = false;
-  export let expanded = false;
-  export let typedError = false;
-  export let active = false;
-  export let onRowClick: () => void = () => {};
+  let expanded = $state(expandedProp);
```
GraceGardner, line 84: For expanded, should I be using bind?

src/lib/components/event/event-summary-row.svelte
```diff
@@ -37,104 +39,148 @@
...
+    if (isLocalActivityMarkerEvent(event))
+      primaryLocalAttribute = decodeLocalActivity(event);
   });
 </script>
 
 <tr
-  class="row dense"
-  id={`${event.id}-${index}`}
+  class="dense flex select-none items-center gap-4 px-2 text-sm no-underline"
+  class:border={failure() || canceled() || terminated()}
```
GraceGardner: I kind of thought this was a little cleaner, but let me know if we would rather put this multiple times with the failure, canceled, an terminated.

src/lib/components/event/event-summary-row.svelte
```diff
@@ -37,104 +39,148 @@
...
 
-  $: ({ workflow, run, namespace } = $page.params);
-  $: href = routeForEventHistoryEvent({
-    eventId: event.id,
-    namespace,
-    workflow,
-    run,
-  });
-  $: attributes = formatAttributes(event);
+  const { workflow, run, namespace } = $page.params;
```
Alex-Tideman: Let's make this const { workflow, run, namespace } = $derived(page.params)

src/lib/components/event/event-summary-row.svelte
```diff
@@ -1,6 +1,8 @@
 <script lang="ts">
   import { fade, slide } from 'svelte/transition';
 
+  import { onMount } from 'svelte';
+
   import { page } from '$app/stores';
```
Alex-Tideman: Should be import { page } from '$app/state'

src/lib/components/event/event-summary-row.svelte
```diff
@@ -37,104 +39,148 @@
...
-  export let onRowClick: () => void = () => {};
+  let expanded = $state(expandedProp);
+  let primaryLocalAttribute = $state<SummaryAttribute | undefined>(undefined);
 
-  $: selectedId = isEventGroup(event)
-    ? Array.from(event.events.keys()).shift()
-    : event.id;
+  const selectedId = $derived(() =>
+    isEventGroup(event) ? Array.from(event.events.keys()).shift() : event.id,
+  );
```
Alex-Tideman: All of these $derived() values shouldn't be functions. For example, 

`const selectedId = $derived(isEventGroup(event) ? Array.from(event.events.keys()).shift() : event.id);`

src/lib/components/event/event-summary-row.svelte
```diff
@@ -37,104 +39,148 @@
...
+    event,
+    group,
+    initialItem,
+    index,
+    compact = false,
+    typedError = false,
+    active = false,
+    expanded: expandedProp = false,
+    onRowClick = () => {},
+  } = $props();
```
andrewzamojc: Suggestion: I don't see the types from the old code like `IterableEvent`. I've been typing `$props` with a separate interface, and marking the ones with a default value as optional with the `?`.

```
interface Props {
  event: IterableEvent;
  group?: EventGroup;
  // etc
}
const { 
  event,
  group = undefined
}: Props = $props();
```

src/lib/components/event/event-summary-row.svelte
```diff
@@ -37,104 +39,148 @@
...
-    : '';
+  const attributes = $derived(() => formatAttributes(event));
 
-  const onLinkClick = () => {
-    expanded = !expanded;
-    onRowClick();
-  };
+  const currentEvent = $derived(() =>
+    isEventGroup(event) ? event.events.get(selectedId()) : event,
+  );
```
Alex-Tideman: Same

` const currentEvent = $derived(isEventGroup(event) ? event.events.get(selectedId()) : event);`

src/lib/components/event/event-summary-row.svelte
```diff
@@ -37,104 +39,148 @@
...
 
-  $: failure = eventOrGroupIsFailureOrTimedOut(event);
-  $: canceled = eventOrGroupIsCanceled(event);
-  $: terminated = eventOrGroupIsTerminated(event);
+  const elapsedTime = $derived(() =>
+    formatDistanceAbbreviated({
+      start: initialItem?.eventTime,
+      end: isEventGroup(event)
+        ? event.initialEvent?.eventTime
+        : currentEvent()?.eventTime,
```
Alex-Tideman: This would then be currentEvent

src/lib/components/event/event-summary-row.svelte
andrewzamojc, line 84: Question: I'm not entirely sure, but the svelte5 docs make it sound like props can be updated inside the component without putting them in state. Is this state necessary? I assume the prop is for setting an "initial" expanded state for the component...

src/lib/components/event/event-summary-row.svelte
GraceGardner: Why did I do this??? Thank you!

src/lib/components/event/event-summary-row.svelte
GraceGardner: Dude don't know what's up with me. Thanks for catching these. I don't know why I did that either.

src/lib/components/event/event-summary-row.svelte
```diff
@@ -263,48 +322,10 @@
...
   >
-    <td class="expanded-cell w-full">
-      <EventDetailsFull {group} event={currentEvent} />
+    <td class="w-full text-sm no-underline" class:border-b-0={typedError}>
+      <EventDetailsFull {group} event={currentEvent()} />
     </td>
   </tr>
 {/if}
-
-<style lang="postcss">
```
andrewzamojc, line 274: Comment: Nice delete!

andrewzamojc (approved):
Looks good to me!

src/lib/components/event/event-summary-row.svelte
GraceGardner: So I completely overcomplicated it... Thanks Andrew. I think my brain was thinking it needed to be reactive because we click on it and then a change happens. So no need for a bind or the state, just the prop. Will change! Thank you!

src/lib/components/event/event-summary-row.svelte
GraceGardner: Sounds good! Thank you!

src/lib/components/event/event-summary-row.svelte
```diff
@@ -37,104 +39,145 @@
...
+
+  const abbrEventTime = $derived(
+    formatDate(currentEvent?.eventTime, $timeFormat, {
+      relative: $relativeTime,
+      abbrFormat: true,
+    }),
+  );
+
+  const onLinkClick = () => {
+    expanded = !expanded;
```
Alex-Tideman, line 237: We'll want to pass the event so we can do event.stopPropagation()

src/lib/components/event/event-summary-row.svelte
GraceGardner: Welp, looks like I did need the state? When I use it just as a prop and try to set it the expand doesn't work

src/lib/components/event/event-summary-row.svelte
GraceGardner: oh geezzz. Sorry! I kept getting errors when I was trying to add it in the html section. This makes so much more sense. I meant to ask about it but spaced. Thanks for catching this!

andrewzamojc: Just noodling / not important: 

I wonder if the need for stopPropagation is because some events higher up are listening for clicks but not checking that `target = currentTarget` or `target = current element` or whatever. Any time I see events eaten lower down the hierarchy I'm always a little curious what's up. Or could it be a svelte thing, or some 3rd party components that don't check properly?

src/lib/components/event/event-summary-row.svelte
```diff
@@ -38,103 +53,209 @@
...
+  };
 
-  $: duration = isEventGroup(event)
-    ? formatDistanceAbbreviated({
-        start: event.initialEvent?.eventTime,
-        end: event.lastEvent?.eventTime,
-        includeMillisecondsForUnderSecond: true,
-      })
-    : '';
+  export const decodeLocalActivity = async (
```
Alex-Tideman: Let's not make this an export if it's in this component. Exports in .svelte files only work in <script module> tags

src/lib/components/event/event-summary-row.svelte
```diff
@@ -38,103 +53,209 @@
...
+      codec: {
+        ...page.data.settings?.codec,
+        endpoint: getCodecEndpoint(page.data.settings),
+        passAccessToken: getCodecPassAccessToken(page.data.settings),
+        includeCredentials: getCodecIncludeCredentials(page.data.settings),
+      },
+    };
+    const accessToken = $authUser.accessToken;
+    const namespace = page.params.namespace;
+    if (!isLocalActivityMarkerEvent(event)) return;
```
Alex-Tideman: We probably don't need this if the onMount does this check already. If we were to put this in another file as an export, I would also move this check to the very top so it would return immediately

src/lib/components/event/event-summary-row.svelte
```diff
@@ -38,103 +53,209 @@
...
-  export let group: EventGroup | undefined = undefined;
-  export let initialItem: IterableEvent | undefined;
-  export let index = 0;
-  export let compact = false;
-  export let expanded = false;
-  export let typedError = false;
-  export let active = false;
-  export let onRowClick: () => void = () => {};
+  interface Props {
+    event: WorkflowEvent;
```
Alex-Tideman: I think this should stay as an IterableEvent. The event could be a pending activity or pending nexus operation, not just a WorkflowEvent. Let me know how I can help if you get stuck on type errors.

src/lib/components/event/event-summary-row.svelte
```diff
@@ -38,103 +53,209 @@
...
+    group?: EventGroup;
```
Alex-Tideman: Let's make this not an optional prop. It's just that it could be undefined

group: EventGroup | undefined;

Alex-Tideman (approved):

PR #2771, opened 2025-06-16
DT-2971-change-codeblock-expand-to-maximize

andrewzamojc:
## Description & motivation 💭 

https://temporalio.atlassian.net/browse/DT-2971

- Replace expand button with maximize button
- add maximize styles as theme extension
- update to svelte5
- refactor reactivity (code mirror state update)

If the refactor is too hard to diff, you can look at individual commits and see adding the Maximizable component separate from the "conver to svelte5 and refactor" commit.

Screenshot 2025-06-16 at 6 41 49 PM

Screenshot 2025-06-16 at 6 41 30 PM

Screenshot 2025-06-16 at 6 41 11 PM

### Design Considerations 🎨 

- i rounded the buttons, I like em but can unround if needed
- The scroll position is maintained after maximizing or minimizing. This just works thanks to the codemirror compartment concept, so the whole thing doesn't redraw.
- focus stays on maximize/minimize icon button so easy to toggle with keyboard.
- escape key minimizes

### How was this tested 👻 

- [x] Manual testing in storybook and cloud-ui
- [ ] E2E tests added
- [ ] Unit tests added

Use the preview link to view the component library and find the Code Block component in the left menu. Go to the story about Max Height, which shows a maximize button because code can be hidden and scrolled. 
- Click the maximize button to toggle.
- Try scrolling when maximized
- Tab into the code block and us up/down to scroll
- Press the button with space bar
- Use esc key to close

In the product
- Click workflows
- Click a workflow and see its code blocks
- Go back to the workflows list
- Click Create Workflow
- Try the input code block, which is editable


src/lib/holocene/code-block.svelte
```diff
@@ -3,242 +3,205 @@
...
-    if (view && (!editable || view.state.doc.toString() !== content)) {
-      resetView(content);
-    }
+  const onMaximize = (value: boolean) => {
+    maximized = value;
   };
 
-  $: content, language, setView();
-
-  $: expandable = !editable && maxHeight && contentHeight(editor) > maxHeight;
```
andrewzamojc, line 209: Annotation: The expandable functionality is replaced with maximized functionality.

src/lib/holocene/code-block.svelte
```diff
@@ -3,242 +3,205 @@
...
-  export { className as class };
-  export let editable = false;
-  export let inline = false;
-  export let language = 'json';
-  export let copyable = true;
-  export let copyIconTitle = '';
-  export let copySuccessIconTitle = '';
-  export let minHeight = undefined;
-  export let maxHeight = undefined;
-  export let label = '';
```
andrewzamojc, line 67: Annotation: Some changes to look for while migrating to svelte5:
- formatJSON extracted to a utility
- extension code is moved to a `custom-extensions.ts` file, stuff like the theme
- extensions are grouped into static or dynamic, so that a "compartment" can update the dynamic ones when props change
- A single create function `createView` constructs a new codemirror view, passing a new EditorState with extensions.

src/lib/holocene/code-block.svelte
```diff
@@ -3,242 +3,202 @@
...
-
-  export let content: string;
-  let className: string = null;
-  export { className as class };
-  export let editable = false;
-  export let inline = false;
-  export let language = 'json';
-  export let copyable = true;
-  export let copyIconTitle = '';
-  export let copySuccessIconTitle = '';
```
andrewzamojc, line 64: Annotation: I feel copyable should be defaulted to false, and when true have the translation strings required. As is, with it default to true, the required strings aren't required, and default to empty strings. Could we make a breaking change to tidy that up, or no?

src/lib/holocene/code-block.svelte
```diff
@@ -3,242 +3,203 @@
...
+  } from '$lib/vendor/codemirror/custom-extensions';
+
+  interface BaseProps extends HTMLAttributes<HTMLDivElement> {
     content: string;
-    language?: 'json' | 'text' | 'shell';
+    language?: EditorLanguage;
     editable?: boolean;
+    copyable?: boolean;
+    copyIconTitle?: string;
+    copySuccessIconTitle?: string;
```
andrewzamojc: Annotation: Although these are new, some usage seems to need it :S Type errors if removed. 

Specifically, there is usage where there is no copyable prop (which defaults to _true_), but translations are provided.

I find the default copyable=true confusing, and harder to type the translations properly. It would be nice to tidy these up and enforce the translations better, but small changes here cause type errors, so any improvement to the types involves fixing a bunch of usage in `ui` and `cloud-ui`.

src/lib/holocene/maximizable/maximizable.svelte
```diff
@@ -0,0 +1,77 @@
...
+  const handleClick = () => {
+    onMaximize(!maximized);
+  };
+</script>
+
+<div
+  class={merge(
+    'relative',
+    maximized
+      ? 'fixed bottom-0 left-0 right-0 top-0 z-10 h-full w-full overflow-y-auto bg-white dark:bg-black'
```
andrewzamojc: Annotation: the maximizing magic

Alex-Tideman: This is super slick, like it a lot. The one thing that we'll need to change is accommodating for the navs, both the top and the left expanded/not expanded

Screenshot 2025-06-16 at 9 18 54 PM

src/lib/holocene/code-block.svelte
```diff
@@ -3,242 +3,205 @@
...
+      parent: editorElement,
+      state: EditorState.create({
+        doc: formatContent(language, content, inline),
+        extensions: [staticExtensions, compartment.of(dynamicExtensions)],
+      }),
+      dispatch(transaction, view) {
         view.update([transaction]);
         if (transaction.docChanged) {
-          dispatch('change', view.state.doc.toString());
+          svelteDispatch('change', view.state.doc.toString());
```
rossedfort: I think as part of a svelte5 upgrade we would just make this a [callback prop](https://svelte.dev/docs/svelte/v5-migration-guide#Event-changes-Component-events), so we'd do `<Codeblock onchange={doSomething} />` instead of `<Codeblock on:change={doSomething} />`.

andrewzamojc: Oh I see! I thought it was doing something more complicated so left it. It would be a breaking change from holocene. How do we feel about those? Is it ok as long as I get a PR ready in `cloud-ui` to update all usage?

src/lib/holocene/maximizable/maximizable.svelte
andrewzamojc: TY, I'll test it out.

src/lib/holocene/code-block.svelte
andrewzamojc: Codeblock now offers and onChange callback that receives a CustomEvent as it did before.

src/lib/holocene/maximizable/maximizable.svelte
andrewzamojc: This is resolved.

src/lib/holocene/code-block.svelte
```diff
@@ -3,242 +3,211 @@
...
+    copySuccessIconTitle?: string;
     inline?: boolean;
     testId?: string;
-    copyable?: boolean;
     minHeight?: number;
     maxHeight?: number;
     label?: string;
     class?: string;
-  };
+    onChange?: (event: CustomEvent<string>) => void;
```
rossedfort: We don't really need to wrap the string parameter in a `CustomEvent`, that was just the old way to do it using `createEventDispatcher`. More here: https://svelte.dev/docs/svelte/v5-migration-guide#Event-changes-Component-events

rossedfort: I might also consider changing the name of this to `onchange` (all lowercase), to match svelte/native html conventions, but it may be something to discuss with the team first.

src/lib/holocene/maximizable/button.svelte
```diff
@@ -0,0 +1,36 @@
...
+  class={merge(
+    'surface-interactive-secondary m-1 border border-[transparent] bg-transparent p-1 focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-primary/70',
+    className,
+  )}
+  onclick={onClick}
+  {...restProps}
+>
+  <Icon
+    title={maximized ? 'minimize' : 'expand'}
+    name={maximized ? 'minimize' : 'expand'}
```
GiantRobots: I have disdain for ternaries can we pull this into a @const or a derived and pass in the title/name in one place

src/lib/holocene/maximizable/maximizable.svelte
```diff
@@ -0,0 +1,89 @@
...
+    }
+  });
+
+  $effect(() => {
+    if (maximized) {
+      document.addEventListener('keydown', escapeListener);
+    } else {
+      document.removeEventListener('keydown', escapeListener);
+    }
+  });
```
GiantRobots: This effect seems weird to me why do we need to add and remove the event listener on maximized (or escapelistener) change

src/lib/holocene/maximizable/maximizable.svelte
```diff
@@ -0,0 +1,89 @@
...
+    class: className = undefined,
+    enabled = true,
+    actions = undefined,
+  }: Props = $props();
+
+  let escapeListener = $state((event: KeyboardEvent) => {
+    if (event.key === 'Escape') {
+      onMaximize(false);
+    }
+  });
```
GiantRobots: I don't think this needs to be state. It can just be a function reference no?

src/lib/holocene/maximizable/maximizable.svelte
```diff
@@ -0,0 +1,89 @@
...
+      document.removeEventListener('keydown', escapeListener);
+    }
+  });
+
+  onMount(() => {
+    // on unmount
+    return () => {
+      document.removeEventListener('keydown', escapeListener);
+    };
+  });
```
GiantRobots: Instead of managing the listener directly I would recommend using `<svelte:body />` it's dope and it will do it for you

https://svelte.dev/docs/svelte/svelte-body

src/lib/holocene/maximizable/maximizable.svelte
```diff
@@ -0,0 +1,89 @@
...
+    }
+  };
+</script>
+
+<div
+  class={merge(
+    'relative',
+    maximized
+      ? 'fixed bottom-0 left-0 right-0 top-0 z-100 h-full w-full overflow-y-auto bg-white dark:bg-black'
+      : '',
```
GiantRobots: Do not use ternaries. Use the cslx stuff for this instead https://svelte.dev/docs/svelte/class#Attributes-Objects-and-arrays

src/lib/holocene/maximizable/maximizable.svelte
```diff
@@ -0,0 +1,89 @@
...
+    className,
+  )}
+  onfocusout={handleFocusOut}
+>
+  {@render children()}
+
+  <div
+    class={merge(
+      maximized ? 'fixed' : 'absolute',
```
GiantRobots: Ternary 🔪

src/lib/holocene/code-block.svelte
```diff
@@ -3,242 +3,253 @@
...
-    }
-    return 0;
-  };
+  onMount(() => {
+    editorView = createEditorView();
+    editorView.contentDOM.onblur = handleEditorBlur;
+    return () => {
+      editorView?.destroy();
+    };
+  });
```
andrewzamojc, line 227: Annotation: When the svelte update lands from @GiantRobots , I'll update this to an `@attach`

src/lib/holocene/code-block.svelte
andrewzamojc: Done.

src/lib/holocene/maximizable/button.svelte
andrewzamojc: Did this.

src/lib/holocene/maximizable/maximizable.svelte
andrewzamojc: using svelte now

src/lib/holocene/maximizable/maximizable.svelte
andrewzamojc: done

src/lib/holocene/maximizable/maximizable.svelte
andrewzamojc: donzo

src/lib/holocene/maximizable/maximizable.svelte
andrewzamojc: did

andrewzamojc:
Fixed up and ready for another look. Feel free to resolve any threads you think are resolved.

Alex-Tideman (approved):

PR #2774, opened 2025-06-17
Upgrade Echo and implement dynamic CORS middleware with Air hot reloading

rossnelson:
## Description & motivation 💭

This PR addresses CORS issues by upgrading Echo and implementing a dynamic CORS middleware system, while also adding Air hot reloading for improved development experience.

The CORS error:
```
/namespaces/abhinavterraformtest2.temporal-dev/workflows:1 Access to fetch at 'https://abhinavterraformtest2.temporal-dev.web.tmprl-test.cloud/api/v1/namespaces/abhinavterraformtest2.temporal-dev/workflow-count?query=GROUP+BY+ExecutionStatus' from origin 'https://staging.thundergun.io/' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: The value of the 'Access-Control-Allow-Origin' header in the response must not be the wildcard '*' when the request's credentials mode is 'include'.
```

**Key Changes:**
- **Security Fix**: Upgraded Echo from v4.9.1 to v4.13.4, removing vulnerable golang-jwt dependency
- **Dynamic CORS**: Implemented custom middleware that evaluates CORS configuration at runtime
- **Development Enhancement**: Added `unsafeAllowAllOrigins` config option for flexible development environments
- **Hot Reloading**: Integrated Air for automatic Go server rebuilds during development
- **Improved DX**: Added verbose mode and comprehensive documentation

Air hot reload and verbose mode:
CleanShot 2025-06-17 at 10 19 33

### Design Considerations 🎨

- Custom CORS middleware is performance-optimized with early returns
- Security-first approach with clear warnings about unsafe configurations
- Backward compatibility maintained for existing CORS configurations

### How was this tested 👻

- [x] Manual testing - Verified CORS behavior in both safe and unsafe modes
- [x] Unit tests added - Comprehensive test suite for CORS middleware
- [x] Air integration tested - Hot reloading works correctly in development

1. **Test CORS Security**: 
   - Set `unsafeAllowAllOrigins: false` and verify only allowed origins work
   - Set `unsafeAllowAllOrigins: true` and verify any origin is accepted

2. **Test Hot Reloading**:
   - Run `pnpm dev` in development mode
   - Modify Go source files and verify automatic rebuilds
   - Use `UI_SERVER_VERBOSE=true pnpm dev` to see detailed logs

3. **Test E2E Mode**:
   - Verify e2e environment still uses traditional build process

### Draft Checklist
- [x] All tests passing
- [x] Documentation updated
- [x] Security considerations addressed

### Merge Checklist
- [x] Code reviewed
- [x] CORS security implications understood
- [x] Air configuration verified

### Any docs updates needed?

- [x] Updated server README with CORS configuration documentation
- [x] Added security warnings for unsafe configurations
- [x] Documented Air hot reloading setup
- [x] Added verbose mode documentation

Depends on: https://github.com/temporalio/saas-components/pull/1930


server/server/cors/cors.go
```diff
@@ -0,0 +1,119 @@
+// The MIT License
+//
+// Copyright (c) 2022 Temporal Technologies Inc.  All rights reserved.
+//
+// Copyright (c) 2020 Uber Technologies, Inc.
```
andrewzamojc, line 5: Question: I'm curious about this code. Did we write it or is it copied from somewhere? If the latter, does a package exist instead?

utilities/ui-server.ts
```diff
@@ -23,13 +23,39 @@ const portForEnv = (env: Environemt) => {
...
-  console.log(`✨ ui-server running in ${env} mode on port ${portForEnv(env)}`);
+  let uiServerProcess: ReturnType<typeof $>;
+
+  if (env === 'development') {
+    // Install Air if not already available
+    try {
+      await $`which air`.quiet();
+    } catch {
+      console.log('📦 Installing Air for hot reloading...');
+      await $`go install github.com/air-verse/air@latest`;
```
andrewzamojc, line 41: Question: I may be missing something in go land. Could this be a dev dependency instead of a scripted install?

andrewzamojc (commented):
Looks pretty good. I left a couple questions.

server/server/cors/cors.go
rossnelson: We wrote it. its kind of a non standard feature to allow a dynamic origin. this is actually bypassing the CORS lib to handle the unsafe dynamic domain logic. 

Also, regarding the lib question, ill point to the [go proverbs](https://go-proverbs.github.io/) -> [A little copying is better than a little dependency.](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=9m28s)

utilities/ui-server.ts
rossnelson: go does not have the concept of dev dependencies. I'm not familiar with the way go land might handle dev deps. IME, since it runs at the system level and isn't an app dependency it typically gets installed as such.

Alex-Tideman (approved):

server/server/cors/cors.go
andrewzamojc: Neat. Sounds good to me!

utilities/ui-server.ts
andrewzamojc: Ack.

andrewzamojc (approved):
LGTM

PR #2778, opened 2025-06-20
Deployment Updates - Pinned Workers and Unversioned current

Alex-Tideman:
…only pinned workers if workflow is pinned

## Description & motivation 💭 
If a workflow is pinned, show only the workers on that pinned Build ID on the Workers tab.

Update __unversioned__ to No Current Version, hide deployed at timestamp for unversioned since we don't know it.

Screenshot 2025-06-20 at 12 03 40 PM
Screenshot 2025-06-20 at 12 21 50 PM

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/worker-table.svelte
+++ b/src/lib/components/worker-table.svelte
@@ -10,7 +10,9 @@
   import { translate } from '$lib/i18n/translate';
   import { type GetPollersResponse } from '$lib/services/pollers-service';
   import { relativeTime, timeFormat } from '$lib/stores/time-format';
+  import { workflowRun } from '$lib/stores/workflow-run';
   import { formatDate } from '$lib/utilities/format-date';
+  import { getBuildIdFromVersion } from '$lib/utilities/get-deployment-build-id';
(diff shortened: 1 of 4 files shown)


src/lib/components/worker-table.svelte
```diff
@@ -21,6 +23,28 @@
...
+  const { workflow } = $derived($workflowRun);
+
+  const deploymentVersion = $derived(
+    workflow?.searchAttributes?.indexedFields?.[
+      'TemporalWorkerDeploymentVersion'
+    ],
+  );
+
+  const versioningBuildId = $derived(
+    workflow?.searchAttributes?.indexedFields?.['TemporalWorkerBuildId'] ||
```
andrewzamojc: Comment: If the `||` is a null fallback, could do `??`. In case buildId is falsy...unless you're trying to also capture empty string.

andrewzamojc (approved):
Looks ok to me.

PR #2782, opened 2025-06-24
hide SDK workflow detail unless there is a value

andrewzamojc:
## Description & motivation 💭 

Update workflow details to hide the SDK label if there is no value. This involved pulling some logic out of SdkLogo and into workflow details. 

The bug (notice SDK label with no value)

Screenshot 2025-06-24 at 4 00 39 PM

The fix without SDK

Screenshot 2025-06-24 at 3 14 31 PM

The fix with SDK

Screenshot 2025-06-24 at 3 14 20 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/lines-and-dots/workflow-details.svelte
+++ b/src/lib/components/lines-and-dots/workflow-details.svelte
@@ -4,11 +4,14 @@
   import { translate } from '$lib/i18n/translate';
   import { fetchWorkflow } from '$lib/services/workflow-service';
   import { isCloud } from '$lib/stores/advanced-visibility';
+  import { fullEventHistory } from '$lib/stores/events';
   import { relativeTime, timeFormat } from '$lib/stores/time-format';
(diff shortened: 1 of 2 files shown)


src/lib/components/lines-and-dots/workflow-details.svelte
```diff
@@ -36,34 +39,42 @@
 
   let latestRunId = $state<string | undefined>(undefined);
 
-  const { namespace } = $derived(page.params);
-  const elapsedTime = $derived(
+  let { namespace } = $derived(page.params);
+  let elapsedTime = $derived(
```
Alex-Tideman: I think these can all stay consts. We aren't mutating the values

andrewzamojc: I put them back to const for $derived 👍

Alex-Tideman (approved):

PR #2790, opened 2025-06-26
Show the related activity events on hover/click in timeline

KATIETOLER:
This is the first step in adding highlighting related activities in the event timeline/summary row. 

when hovering on an activity event row, a user can see a visual indicator on the related event rows 
Screenshot 2025-06-27 at 12 27 46 PM
Screenshot 2025-06-27 at 12 27 59 PM

---
## Next steps:
- [ ] do the same thing for workflows
- [ ] slight dim on the other event content

### DIFF
--- a/src/lib/components/event/event-summary-row.svelte
+++ b/src/lib/components/event/event-summary-row.svelte
@@ -427,10 +427,10 @@
 
 <style lang="postcss">
   tr[data-testid='event-summary-row'].active {
-    @apply surface-table-header;
+    @apply surface-table-related-hover;
   }
 
   tr[data-testid='event-summary-row'].active:hover {
-    @apply surface-interactive;
+    @apply surface-table-header;
   }
 </style>
--- a/src/lib/components/event/event-summary-row.svelte
+++ b/src/lib/components/event/event-summary-row.svelte
@@ -424,3 +424,13 @@
     </td>
   </tr>
 {/if}
+
+<style lang="postcss">
+  tr[data-testid='event-summary-row'].active {
+    @apply bg-slate-800;
+  }
+
+  tr[data-testid='event-summary-row'].active:hover {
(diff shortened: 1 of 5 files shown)


src/lib/components/event/event-summary-row.svelte
```diff
@@ -409,3 +424,13 @@
...
 {/if}
+
+<style lang="postcss">
+  tr[data-testid='event-summary-row'].active {
+    @apply bg-slate-800;
+  }
+
+  tr[data-testid='event-summary-row'].active:hover {
+    @apply border-l-4 border-slate-600 bg-slate-300;
+  }
```
andrewzamojc, line 435: Suggestion: Any reason not to do this with tailwind? Instead of conditionally applying the `.active` class and styling it here, you could probably modify you're first class entry

```
class={[
  // existing
  'hover:cursor-pointer',
  // new
  hasRelatedActivities(group, hoveredEventId) && 'bg-slate-800 hover:bg-slate-300 etc'
]}
```

or something like that.

src/lib/components/event/event-summary-table.svelte
```diff
@@ -127,6 +129,7 @@
         />
       {:else}
         <EventSummaryRow
+          bind:hoveredEventId
```
andrewzamojc, line 132: Comment: Pretty slick solution.

andrewzamojc (commented):
Code looks pretty good to me! I'll try it shortly.

Alex-Tideman:
We'll want a different color too for light mode

Screenshot 2025-06-26 at 2 49 10 PM

andrewzamojc (approved):
Works great! LGTM pending the light mode comment from Alex.

Alex-Tideman (approved):

PR #2796, opened 2025-06-30
User metadata

GraceGardner:
### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/workflow/metadata/workflow-current-details.svelte
+++ b/src/lib/components/workflow/metadata/workflow-current-details.svelte
@@ -15,6 +15,11 @@
   const { namespace } = $derived(page.params);
   const { workflow } = $derived($workflowRun);
   const currentDetails = $derived($workflowRun?.metadata?.currentDetails || '');
+  const summary = $derived($workflowRun?.userMetadata?.summary);
+  const details = $derived($workflowRun?.userMetadata?.details);
+  const borderClass = $derived(() =>
+    summary || details ? 'border-l border-subtle' : '',
+  );
 
   let loading = $state(false);
 
@@ -35,6 +40,7 @@
         $authUser?.accessToken,
       );
       $workflowRun.metadata = metadata;
+      console.log('Current details fetched:', metadata.currentDetails);
     } catch (error) {
       console.error('Error fetching current details:', error);
     } finally {
@@ -45,11 +51,15 @@
   onMount(() => {
     fetchCurrentDetails();
   });
+
+  $inspect(currentDetails);
 </script>
 
-<div class="flex flex-1 flex-col gap-2">
+<div
+  class={`surface-background flex h-full flex-1 flex-col gap-2 border-l border-subtle p-6 ${borderClass}`}
+>
   <div class="flex items-center justify-between">
-    <h3>{translate('workflows.current-details')}</h3>
+    <h3 class="pl-6 pt-6">{translate('workflows.current-details')}</h3>
     {#if workflow.isRunning}
       <Tooltip text={translate('workflows.update-details')} left>
         <Button
@@ -63,6 +73,10 @@
     {/if}
   </div>
   {#key currentDetails}
-    <Markdown content={currentDetails} />
+    <Markdown
+      className="p-2"
+      overrideTheme="background"
+      content={currentDetails}
+    />
   {/key}
 </div>
--- a/src/lib/components/workflow/metadata/workflow-current-details.svelte
+++ b/src/lib/components/workflow/metadata/workflow-current-details.svelte
@@ -1,7 +1,8 @@
 <script lang="ts">
-  import { page } from '$app/stores';
+  import { onMount } from 'svelte';
 
-  import AccordionLight from '$lib/holocene/accordion/accordion-light.svelte';
+  import { page } from '$app/state';
+
   import Button from '$lib/holocene/button.svelte';
   import Icon from '$lib/holocene/icon/icon.svelte';
   import Markdown from '$lib/holocene/monaco/markdown.svelte';
@@ -11,18 +12,17 @@
   import { authUser } from '$lib/stores/auth-user';
   import { workflowRun } from '$lib/stores/workflow-run';
 
-  $: ({ namespace } = $page.params);
-  $: ({ workflow } = $workflowRun);
-  $: currentDetails = $workflowRun?.metadata?.currentDetails || '';
-  $: closedWithoutDetails = !workflow.isRunning && !currentDetails;
+  const { namespace } = $derived(page.params);
+  const { workflow } = $derived($workflowRun);
+  const currentDetails = $derived($workflowRun?.metadata?.currentDetails || '');
 
-  let loading = false;
+  let loading = $state(false);
 
   const fetchCurrentDetails = async () => {
     if (loading) return;
     loading = true;
     try {
-      const { settings } = $page.data;
+      const { settings } = page.data;
       const metadata = await getWorkflowMetadata(
         {
           namespace,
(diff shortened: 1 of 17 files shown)


Alex-Tideman:
Screenshot 2025-07-01 at 1 35 16 PM

I think we should always split this, and show the empty state for Summary / Details

Alex-Tideman:
Screenshot 2025-07-01 at 1 37 09 PM

Purely a design thing... but this feels a little bland. I'm not totally sure what it needs but doesn't feel polished.

src/routes/(app)/namespaces/[namespace]/workflows/[workflow]/[run]/memo/+page.svelte
```diff
@@ -0,0 +1,35 @@
...
+  import { workflowRun } from '$lib/stores/workflow-run';
+
+  const { workflow } = $derived($workflowRun);
+</script>
+
+<PageTitle
+  title={`${translate('workflows.memo-tab')} | ${page.params.workflow}`}
+  url={page.url.href}
+/>
+<WorkflowPaddedLayout>
```
Comment (Alex-Tideman) [inline src/routes/(app)/namespaces/[namespace]/workflows/[workflow]/[run]/memo/+page.svelte:14]: Let's bundle up the code between WorkflowPaddedLayout into a page component.

```
<WorkflowPaddedLayout>
 <WorkflowMemo />
</WorkflowPaddedLayout>
```

That way we don't have to duplicate this logic in Cloud and can just import in WorkflowMemo (see the other pages like query/+page.svelte).

src/routes/(app)/namespaces/[namespace]/workflows/[workflow]/[run]/search-attributes/+page.svelte
```diff
@@ -0,0 +1,40 @@
...
+
+  const { workflow } = $derived($workflowRun);
+</script>
+
+<PageTitle
+  title={`${translate('workflows.search-attributes-tab')} | ${page.params.workflow}`}
+  url={page.url.href}
+/>
+<WorkflowPaddedLayout>
+  <div class="mt-4 flex flex-col gap-2">
```
Comment (Alex-Tideman) [inline src/routes/(app)/namespaces/[namespace]/workflows/[workflow]/[run]/search-attributes/+page.svelte]: Same here: 

```
<WorkflowPaddedLayout>
 <WorkflowSearchAttributes />
</WorkflowPaddedLayout>
```

src/routes/(app)/namespaces/[namespace]/workflows/[workflow]/[run]/user-metadata/+page.svelte
```diff
@@ -0,0 +1,49 @@
...
+  const eventGroups = $derived(
+    groupEvents(
+      relevantEvents,
+      'ascending',
+      workflow?.pendingActivities,
+      workflow?.pendingNexusOperations,
+    ),
+  );
+</script>
+
```
Comment (Alex-Tideman) [inline src/routes/(app)/namespaces/[namespace]/workflows/[workflow]/[run]/user-metadata/+page.svelte:8]: Same here. Let's put all this logic in <WorkflowUserMetadata /> page and add PageTitle here


```
<PageTitle
  title={`${translate('workflows.user-metadata-tab')} | ${page.params.workflow}`}
  url={page.url.href}
/>
<WorkflowPaddedLayout>
 <WorkflowUserMetadata />
</WorkflowPaddedLayout>

```

src/routes/(app)/namespaces/[namespace]/workflows/[workflow]/[run]/user-metadata/+page.svelte
```diff
@@ -0,0 +1,49 @@
...
+  import WorkflowSummaryAndDetails from '$lib/components/workflow/metadata/workflow-summary-and-details.svelte';
+  import { translate } from '$lib/i18n/translate';
+  import { groupEvents } from '$lib/models/event-groups';
+  import { fullEventHistory } from '$lib/stores/events';
+  import { workflowRun } from '$lib/stores/workflow-run';
+
+  const { workflow } = $derived($workflowRun);
+
+  const relevantEvents = $derived(
+    $fullEventHistory.filter((event) => {
```
Comment (Alex-Tideman) [inline src/routes/(app)/namespaces/[namespace]/workflows/[workflow]/[run]/user-metadata/+page.svelte]: Let's not filter the events quite yet. Let's create the event groups first, then filter the groups by group.userMetadata?.summary. That way we have the full status of the group. How it works now is that it filters it down to only Scheduled/Started events and then creates groups, which won't include the Completed event with the final status

src/lib/components/workflow/metadata/metadata-events.svelte
```diff
@@ -0,0 +1,84 @@
...
+
+    {#each groups as group}
+      <TableRow class="hover:bg-interactive-table-hover">
+        <td class="px-3 py-3 text-left">
+          <div class="flex flex-col gap-1 capitalize">
+            {group.category}
+          </div>
+        </td>
+        <td class="px-3 py-3 text-left">
+          <Badge type={getBadgeType(group.finalClassification)} class="text-xs">
```
Alex-Tideman: Should be able to use the <WorkflowStatus /> component here.

src/lib/components/workflow/metadata/workflow-current-details.svelte
```diff
@@ -1,28 +1,28 @@
...
+  const { workflow } = $derived($workflowRun);
+
+  const currentDetails = $derived($workflowRun?.metadata?.currentDetails || '');
 
-  let loading = false;
+  let loading = $state(false);
 
   const fetchCurrentDetails = async () => {
     if (loading) return;
     loading = true;
```
Alex-Tideman, line 26: Let's add a timestamp when they fetch the current details and show it to the right of the "Press the R button...". That way they have some feedback if it actually was fetched

src/lib/holocene/monaco/markdown.svelte
```diff
@@ -40,13 +48,13 @@
   const templatedContent = replaceTemplate(content);
 </script>
 
-<section class="h-full w-full" in:fade={{ duration: 1000 }}>
+<section class={`h-full w-full ${className}`} in:fade={{ duration: 1000 }}>
```
rossedfort: nit: using template literal interpolation here isn't necessary

```suggestion
<section class="h-full w-full {className}" in:fade={{ duration: 1000 }}>
```
In fact, we should probably use `twMerge` for stuff like this instead.

src/lib/holocene/monaco/markdown.svelte
```diff
@@ -40,13 +48,13 @@
...
   {#key theme}
     <iframe
       bind:this={iframe}
       on:load={resizeIframe}
       title="output"
-      src="/render?content={encodeURIComponent(templatedContent)}&theme={theme}"
+      src={`/render?content=${encodeURIComponent(templatedContent)}&theme=${theme}&overrideTheme=${overrideTheme}`}
```
rossedfort: Same here, can be

```suggestion
      src="/render?content={encodeURIComponent(templatedContent)}&theme={theme}&overrideTheme={overrideTheme}"
```

src/routes/(app)/namespaces/[namespace]/workflows/[workflow]/[run]/user-metadata/+page.svelte
Comment (GraceGardner) [reply src/routes/(app)/namespaces/[namespace]/workflows/[workflow]/[run]/user-metadata/+page.svelte:8]: I made the WorkflowUserMetadata into a component, but opted to not use the WorkflowPaddedLayout because the formatting is a little bit different.

src/lib/components/workflow/metadata/metadata-events.svelte
```diff
@@ -0,0 +1,45 @@
...
+  const metadataGroups = $derived(
+    groupEvents(
+      $fullEventHistory,
+      'ascending',
+      workflow?.pendingActivities,
+      workflow?.pendingNexusOperations,
+    ).filter((group) => group.userMetadata?.summary),
+  );
+
+  $inspect('Metadata groups:', metadataGroups);
```
Alex-Tideman: 🗑️

GraceGardner: Dang it... I searched for console logs at the end and forgot about inspect. Thank you!

src/lib/components/workflow/metadata/workflow-current-details.svelte
```diff
@@ -35,40 +36,63 @@
...
+            variant="ghost"
+            on:click={fetchCurrentDetails}
+            disabled={loading}
+          >
+            <Icon name="retry" />
+          </Button>
+        </div>
+        {#if lastFetched}
+          <p class="text-xs text-secondary">
+            {lastFetched.toLocaleTimeString()}
```
Alex-Tideman: let's use formatDate() so it uses the timeFormat like all our other timestamps

PR #2800, opened 2025-07-01
Check for localActivity if compact mode

Alex-Tideman:
## Description & motivation 💭 
Decode Local Activity activity type in Compact mode.

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/event/event-summary-row.svelte
+++ b/src/lib/components/event/event-summary-row.svelte
@@ -250,6 +250,11 @@
   onMount(async () => {
     if (isLocalActivityMarkerEvent(event)) {
       primaryLocalAttribute = await decodeLocalActivity(event);
+    } else if (
+      isEventGroup(event) &&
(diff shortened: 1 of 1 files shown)


src/lib/components/event/event-summary-row.svelte
```diff
@@ -250,6 +250,11 @@
   onMount(async () => {
     if (isLocalActivityMarkerEvent(event)) {
       primaryLocalAttribute = await decodeLocalActivity(event);
+    } else if (
+      isEventGroup(event) &&
+      isLocalActivityMarkerEvent(event.initialEvent)
+    ) {
```
GraceGardner, line 256: DUDE! I should have thought of this! Thanks for the update. This is a nice fallback. Nice catch FR!

GraceGardner (approved):

PR #2801, opened 2025-07-01
Fix bug: image doesn't switch modes on table empty state

KATIETOLER:
DT-2917
Bug in dark mode: When no filter results in Workflow list table , the image is in light mode

UI:
Screenshot 2025-07-10 at 1 15 46 PM
Screenshot 2025-07-10 at 1 15 55 PM

Cloud:
Screenshot 2025-07-15 at 4 06 25 PM
Screenshot 2025-07-15 at 4 06 33 PM


src/lib/components/workflow/workflows-summary-configurable-table/table-empty-state.svelte
```diff
@@ -75,10 +74,6 @@
     {/if}
   </div>
   <div class="bg-[#DDD6FE]">
-    <img
-      src={query ? noResultsImages : emptyImage}
-      alt=""
-      class="aspect-auto"
-    />
+    <img src={query || emptyImage} alt="" class="aspect-auto" />
```
rossedfort: How will this behave if there is a query? If we only want to show the `emptyImage` when there is no query, we should wrap the whole `` in `{#if !query}`.

Alex-Tideman: Yeah we don't want to use the query as the src of the img, it will try to find it

Screenshot 2025-07-02 at 9 07 17 AM

src/lib/components/workflow/workflows-summary-configurable-table/table-empty-state.svelte
```diff
@@ -75,10 +74,8 @@
     {/if}
   </div>
   <div class="bg-[#DDD6FE]">
-    <img
-      src={query ? noResultsImages : emptyImage}
-      alt=""
-      class="aspect-auto"
-    />
+    {#if !query}
+      <img src={emptyImage} alt="" class="aspect-auto" />
```
laurakwhit: > Bug in dark mode: When no results in Workflow list table, the image is in light mode

I might need some more context here, but the original ask ⬆️ seems to refer to fixing the image in dark mode. To do that we don't need to update the query logic here but we do need to swap `empty-state-dark_2x.png` with `empty-state-light_2x.png` based on the mode. 

If we do want to remove the image, then based on the designs linked in that task I think we can remove this entirely and do something like this instead ⬇️ 

Screenshot 2025-07-08 at 2 56 53 PM

Let's sync with @tigernaut and align on the UI we want here.

src/lib/components/workflow/workflows-summary-configurable-table/table-empty-state.svelte
```diff
@@ -6,7 +6,6 @@
   import { translate } from '$lib/i18n/translate';
   import { workflowError } from '$lib/stores/workflows';
   import emptyImage from '$lib/vendor/empty-state-dark_2x.png';
-  import noResultsImages from '$lib/vendor/empty-state-light_2x.png';
```
laurakwhit, line 9: If we're no longer using this image anywhere we might as well remove it from `src/lib/vendor` entirely 🗑️

src/lib/components/workflow/workflows-summary-configurable-table/table-empty-state.svelte
tigernaut: this is looking great. we can remove the box around the "no workflow..." text. we will want to have an image that will swap light/dark mode. we don't have those visuals yet, we'll be working with brand to get those.

laurakwhit: @tigernaut we currently have [a light](https://github.com/temporalio/ui/blob/main/src/lib/vendor/empty-state-light_4x.png) and [a dark](https://github.com/temporalio/ui/blob/main/src/lib/vendor/empty-state-dark_2x.png) mode image. Do we want to remove the image here completely? Or just swap them based on light vs. dark mode?

KATIETOLER: This PR is for the `results` when a user uses the filter and there are no results. From what I can tell this is the UI when there are no workflows. The 'no workflows' seems to also need a fix though for light mode as well so I can add that to this or maybe open a separate PR for this issue as well. 

I thought @tigernaut we said we should just remove that image for now? Should I remove the 'no workflows' image as well? I can do whatever though! Sorry if I'm being dense!

src/lib/components/workflow/workflows-summary-configurable-table/table-empty-state.svelte
KATIETOLER: I will remove this if we decide to remove it from the `no workflows` page as well! :)

src/lib/components/workflow/workflows-summary-configurable-table/table-empty-state.svelte
tigernaut: @laurakwhit @KATIETOLER no longer want to remove the image here. 
we did initially want to remove the image. but we've had some more recent convos about bringing more brand into the product so the desire (currently) is to have the image change to its respective mode. what the new visual will be is TBD as it depends on brand and we don't have an updated asset as of yet. here's the link in  [figma](https://www.figma.com/design/TYX2RBVoj8Bv8wkzL1eBeM/Patterns?node-id=91-22543&t=LLrrDtggjbsLuMVQ-11)

less containers on the empty state would help clean it up visually. it would be great if we could achieve something like this for the empty state:
Screenshot 2025-07-10 at 12 20 51 PM
vs
Screenshot 2025-07-10 at 12 14 45 PM

KATIETOLER: @tigernaut  @laurakwhit 
 I added the original images back and made them switch with light/darkmode. It should be pretty easy to switch these images out when we get the new branding images!  Here's some SC's below if either of ya'll have any feedback/thoughts.

KATIETOLER: <p float="left">
Screenshot 2025-07-10 at 1 15 46 PM
Screenshot 2025-07-10 at 1 15 55 PM
Screenshot 2025-07-10 at 1 16 07 PM
Screenshot 2025-07-10 at 1 16 14 PM
</p>

KATIETOLER: Sorry for 1 million sc's heres a few from cloud - looks the same I think. 
<p float="left">
Screenshot 2025-07-10 at 1 43 07 PM
Screenshot 2025-07-10 at 1 43 14 PM
</p>

src/lib/components/workflow/workflows-summary-configurable-table/table-empty-state.svelte
```diff
@@ -5,11 +5,15 @@
   import Link from '$lib/holocene/link.svelte';
   import { translate } from '$lib/i18n/translate';
   import { workflowError } from '$lib/stores/workflows';
-  import emptyImage from '$lib/vendor/empty-state-dark_2x.png';
-  import noResultsImages from '$lib/vendor/empty-state-light_2x.png';
+  import { useDarkMode } from '$lib/utilities/dark-mode';
+  import emptyImageDark from '$lib/vendor/empty-state-dark_2x.png';
+  import emptyImageLight from '$lib/vendor/empty-state-light_2x.png';
```
laurakwhit, line 10: We should probably preload both images in the `<svelte:head>`

src/lib/components/workflow/workflows-summary-configurable-table/table-empty-state.svelte
```diff
@@ -74,11 +78,13 @@
       </ul>
     {/if}
   </div>
-  <div class="bg-[#DDD6FE]">
+  <div
+    class="flex h-64 items-center justify-center overflow-hidden bg-[#DDD6FE] xl:h-auto xl:flex-1"
+  >
     <img
       src={query ? noResultsImages : emptyImage}
```
laurakwhit: We're using the same image regardless of query, so we can simplify this 
```suggestion
      src={emptyImage}
```

src/lib/components/workflow/workflows-summary-configurable-table/table-empty-state.svelte
```diff
@@ -74,11 +78,13 @@
       </ul>
     {/if}
   </div>
-  <div class="bg-[#DDD6FE]">
+  <div
+    class="flex h-64 items-center justify-center overflow-hidden bg-[#DDD6FE] xl:h-auto xl:flex-1"
```
laurakwhit: I think the flex broke some things here ⬇️  

| Before | After |
|-|-|
|Screenshot 2025-07-10 at 2 12 13 PM|Screenshot 2025-07-10 at 2 13 29 PM|

I'd leave the rest of the styling as is for now and just fix the light vs. dark image.

src/lib/components/workflow/workflows-summary-configurable-table/table-empty-state.svelte
laurakwhit: @KATIETOLER looks like that light mode in Cloud still shows the dark image. Might need to verify you're actually viewing Cloud with your changes applied.

KATIETOLER: Yep. I screenshotted thundergun. Sorry! 
Screenshot 2025-07-15 at 4 06 25 PM
Screenshot 2025-07-15 at 4 06 33 PM

src/lib/components/workflow/workflows-summary-configurable-table/table-empty-state.svelte
```diff
@@ -75,10 +72,17 @@
     {/if}
   </div>
   <div class="bg-[#DDD6FE]">
+    {#if !query}
+      <img
+        src={noResultsImages}
+        alt=""
+        class="max-h-full max-w-full object-contain"
+      />
+    {/if}
```
laurakwhit: Definitely don't want to render two images in the case of no query

```suggestion
```

src/lib/components/workflow/workflows-summary-configurable-table/table-empty-state.svelte
```diff
@@ -20,15 +23,9 @@
   ];
 </script>
 
-<svelte:head>
-  <link rel="preload" as="image" href={emptyImage} />
-</svelte:head>
```
laurakwhit, line 25: Don't think we want to remove the preload here

PR #2805, opened 2025-07-07
Check certFile modification time

ndtretyak:
## Description & motivation 💭 
Use the modification time of `certFile` instead of `keyFile`, since `keyFile` may remain unchanged when the certificate is renewed using the same key.

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [x] Unit tests added


rossnelson:
Thanks for this fix @ndtretyak — the analysis is spot-on. Since this has been idle for a bit, I've opened #3178 with the same changes so we can get it merged. Credit to you in the PR description.

PR #2814, opened 2025-07-15
New link extraction

Alex-Tideman:
## Description & motivation 💭 
Add check for completionCallback and attachedCompletionCallback links. 

There's really not a better way to do this until the TS SDK supports these new fields and we can use the types it generates. Right now only Go SDK supports this. 

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


andrewzamojc (approved):
LGTM

PR #2817, opened 2025-07-17
Fix dark mode image swap on table empty state

KATIETOLER:
DT-2917
Bug in dark mode: When no filter results in Workflow list table , the image is in light mode

Cloud
Screenshot 2025-07-15 at 4 06 25 PM
Screenshot 2025-07-15 at 4 06 33 PM

UI
Screenshot 2025-07-15 at 4 10 02 PM
Screenshot 2025-07-15 at 4 10 10 PM

### DIFF
--- a/src/lib/components/workflow/workflows-summary-configurable-table/table-empty-state.svelte
+++ b/src/lib/components/workflow/workflows-summary-configurable-table/table-empty-state.svelte
@@ -24,7 +24,8 @@
 </script>
 
 <svelte:head>
-  <link rel="preload" as="image" href={noResultsImages} />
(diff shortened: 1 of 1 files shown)


src/lib/components/workflow/workflows-summary-configurable-table/table-empty-state.svelte
```diff
@@ -21,14 +24,11 @@
 </script>
 
 <svelte:head>
-  <link rel="preload" as="image" href={emptyImage} />
+  <link rel="preload" as="image" href={noResultsImages} />
```
Alex-Tideman: I think we'll need two preload link tags for each emptyImageDark and emptyImageLight

PR #2822, opened 2025-07-21
Decode Failures if custom FailureConverter configured

Alex-Tideman:
## Description & motivation 💭 
If a user sets up a custom [FailureConverter](https://typescript.temporal.io/api/interfaces/common.FailureConverter), the encoded last failures were not decoded. This PR adds a check for encodedAttributes and if it exists, decode the last failure. It also adds a #key on the attempt number so the failure is re-decoded on a new failure.

Because encodedAttributes can be anything a user sets, I'm typing activity.lastFailure as PotentiallyDecodable and letting PayloadDecoder attempt the decode as best as possible.

Screenshot 2025-07-21 at 12 25 44 PM

## Testing 🧪 
To test:

Add a custom FailureConverter to the samples-typescript/encryption project

data-converter.ts
```
async function createDataConverter(): Promise<DataConverter> {
  return {
    payloadCodecs: [await EncryptionCodec.create('test-key-id')],
    failureConverterPath: require.resolve('./failure-converter.ts'),
  };
}
```

failure-converter.ts
```
import { DefaultFailureConverter } from '@temporalio/common';

export const failureConverter = new DefaultFailureConverter({
  encodeCommonAttributes: true,
});
```

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/workflow/pending-activity/pending-activity-card.svelte
+++ b/src/lib/components/workflow/pending-activity/pending-activity-card.svelte
@@ -16,6 +16,7 @@
   import { workflowRun } from '$lib/stores/workflow-run';
   import type { PendingActivity } from '$lib/types/events';
   import { activityCommandsEnabled } from '$lib/utilities/activity-commands-enabled';
+  import { type PotentiallyDecodable } from '$lib/utilities/decode-payload';
   import { formatDate } from '$lib/utilities/format-date';
   import {
     formatAttemptsLeft,
@@ -42,6 +43,8 @@
       page.params.namespace,
     ) && isRunning,
   );
+
+  $inspect('activity.lastFailure: ', activity.lastFailure);
 </script>
 
 <div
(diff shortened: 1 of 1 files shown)


src/lib/components/workflow/pending-activity/pending-activity-card.svelte
```diff
@@ -188,14 +189,32 @@
...
-        <CodeBlock
-          content={stringifyWithBigInt(
-            omit(activity.lastFailure, 'stackTrace'),
-          )}
-          maxHeight={384}
-          copyIconTitle={translate('common.copy-icon-title')}
-          copySuccessIconTitle={translate('common.copy-success-icon-title')}
-        />
+        {#if activity.lastFailure?.encodedAttributes}
+          {#key activity.attempt}
```
laurakwhit, line 193: Shouldn't need this with [this change](https://github.com/temporalio/ui/pull/2825/files#diff-f5df298e3cf10bab6f20ce354a1c5e425bf2f7e174a88ca006a11fd0e21d8f83R33), but can update in my PR.

laurakwhit (approved):

src/lib/components/workflow/pending-activity/pending-activity-card.svelte
Alex-Tideman: 👍

PR #2825, opened 2025-07-22
Fix initial load of JSON view not rendering event

laurakwhit:
## Description & motivation 💭 
Updates `PayloadDecoder` to use Svelte 5 syntax and updates the `decodedPayload` when the value changes.

### How was this tested 👻 

- [X] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

* In the JSON event history view when either refreshing the page or linking to it 
    - [ ] Verify the initial event is shown and does not appear as `{}`

### Issue(s) closed 
`DT-3138`

### DIFF
--- a/src/lib/components/event/payload-decoder.svelte
+++ b/src/lib/components/event/payload-decoder.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import type { Snippet } from 'svelte';
+  import { onMount, type Snippet } from 'svelte';
 
   import { page } from '$app/stores';
 
@@ -30,7 +30,7 @@
   let keyedValue = key && value?.[key] ? value[key] : value;
   let decodedValue = $state(stringifyWithBigInt(keyedValue));
 
-  $effect(() => {
+  onMount(() => {
     decodePayloads(value);
   });
 
--- a/src/lib/components/event/payload-decoder.svelte
+++ b/src/lib/components/event/payload-decoder.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import { onMount } from 'svelte';
+  import type { Snippet } from 'svelte';
 
   import { page } from '$app/stores';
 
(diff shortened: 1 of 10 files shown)


src/lib/components/event/payload-decoder.svelte
```diff
@@ -18,18 +18,19 @@
...
+
+  let { children, value, key = '', onDecode }: Props = $props();
 
   let keyedValue = key && value?.[key] ? value[key] : value;
-  let decodedValue = stringifyWithBigInt(keyedValue);
+  let decodedValue = $state(stringifyWithBigInt(keyedValue));
 
-  onMount(() => {
+  $effect(() => {
     decodePayloads(value);
```
Alex-Tideman, line 34: I do have some concerns moving to $effect() here. In the past with Svelte4 this was an issue because the value changed due to it being JSON.stringified()

https://github.com/temporalio/ui/pull/2750

Alex-Tideman: We were re-rendering values when new events came in even when the payload hasn't changed

laurakwhit: Hmm will look into this, but reached for `$effect` because the value is being updated after `onMount` in the `workflow-json-navigator`. Maybe there's something to address in that component instead 🤔

Alex-Tideman: Yeah let's leave it onMount in palyoad-decoder and see if the fix can be in json-navigator


https://github.com/user-attachments/assets/61af8cb9-3acc-47c2-a4a2-21a17b8e622a

laurakwhit: Reverted to use `onMount` and instead added a check for `events.length > 0` to the `<WorkflowJsonNavigator/>` [here](https://github.com/temporalio/ui/pull/2825/commits/ede6cd961b3a95bd8569408c4e47f4df322c2e57) ✅

PR #2828, opened 2025-07-25
Add pagination to Archived Workflows

Alex-Tideman:
## Description & motivation 💭 
Previously we fetched 10,000 archived workflows at once and then did not allow pagination. This changes the behavior to paginate the archived workflows api with proper page size and let's the user page through all archived workflows.

The querying UI - which wasn't correctly working - has been replaced with the manual query input so AWS/Google/Custom archival providers can still be queried with their respective query language. If default archival (file system) is used, the user can use the row filtering as well.

Screenshot 2025-07-25 at 9 28 25 AM
Screenshot 2025-07-25 at 9 28 33 AM
Screenshot 2025-07-25 at 9 35 38 AM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


rossnelson (approved):
its working

PR #2758, opened 2025-06-03
Billable Actions and Event History Table / Timeline updates

Alex-Tideman:
## Description & motivation 💭 

For Cloud only, labels events in the Event History with Billable Action count. Includes total Billable Action count in the workflow header.

Other changes:

* Makes the Event History table an actual table
* Moves the Lines/Dots out of the Event History table and only shows them when in “Expanded” mode
* Makes Event Details (expanded event row or expanded in Timeline) to be consistently styled cards
* Hides certain duplicate fields on Event Details (e.g. scheduled/started/initiated event IDs, Namespace, Namespace ID)

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Merge Checklist 

- [ ] Waiting on confirmation from Cloud team on counts

### Issue(s) closed 
`DT-2916`
`DT-2982`
`DT-2985`

PR #2830, opened 2025-07-25
DT-3150 - refactor paginated table to use `<Table />` component

rossedfort:
## Description & motivation 💭 
- Refactor paginated table to use `<Table />` component
- Remove "split" variant for paginated table
- Add "bordered" prop to `<Table />` to support only adding a border when the table isn't wrapped by an element with a border
  - Also added to `<SkeletonTable />`
 - Add deprecation comment to `<ApiPagination />`
### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/holocene/api-pagination.svelte
+++ b/src/lib/holocene/api-pagination.svelte
@@ -1,3 +1,7 @@
+<!--
+@component ApiPagination
+@deprecated This component is deprecated. Use `src/lib/holocene/table/paginated-table/api-paginated.svelte` instead.
+-->
 <script lang="ts">
   import type { HTMLAttributes } from 'svelte/elements';
 
--- a/src/lib/holocene/skeleton/table.svelte
+++ b/src/lib/holocene/skeleton/table.svelte
@@ -13,7 +13,7 @@
   export { className as class };
 </script>
 
-<Table class="w-full table-fixed {className}" variant="fancy">
+<Table class="w-full table-fixed {className}" variant="primary">
   <TableHeaderRow slot="headers" class="h-8">
     <slot name="headers">
       {#each Array.from(new Array(columns)) as _column, i}
--- a/src/lib/holocene/skeleton/table.svelte
+++ b/src/lib/holocene/skeleton/table.svelte
@@ -9,11 +9,10 @@
   export let rows = 10;
   export let columns = 4;
   export let columnWidths: number[] = new Array(columns).fill(100 / columns);
-  export let className = '';
-  export { className as class };
+  export let bordered: boolean = true;
 </script>
 
-<Table class="w-full table-fixed {className}" variant="primary">
+<Table class="w-full" fixed variant="primary" {bordered}>
   <TableHeaderRow slot="headers" class="h-8">
     <slot name="headers">
       {#each Array.from(new Array(columns)) as _column, i}
--- a/src/lib/holocene/table/paginated-table/index.svelte
+++ b/src/lib/holocene/table/paginated-table/index.svelte
@@ -5,7 +5,6 @@
   type Item = $$Generic;
 
   export let visibleItems: Item[];
-  export let variant: 'primary' | 'split' = 'primary';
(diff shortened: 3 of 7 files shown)


src/lib/holocene/table/paginated-table/index.svelte
```diff
@@ -1,15 +1,15 @@
 <script lang="ts">
-  import ProgressBar from '$lib/holocene/progress-bar.svelte';
   import SkeletonTable from '$lib/holocene/skeleton/table.svelte';
+  import Table from '$lib/holocene/table/table.svelte';
 
   type Item = $$Generic;
 
   export let visibleItems: Item[];
-  export let variant: 'primary' | 'split' = 'primary';
```
rossedfort, line 8: @Alex-Tideman do you know if this variant is still used anywhere or if it's safe to remove? I searched both oss and cloud and couldn't find any instances where we use `"split"`.

laurakwhit: I think we removed our only instance of it [here](https://github.com/temporalio/ui/pull/2758/files#diff-0bfea6afef040a85292189cfa9f08a90920a25cf34281383e894b6fa8be5b2f0L60), so should be fine to remove 🗑️

src/lib/holocene/api-pagination.svelte
```diff
@@ -1,3 +1,7 @@
+<!--
+@component ApiPagination
+@deprecated This component is deprecated. Use `src/lib/holocene/table/paginated-table/api-paginated.svelte` instead.
```
laurakwhit, line 3: 🎉

src/lib/holocene/skeleton/table.svelte
```diff
@@ -9,11 +9,10 @@
   export let rows = 10;
   export let columns = 4;
   export let columnWidths: number[] = new Array(columns).fill(100 / columns);
-  export let className = '';
-  export { className as class };
+  export let bordered: boolean = true;
```
laurakwhit, line 12: NIT: No strong opinions on noun vs. adjective props, but maybe something we should align on for a code style guide. E.g. `border` vs. `bordered`.

src/lib/holocene/table/paginated-table/index.svelte
Alex-Tideman: Yup 🗑️ it

src/lib/holocene/table/paginated-table/index.svelte
```diff
@@ -1,15 +1,15 @@
...
   export let loading = false;
   export let updating = false;
   export let maxHeight = '';
   export let fixed = false;
+  export let id: string = null;
```
laurakwhit: Should we just extend the `HTMLTableAttributes` here instead of defining `id`?

src/lib/holocene/skeleton/table.svelte
rossedfort: In naming this I was thinking along the lines of props/attributes like `disabled` or `required` that imply truthiness about a property/state of the component. Also similar to how we use `expanded` elsewhere.

laurakwhit: Yeah was mostly just musing on what might be a "rule" we can align on for everywhere. We have `disabled` but then also `isEnabled` in places or `active` vs. `isActive` for example. I'll bring to the rest of the team for discussion.

src/lib/holocene/table/paginated-table/index.svelte
rossedfort: In this case maybe not the most useful, because this component is really only ever used when wrapped by `src/lib/holocene/table/paginated-table/api-paginated.svelte` or `src/lib/holocene/table/paginated-table/paginated.svelte`, and those components don't extended html table attributes. That said, they probably should. I've at least made the change to this component for now.

Alex-Tideman (approved):

PR #2832, opened 2025-07-25
Link to target Task Queue, show if any active workers, add Nexus Handler column

Alex-Tideman:
…exus handler column

## Description & motivation 💭 
Updates for Nexus Endpoint:
1. Link to target Task Queue page
2. Show alert with count of pollers to alert user if no pollers are active on the task queue - Same as Start Workflow page
3. Add Nexus Handler column on the Worker Table to show if poller handles Nexus tasks.

Screenshot 2025-07-25 at 2 20 08 PM
Screenshot 2025-07-25 at 2 16 36 PM
Screenshot 2025-07-25 at 2 19 05 PM

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Draft Checklist 

- [ ] Need to check if the pollers-service logic for Nexus is correct.

### DIFF
--- a/src/lib/components/worker-table.svelte
+++ b/src/lib/components/worker-table.svelte
@@ -122,6 +122,9 @@
     <th class="w-2/12 text-center">
       <p class="text-center">{translate('workflows.activity-handler')}</p>
     </th>
+    <th class="w-2/12 text-center">
+      <p class="text-center">{translate('workflows.nexus-handler')}</p>
+    </th>
   </TableHeaderRow>
   {#each pollers as poller}
     {@const deployment = getPollerDeploymentName(poller)}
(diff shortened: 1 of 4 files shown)


src/lib/components/worker-table.svelte
```diff
@@ -122,6 +120,9 @@
     <th class="w-2/12 text-center">
       <p class="text-center">{translate('workflows.activity-handler')}</p>
     </th>
+    <th class="w-2/12 text-center">
+      <p class="text-center">{translate('workflows.nexus-handler')}</p>
+    </th>
```
rossedfort, line 125: Do we need this `<p />` inside of a `<th />`?

rossedfort (approved):

src/lib/components/worker-table.svelte
Alex-Tideman: Yeah just so we can actually text-center. The table default css will override the <th class="text-center">, something to think about for future tabling.

rossedfort: Ah makes sense. Yeah instead of styling globally if we created components for each table element, and used `twMerge` / `clsx`, that would allow us to do `<Table.HeaderCell class="text-center" />` 👍

PR #2840, opened 2025-07-29
Limit workflow execution query length to 15k characters

laurakwhit:
## Description & motivation 💭 
* Prevents queries over 15k from being 
     * sent as a request in the URL
     * pasted into the workflow search input 
* Moves the `maxLength` outside of the `Input` component and inline with the `hintText` below
* Replaces `Spinner` icon with solid version

| Before | After |
|-|-|
|Screenshot 2025-07-29 at 3 38 07 PM|Screenshot 2025-07-29 at 3 38 34 PM|
|Screenshot 2025-07-29 at 5 17 30 PM|Screenshot 2025-07-29 at 5 18 06 PM|

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

* On a clearable `Input` that has `maxLength` set
   - [ ] Verify clicking the `x` clear button clears the value
   - [ ] Verify the `maxLength` is visible below the input on focus (and if `hideCount` is `false`)

---

* On the `/workflows` page
   - [ ] Verify the `Filter` button is disabled if the query in the URL is over 15k characters
   - [ ] Verify a value can not be pasted in the workflow search input that is over 15k characters
 
### Issue(s) closed 
`DT-3119`


src/lib/utilities/request-from-api.ts
```diff
@@ -80,6 +82,11 @@ export const requestFromAPI = async <T>(
       ...nextPageToken,
     });
   }
+
+  const queryString = query.toString();
+  if (queryString.length >= MAX_QUERY_LENGTH) {
+    query = new URLSearchParams(queryString.substring(0, MAX_QUERY_LENGTH - 1));
```
laurakwhit: Not sure if this is really what we want to be doing here. Seems like the API should maybe just return an error instead. Maybe we can add something via middleware? Open to suggestions!

Alex-Tideman: Gah ya this is tricky. I don't think we wanna do this because it will almost certainly still result in an error with hacking off the end of the query. I think the spirit of ticket is to prevent the request from ever being sent. 

It's really only going to happen on our workflow list query, so I'd suggest we do the check there (onMount with the query param and when triggering via manual query a la copy/paste most likely) and throw error text on the input and prevent triggering the query

laurakwhit: I think it would also happen for the workflow count query, no? Maybe safest to keep something that checks all the requests 🤔 

What do you think about a custom response [here](https://github.com/temporalio/ui/pull/2840/commits/304dea9526645f8ebd0435dd05d1e7ac0d2a8ebf). That way the error would show up in the UI like this on the `/workflows` page and the request wouldn't be sent ⬇️ 
Screenshot 2025-07-30 at 3 35 44 PM

src/lib/holocene/input/input.svelte
```diff
@@ -173,15 +164,34 @@
...
 
-  <span
-    class="hint-text inline-block"
-    class:invalid={!valid}
-    class:error
-    class:hidden={!hintText}
-    role={error ? 'alert' : null}
+  <div
+    class="flex justify-between gap-2"
+    class:hidden={!hintText && (!maxLength || disabled || hideCount)}
```
Alex-Tideman, line 169: We should make this inline-flex so it doesn't change the height

src/lib/components/workflow/workflow-advanced-search.svelte
```diff
@@ -80,7 +80,8 @@
       on:clear={handleClearInput}
       bind:value={manualSearchString}
       maxLength={MAX_QUERY_LENGTH}
-      hideCount={manualSearchString?.length < MAX_QUERY_LENGTH}
+      hideCount={!manualSearchString ||
```
laurakwhit, line 83: > Looks like the max appears on focus

@Alex-Tideman should be fixed now, thanks for catching!

Alex-Tideman:
Screenshot 2025-08-05 at 3 20 45 PM
Screenshot 2025-08-05 at 3 20 58 PM

Screenshot 2025-08-05 at 3 23 28 PM
Screenshot 2025-08-05 at 3 24 18 PM


So it does at height to the input, wondering if we can make the count smaller text

src/lib/holocene/input/input.svelte
```diff
@@ -178,7 +178,7 @@
     </span>
     {#if maxLength && !disabled && !hideCount}
       <span
-        class="invisible text-right text-sm font-medium tracking-widest group-focus-within:visible"
+        class="invisible text-right text-xs tracking-widest group-focus-within:visible"
```
laurakwhit, line 181: > So it does at height to the input, wondering if we can make the count smaller text


@Alex-Tideman updated size for both `Input` and `Textarea` to be `text-xs` ⬇️ 

Screenshot 2025-08-05 at 3 37 11 PM
Screenshot 2025-08-05 at 3 36 44 PM

src/lib/utilities/request-from-api.ts
```diff
@@ -90,7 +90,9 @@ export const requestFromAPI = async <T>(
       options = await withAuth(options, isBrowser);
     }
 
-    const queryIsTooLong = query.toString().length > MAX_QUERY_LENGTH;
+    const queryIsTooLong = [...query.values()].some(
+      (value) => value.length > MAX_QUERY_LENGTH,
+    );
```
laurakwhit, line 95: Decided to use `values()` instead of `toString()` so the character count should be a little more accurate here. But will still be a little off for the `/worklfow-count` request because we add a "GROUP BY ExecutionStatus" in addition to whatever is in the query input.

Alex-Tideman: 👍

PR #2845, opened 2025-08-05
Add workflowDetailsAddition snippet support to workflow header

rossnelson:
## Summary
- Add optional snippet parameter to WorkflowHeader and WorkflowRunLayout components
- Enable custom content injection in workflow details section
- Provide example implementation with namespace tags display

## Test plan
- [ ] Verify workflow header renders correctly with additional content
- [ ] Test that existing workflow pages continue to work without changes
- [ ] Confirm snippet renders in correct position within workflow details


rossnelson:
There is some thoughts on a different design. 

image

Pausing on this until the decision is made.

PR #2848, opened 2025-08-06
Add local activity to timeline

GraceGardner:
## Description & motivation 💭 

So I've pulled out the decoding logic for the local activity. 

The thing I'm stuck on is that the local activities are showing in the time line event... Should the text next to the little icon not be the activity type text, but instead be "Local Activity" 

Local activity in the Timeline view 

Screenshot 2025-08-25 at 12 11 21 PM

Local Activity in the Event Summary Row. 

Screenshot 2025-08-25 at 12 11 35 PM

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
DT-3026


src/lib/components/event/event-summary-row.svelte
```diff
@@ -251,13 +188,23 @@
   };
 
   onMount(async () => {
+    let targetEvent: WorkflowEvent | undefined;
+
     if (isLocalActivityMarkerEvent(event)) {
-      primaryLocalAttribute = await decodeLocalActivity(event);
+      targetEvent = event as WorkflowEvent;
```
laurakwhit: Not sure we want this `as WorkflowEvent` here. Can we change the type on the event arg in decodeLocalActivity to `IterableEvent`?

src/lib/components/lines-and-dots/svg/timeline-graph-row.svelte
```diff
@@ -191,7 +226,11 @@
     <Dot
       point={[x, y]}
       classification={group.eventList[index]?.classification}
-      icon={pauseTime && index !== 0 ? 'pause' : CategoryIcon[group.category]}
+      icon={pauseTime && index !== 0
+        ? 'pause'
+        : decodedLocalActivity
+          ? CategoryIcon['local-activity']
+          : CategoryIcon[group.category]}
```
laurakwhit, line 232: I think it's fine to leave the text next to the icon as is in the timeline, but wondering about the "Marker Recorded" on the expanded row ⬇️ 

Screenshot 2025-08-13 at 9 53 41 AM

Will let @Alex-Tideman weight-in though.

src/lib/components/event/event-summary-row.svelte
GraceGardner: Thanks Laura, I keep mixing these two up. You're right, it should be the IterableEvent.

src/lib/components/lines-and-dots/svg/timeline-graph-row.svelte
Alex-Tideman: Yeah that should say Local Activity

src/lib/components/event/event-summary-row.svelte
GraceGardner: Turns out we do need WorkflowEvent, that's where we can get the marker Local Activity from. If we use IterableEvent we get Undefined. The original implementation also used WorkflowEvent.

laurakwhit: `IterableEvent` (which is what the event is being typed as) includes `WorkflowEvent`. If we should only be getting a `WorkflowEvent` here we might want some type narrowing or additional conditions to check for that.

src/lib/components/lines-and-dots/svg/timeline-graph-row.svelte
```diff
@@ -162,7 +195,7 @@
         prefix={isActivityTaskScheduledEvent(group.initialEvent)
           ? group?.displayName
           : ''}
-        fallback={group?.displayName}
+        fallback={decodedLocalActivity ? 'Local Activity' : group?.displayName}
```
laurakwhit:
```suggestion
        fallback={decodedLocalActivity ? translate('events.category.local-activity') : group?.displayName}
```

src/lib/components/event/event-summary-row.svelte
```diff
@@ -252,12 +189,23 @@
...
     if (isLocalActivityMarkerEvent(event)) {
-      primaryLocalAttribute = await decodeLocalActivity(event);
+      primaryLocalAttribute = await decodeLocalActivity(
+        event as WorkflowEvent,
+        {
+          namespace: page.params.namespace,
+          settings: page.data.settings,
+          accessToken: $authUser.accessToken,
+        },
+      );
```
laurakwhit: We can get rid of that `as WorkflowEvent` since `decodeLocalActivity` accepts an `IterableEvent` now

```suggestion
      primaryLocalAttribute = await decodeLocalActivity(event, {
        namespace: page.params.namespace,
        settings: page.data.settings,
        accessToken: $authUser.accessToken,
      });
```

src/lib/components/lines-and-dots/svg/timeline-graph-row.svelte
```diff
@@ -37,6 +47,29 @@
   $: pendingActivity = group?.pendingActivity;
   $: pauseTime = pendingActivity && pendingActivity.pauseInfo?.pauseTime;
 
+  let decodedLocalActivity: SummaryAttribute | undefined;
+
+  onMount(async () => {
+    if (hasLocalActivityMarker(group)) {
+      const localActivityEvent = getLocalActivityMarkerEvent(group);
```
laurakwhit: Is there a reason for both `hasLocalActivityMarker` and `getLocalActivityMarkerEvent`? Seems like we could just determine if there is a local activity marker event with `getLocalActivityMarkerEvent` (especially since there is already a `if (localActivityEvent)` condition below) instead of iterating over the group twice.

laurakwhit (approved):
Some small comments below, but otherwise looks good!

src/lib/utilities/decode-local-activity.ts
```diff
@@ -0,0 +1,107 @@
...
+
+export const hasLocalActivityMarker = (
+  eventOrGroup: WorkflowEvent | EventGroup,
+): boolean => {
+  if ('eventList' in eventOrGroup) {
+    return eventOrGroup.eventList.some(isLocalActivityMarkerEvent);
+  } else {
+    return isLocalActivityMarkerEvent(eventOrGroup);
+  }
+};
```
laurakwhit: Can remove this if no longer using 🗑️ 

```suggestion
```

PR #2850, opened 2025-08-07
Add core context with getUserIdentifier

laurakwhit:
## Description & motivation 💭 
We were using the $authUser.email in the identity field for certain requests. This PR adds a `getUserIdentifier` to context that is set in the root layout so that identity can be something other than email and is accessible to all child components.

This PR also adds the identity field to the Start Workflow request.

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Merge Checklist 
If an auth user is set and has an email
  - [ ] The email should show up in the placeholder text and request when
      - cancelling workflows
      - reseting workflows
      - terminating workflows
  - [ ] The email should show up in the request when
     - creating a schedule
     - editing a schedule
     - pausing/unpausing a schedule
     - deleting a schedule
     - triggering a schedule immediately 
     - backfilling a schedule
     - starting a workflow
### Issue(s) closed 
Related to `DT-3141`
`DT-3158`

### DIFF
--- a/src/lib/pages/schedule-view.svelte
+++ b/src/lib/pages/schedule-view.svelte
@@ -43,6 +43,7 @@
   import { relativeTime, timeFormat } from '$lib/stores/time-format';
   import { refresh, workflowCount } from '$lib/stores/workflows';
   import type { OverlapPolicy } from '$lib/types/schedule';
+  import { getIdentity } from '$lib/utilities/core-context';
   import { decodeURIForSvelte } from '$lib/utilities/encode-uri';
   import { formatDate, getUTCString } from '$lib/utilities/format-date';
   import {
@@ -56,6 +57,8 @@
   let namespace = $page.params.namespace;
   let scheduleId = $page.params.schedule;
   let workflowQuery = `TemporalScheduledById="${scheduleId}"`;
+
+  const identity = getIdentity();
 
   const parameters = {
     namespace,
@@ -119,7 +122,7 @@
     error = '';
     try {
       $loading = true;
-      await deleteSchedule({ namespace, scheduleId });
+      await deleteSchedule({ identity, namespace, scheduleId });
       deleteConfirmationModalOpen = false;
       setTimeout(() => {
         $loading = false;
@@ -137,11 +140,13 @@
   const handlePause = async (schedule: DescribeScheduleResponse) => {
     schedule?.schedule?.state?.paused
       ? await unpauseSchedule({
+          identity,
           namespace,
           scheduleId,
           reason,
         })
       : await pauseSchedule({
+          identity,
           namespace,
           scheduleId,
           reason,
@@ -159,6 +164,7 @@
   const handleTriggerImmediately = async () => {
     scheduleUpdating = true;
     await triggerImmediately({
+      identity,
       namespace,
       scheduleId,
       overlapPolicy: $overlapPolicy,
@@ -232,6 +238,7 @@
     });
 
     await backfillRequest({
+      identity,
       namespace,
       scheduleId,
       overlapPolicy: $overlapPolicy,
--- a/src/lib/services/workflow-service.ts
+++ b/src/lib/services/workflow-service.ts
@@ -129,6 +129,7 @@ type TerminateWorkflowOptions = {
   namespace: string;
   reason: string;
   first: string | undefined;
+  identity: string;
 };
 
 export type ResetWorkflowOptions = {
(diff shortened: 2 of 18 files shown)


src/lib/services/workflow-service.ts
```diff
@@ -469,6 +470,7 @@ export async function resetWorkflow({
     workflowTaskFinishEventId: eventId,
     requestId: v4(),
```
GiantRobots, line 471: Use https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID instead of v4 wherever possible so we can kill our guid dependency.

src/lib/types/schedule.ts
```diff
@@ -33,6 +33,7 @@ export type SchedulePreset =
 export type ScheduleOffsetUnit = 'days' | 'hrs' | 'min' | 'sec';
 
 export type ScheduleActionParameters = {
+  identity?: string;
```
GiantRobots, line 36: This being allowable to undefined is potentially fucky I think? It's not immediately obvious to me why we wouldn't always supply identity

Alex-Tideman: Local development won't have an identity (no authUser)

src/lib/utilities/workflow-actions.ts
```diff
@@ -21,10 +21,10 @@ export const getPlacholder = (action: Action, email?: string): string => {
...
   }
 
-  return email
-    ? translate('workflows.workflow-action-reason-placeholder-with-email', {
+  return identity
+    ? translate('workflows.workflow-action-reason-placeholder-with-identity', {
         action: translatedAction,
-        email,
+        identity,
       })
```
GiantRobots: I wonder if we should be using i18n plurals support instead of ternary operators in general. Just a thought not a necessary change https://www.i18next.com/translation-function/plurals

src/routes/(app)/+layout.svelte
```diff
@@ -219,6 +220,10 @@
   afterNavigate(() => {
     document.getElementById('content')?.scrollTo(0, 0);
   });
+
+  setCoreContext({
+    getUserIdentifier: () => $authUser.email || '',
```
GiantRobots, line 225: Did we verify this is the earliest to set this? I assume that this flow was tested I just have ptsd from trying to do this for auth stuff.

laurakwhit: We're setting the authUser in the layout load function here, so this would be the earliest we could set this in a .svelte component.

src/lib/pages/schedule-view.svelte
```diff
@@ -137,11 +140,13 @@
   const handlePause = async (schedule: DescribeScheduleResponse) => {
     schedule?.schedule?.state?.paused
       ? await unpauseSchedule({
+          identity,
```
GiantRobots, line 143: Can this ternary be an if statement instead it feels like it would be clearer what is happening.

src/lib/services/workflow-service.ts
```diff
@@ -639,6 +642,7 @@ export async function startWorkflow({
               ...setSearchAttributes(searchAttributes),
             },
           },
+    ...(identity ? { identity } : {}),
```
GiantRobots: In my ever long quest to remove all ternaries I think this works the same using the && operator, dunno if it's any better tho

```typescript
searchAttributes: searchAttributes.length > 0 && {
  indexedFields: {
    ...setSearchAttributes(searchAttributes),
  },
},
...(identity && { identity }),`

GiantRobots: oh lol we use `...(identity && { identity }),` in one of the other files

src/lib/services/workflow-service.ts
laurakwhit: Added a follow-up task for this under core improvements.

src/lib/services/workflow-service.ts
laurakwhit: [Updated](https://github.com/temporalio/ui/pull/2850/commits/d3ea67675743769f0b61b7a32dabe0cbb53a2666) ✅

src/lib/utilities/workflow-actions.ts
laurakwhit: [Updated](https://github.com/temporalio/ui/pull/2850/commits/773481a2ce81ab3b693bb92e14b7fa30652725ca) ✅

PR #2868, opened 2025-08-19
DT-3233 - add more language support to codeblock

rossedfort:
## Description & motivation 💭 
- Adds Go, TypeScript, Python, PHP, Java, .NET and Ruby language support to codeblock
- Adds Stories for each of the new languages with applicable code snippets
- Adds optional Tabs to codeblock
- fix Link icon

| before | after | 
| --- | --- |
|Screenshot 2025-08-26 at 10 10 14 AM|Screenshot 2025-08-26 at 10 20 43 AM|

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/holocene/code-block.stories.svelte
```diff
@@ -188,3 +208,126 @@
...
+
+<Story name="With Header">
+  <CodeBlock
+    copyable
+    language="typescript"
+    tabs={['File A', 'File B']}
+    bind:activeTab
+    content={hidden
+      ? content[activeTab]
+      : content[activeTab].replace('***', 'secret')}
```
andrewzamojc, line 321: Comment: Neat tabs. I like the simple interface.

src/lib/holocene/code-block.svelte
```diff
@@ -227,7 +234,44 @@
...
+            {copyIconTitle}
+            {copySuccessIconTitle}
+            class="m-0 rounded-full text-secondary"
+            on:click={handleCopy}
+            copied={$copied}
+          />
+        {/if}
+      </div>
+    </div>
+  {/if}
```
andrewzamojc, line 274: Possible change: Do the tabs and header actions play well with the maximizable and existing copy button? It seems kind of separate in the code and I can't quite do it in storybook. Ideally the tabs are maximizable too ~and the copy button isn't repeated.~

andrewzamojc (commented):
Cool looking tabbed ui. I left a comment that suggests a change. Let me know what you think.

rossedfort:
> You sure this passes accessibility?

Good call, I didn't check

src/lib/holocene/code-block.svelte
rossedfort: Good question, I didn't check on that behavior.

Alex-Tideman (approved):

PR #2870, opened 2025-08-19
Always decode last failure

Alex-Tideman:
## Description & motivation 💭 
Previously I was only decoding if failure converter was setup. However last failure details can also be a payload. So just always decode it. It should render the last failure correctly if no payloads. It should also update after every attempt.

To test, add ApplicationFailure in activity

```
      throw ApplicationFailure.create({
        message: "Error message",
        type: "ErrorType",
        details: [{ failure: `Failed at ${Date.now()}` }],
      });
```

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


rossnelson (approved):
sweet 👍 

https://github.com/user-attachments/assets/4fb11477-6130-41e5-9629-8e909e9b66d3

PR #2872, opened 2025-08-20
Add hidden edit btn and default route prop

KATIETOLER:
This adds the ability to change the route on the `back` breadcrumb on the nexus details page while still keeping the original functionality.

### DIFF
--- a/src/lib/pages/nexus-endpoint.svelte
+++ b/src/lib/pages/nexus-endpoint.svelte
@@ -17,12 +17,11 @@
   let {
     endpoint,
     editDisabled = false,
-    editHidden = false,
-    backHref = routeForNexus,
+    backHref = routeForNexus(),
     taskQueueStatus,
   }: {
     endpoint: Endpoint;
-    backHref?: () => string;
+    backHref: string;
     editDisabled?: boolean;
     editHidden?: boolean;
     taskQueueStatus?: Snippet;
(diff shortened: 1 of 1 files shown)


src/lib/pages/nexus-endpoint.svelte
```diff
@@ -17,17 +17,21 @@
   let {
     endpoint,
     editDisabled = false,
+    editHidden = false,
+    backHref = routeForNexus,
     taskQueueStatus,
   }: {
     endpoint: Endpoint;
+    backHref?: () => string;
```
Alex-Tideman: Let's make this a string not a function.

So default prop would be

`backHref = routeForNexus()`

src/lib/pages/nexus-endpoint.svelte
```diff
@@ -38,6 +42,7 @@
       </h1>
       <Button
         href={routeForNexusEndpointEdit(endpoint.id)}
+        hidden={editHidden}
```
Alex-Tideman: I don't believe this will hide the button. We don't have a hidden prop. I think it's fine to just use editDisabled and not hide this at all

andrewzamojc: The `hidden` attribute might be passing through to the element via `HTMLAnchorAttributes`. Disabled could work too, although a tooltip would be nice when disabled.

src/lib/pages/nexus-endpoint.svelte
```diff
@@ -17,17 +17,20 @@
   let {
     endpoint,
     editDisabled = false,
+    backHref = routeForNexus(),
     taskQueueStatus,
   }: {
     endpoint: Endpoint;
+    backHref?: string;
     editDisabled?: boolean;
+    editHidden?: boolean;
```
laurakwhit: Is this being used anywhere?

KATIETOLER: nooo, i'll take i out!

PR #2883, opened 2025-08-28
Show running workflow link

GraceGardner:
## Description & motivation 💭 
If a workflow has already been started, the error message will now also display a link to the workflow the has already been started so that user can navigate to it. 

Screenshot 2025-08-27 at 10 30 23 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
DT-3025

### DIFF
--- a/src/lib/pages/start-workflow.svelte
+++ b/src/lib/pages/start-workflow.svelte
@@ -64,6 +64,26 @@
 
   let searchAttributes: SearchAttributeInput[] = [];
 
+  $: errorWorkflowDetails = extractWorkflowFromError(error);
+
+  function extractWorkflowFromError(errorMessage: string): {
+    workflowId?: string;
+    runId?: string;
+  } {
+    if (!errorMessage) return {};
+    const workflowIdMatch = errorMessage.match(
+      /workflow[_\s]*id[:\s=]+['"]*([^'"\s,;)]+)['"]*\s*/i,
+    );
+    const runIdMatch = errorMessage.match(
+      /run[_\s]*id[:\s=]+['"]*([^'"\s,;)]+)['"]*\s*/i,
+    );
+
+    return {
+      workflowId: workflowIdMatch?.[1]?.trim(),
+      runId: runIdMatch?.[1]?.trim(),
+    };
+  }
+
   $: taskQueueParam = $page.url.searchParams.get('taskQueue');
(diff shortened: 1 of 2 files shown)


src/lib/pages/start-workflow.svelte
```diff
@@ -64,6 +64,29 @@
...
+  $: errorWorkflowDetails = extractWorkflowFromError(error);
+
+  function extractWorkflowFromError(errorMessage: string): {
+    workflowId?: string;
+    runId?: string;
+  } {
+    if (!errorMessage) return {};
+
+    const workflowIdMatch = errorMessage.match(
+      /workflow[_\s]*id[:\s=]+['"]*([^'"\s,;).]+)['"]*[.\s]*/i,
```
Alex-Tideman: This regex isn't going to work because workflowIds/runIds can have periods in them.

So something like workflowId=temporal.canary.cron-workflow.sanity-2025-08-28T09 would go to just temporal/runId

Might need to try to split the errorMessage at WorkflowId: and RunId:

GraceGardner: Okay, I added in the string functions instead of the regex. I think it's better now, but let me know what you think and if there are any other edge cases I should be keeping an eye out for. Thanks boss!

src/lib/pages/start-workflow.svelte
```diff
@@ -64,6 +64,38 @@
...
+      const runPart = runParts[1].slice(0, -1).trim();
+      if (runPart) {
+        runId = runPart;
+      }
+    }
+    return {
+      workflowId,
+      runId,
+    };
+  }
```
Alex-Tideman: This might be a little cleaner way to do this with regex:

```
  function extractWorkflowFromError(errorMessage: string): {
    workflowId?: string;
    runId?: string;
  } {
    if (!errorMessage) return {};

    const match = errorMessage.match(/WorkflowId: (.+?), RunId: (.+?)\.?$/);
    if (!match) return {};
    const workflowId = match[1]?.trim(),
    const runId = match[2]?.trim()

    if (!workflowId || !runId) return {};
    return {
      workflowId: match[1].trim(),
      runId: match[2].trim(),
    };
  }
```

PR #2885, opened 2025-08-29
Small UI fixes and improvements

laurakwhit:
## Description & motivation 💭 

* Make `Drawer` have `w-screen` on small screens
* Add `position` prop with `left` and `right` options to the `Select` component
* Update `ToggleButton` to just use the `Button` component
* Fix gap between label and input in `Combobox` and `Select` components to match other similar components
* Fix background gap ⬇️ next to empty state image on extra wide screens
Screenshot 2025-08-06 at 5 33 29 PM

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/workflow/workflows-summary-configurable-table/table-empty-state.svelte
+++ b/src/lib/components/workflow/workflows-summary-configurable-table/table-empty-state.svelte
@@ -79,7 +79,7 @@
     {/if}
   </div>
   <div
-    class="bg-[#DDD6FE] bg-contain bg-no-repeat xl:flex xl:min-h-full xl:flex-1 xl:items-stretch xl:justify-center"
+    class="bg-[#DDD6FE] bg-contain bg-no-repeat xl:flex xl:min-h-full xl:max-w-[1050px] xl:flex-1 xl:items-stretch xl:justify-center"
     style="background-image: url({noResultsImages});"
   >
     <img
--- a/src/lib/holocene/combobox/combobox.svelte
+++ b/src/lib/holocene/combobox/combobox.svelte
@@ -422,96 +422,105 @@
 </script>
 
 <MenuContainer {open} on:close={handleMenuClose}>
-  <Label class="pb-1" hidden={labelHidden} {required} {label} for={id} />
-
-  <div
-    class={merge(
-      comboboxStyles({ variant }),
-      !valid &&
-        variant === 'default' &&
-        'border border-danger text-danger focus-within:ring-danger/70',
-      disabled && 'opacity-50',
-      className,
-    )}
-    class:disabled
-  >
-    {#if leadingIcon}
-      <Icon class="ml-2 shrink-0" name={leadingIcon} />
-    {/if}
+  <div class="flex flex-col gap-1">
+    <Label hidden={labelHidden} {required} {label} for={id} />
     <div
       class={merge(
-        'input-wrapper',
-        multiselect && 'gap-1',
-        multiselect && 'm-1',
-        leadingIcon && multiselect && 'ml-2',
+        comboboxStyles({ variant }),
+        !valid &&
+          variant === 'default' &&
+          'border border-danger text-danger focus-within:ring-danger/70',
+        disabled && 'opacity-50',
+        className,
       )}
+      class:disabled
     >
-      {#if multiselect && isArrayValue(value) && value.length > 0}
-        {#if displayChips}
-          {#each value.slice(0, chipLimit) as v}
-            <Chip
-              on:remove={() => removeOption(v)}
-              removeButtonLabel={removeChipLabel}>{v}</Chip
-            >
-          {/each}
-          {#if value.length > chipLimit}
-            <p>+{value.slice(chipLimit).length}</p>
-          {/if}
-        {:else}
-          <Badge>{numberOfItemsSelectedLabel(value.length)}</Badge>
-        {/if}
+      {#if leadingIcon}
+        <Icon class="ml-2 shrink-0" name={leadingIcon} />
       {/if}
-      <input
-        {id}
-        {placeholder}
-        {required}
-        {readonly}
-        {disabled}
-        type="text"
-        value={displayValue}
-        class:disabled
+      <div
         class={merge(
-          'combobox-input',
-          multiselect
-            ? value.length > 0 || leadingIcon
-              ? 'indent-0'
-              : 'indent-1'
-            : 'indent-2',
-          className,
+          'input-wrapper',
+          multiselect && 'gap-1',
+          multiselect && 'm-1',
+          leadingIcon && multiselect && 'ml-2',
         )}
-        role="combobox"
-        autocomplete="off"
-        autocapitalize="off"
-        spellcheck="false"
-        data-lpignore="true"
-        data-1p-ignore="true"
-        aria-controls="{id}-listbox"
-        aria-expanded={$open}
-        aria-required={required}
-        aria-autocomplete="list"
-        on:focus|stopPropagation={openList}
-        on:input|stopPropagation={handleInput}
-        on:keydown|stopPropagation={handleInputKeydown}
-        on:click|stopPropagation={handleInputClick}
-        data-testid={$$props['data-testid'] ?? id}
-        bind:this={inputElement}
-        {...$$restProps}
-      />
-    </div>
-    {#if $$slots.action}
-      <div class="ml-1 flex h-full items-start border-l border-subtle p-0.5">
-        {#if actionTooltip}
-          <Tooltip text={actionTooltip} right>
-            <slot name="action" />
-          </Tooltip>
-        {:else}
-          <slot name="action" />
+      >
+        {#if multiselect && isArrayValue(value) && value.length > 0}
+          {#if displayChips}
+            {#each value.slice(0, chipLimit) as v}
+              <Chip
+                on:remove={() => removeOption(v)}
+                removeButtonLabel={removeChipLabel}>{v}</Chip
+              >
+            {/each}
+            {#if value.length > chipLimit}
+              <p>+{value.slice(chipLimit).length}</p>
+            {/if}
+          {:else}
+            <Badge>{numberOfItemsSelectedLabel(value.length)}</Badge>
+          {/if}
         {/if}
+        <input
+          {id}
+          {placeholder}
+          {required}
+          {readonly}
+          {disabled}
+          type="text"
+          value={displayValue}
+          class:disabled
+          class={merge(
+            'combobox-input',
+            multiselect
+              ? value.length > 0 || leadingIcon
+                ? 'indent-0'
+                : 'indent-1'
+              : 'indent-2',
+            className,
+          )}
+          role="combobox"
+          autocomplete="off"
+          autocapitalize="off"
+          spellcheck="false"
+          data-lpignore="true"
+          data-1p-ignore="true"
+          aria-controls="{id}-listbox"
+          aria-expanded={$open}
+          aria-required={required}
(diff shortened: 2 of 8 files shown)


src/lib/components/workflow/workflows-summary-configurable-table/table-empty-state.svelte
```diff
@@ -79,7 +79,7 @@
     {/if}
   </div>
   <div
-    class="bg-[#DDD6FE] bg-contain bg-no-repeat xl:flex xl:min-h-full xl:flex-1 xl:items-stretch xl:justify-center"
+    class="bg-[#DDD6FE] bg-contain bg-no-repeat xl:flex xl:min-h-full xl:max-w-[1050px] xl:flex-1 xl:items-stretch xl:justify-center"
```
GiantRobots, line 82: What is the 1050 magic number here? 1/2 of the total screen width of xl?

I kind of like static max-w's because it keeps the width stationary when it hits max instead of always expanding/minimizing so I think I prefer this to like `max-w-[50dvw]`

Nothing to do thinking out loud

src/lib/holocene/combobox/combobox.svelte
```diff
@@ -422,133 +422,134 @@
...
-        multiselect && 'm-1',
-        leadingIcon && multiselect && 'ml-2',
+        comboboxStyles({ variant }),
+        !valid &&
+          variant === 'default' &&
+          'border border-danger text-danger focus-within:ring-danger/70',
+        disabled && 'opacity-50',
+        className,
       )}
+      class:disabled
```
GiantRobots: The mix match of disableds here is a lil weird, should disabled go on the `disabled && ''` above

src/lib/holocene/combobox/combobox.svelte
```diff
@@ -422,133 +422,134 @@
...
-          multiselect
-            ? value.length > 0 || leadingIcon
-              ? 'indent-0'
-              : 'indent-1'
-            : 'indent-2',
-          className,
+          'input-wrapper',
+          multiselect && 'gap-1',
+          multiselect && 'm-1',
+          leadingIcon && multiselect && 'ml-2',
```
GiantRobots, line 445: Gorgeous removal of the ternary 10/10

src/lib/holocene/combobox/combobox.svelte
```diff
@@ -422,133 +422,134 @@
...
+          type="text"
+          value={displayValue}
+          class:disabled
+          class={merge(
+            'combobox-input',
+            multiselect
+              ? value.length > 0 || leadingIcon
+                ? 'indent-0'
+                : 'indent-1'
+              : 'indent-2',
```
GiantRobots, line 477: I wonder if we can turn this into not a ternary but I see the logic here being useful. Also not sure about why we're `indent-ing`

src/lib/holocene/combobox/combobox.svelte
```diff
@@ -422,133 +422,134 @@
...
-            disabled={hrefDisabled}
-            leadingIcon="external-link"
+          {/if}
+        </div>
+      {/if}
+      {#if showChevron}
+        <button
+          type="button"
+          class="hover:bg-gray-100 flex h-full items-center rounded pr-2 focus:outline-none"
+          on:click|stopPropagation={() => ($open ? closeList() : openList())}
```
GiantRobots, line 536: Why are we stopping the Propagation on this?

src/lib/holocene/toggle-button/toggle-button.svelte
```diff
@@ -1,99 +1,54 @@
...
-  data-track-intent="toggle"
-  data-track-text="*textContent*"
+<Button
   on:click
-  role="button"
-  tabindex="0"
+  class={merge(
+    (href ? $page.url.pathname.includes(base) : active) &&
+      'bg-interactive-secondary-active',
+    group && '[&:not(:last-child)]:border-r-0',
```
GiantRobots, line 45: I think this ternary can be expressed like `href  && ($page.url.pathname.includes(base) || active) && 'css'`

src/lib/components/workflow/workflows-summary-configurable-table/table-empty-state.svelte
laurakwhit: I just set it to the max width of the image before it started showing the gap.

src/lib/holocene/combobox/combobox.svelte
laurakwhit: GitHub diff might indicate otherwise, but I didn't actually touch any of the these. Just wrapped stuff in a `<div />`.

But agreed. I think we can probably just delete that `class:disabled`, it's likely a relic that isn't being used.

src/lib/holocene/toggle-button/toggle-button.svelte
```diff
@@ -1,99 +1,54 @@
...
+    className,
+  )}
+  data-track-name="toggle-button"
+  {variant}
+  href={href ? href + $page.url.search : null}
```
GiantRobots, line 50: This href statement confuses me. Could probably again abuse the other syntax like...

```javascript
href && href + $page.url.search
```

But I'm unsure why we need to add the search param of the current page?

src/lib/holocene/toggle-button/toggle-button.svelte
```diff
@@ -1,99 +1,54 @@
...
-  }
-
-  .toggle-button.disabled {
-    @apply cursor-not-allowed opacity-50;
-  }
-
-  .group:not(:last-child) {
-    @apply border-r-0;
-  }
-</style>
```
GiantRobots, line 99: Rip and tear all the `@apply`s until it's done so good 🪩

src/lib/holocene/drawer.svelte
```diff
@@ -71,9 +71,9 @@
       position === 'bottom' && 'bottom-0 left-0 right-0 border-t',
       position === 'right' && 'right-0 top-0 h-full border-l',
       dark && 'bg-black text-off-white',
+      position === 'right' && 'w-screen sm:max-w-fit',
```
GiantRobots: I like moving this into the classValues stuff, should this be in the position === right on line 72?

src/lib/holocene/combobox/combobox.svelte
```diff
@@ -422,133 +422,134 @@
...
+          data-lpignore="true"
+          data-1p-ignore="true"
+          aria-controls="{id}-listbox"
+          aria-expanded={$open}
+          aria-required={required}
+          aria-autocomplete="list"
+          on:focus|stopPropagation={openList}
+          on:input|stopPropagation={handleInput}
+          on:keydown|stopPropagation={handleInputKeydown}
+          on:click|stopPropagation={handleInputClick}
```
GiantRobots, line 493: I don't like seeing stopPropagations do we have a good reason for these? Asking because you have context you don't have to remove it just trying to figure out why we have it all over

src/lib/holocene/drawer.svelte
laurakwhit: Will consolidate ✔️

src/lib/holocene/combobox/combobox.svelte
GiantRobots: Yeah I'd probably split this into two of those at the least, the 0 or 1 is an okay place for a ternary (still not my fav) but we could do like 

```javascript
multiselect && value.length > 0 ? 'ident-0' : 'ident-1'
!multiselect && leadingIcon && 'ident-2' 

# maybe looking at ident-0 you only really need like
multiselect && value.length && 'ident-1'

# or if you prefer the .length being 
multiselect && value.length > 0 && 'ident-1'
```

src/lib/components/lines-and-dots/event-type-filter.svelte
```diff
@@ -159,6 +162,7 @@
           checked={$eventTypeFilter.some((type) => type === option.value)}
           label={option.label}
           labelHidden
+          class="mt-px"
```
GiantRobots, line 165: What do these mt-px classes do? I'm confused maybe set mt to 0? If that's the case maybe we do mt-0 instead?

src/lib/components/workflow/workflows-summary-configurable-table/table-empty-state.svelte
GiantRobots: Hmm if that's the case maybe we should be using `m-x-fit`

GiantRobots: (assuming nothing is using 100% widths)

src/lib/components/lines-and-dots/event-type-filter.svelte
laurakwhit: Aligns the checkbox center with the text. 

| Before | After |
|-|-|
|Screenshot 2025-08-29 at 12 48 16 PM|Screenshot 2025-08-29 at 12 53 56 PM|

src/lib/components/workflow/workflows-summary-configurable-table/table-empty-state.svelte
laurakwhit: Don't think that will work with the `background-image` (which is what we're using on xl screens).

src/lib/holocene/combobox/combobox.svelte
laurakwhit: I'm not 100% sure the reason for these... Definitely something we should revisit in a lot of our Holocene components.

src/lib/components/lines-and-dots/event-type-filter.svelte
rossedfort: Is this something that can be done at a lower level, eg. in the Checkbox or MenuItem component? Or is the issue isolated to this event type filter component?

src/lib/components/workflow/workflows-summary-configurable-table/table-empty-state.svelte
rossedfort: I wonder if `bg-[100%_auto]` would be better here

rossedfort: (Instead of `bg-contain` and `xl:max-w-[1050px]`)

laurakwhit: Using `bg-[100%_auto]` means the image can get stretched ⬇️ 

Screenshot 2025-09-03 at 9 21 29 AM

vs. with a max-width and `bg-contain` it keeps it's proportions and the left side with white space gets bigger instead

Screenshot 2025-09-03 at 9 23 03 AM

src/lib/components/lines-and-dots/event-type-filter.svelte
laurakwhit: I think it's likely isolated to this event type filter component. Or at least, only places where there's text with font size 14 (which means the element has a height of 20px) next to a `Checkbox` component (which has a height of 18px) and they have `align-items: flex-start;` set.

src/lib/components/workflow/workflows-summary-configurable-table/table-empty-state.svelte
rossedfort: True, I just figured it would be less brittle than a magic number

src/lib/holocene/combobox/combobox.svelte
Alex-Tideman: Probably because we don't want to submit any form on the page current viewing. Like if you on schedules create, stop propagation of form

PR #2891, opened 2025-09-03
Modify tabs list

GraceGardner:
## Description & motivation 💭 
Converts tab and tablist to runes mode and adds borderBottom as an optional prop. 

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/holocene/tab/tab-list.svelte
+++ b/src/lib/holocene/tab/tab-list.svelte
@@ -7,25 +7,19 @@
   interface Props extends HTMLAttributes<HTMLDivElement> {
     label: string;
     class?: string;
-    borderBottom?: boolean;
     children: Snippet;
   }
 
   let {
     label,
     class: className = '',
-    borderBottom = false,
     children,
     ...restProps
   }: Props = $props();
 </script>
 
 <div
-  class={merge(
-    'flex flex-row gap-4',
-    borderBottom && 'border-b border-subtle',
-    className,
-  )}
+  class={merge('flex flex-row gap-4 border-b border-subtle', className)}
   role="tablist"
   aria-label={label}
   {...restProps}
--- a/src/lib/holocene/tab/tab-list.svelte
+++ b/src/lib/holocene/tab/tab-list.svelte
@@ -1,29 +1,34 @@
 <script lang="ts">
   import type { HTMLAttributes } from 'svelte/elements';
 
+  import type { Snippet } from 'svelte';
   import { twMerge as merge } from 'tailwind-merge';
 
-  interface $$Props extends HTMLAttributes<HTMLDivElement> {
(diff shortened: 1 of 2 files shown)


src/lib/holocene/tab/tab.svelte
```diff
@@ -16,72 +17,68 @@
...
+  const tabClasses = $derived(
+    [
+      'flex cursor-pointer items-center gap-1 whitespace-nowrap border-b border-transparent text-sm font-medium outline-none focus-visible:ring-2 focus-visible:ring-primary/70',
+      isActive && 'border-brand text-brand',
+      disabled && 'cursor-not-allowed opacity-50',
+      !disabled && 'hover:text-brand',
+    ]
+      .filter(Boolean)
+      .join(' '),
+  );
```
rossedfort: I would use `twMerge` for this

src/lib/holocene/tab/tab-list.svelte
```diff
@@ -1,29 +1,34 @@
...
+    children: Snippet;
   }
 
-  export let label: string;
-  let className = '';
-  export { className as class };
+  let {
+    label,
+    class: className = '',
+    borderBottom = false,
```
Alex-Tideman: I would say this would just be a class passed in, let's avoid a specific class prop.

rossedfort: I disagree 😂 I think because this is a Holocene component we should limit the amount of customization. The nice part of making it a `borderBottom` prop is that I don't have to think about what width or color the border should be, I just know it should have a border. Otherwise, we could end up with some tabs that have larger/smaller borders, or borders of different colors.

Alex-Tideman: Cool with that as long as we remove the class prop

src/lib/holocene/tab/tab.svelte
```diff
@@ -47,41 +55,25 @@
...
   type={href ? undefined : 'button'}
   role="tab"
-  class="tab"
+  class={merge(
+    'flex cursor-pointer items-center gap-1 whitespace-nowrap border-b border-transparent text-sm font-medium outline-none focus-visible:ring-2 focus-visible:ring-primary/70',
+    isActive && 'border-brand text-brand',
+    disabled && 'cursor-not-allowed opacity-50',
+    !disabled && 'hover:text-brand',
+  )}
+  style="line-height: 2rem"
```
rossedfort: FYI the tailwind classes for line height are `leading-*`. Let's use that here. https://tailwindcss.com/docs/line-height

GraceGardner: Oh good catch! Thank you.

rossedfort (approved):

PR #2898, opened 2025-09-09
Default to system preference for light/dark/system chooser

GiantRobots:
### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/utilities/dark-mode/dark-mode.ts
+++ b/src/lib/utilities/dark-mode/dark-mode.ts
@@ -6,7 +6,7 @@ export type DarkModePreference = boolean | 'system';
 
 export const useDarkModePreference = persistStore<DarkModePreference>(
   'dark mode',
-  !!import.meta.env.VITE_DARK_MODE,
+  'system',
   true,
 );
 


src/lib/utilities/dark-mode/dark-mode.ts
```diff
@@ -6,7 +6,7 @@ export type DarkModePreference = boolean | 'system';
 
 export const useDarkModePreference = persistStore<DarkModePreference>(
   'dark mode',
-  !!import.meta.env.VITE_DARK_MODE,
```
laurakwhit, line 9: Do we need `VITE_DARK_MODE` [here](https://github.com/temporalio/ui/blob/main/src/env.d.ts#L8) anymore or can we remove it?

GiantRobots: Pretty sure we can kill it the commit message was
```* Add hidden/optional dark mode
```
it's not hidden anymore so seems the thing to nix it

PR #2901, opened 2025-09-10
rm postcss

GraceGardner:
## Description & motivation 💭 
Switch from postcss to twmerge for classes in the PaginatedTable component. 

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/holocene/table/paginated-table/index.svelte
+++ b/src/lib/holocene/table/paginated-table/index.svelte
@@ -1,18 +1,21 @@
 <script lang="ts">
   import type { HTMLAttributes } from 'svelte/elements';
 
+  import { twMerge as merge } from 'tailwind-merge';
+
   import SkeletonTable from '$lib/holocene/skeleton/table.svelte';
   import Table from '$lib/holocene/table/table.svelte';
 
   type Item = $$Generic;
 
-  interface $$Props extends HTMLAttributes<HTMLTableElement> {
+  type $$Props = HTMLAttributes<HTMLTableElement> & {
     visibleItems: Item[];
     loading?: boolean;
     updating?: boolean;
     maxHeight?: string;
     fixed?: boolean;
-  }
+    class?: string;
+  };
 
   export let visibleItems: Item[];
   export let loading = false;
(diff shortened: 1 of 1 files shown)


src/lib/holocene/table/paginated-table/index.svelte
```diff
@@ -1,34 +1,43 @@
...
 
-  interface $$Props extends HTMLAttributes<HTMLTableElement> {
+  type $$Props = HTMLAttributes<HTMLTableElement> & {
     visibleItems: Item[];
     loading?: boolean;
     updating?: boolean;
     maxHeight?: string;
     fixed?: boolean;
-  }
+    class?: string;
```
rossedfort: I think the idea is to use the `ClassValue` type here, which supports arrays, objects, etc. but will defer to @GiantRobots

GiantRobots: `ClassNameValue` from the twmerge import is what we want (twmerge supports all of the cslx goodness... except objects. So we want to use what their api allows so we don't have to do more work subbing out to cslx, then merge, etc. objects are fine to not use because you can use arrays and you basically just invert the declarations)

src/lib/holocene/table/paginated-table/index.svelte
```diff
@@ -1,34 +1,43 @@
...
 
+  import { twMerge as merge } from 'tailwind-merge';
+
   import SkeletonTable from '$lib/holocene/skeleton/table.svelte';
   import Table from '$lib/holocene/table/table.svelte';
 
   type Item = $$Generic;
 
-  interface $$Props extends HTMLAttributes<HTMLTableElement> {
+  type $$Props = HTMLAttributes<HTMLTableElement> & {
```
GiantRobots: Why the change from interface to type?

GraceGardner: I was trying to switch from legacy over to runes and then decided to go back when it went bad and there were apparently casualties along the way.

src/lib/holocene/table/paginated-table/index.svelte
```diff
@@ -12,6 +19,7 @@
     updating?: boolean;
     maxHeight?: string;
     fixed?: boolean;
+    class?: ClassValue;
```
GiantRobots: Use `ClassNameValue` from twmerge

src/lib/holocene/table/paginated-table/index.svelte
```diff
@@ -20,15 +28,21 @@
...
+    ? tableContainer.offsetTop + 32
     : 0;
 </script>
 
 <div
-  class="paginated-table-wrapper"
+  class={merge(
+    'surface-primary min-h-[154px] grow overflow-auto border border-subtle',
+    className,
+  )}
```
GiantRobots, line 40: Just use the regular merge with the different type on classes and we gucci

package.json
```diff
@@ -79,19 +79,20 @@
   "dependencies": {
     "@codemirror/autocomplete": "^6.17.0",
     "@codemirror/commands": "^6.6.0",
+    "@codemirror/lang-go": "^6.0.1",
+    "@codemirror/lang-java": "^6.0.2",
     "@codemirror/lang-json": "^6.0.1",
+    "@codemirror/lang-php": "^6.0.2",
```
GiantRobots, line 85: Probably don't need these codemirror addons?

rossedfort: Looks like they just got alphabetized

GraceGardner: Yeah, this happened when I ran install for some reason. But Those aren't new additions.

PR #2902, opened 2025-09-10
Fix down arrow on Combobox

laurakwhit:
## Description & motivation 💭 
Menu options were being reset after entering a filter value in the `Combobox` component and then hitting the down arrow to navigate to a menu option.

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

In the `Combobox` component > enter a filter value > hit the down arrow
 - [ ] menu options should still reflect the filter value
 - [ ] user should be able to navigate to the menu options via the up/down arrows

### Issue(s) closed 
`DT-3290`

### DIFF
--- a/src/lib/holocene/combobox/combobox.svelte
+++ b/src/lib/holocene/combobox/combobox.svelte
@@ -221,6 +221,7 @@
   }
 
   const openList = () => {
+    if ($open) return;
     $open = true;
     filterValue = '';
     inputElement.focus();
@@ -364,6 +365,7 @@
         closeList();
         break;
       case 'ArrowDown':
+        openList();
         focusFirstOption();
         break;
       case 'Enter':


src/lib/holocene/combobox/combobox.svelte
```diff
@@ -364,7 +364,6 @@
         closeList();
         break;
       case 'ArrowDown':
-        openList();
```
laurakwhit: With the addition of [this line of code](https://github.com/temporalio/ui/pull/2431/files#diff-a3db663fd01dcc941f639c835078a17e5ca9d85a8c6be95c615f59355555d827R171) we’re clearing the `filterValue` on `openList()`. 

~I think it's safe to remove `openList()` from `ArrowDown` here. All it will change is that a user can't hit `ESC` and then `ArrowDown` to open the menu again.~ Actually we're calling `openList()` [on:focus](https://github.com/temporalio/ui/blob/main/src/lib/holocene/combobox/combobox.svelte#L492) as well so if the user goes back to the input after navigating to a menu option it will also clear the `filterValue`. Adding a [check for if the menu is already open](https://github.com/temporalio/ui/pull/2902/commits/9db63417a786fa30f82047412148ae438d964ca4) instead.

PR #2920, opened 2025-09-18
remove monaco in favor of codemirror

rossedfort:
## Description & motivation 💭 

Remove monaco-editor and use codemirror instead for nexus desccription

https://github.com/user-attachments/assets/1fbe0a48-992f-483b-9ab6-7678abd44dce

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

- Test description is still saved when creating a Nexus Endpoint
- Test summary and details are still saved when starting a Workflow


src/lib/holocene/markdown-editor/markdown-editor.svelte
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  let { content = $bindable('') }: Props = $props();
+  let editorElement: HTMLElement | null = $state(null);
+  let editorView: EditorView | null = $state(null);
+  let activeTab: 'edit' | 'preview' = $state('edit');
+
+  const createEditorView = () => {
+    return new EditorView({
+      parent: editorElement,
```
GiantRobots: You could make this function pure if you passed editorElement in instead of depending on the state being declared already.

src/lib/holocene/markdown-editor/markdown-editor.svelte
```diff
@@ -0,0 +1,80 @@
...
+      state: EditorState.create({
+        doc: content,
```
GiantRobots: Same here but unsure if EditorState updates content, but you could purify the func by passing this in as well, the rest seem to be options

src/lib/holocene/markdown-editor/markdown-editor.svelte
```diff
@@ -0,0 +1,80 @@
...
+          lineNumbers(),
+          getEditorThemeWithLineNumbers($useDarkMode, false),
+        ],
+      }),
+      dispatch(transaction) {
+        editorView.update([transaction]);
+        if (transaction.docChanged) {
+          content = editorView.state.doc.toString();
+        }
+      },
```
GiantRobots: Lol I got to here and it seems less feasible but I'm curious when this function ends up getting called in the lifecycle of codemirror. 

Probably ignore the comments above.

src/lib/holocene/markdown-editor/markdown-editor.svelte
```diff
@@ -0,0 +1,80 @@
...
+  const setActiveTab = async (tab: 'edit' | 'preview') => {
+    activeTab = tab;
+
+    if (activeTab === 'edit') {
+      // because we're mounting/unmounting the editor element, we have to wait until
+      // the element is re-mounted before re-creating the editorView
+      await tick();
+      editorView = createEditorView();
+    }
+  };
```
GiantRobots, line 18: If you made an editor component and used it in the if statement wouldn't this happen automagically without you having to manage and use await tick? Also the onMount would naturally move to the new editor component.

src/lib/holocene/markdown-editor/markdown-preview.svelte
```diff
@@ -0,0 +1,65 @@
...
+  const previewTheme = $derived($useDarkMode ? 'dark' : 'light');
+  const previewPath = $derived(
+    `/render?content=${encodeURIComponent(templatedContent)}&theme=${previewTheme}&overrideTheme=${overrideTheme}`,
+  );
+</script>
+
+<section class={twMerge('h-full w-full', className)}>
+  <iframe
+    bind:this={iframe}
+    onload={resizeIframe}
```
GiantRobots: Do we care if the window size is changed  and need to update this? Should we use a resizeObserver (or whatever the window version is) on this to keep things properly laid out?

src/lib/holocene/markdown-editor/markdown-editor.svelte
```diff
@@ -0,0 +1,80 @@
...
+        extensions: [
+          markdown(),
+          lineNumbers(),
+          getEditorThemeWithLineNumbers($useDarkMode, false),
+        ],
+      }),
+      dispatch(transaction) {
+        editorView.update([transaction]);
+        if (transaction.docChanged) {
+          content = editorView.state.doc.toString();
```
andrewzamojc: Comment: Neat using bind for content vs the code-block's onchange callback.

src/lib/holocene/markdown-editor/markdown-editor.svelte
```diff
@@ -0,0 +1,80 @@
...
+
+  const createEditorView = () => {
+    return new EditorView({
+      parent: editorElement,
+      state: EditorState.create({
+        doc: content,
+        extensions: [
+          markdown(),
+          lineNumbers(),
+          getEditorThemeWithLineNumbers($useDarkMode, false),
```
andrewzamojc: Possible change: Since this extension is dynamic based on dark mode value, you may need a code-mirror `compartment` to have it update. I'm basing this off the code-block component's code, but haven't tested your dark mode yet.

compartment: https://github.com/temporalio/ui/blob/73c8389410e6ba9abb1f959ba27ccbdddf66cc29/src/lib/holocene/code-block.svelte#L169

keeping them up to date:
https://github.com/temporalio/ui/blob/73c8389410e6ba9abb1f959ba27ccbdddf66cc29/src/lib/holocene/code-block.svelte#L181

src/lib/vendor/codemirror/custom-extensions.ts
```diff
@@ -26,34 +26,70 @@ export type EditorLanguage =
...
+
+const headerStyles = (header: boolean) =>
+  header
+    ? {}
+    : {
+        borderWidth: '1px',
+        borderColor: css('--color-border-subtle'),
+      };
+
+export const getEditorTheme = (isDark: boolean, header) =>
```
andrewzamojc, line 59: Question: Code-block uses this function. Is this change ok for that component too?

src/lib/holocene/markdown-editor/markdown-preview.svelte
```diff
@@ -0,0 +1,65 @@
...
+  };
+
+  const templatedContent = $derived(replaceTemplate(content));
+  const previewTheme = $derived($useDarkMode ? 'dark' : 'light');
+  const previewPath = $derived(
+    `/render?content=${encodeURIComponent(templatedContent)}&theme=${previewTheme}&overrideTheme=${overrideTheme}`,
+  );
+</script>
+
+<section class={twMerge('h-full w-full', className)}>
```
andrewzamojc: Suggestion: The preview could use a border imo, so it sort of looks like the wrapping component is still there, just switching what's inside. If the preview panel has a max height you could even make it a tab stop, so up/down arrows scroll it.

src/lib/holocene/markdown-editor/markdown-preview.svelte
```diff
@@ -0,0 +1,65 @@
...
+  <iframe
+    bind:this={iframe}
+    onload={resizeIframe}
+    title="output"
+    class="w-full"
+    src={previewPath}
+  ></iframe>
+</section>
```
andrewzamojc: Suggestion: Since these are holocene components, a story in storybook would be nice for quick preview, e.g. to test what I see in this PR.

andrewzamojc (commented):
The implementation is looking good. I left a couple technical suggestions for the component, and a request for a storybook story so I can use the preview link to try the component :)

src/lib/vendor/codemirror/custom-extensions.ts
andrewzamojc: I tested code-block in storybook and don't see anything weird, so I assume its aok :)

src/lib/holocene/markdown-editor/markdown-editor.svelte
rossedfort: So I'm not sure we even need to be passing a `darkMode` parameter to `getEditorTheme` or `getEditorThemeWithLineNumbers`. Looking at the docs for `EditorView.theme`, it says `dark` will override styles from the "base themes" which I don't believe we're using

https://codemirror.net/docs/ref/#view.EditorView^theme

https://github.com/user-attachments/assets/c28b9649-4666-4331-a9bb-61e9a4bbe93e

rossedfort: And because those fn's use css variables, the variables update with the mode change

src/lib/vendor/codemirror/custom-extensions.ts
rossedfort: Yes, I added a new function to build the theme for the markdown editor (with line numbers), and both theme functions inherit from that base theme. There should be no net change here.

src/lib/holocene/markdown-editor/markdown-preview.svelte
rossedfort: Ah, good catch. I added the border but must have removed it at some point. I only added the border in the editor context
Screenshot 2025-09-18 at 1 00 28 PM
but not when the preview is used separately, such as on the User Metadata tab of the workflow details page.
Screenshot 2025-09-18 at 12 59 24 PM

src/lib/holocene/markdown-editor/markdown-preview.svelte
rossedfort: Good question, this was existing code I copied over from [src/lib/holocene/monaco/markdown.svelte](https://github.com/temporalio/ui/pull/2920/files#diff-224b769364bab6f5d9cc0529ec36ffb2d0c32a7e38bb7212e39759cb719c6276), so I didn't want to touch it.

rossedfort: It seems to work fine

https://github.com/user-attachments/assets/cd4166f8-8d46-4e55-aa7a-d557a681cf4f

src/lib/holocene/markdown-editor/markdown-preview.svelte
rossedfort: Good call, I added a story, but we can't test the preview functionality because our sveltekit app hosts the `/render` endpoint that renders the markdown, and we don't run the app during storybook.

src/lib/holocene/markdown-editor/markdown-editor.svelte
andrewzamojc: Sweet. Seems like it's working. I think your line of thinking makes sense re css variables.

src/lib/holocene/markdown-editor/markdown-preview.svelte
andrewzamojc: Thanks! Too bad about preview, but ya makes sense if render is in the app.

andrewzamojc (approved):
LGTM

PR #2931, opened 2025-09-29
Saved Workflow Query Views

Alex-Tideman:
## Description & motivation 💭 
This PR provides users with preset "system" queries and custom "user" queries that are saved to localStorage. User's can quickly navigate to queries they have previously run without needing to rebuild the queries from scratch, saving time and giving quicker results. Users can also copy/edit/discard the saved queries and also share them with other users. The Saved Query left nav can be expanded and collapsed based on the user preference. There is a limit of 20 saved queries at this time, not a system limitation but to prevent user frustration should they clear localStorage and wipe the saved queries. The refresh rate for workflow counts was significantly reduced from 5 seconds to 1 minute, which helps with the load on visibility store.

Also included is a more streamlines UI for applying filters. Each filter is it's own dropdown menu that makes it easier to edit on the fly.

Screenshot 2025-09-29 at 2 45 33 PM
Screenshot 2025-09-29 at 2 45 48 PM
Screenshot 2025-09-29 at 2 46 02 PM
Screenshot 2025-09-29 at 2 46 15 PM
Screenshot 2025-09-29 at 2 46 27 PM
Screenshot 2025-09-29 at 2 47 19 PM

### How was this tested 👻 

- [x] Manual testing
- [x] E2E tests added
- [x] Unit tests added

### DIFF
--- a/src/lib/components/workflow/filter-bar/dropdown-filter-chip.svelte
+++ b/src/lib/components/workflow/filter-bar/dropdown-filter-chip.svelte
@@ -452,6 +452,7 @@
               type="number"
               placeholder="Enter number..."
               disabled={isNullFilter}
+              step="any"
               bind:value={localFilter.value}
             />
           </div>


bilal-karim:
Nit: Can we reduce this gap to 2?

Screenshot 2025-09-30 at 12 43 39 PM

---

Nit: Can we reduce this gap to 1.5 or increase the padding of the filter bar container to 2? Both should be the same so the spacing around and between elements is even.

Screenshot 2025-09-30 at 12 45 28 PM

---

Nit: The tooltips in this UX have rounded corners, but the tooltips on left vertical nav do not have rounded corners.

Screenshot 2025-09-30 at 12 47 37 PM

Screenshot 2025-09-30 at 12 47 44 PM

src/lib/components/search-attribute-filter/search-attribute-menu.svelte
```diff
@@ -30,6 +30,9 @@
 
   export let filters: SearchAttributeFilter[];
   export let options: SearchAttributeOption[];
+  export let showOptions = workflowRoutePattern.match(
+    window?.location?.pathname,
```
rossedfort: Does this need to be reactive? If so (or maybe even if not) we could use `$derived` and `('$app/state').page.url.pathname`

src/lib/components/workflow/filter-bar/dropdown-filter-chip.svelte
rossedfort, line 1: I might consider creating a component for each type of filter value so this component isn't so large.

src/lib/components/workflow/filter-bar/search.svelte
```diff
@@ -0,0 +1,19 @@
...
+  };
+  let { query = $bindable() }: Props = $props();
+</script>
+
+<Input
+  id="power-filter-query"
+  type="search"
+  label="Search"
+  labelHidden
+  placeholder="failed workflows in the last 15 minutes"
```
rossedfort: Is this a valid query? Is this component used anywhere?

src/lib/components/workflow/workflows-summary-configurable-table.svelte
```diff
@@ -80,6 +80,7 @@
     nextButtonLabel={translate('common.next')}
     previousButtonLabel={translate('common.previous')}
     emptyStateMessage={translate('workflows.empty-state-title')}
+    maxHeight="73dvh"
```
rossedfort: Should we use the css var here set in src/app.css?

src/lib/holocene/input/input.svelte
```diff
@@ -70,6 +71,7 @@
 
   let className = '';
   export { className as class };
+  export let inputClass = '';
```
rossedfort: We're adding this class to the input container, should we name it `inputContainerClass` for clarity?

src/lib/holocene/modal.svelte
```diff
@@ -125,19 +125,25 @@
...
-          variant={confirmType}
-          {loading}
-          disabled={confirmDisabled || loading}
-          data-testid="confirm-modal-button"
-          type="submit">{confirmText}</Button
+
+    <div class="flex items-center justify-between p-6">
+      <slot name="footer">
+        <div></div>
+      </slot>
```
rossedfort, line 132: Why the empty div?

src/lib/holocene/tab-buttons/tab-buttons.svelte
```diff
@@ -1,8 +1,11 @@
...
+  import { setContext, type Snippet } from 'svelte';
   setContext<boolean>('group', true);
+
+  let { children, class: className }: { children?: Snippet; class?: string } =
+    $props();
 </script>
 
-<div class="flex">
-  <slot />
+<div class="flex {className}" role="tablist">
```
rossedfort: We should use `twMerge` here

src/lib/pages/saved-query-views.svelte
```diff
@@ -0,0 +1,499 @@
...
+{#if showTooltip && tooltipText}
+  <div
+    use:portal
+    class="pointer-events-none z-[9999] inline-block select-none rounded-md bg-slate-800 p-2 text-xs text-slate-50 opacity-95"
+    style={`position: fixed; top: ${tooltipY}px; left: ${tooltipX}px; transform: translateY(-50%); max-width: 280px;`}
+    role="tooltip"
+  >
+    {tooltipText}
+  </div>
+{/if}
```
rossedfort, line 499: Why not use our existing tooltip component here? If this is better, should we refactor tooltip to use this code?

src/lib/holocene/tab-buttons/tab-buttons.svelte
Alex-Tideman: 👍 done

src/lib/holocene/modal.svelte
Alex-Tideman: This is so the confirm/cancel buttons stay to the right with the justify-between

src/lib/holocene/input/input.svelte
Alex-Tideman: 👍 done

src/lib/components/workflow/workflows-summary-configurable-table.svelte
Alex-Tideman: 👍 done

src/lib/components/workflow/filter-bar/search.svelte
Alex-Tideman: Deleted this file, it was for prototyping the natural language query but it's not used anywhere.

src/lib/components/search-attribute-filter/search-attribute-menu.svelte
Alex-Tideman: I removed it, we don't need this anymore since I'll only show the child workflows option when it's the Parent Workflows query

src/lib/components/workflow/filter-bar/dropdown-filter-chip.svelte
Alex-Tideman: Thought about it, and that's how it was before. Ultimately makes it a lot harder to grok what's going on when you have jump around that many files. They are all pretty closely tied together and I'd rather sacrifice larger file for readability.

src/lib/pages/saved-query-views.svelte
Alex-Tideman: Very fair question. I decided not to use our Tooltip for a couple of reasons. 
1. I wanted the parent container to be overflow-auto, and using our current Tooltip that would cut it off and not show it.
2. I am wrapping this tooltip around a large container of things, button + more buttons in a parent div. When using our Tooltip, it centers and styles the element in ways I didn't want.

So I think of it this way:
1. Current Tooltip is used for small, non-nested elements like some text or icon or a single Button.
2. This is a special tooltip that pops out of the element and needs to adjust it's position dynamically. Could it be it's own component? Definitely. Not sure yet we need to do that, but we can if we need to. 

I think we could make the Tooltip work for both situations, but I tried it and it got really complex and felt it was safer to make this special one just baked into here.

rossedfort (approved):

src/lib/components/workflow/filter-bar/dropdown-filter-chip.svelte
```diff
@@ -0,0 +1,553 @@
...
+        {:else if isNumberFilter(localFilter)}
+          <div class="space-y-3">
+            {@render conditionalButtons(numberConditionalOptions)}
+            <Input
+              id={`${controlsId}-number`}
+              label="Value"
+              type="number"
+              placeholder="Enter number..."
+              disabled={isNullFilter}
+              bind:value={localFilter.value}
```
laurakwhit, line 456: Should this have a min value? And do we want to allow decimals as well?

```suggestion
              bind:value={localFilter.value}
              min={0}
```

Alex-Tideman: Here's the docs from https://docs.temporal.io/search-attribute

Double is backed up by scaled_float Elasticsearch type with scale factor 10000 (4 decimal digits).
Datetime is backed up by date type with milliseconds precision in Elasticsearch 6 and date_nanos type with nanoseconds precision in Elasticsearch 7.
Int is 64-bit integer (long Elasticsearch type).

So I think we want to allow negatives and floats.

Alex-Tideman: Added steps="any" to support decimals

Screenshot 2025-10-08 at 8 08 13 AM
Screenshot 2025-10-08 at 8 08 27 AM
Screenshot 2025-10-08 at 8 14 02 AM

PR #2936, opened 2025-10-06
move editdisabled into util

GraceGardner:
## Description & motivation 💭 
Turn editDisabled into a util function. 

## Testing 🧪 
Had Claude create some unit tests 

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
DT-3050

### DIFF
--- a/src/routes/(app)/nexus/+page.svelte
+++ b/src/routes/(app)/nexus/+page.svelte
@@ -9,8 +9,7 @@
   export let data: PageData;
 
   $: ({ endpoints } = data);
-  $: createDisabled = $page.data.settings?.disableWriteActions ?? false;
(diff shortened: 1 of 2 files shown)


src/lib/pages/schedule-view.svelte
```diff
@@ -115,8 +115,7 @@
     ];
 
   let coreUser = coreUserStore();
-  $: editDisabled =
```
laurakwhit, line 118: I think this is fine as is. `DT-3050` looks like it's referring to some logic in Cloud (not OSS). Left a comment on the task and I think we can close this PR.

PR #2961, opened 2025-10-20
Add catchup window to schedule details

laurakwhit:
## Description & motivation 💭 
Adds `catchupWindow`, `pauseOnFailure` and  `keepOriginalWorkflowId` under "Advanced Settings" on the schedule details page.

Screenshot 2025-10-21 at 11 45 25 AM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
`DT-3391`

### Any docs updates needed? 

[DT-3391]: https://temporalio.atlassian.net/browse/DT-3391?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/components/schedule/schedule-advanced-settings.svelte
+++ b/src/lib/components/schedule/schedule-advanced-settings.svelte
@@ -46,21 +46,29 @@
     <li>
       {translate('schedules.catchup-window')}
       <span
-        >{policies?.catchupWindow !== undefined
+        >{policies?.catchupWindow != null
           ? formatDuration(policies.catchupWindow.toString())
           : translate('common.none')}
       </span>
     </li>
-    {#if policies?.pauseOnFailure}
+    {#if policies?.pauseOnFailure != null}
       <li>
         {translate('schedules.pause-on-failure')}
-        <span>{translate('common.enabled')}</span>
+        <span
+          >{policies.pauseOnFailure
+            ? translate('common.true')
+            : translate('common.false')}</span
+        >
       </li>
     {/if}
-    {#if policies?.keepOriginalWorkflowId}
+    {#if policies?.keepOriginalWorkflowId != null}
       <li>
         {translate('schedules.keep-original-workflow-id')}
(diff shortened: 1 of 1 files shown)


src/lib/components/schedule/schedule-advanced-settings.svelte
```diff
@@ -42,6 +43,26 @@
...
+        {translate('schedules.pause-on-failure')}
+        <span>{translate('common.enabled')}</span>
+      </li>
+    {/if}
+    {#if policies?.keepOriginalWorkflowId}
+      <li>
+        {translate('schedules.keep-original-workflow-id')}
+        <span>{translate('common.enabled')}</span>
+      </li>
+    {/if}
```
rossedfort: Should we always render these rows and instead show "Disabled" if the value is falsey?

laurakwhit: Based on the description ⬇️  is seems like a value of false is similar to having nothing set at all, so I just left it off entirely. 

Screenshot 2025-10-21 at 10 35 57 AM


Maybe we should not show if they are `undefined` but show `Disabled` if they are set to `false` 🤔

laurakwhit: cc @Alex-Tideman for thoughts

Alex-Tideman: I think it might be good to show false here and don't show if null. If set as false, it seems like an intentional thing

laurakwhit: [Updated](https://github.com/temporalio/ui/pull/2961/commits/053c56186ace0ebcdfb9a6a6afb5db905ed8eec1) ✅

PR #2966, opened 2025-10-21
Refresh token POC

Alex-Tideman:
### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/server/server/auth/auth.go
+++ b/server/server/auth/auth.go
@@ -100,7 +100,27 @@ func SetUser(c echo.Context, user *User) error {
 	if rt := user.OAuth2Token.RefreshToken; rt != "" {
 		log.Println("[Auth] Setting refresh token cookie")
 
-		refreshMaxAge := int((30 * 24 * time.Hour).Seconds())
+		// Calculate MaxAge from OAuth2 token expiry.
+		// IMPORTANT: When a refresh token is issued, oauth2.Token.Expiry typically
+		// reflects the refresh token's lifetime, not the access token's lifetime.
+		// This is standard behavior in OIDC flows with offline_access scope.
+		// See: https://pkg.go.dev/golang.org/x/oauth2#Token
+		var refreshMaxAge int
+		if user.OAuth2Token.Expiry.IsZero() {
+			// Fallback: if IdP doesn't provide expiry, use 7 days
+			refreshMaxAge = int((7 * 24 * time.Hour).Seconds())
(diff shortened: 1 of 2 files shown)


src/lib/utilities/request-from-api.ts
```diff
@@ -93,15 +94,35 @@ export const requestFromAPI = async <T>(
...
+      const refreshed = await refreshTokens();
+      if (refreshed) {
+        options = withSecurityOptions(init.options, isBrowser);
+        if (!endpoint.endsWith('api/v1/settings')) {
+          options = await withAuth(options, isBrowser);
+        }
+        response = await makeRequest();
+        status = response.status;
+        statusText = response.statusText;
+      }
```
rossnelson, line 125: In the past I've implemented a concept called "interceptors" almost like middleware or maybe more similar to hooks. Basically an abstraction of what this is doing.

rossnelson:
@claude review this PR

server/server/auth/auth.go
```diff
@@ -82,6 +97,43 @@ func SetUser(c echo.Context, user *User) error {
...
+func validateJWT(ctx context.Context, tokenString string) error {
+	if tokenString == "" {
+		log.Println("[JWT Validation] Token is empty, skipping validation")
+		return nil
+	}
+
+	if tokenVerifier == nil {
+		log.Println("[JWT Validation] No verifier configured, skipping validation")
+		return nil
+	}
```
asmclean, line 196: These types of cases seem dangerous to me.  Return values for successful token verification should be explicitly different from when a token value is missing or configuration is improper/incomplete, IMO.

asmclean (commented):
I haven't had a chance to go through all of this in detail, but the biggest thing that really jumps out at me is that it appears we're just base 64 encoding JSON data and stuffing that into cookies as part of this stateless BFF-ish pattern, which seems especially dicey with refresh tokens involved.  This is considered a `SHOULD` in this current draft just for access tokens: https://datatracker.ietf.org/doc/html/draft-ietf-oauth-browser-based-apps#section-6.1.3.2-6  (The recommendations for cookies in that same section, e.g. using `__Host`, are probably worth consideration, too.)  Doing this would likely be a performance hit, but it would remove the need to store and continuously validate the ID token.

rossnelson (approved):

PR #2975, opened 2025-10-22
remove scroll up/down buttons in main content

rossedfort:
## Description & motivation 💭 
In preparation for launching userflow in the cloud UI, the "scroll to top"/"scroll to bottom" buttons needed to be moved to the left side of the main content container, as userflow's "Resource Center" will be placed in the bottom right. We opted instead to just remove the buttons because they are not heavily used

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/holocene/scroll-to-container.svelte
```diff
@@ -1,17 +1,40 @@
...
+    scrollToBottomAriaLabel,
+  }: Props = $props();
 </script>
 
-<div id="scroll-container" class={$$props.class}>
+<div
+  id="scroll-container"
+  class={twMerge(
+    'fixed bottom-5 z-50 hidden w-auto gap-2 transition-all md:flex',
+    $navOpen ? 'left-[196px]' : 'left-20',
```
andrewzamojc: Question: Is there any way to anchor this to the content area, rather than fixed and having to shift it for the nav by a magic number? Like, if the content area is position relative, then this could be absolute, and it would stick to the bottom left of the content automatically.

No need to block on this thought. Just curious what you think :)

PR #2977, opened 2025-10-23
Add gradient for completed Activities with retries

Alex-Tideman:
## Description & motivation 💭 
Replace the green/red stripes with gradient

**Current**
Screenshot 2025-10-23 at 4 47 08 PM

**Update**
Screenshot 2025-10-23 at 4 41 10 PM
Screenshot 2025-10-23 at 4 41 23 PM
Screenshot 2025-10-23 at 4 43 00 PM
Screenshot 2025-10-23 at 4 43 11 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Draft Checklist 

- [x] Validate where all the green/red Timeline updates will go


src/lib/components/workflow/relationships/workflow-family-node-tree.svelte
```diff
@@ -456,45 +456,3 @@
...
-  }
-
-  .Canceled {
-    @apply fill-green-200;
-  }
-
-  .ContinuedAsNew {
-    @apply fill-green-200;
-  }
-</style>
```
GiantRobots, line 500: Rip and tear 🤘🏼

src/lib/components/lines-and-dots/svg/line.svelte
```diff
@@ -14,33 +14,35 @@
...
-  stroke-dasharray={pending ? '3' : strokeDasharray}
-  x1={Math.max(0, x1)}
-  x2={Math.max(0, x2)}
-  {y1}
-  {y2}
-/>
+    <div
+      class="h-full w-full"
+      style="background: linear-gradient(255deg, #1FF1A5 0%, #F55 100%);"
+    ></div>
```
GiantRobots, line 30: Can you leave a brief synopsis of why this instead of a `<linearGradient>` stroke or sumpin

Alex-Tideman: <linearGradient> defs aren't being respected on the line, I think because of how deeply nested the timeline elements are and the fact that I used ```<g>``` groups and ```<foreignObjects>```. Basically the definition of linearGradient isn't applied because it's overridden somewhere in the svg. Tried moving it to a bunch of different locations, but nothing worked. So... opted for simplicity with a little hit of a DOM node.

PR #2978, opened 2025-10-23
Set day as max unit in formatDate and getDuration

laurakwhit:
## Description & motivation 💭 
Updates `formatDate` (for `relative` time) and `getDuration` to use days as the maximum unit unless `flexibleUnits` is true.

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

### Issue(s) closed 
`DT-3349`

### DIFF
--- a/src/lib/utilities/format-time.test.ts
+++ b/src/lib/utilities/format-time.test.ts
@@ -97,12 +97,16 @@ describe('getDuration', () => {
     expect(distance).toBe('11 days, 5 hours, 5 minutes, 45 seconds');
     expect(abbvDistancer).toBe('11d 5h 5m 45s');
   });
-  it('should get months/days/hours/minutes/seconds duration of a start and end date', () => {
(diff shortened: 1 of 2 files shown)


src/lib/utilities/format-date.ts
```diff
@@ -52,9 +47,11 @@ export function formatDate(
 
     if (timeFormat === 'local') {
       if (relative)
-        return relativeStrict
-          ? formatDistanceToNowStrict(parsed) + ` ${relativeLabel}`
-          : formatDistanceToNow(parsed) + ` ${relativeLabel}`;
+        return (
+          formatDistanceToNowStrict(parsed, {
```
laurakwhit, line 51: Removed `formatDistanceToNow` here as it feels like we should probably just be using `formatDistanceToNowStrict`. 

Unless there's some context as to why we were previously defaulting to `formatDistanceToNow`?

PR #2980, opened 2025-10-24
Allow custom keyword SAs to be quick filterable

Alex-Tideman:
## Description & motivation 💭 
Allow custom Keyword Search Attributes to be filterable in workflow list row

https://github.com/user-attachments/assets/04de1bd2-16fc-4f2b-bbf9-b2aa3f53e109

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/workflow/workflows-summary-configurable-table/table-body-cell.svelte
+++ b/src/lib/components/workflow/workflows-summary-configurable-table/table-body-cell.svelte
@@ -157,6 +157,7 @@
         {filterOrCopyButtonsVisible}
         attribute={label}
         value={content}
+        type={$customSearchAttributes[label]}
       />
     {/if}
   </td>
--- a/src/lib/components/workflow/workflows-summary-configurable-table/filterable-table-cell.svelte
+++ b/src/lib/components/workflow/workflows-summary-configurable-table/filterable-table-cell.svelte
@@ -6,7 +6,10 @@
   import { translate } from '$lib/i18n/translate';
   import type { SearchAttributeFilter } from '$lib/models/search-attribute-filters';
   import { workflowFilters } from '$lib/stores/filters';
-  import { SEARCH_ATTRIBUTE_TYPE } from '$lib/types/workflows';
+  import {
+    SEARCH_ATTRIBUTE_TYPE,
+    type SearchAttributeType,
+  } from '$lib/types/workflows';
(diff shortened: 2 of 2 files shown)


src/lib/components/workflow/workflows-summary-configurable-table/table-body-cell.svelte
```diff
@@ -147,6 +147,23 @@
       />
     {/if}
   </td>
+{:else if isCustomSearchAttribute(label) && workflowIncludesSearchAttribute(workflow, label) && $customSearchAttributes[label] === SEARCH_ATTRIBUTE_TYPE.KEYWORD && typeof workflow.searchAttributes.indexedFields[label] === 'string'}
```
laurakwhit: What do you think about just adding an `isCustomKeywordOrTextAttribute` to the `{#if filterableLabels.includes(label)}` block? Something like this ⬇️ 

```js
  let isCustomKeywordOrTextAttribute = $derived(
    isCustomSearchAttribute(label) &&
      ($customSearchAttributes[label] === SEARCH_ATTRIBUTE_TYPE.KEYWORD ||
        $customSearchAttributes[label] === SEARCH_ATTRIBUTE_TYPE.TEXT) &&
      typeof workflow.searchAttributes.indexedFields?.[label] === 'string',
  );
  ...
  {#if filterableLabels.includes(label) || isCustomKeywordOrTextAttribute}
   ...
  {:else if label === 'Versioning Behavior'}
   ...
  {:else}
       <FilterableTableCell
         {filterOrCopyButtonsVisible}
         attribute={label}
         value={workflow.searchAttributes.indexedFields[label]}
       />
   {/if}
```

That way we don't have to make updates in two places if we make changes to the `<td>` in the future. Also, we can  check for `SEARCH_ATTRIBUTE_TYPE.TEXT` as well.

Alex-Tideman: Yup like that better. We do only set Keyword as the type when filter is clicked, but maybe can look at the attribute to pick what it should be

laurakwhit: Ah yeah, good catch. If we can look at the attribute for the type that would be ideal. ~I have some of these changes locally, want me to just push them up?~  Pushed up some changes.

PR #2984, opened 2025-10-27
feat(search-attributes): add disableTypeForExisting prop

rossnelson:
## Summary

Adds an optional `disableTypeForExisting` prop to the SearchAttributesForm component that disables type selection for existing search attributes while keeping new attributes editable.

Also removes the `taintedMessage` prop from the superforms config. https://superforms.rocks/concepts/tainted#tainted-fields

## Changes

- Add `disableTypeForExisting` optional boolean prop to SearchAttributesForm and SearchAttributesFormContent
- Track initial attribute names using a derived Set
- Conditionally disable type Select component for attributes that exist in initialAttributes
- New attributes added via "Add Attribute" button remain fully editable
- Also fixes tainted badge display in codec server and search attributes forms

## Usage

```svelte
<SearchAttributesForm
  disableTypeForExisting={true}
  initialAttributesPromise={attributesPromise}
  onSave={handleSave}
  getSupportedTypes={getTypes}
/>
```

When `disableTypeForExisting` is true, the type dropdown will be disabled for any attribute whose name matches an attribute in `initialAttributes`.

### DIFF
--- a/src/lib/components/search-attributes-form/search-attributes-form-content.svelte
+++ b/src/lib/components/search-attributes-form/search-attributes-form-content.svelte
@@ -7,15 +7,14 @@
   import TaintedBadge from '$lib/components/form/tainted-badge.svelte';
   import Button from '$lib/holocene/button.svelte';
   import Card from '$lib/holocene/card.svelte';
-  import Input from '$lib/holocene/input/input.svelte';
-  import Option from '$lib/holocene/select/option.svelte';
-  import Select from '$lib/holocene/select/select.svelte';
   import { translate } from '$lib/i18n/translate';
 
   import type {
     SearchAttributeDefinition,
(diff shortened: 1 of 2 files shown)


src/lib/components/search-attributes-form/search-attributes-form-content.svelte
```diff
@@ -178,7 +184,9 @@
                   index: index + 1,
                 })}
                 labelHidden
-                disabled={$submitting}
+                disabled={$submitting ||
+                  (disableTypeForExisting &&
+                    initialAttributeNames.has(attribute.name))}
```
rossedfort: nit: this is maybe just complex enough to pull out into a `$derived`

rossedfort (approved):

PR #2986, opened 2025-10-28
Fix handleFocusOut on maximize to allow highlighting text to copy

Alex-Tideman:
## Description & motivation 💭 
Users want to be able to highlight and copy text when maximized. handleFocusOut would minimized the screen whenever a user clicked into the maximized CodeBlock.

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/holocene/maximizable/maximizable.svelte
+++ b/src/lib/holocene/maximizable/maximizable.svelte
@@ -29,6 +29,17 @@
   const handleClick = () => {
     maximized = !maximized;
   };
+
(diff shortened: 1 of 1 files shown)


andrewzamojc (approved):
Your new logic works when I test in the app! In Storybook when tabbing away, it ends up on a non-Element so the logic fails, but I'm ok assuming that's because of Storybook's frames and such. Very unlikely to be a problem IRL.

PR #2992, opened 2025-10-30
fix(search-attributes): use $derived for reactive props pattern

rossnelson:
## Summary
Fix Svelte 5 reactivity warnings in search attributes form by using `$derived` for prop composition instead of direct bindings. Follows the same pattern as tags implementation in cloud-ui.

## Changes
- Replace direct prop bindings with `$derived` props objects  
- Create `inputProps` and `selectProps` with inline callbacks
- Spread derived props into Input and Select components
- Maintain same external interface (onNameChange/onTypeChange callbacks)

## Technical Details
The issue was that child components were attempting to bind directly to props (`bind:value={attribute.name}`), which doesn't work in Svelte 5 and causes `binding_property_non_reactive` warnings.

The solution uses `$derived` to compose props objects that include both values and event handlers, following the established pattern from `tags/manager/tag-row.svelte` in cloud-ui.

## Test Plan
- [x] Lint and type checks pass
- [x] Verify form tracks changes correctly in Storybook
- [ ] Test on beta settings page that attribute renaming works without console warnings

Related to cloud-ui changes for search attributes rename support.

### DIFF
--- a/src/lib/components/search-attributes-form/search-attribute-row.svelte
+++ b/src/lib/components/search-attributes-form/search-attribute-row.svelte
@@ -1,4 +1,6 @@
 <script lang="ts">
+  import type { FormEventHandler } from 'svelte/elements';
+
   import Button from '$lib/holocene/button.svelte';
   import Input from '$lib/holocene/input/input.svelte';
   import Option from '$lib/holocene/select/option.svelte';
@@ -44,31 +46,9 @@
 
   const hasError = $derived(!!error);
 
-  const inputProps = $derived({
-    id: `attribute-name-${index}`,
-    value: name,
-    label: translate('search-attributes.attribute-label', {
-      index: index + 1,
-    }),
-    labelHidden: true,
-    disabled: submitting,
-    error: hasError,
-    oninput: (e: Event) =>
-      onNameChange((e.currentTarget as HTMLInputElement).value),
-  });
-
-  const selectProps = $derived({
-    id: `attribute-type-${index}`,
-    value: type,
-    label: translate('search-attributes.type-label', {
-      index: index + 1,
-    }),
-    labelHidden: true,
-    disabled: isTypeDisabled,
-    placeholder: translate('search-attributes.select-type-placeholder'),
-    onchange: (e: Event) =>
-      onTypeChange((e.currentTarget as HTMLSelectElement).value),
-  });
+  const handleNameInput: FormEventHandler<HTMLInputElement> = (e) => {
+    onNameChange(e.currentTarget.value);
+  };
 </script>
 
 <div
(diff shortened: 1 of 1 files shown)


src/lib/components/search-attributes-form/search-attribute-row.svelte
```diff
@@ -32,44 +33,52 @@
...
+  const inputProps = $derived({
+    id: `attribute-name-${index}`,
+    value: name,
+    label: translate('search-attributes.attribute-label', {
+      index: index + 1,
+    }),
+    labelHidden: true,
+    disabled: submitting,
+    error: hasError,
+    oninput: (e: Event) =>
```
rossedfort: nit: I believe you can use the `FormEventHandler` type to avoid coercion here, eg. `(e: FormEventHandler<HTMLInputElement>)`

src/lib/components/search-attributes-form/search-attribute-row.svelte
```diff
@@ -32,44 +33,52 @@
...
+  const selectProps = $derived({
+    id: `attribute-type-${index}`,
+    value: type,
+    label: translate('search-attributes.type-label', {
+      index: index + 1,
+    }),
+    labelHidden: true,
+    disabled: isTypeDisabled,
+    placeholder: translate('search-attributes.select-type-placeholder'),
+    onchange: (e: Event) =>
```
rossedfort: Same here

src/lib/components/search-attributes-form/search-attribute-row.svelte
```diff
@@ -32,44 +33,52 @@
...
-      (disableTypeForExisting &&
-        attribute.name &&
-        initialAttributeNames.has(attribute.name)),
+      (disableTypeForExisting && name && initialAttributeNames.has(name)),
   );
 
   const hasError = $derived(!!error);
+
+  const inputProps = $derived({
+    id: `attribute-name-${index}`,
```
andrewzamojc: Question: Why all the props? Does id give the problem, or just the `bind:value` one?

src/lib/components/search-attributes-form/search-attribute-row.svelte
andrewzamojc: Question: The way I'm reading the issue, it seems that a non-bindable prop can't drive a bindable value. Are you sure you didn't resolve the issue but using value as non-bindable, and relying on `oninput` to the value? If so, that might work without this derived object. 

TLDR, would this work?
```
<Input
  value={attribute.name}
  oninput={updateName}
/>
```

rossnelson: You aren't wrong. I threw this in there because it mimics the pattern used in tags. The derived object felt like a cleaner way to define the input attributes but maybe I'm overthinking it. I can remove it if its overkill.

src/lib/components/search-attributes-form/search-attribute-row.svelte
rossnelson: same response: https://github.com/temporalio/ui/pull/2992/files/3438ce399ddb51e046be55e10ad221682887ea3d#r2478719584

PR #2996, opened 2025-10-30
chore(docker): inline docker-builds base image dependencies

rossnelson:
## Motivation

ACT is deprecating the docker-builds repo, which currently provides the `temporalio/base-server` and `temporalio/base-builder` images used in the UI server Dockerfile. To prepare for this deprecation, we need to inline these dependencies directly into our Dockerfile.

## Implementation

The Dockerfile now uses standard base images and builds the necessary tooling:

**Builder stage**: Uses `golang:1.24-alpine3.22` directly with make, git, and curl installed via apk. This replaces the `temporalio/base-builder` image which provided the same packages.

**Dockerize builder**: Added a dedicated stage to build the dockerize v0.9.2 utility from source. Previously this was included in the base-server image.

**Server stage**: Uses `alpine:3.22` as the base with ca-certificates, tzdata, bash, and curl installed. Copies the dockerize binary from the builder stage. This replaces the `temporalio/base-server` image.

All functionality remains identical, but the Dockerfile is now self-contained without external dependencies on the docker-builds repo.

### DIFF
--- a/server/Dockerfile
+++ b/server/Dockerfile
@@ -1,7 +1,10 @@
-ARG BASE_SERVER_IMAGE=temporalio/base-server:1.15.13
-ARG BASE_BUILDER_IMAGE=temporalio/base-builder:1.15.6
+FROM golang:1.24-alpine3.22 AS server-builder
 
-FROM ${BASE_BUILDER_IMAGE} AS server-builder
+RUN apk upgrade --no-cache
+RUN apk add --no-cache \
+    make \
+    git \
+    curl
 
 WORKDIR /home/server-builder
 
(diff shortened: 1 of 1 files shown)


server/Dockerfile
```diff
@@ -12,9 +15,28 @@ COPY . ./
 
 RUN make build-server
 
+##### Dockerize builder #####
+
+FROM golang:1.24-alpine3.22 AS dockerize-builder
+
+ARG DOCKERIZE_VERSION=v0.9.2
+RUN go install github.com/jwilder/dockerize@${DOCKERIZE_VERSION}
+RUN cp $(which dockerize) /usr/local/bin/dockerize
```
Alex-Tideman, line 24: Do we still need to support dockerize?

rossnelson: well, i assumed we did since the base image was using it. but ill check.

rossnelson: i think we do: https://github.com/temporalio/ui/blob/main/server/docker/start-ui-server.sh#L8

Alex-Tideman: So that was for old templating, there's that if statement around it.

https://github.com/temporalio/ui/blob/9f00bdf34de0acd8654053873f6f77c9240e2da7/server/docker/README.md?plain=1#L39

I think we wanna get rid of that dependency, just not sure if we can rip the bandaid yet

rossnelson: Feedback from Rob is this will be fully deprecated shortly. We can keep dockerize until further notice.

PR #2999, opened 2025-10-31
Payload retries

GiantRobots:
### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/event/payload-decoder.svelte
+++ b/src/lib/components/event/payload-decoder.svelte
@@ -23,15 +23,19 @@
     key?: string;
     onDecode?: (decodedValue: string) => void;
     children: Snippet<[decodedValue: string]>;
+    error?: Snippet<[retry: () => Promise<string>, err: unknown]>;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    loading?: Snippet<[keyedVal: any]>;
   }
 
-  let { children, value, key = '', onDecode }: Props = $props();
+  let { children, value, key = '', onDecode, error, loading }: Props = $props();
 
   let keyedValue = key && value?.[key] ? value[key] : value;
-  let decodedValue = $state(stringifyWithBigInt(keyedValue));
+
+  let decodeValuePromise = $state<Promise<string> | null>(null);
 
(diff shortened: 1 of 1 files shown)


src/lib/components/event/payload-decoder.svelte
```diff
@@ -61,14 +70,36 @@
...
       if (onDecode) {
         onDecode(decodedValue);
       }
+      return decodedValue;
     } catch (e) {
       console.error('Could not decode payloads');
+      // hmm before this just ate the error we want to throw this to get an error here
+      // but maybe this is leaking information to the users? But it also might be good
+      // to allow that? Think about this harder maybe ask app sec about if this is an okay
+      // design choice
```
Alex-Tideman, line 83: We always want to return the original encoded value if decoding fails

src/lib/components/event/payload-decoder.svelte
```diff
@@ -61,14 +70,36 @@
...
+
+  <Button on:click={retry}>Retry Decoding</Button>
+{/snippet}
+
+{#await decodeValuePromise}
+  {@render loading?.(keyedValue)}
+{:then decoded}
+  {@render children(decoded)}
+{:catch err}
+  {@render error?.(() => decodePayloads(value), err)}
```
Alex-Tideman, line 104: This... confuses me? 

So you'd pass something like

```
{#snippet error(retry, err)}
  {#await retry} ...
  {:catch}
   {err}
  {/await}
{/snippet}
```

src/lib/components/event/payload-decoder.svelte
```diff
@@ -61,14 +70,36 @@
...
+{/await}
```
Alex-Tideman, line 105: @claude explain this

src/lib/components/event/payload-decoder.svelte
GiantRobots: retry is a function so you can call it in the error component if you like should have probably explicitly set the promise above to the decode payloads so it would retry

PR #3006, opened 2025-11-04
Truncate runid

GraceGardner:
## Description & motivation 💭 
Truncates the workflow run-id.

Screenshot 2025-11-04 at 11 54 08 AM

### How was this tested 👻 
I had started writing tests and was going to turn the truncate function into a utility, but realized that we only use it in two places and the number of periods isn't the same. It's also simple enough logic that we probably don't really NEED to do it? IDK, let me know if we should. 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
DT-3375

### DIFF
--- a/src/lib/components/workflow/workflows-summary-configurable-table/filterable-table-cell.svelte
+++ b/src/lib/components/workflow/workflows-summary-configurable-table/filterable-table-cell.svelte
@@ -62,9 +62,7 @@
 
 {#if isRunId}
   <Tooltip text={value} top class="min-w-0">
-    <Link {href} class="tracking-tightest cursor-help text-sm"
-      >{truncateRunId(value)}</Link
-    >
+    <Link {href} class="cursor-help">{truncateRunId(value)}</Link>
   </Tooltip>
 {:else if href}
   <Link {href}>{value}</Link>


src/lib/components/workflow/workflows-summary-configurable-table/filterable-table-cell.svelte
```diff
@@ -27,6 +28,16 @@
...
 
+  const truncateRunId = (runId: string): string => {
+    if (runId.length > 11) {
+      return `${runId.slice(0, 4)}....${runId.slice(-4)}`;
+    }
+    return runId;
+  };
+
+  const isRunId = attribute === 'RunId';
+  const displayValue = isRunId ? truncateRunId(value) : value;
```
andrewzamojc: Question: It's it just me or is it sort of weird to have table-cell know about something so data specific as the run id and its truncation? I can't tell if it matters, but maybe I'm thinking the truncated value should be passed in so cell can "just render". If I'm making a mess of an otherwise quick fix, ignore me hehe.

Alex-Tideman: Normally I'd agree, but this is the workflows-summary-configurable-table/filterable-table-cell... so it's pretty specific to workflows and not a generic table cell.

src/lib/components/workflow/workflows-summary-configurable-table/filterable-table-cell.svelte
```diff
@@ -27,6 +28,16 @@
     type = SEARCH_ATTRIBUTE_TYPE.KEYWORD,
   }: Props = $props();
 
+  const truncateRunId = (runId: string): string => {
+    if (runId.length > 11) {
+      return `${runId.slice(0, 4)}....${runId.slice(-4)}`;
```
Alex-Tideman: Let's use 3 ..., not 4

src/lib/components/workflow/workflows-summary-configurable-table/filterable-table-cell.svelte
```diff
@@ -46,11 +57,23 @@
...
 
-    updateQueryParamsFromFilter($page.url, $workflowFilters);
+    updateQueryParamsFromFilter(page.url, $workflowFilters);
   };
 </script>
 
-{#if href}
+{#if isRunId}
+  <Tooltip text={value} top class="min-w-0">
+    {#if href}
```
Alex-Tideman: I don't think we need this if statement. RunIds will always have href

src/lib/components/workflow/workflows-summary-configurable-table/filterable-table-cell.svelte
```diff
@@ -46,11 +57,23 @@
...
+      <Link {href} class="tracking-tightest cursor-help font-mono text-sm"
```
Alex-Tideman: Lets not use font-mono

src/lib/components/workflow/workflows-summary-configurable-table/filterable-table-cell.svelte
GraceGardner: Okay awesome! Thanks Tideman

src/lib/components/workflow/workflows-summary-configurable-table/filterable-table-cell.svelte
GraceGardner: Sounds good, I know that 3 is more standard and we use 3 in the namespace, but there were 4 in the designs so I wasn't sure if that was a big deal.

src/lib/components/workflow/workflows-summary-configurable-table/filterable-table-cell.svelte
```diff
@@ -46,11 +56,17 @@
...
 
-    updateQueryParamsFromFilter($page.url, $workflowFilters);
+    updateQueryParamsFromFilter(page.url, $workflowFilters);
   };
 </script>
 
-{#if href}
+{#if isRunId}
+  <Tooltip text={value} top class="min-w-0">
+    <Link {href} class="tracking-tightest cursor-help text-sm"
```
Alex-Tideman: I don't think we need any of these classes

andrewzamojc: I was just looking at these too. I don't think we usually put cursor-help on tooltips.

GraceGardner: We do normally have the cursor-help. That was me just follow patterns. I was attempting to tighten up the spacing between the periods, but it's not actually doing anything.

PR #3021, opened 2025-11-14
DT-3307 - add custom timestamp formats

rossedfort:
## Description & motivation 💭 
- Allow users to select from 3 timestamp formats: short, medium, and long
- Create `<Timestamp />` component to minimize direct calls to `formatDate`

Screenshot 2025-11-17 at 10 44 21 AM

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/utilities/format-date.test.ts
+++ b/src/lib/utilities/format-date.test.ts
@@ -9,6 +9,7 @@ import {
   isValidDate,
 } from './format-date';
 
+// // force GH action runners to use en-US and 12-hour clocks starting at 0:00
 const DateTimeFormat = Intl.DateTimeFormat;
 vi.spyOn(global.Intl, 'DateTimeFormat').mockImplementation(
   (_, options) =>
--- a/src/lib/utilities/query/search-attribute-filter.test.ts
+++ b/src/lib/utilities/query/search-attribute-filter.test.ts
@@ -30,6 +30,7 @@ const store = writable<SearchAttributes>({
   CustomC: 'String',
 });
 
+// force GH action runners to use en-US and 12-hour clocks starting at 0:00
 const DateTimeFormat = Intl.DateTimeFormat;
 vi.spyOn(global.Intl, 'DateTimeFormat').mockImplementation(
   (_, options) =>
--- a/tests/integration/workflows-search-attribute-filter.desktop.spec.ts
+++ b/tests/integration/workflows-search-attribute-filter.desktop.spec.ts
@@ -48,7 +48,7 @@ test('it should update the datetime filter based on the selected timezone', asyn
 
   await expect(
     page.getByRole('button', {
-      name: 'CloseTime >= 12/25/25, 04:00:00.00 PST',
+      name: 'CloseTime >= 12/25/25, 4:00:00.00 AM PST',
     }),
   ).toBeVisible();
 
@@ -63,7 +63,7 @@ test('it should update the datetime filter based on the selected timezone', asyn
 
   await expect(
     page.getByRole('button', {
-      name: 'CloseTime >= 12/25/25, 05:00:00.00 MST',
+      name: 'CloseTime >= 12/25/25, 5:00:00.00 AM MST',
     }),
   ).toBeVisible();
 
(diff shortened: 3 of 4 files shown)


src/lib/components/timestamp.svelte
```diff
@@ -0,0 +1,66 @@
+<script lang="ts">
```
Alex-Tideman, line 1: Any reason to put this in components instead of holocene? Feels like we'll use it in cloud

src/lib/stores/time-format.ts
```diff
@@ -6,6 +6,20 @@ import { getLocalTimezone } from '$lib/utilities/format-date';
 
 type TimeFormatTypes = 'relative' | 'absolute';
 
+export const timestampFormats = {
+  short: 'do MMM yyyy H:mm:ss.SS',
+  medium: 'yyyy-MM-dd z HH:mm:ss.SS',
+  long: 'MMMM do yyyy, hh:mm:ss.SS a z',
+  abbreviated: 'yyyy-MM-dd HH:mm:ss a',
+  abbreviatedWithoutSeconds: 'yyyy-MM-dd HH:mm a',
+} as const;
```
Alex-Tideman: Wondering if we should use 'full' | 'long' | 'medium' | 'short' and map to Intl.DateTimeFormat. What I don't know is if our timezone's locale map correctly.

```
new Intl.DateTimeFormat('en-US', {
  year: 'numeric',
  month: 'long',
  day: 'numeric'
}).format(new Date()); // "November 17, 2025"

// Short date
new Intl.DateTimeFormat('en-US', {
  year: '2-digit',
  month: '2-digit',
  day: '2-digit'
}).format(new Date()); // "11/17/25"

// Date + Time
new Intl.DateTimeFormat('en-US', {
  dateStyle: 'short',
  timeStyle: 'short'
}).format(new Date()); // "11/17/25, 2:30 PM"
Key options:

dateStyle: 'full' | 'long' | 'medium' | 'short'
timeStyle: 'full' | 'long' | 'medium' | 'short'
```

src/lib/components/timestamp.svelte
rossedfort: We can still import it in cloud from `@temporalio/ui/components`, right? The reason I didn't put it in Holocene is it's more of a utility component than a true UI component.

src/lib/stores/time-format.ts
rossedfort: Fixed. date formats now are:

`11/18/25, 09:59:16.00 MST`
`Nov 18, 2025, 09:59:35.00 MST`
`November 18, 2025 at 9:59:43.00 AM MST`

The short and medium formats use 24 hour clock, so PM times will be 12, 13, 14, etc. and no `AM`/`PM` indicator.

src/lib/components/timestamp.svelte
Alex-Tideman: Yeah totally, just curious. Makes sense.

src/lib/stores/time-format.ts
Alex-Tideman: What is the default? Medium? Just might be weird for us 12-hour folks to have that now be 24

rossedfort: Yes, medium is the default. It was that way before, right? https://github.com/temporalio/ui/blob/main/src/lib/utilities/format-date.ts#L18

`HH` is 24 clock https://date-fns.org/v4.1.0/docs/format

rossedfort: I'm good with either 24 or 12, although I think you're right that 12 might make more sense for most folks. Just was trying to keep things somewhat aligned with the previous implementation.

Alex-Tideman: Can we do it based on locale? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat#hour12

rossedfort: I guess leaving `hour12` unspecified would do it based on locale, we can do that.

PR #3030, opened 2025-11-20
Clear Schedule timeouts

Alex-Tideman:
## Description & motivation 💭 
Clear timeouts to prevent memory leaks in schedules. Also put full spec in one CodeBlock, don't create a CodeBlock per single spec. This was the real issue in the browser locking up

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/schedule/schedule-frequency.svelte
+++ b/src/lib/components/schedule/schedule-frequency.svelte
@@ -33,7 +33,7 @@
         testId="schedule-calendar"
         language="json"
         content={stringifyWithBigInt(frequency, null, 2)}
-        {...frequency.length > 1 ? { maxHeight: 300 } : {}}
+        {...frequency.length > 1 ? { maxHeight: 600 } : {}}
       />
     </div>
   </div>


src/lib/stores/schedules.ts
```diff
@@ -85,6 +85,9 @@ const setBodySpec = (
   }
 };
 
+let createTimeout: ReturnType<typeof setTimeout>;
+let editTimeout: ReturnType<typeof setTimeout>;
```
laurakwhit, line 89: Could we add a "cleanup" function here that clears the timeouts and then export it and call it where `submitCreateSchedule` and `submitEditSchedule` are used?

Alex-Tideman: Good call

GiantRobots: How do you feel about abstracting it mildly to like 

```typescript
  const createTimeout = (callback: () => void, delay: number) => {
    const timeoutId = setTimeout(callback, delay);
    return () => clearTimeout(timeoutId);
  };
```

I think this method would be useful in onMount functions in components as well if they're used for easier cleanup ex.

```typescript
onMount(()=>{

return createTimeout(() => {
      goto(routeForSchedule({ namespace, scheduleId: name }));
      error.set('');
      loading.set(false);
    }, 2000);
})
  
```

Alex-Tideman: Will refactor to this

PR #3050, opened 2025-12-05
Time range bug fix

KATIETOLER:
## Bug
 - Bug: Multiple filter instances were sharing global stores, causing
  changes in one filter to affect all others
  - UX Issue: Simply using local state would lose the nice "default from
  last used" behavior

## Solution

  Two-way data flow between local and global stores:

  1. On Mount → Copy global stores to local state (get "last used" defaults)
  2. During Editing → Only local state changes (filters are independent)
  3. On Apply → Update both filter value AND global stores (set defaults for
   next filter)

### DIFF
--- a/src/lib/components/search-attribute-filter/datetime-filter.svelte
+++ b/src/lib/components/search-attribute-filter/datetime-filter.svelte
@@ -51,8 +51,13 @@
   let localEndMinute = $state('');
   let localEndSecond = $state('');
 
+  const error = (x: string) => {
+    if (x) return isNaN(Number(x)) || isNaN(parseFloat(x));
+    return false;
+  };
+
   const isTimeRange = $derived($filter.conditional === 'BETWEEN');
-  const selectedTime = $derived(getSelectedTimezone($timeFormat));
+  const selectedTime = $derived(getSelectedTimezone($timeFormat ?? 'UTC'));
 
   const useBetweenDateTimeQuery = $derived(
     isTimeRange || !$supportsAdvancedVisibility,
@@ -137,11 +142,6 @@
     }
 
     handleSubmit();
-  };
-
-  const error = (x: string) => {
-    if (x) return isNaN(Number(x)) || isNaN(parseFloat(x));
-    return false;
   };
 </script>
 


src/lib/components/search-attribute-filter/datetime-filter.svelte
```diff
@@ -40,6 +40,30 @@
...
+  let localStartSecond = '';
+
+  let localEndDate = startOfDay(new Date());
+  let localEndHour = '';
+  let localEndMinute = '';
+  let localEndSecond = '';
+
+  // Initialize local state from global stores to get "last used" defaults
+  onMount(() => {
+    startDate.subscribe((value) => (localStartDate = value))();
```
andrewzamojc: I'm no expert on stores, but this seems like it'll stay subscribed. Like, if the store updates, then these will run again. I can't tell if that's what  you want, or if it's a problem...

src/lib/components/search-attribute-filter/datetime-filter.svelte
```diff
@@ -40,6 +40,30 @@
...
+    startHour.subscribe((value) => (localStartHour = value))();
+    startMinute.subscribe((value) => (localStartMinute = value))();
+    startSecond.subscribe((value) => (localStartSecond = value))();
+
+    endDate.subscribe((value) => (localEndDate = value))();
+    endHour.subscribe((value) => (localEndHour = value))();
+    endMinute.subscribe((value) => (localEndMinute = value))();
+    endSecond.subscribe((value) => (localEndSecond = value))();
```
rossedfort: I think you should remove the trailing `()` after the subscribe call.

For context, `store.subscribe()` returns an "unsubscriber" function, which you typically would call on an `onDestroy` when the component un-mounts. Effectively this code is subscribing to changes to the store value, then immediately un-subscribing.

What do you intend for this code to do? Continually update the "local" values with the store value? Or just update the "local" values once?

src/lib/components/search-attribute-filter/datetime-filter.svelte
rossedfort: This will actually do the opposite. Typically you subscribe to a store because you _do_ want to run a callback on every change. The extra `()` here calls the "unsubscribe" function that's returned by `store.subscribe()`, which negates the purpose of subscribing.

andrewzamojc: OH cool! Neat technique.

rossedfort: @andrewzamojc I don't believe this is a pattern we should be doing 😅 It's effectively the same as calling `get()` from `'svelte/store'` https://svelte.dev/docs/svelte/stores#svelte-store-get

GiantRobots: Alternatively you could just do `localStartDate = $startDate` because that's what svelte does automagically with stores when you use $

PR #3061, opened 2025-12-15
Add schedule link in Workflow Details

Alex-Tideman:
## Description & motivation 💭 
Show link to schedule in Workflow Details if workflow was scheduled. Easily link out to schedule from a workflow

Screenshot 2025-12-15 at 10 10 59 AM
Screenshot 2025-12-15 at 10 11 58 AM

### Design Considerations 🎨 
Because the 3 column could be fairly long, I've made the 3 column for links only and a 4th column for history size/sdk/actions

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/lines-and-dots/workflow-details.svelte
+++ b/src/lib/components/lines-and-dots/workflow-details.svelte
@@ -21,6 +21,7 @@
   import { getSDKandVersion } from '$lib/utilities/get-sdk-version';
   import { isWorkflowTaskCompletedEvent } from '$lib/utilities/is-event-type';
   import {
+    routeForSchedule,
     routeForTaskQueue,
     routeForWorkerDeployment,
     routeForWorkflow,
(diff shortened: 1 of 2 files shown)


src/lib/components/lines-and-dots/workflow-details.svelte
```diff
@@ -225,15 +230,25 @@
   {/if}
 
   <DetailListColumn>
-    {#if workflow?.parent}
+    {#if scheduleId}
+      <DetailListLabel>{translate('workflows.scheduled-by')}</DetailListLabel>
+      <DetailListLinkValue
```
laurakwhit, line 235: Would it also be useful to add a filter for "Scheduled By ID" somewhat similar to what we have for the "Workflow Type"?

Alex-Tideman: Maybe/probably? Schedules usually have one-to-one mapping to a workflow but the list is more useful than the Schedule view which only shows last 5 runs. We do have the View All Runs on the Schedule view page that does this, but it's one extra click. Tots can add it if we get that feedback

laurakwhit: Went ahead and added a draft PR for this [here](https://github.com/temporalio/ui/pull/3063), but we can hold off for more feedback/design input if we want.

PR #3064, opened 2025-12-15
use table for nexus registry

GraceGardner:
### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/routes/(app)/nexus/[id]/+page.svelte
+++ b/src/routes/(app)/nexus/[id]/+page.svelte
@@ -19,13 +19,15 @@
   title={translate('nexus.nexus-endpoint', { id: $page.params.id })}
   url={$page.url.href}
 />
-<div class="flex flex-col gap-4">
-  <Link href={routeForNexus()} icon="chevron-left">
-    {translate('nexus.back-to-endpoints')}
-  </Link>
-  <NexusEndpoint {endpoint}>
-    {#snippet taskQueueStatus()}
-      <TaskQueueStatus {endpoint} />
-    {/snippet}
-  </NexusEndpoint>
-</div>
+{#if endpoint}
+  <div class="flex flex-col gap-4">
+    <Link href={routeForNexus()} icon="chevron-left">
+      {translate('nexus.back-to-endpoints')}
+    </Link>
+    <NexusEndpoint {endpoint}>
+      {#snippet taskQueueStatus()}
+        <TaskQueueStatus {endpoint} />
+      {/snippet}
+    </NexusEndpoint>
+  </div>
+{/if}
--- a/src/routes/(app)/nexus/[id]/edit/+page.svelte
+++ b/src/routes/(app)/nexus/[id]/edit/+page.svelte
@@ -21,12 +21,12 @@
 
   import type { LayoutData } from '../$types';
 
-  export let data: LayoutData;
+  let { data }: { data: LayoutData } = $props();
 
-  $: ({ endpoint } = data);
+  const { endpoint } = $derived(data);
 
-  let error: NetworkError | undefined = undefined;
-  let loading = false;
+  let error = $state<NetworkError | undefined>(undefined);
+  let loading = $state(false);
 
   const onUpdate = async () => {
     if (!endpoint) return;
@@ -42,7 +42,7 @@
         input: JSON.stringify(body.spec.descriptionString),
         encoding: 'json/plain',
       });
-      body.spec.description = payloads[0];
+      body.spec.description = payloads?.[0];
 
       delete body.spec.allowedCallerNamespaces;
       delete body.spec.descriptionString;
@@ -75,9 +75,11 @@
     }
   };
 
-  $: targetNamespaceList = $namespaces.map((namespace) => ({
-    namespace: namespace.namespaceInfo.name,
-  }));
+  const targetNamespaceList = $derived(
+    $namespaces.map((namespace) => ({
+      namespace: namespace.namespaceInfo?.name ?? '',
+    })),
+  );
 </script>
 
 <PageTitle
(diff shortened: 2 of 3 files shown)


src/lib/pages/nexus-endpoints.svelte
```diff
@@ -1,47 +1,68 @@
...
+    rowColumns,
+    children,
+  }: {
+    endpoints: NexusEndpoint[];
+    searchPlaceholder?: string;
+    createDisabled?: boolean;
+    headerColumns?: Snippet;
+    rowColumns?: Snippet<[NexusEndpoint]>;
+    children?: Snippet;
+  } = $props();
```
rossnelson: Minor suggestion, move the props into a type for clarity:

```
  type Props = {
    endpoints: NexusEndpoint[];
    searchPlaceholder?: string;
    createDisabled?: boolean;
    headerColumns?: Snippet;
    rowColumns?: Snippet<[NexusEndpoint]>;
    children?: Snippet;
  };

  let { endpoints = [], ... }: Props = $props();
```

src/lib/pages/nexus-endpoints.svelte
rossnelson: Do we need this here:

```
href={createHref ?? routeForNexusEndpointCreate()}
```

src/lib/pages/nexus-edit-endpoint.svelte
```diff
@@ -1,16 +1,13 @@
...
-  import Link from '$lib/holocene/link.svelte';
   import Modal from '$lib/holocene/modal.svelte';
   import { translate } from '$lib/i18n/translate';
   import NexusForm from '$lib/pages/nexus-form.svelte';
   import type { NetworkError } from '$lib/types/global';
   import type { NexusEndpoint } from '$lib/types/nexus';
-  import { routeForNexusEndpoint } from '$lib/utilities/route-for';
 
   export let endpoint: NexusEndpoint;
   export let targetNamespaceList: { namespace: string }[] = [];
```
Alex-Tideman: Let's svelte5 this file too while we are in here

src/lib/pages/nexus-endpoints.svelte
```diff
@@ -1,121 +1,179 @@
...
-              {#if endpoint.createdTime}
-                <Timestamp
-                  as="p"
-                  class="text-xs text-secondary"
-                  dateTime={endpoint.createdTime}
+            </td>
+
+            <td class="px-2">
+              {#if !endpoint.spec?.allowedCallerNamespaces?.length}
+                <span class="text-sm text-secondary">0 Namespaces</span>
```
Alex-Tideman: We'll want to wrap this whole cell / column in {#if endpoint.spec?.allowedCallerNamespaces}

In OSS, there's no such thing of an allowedCallerNamespaces field. All namespaces can be the caller. So we should hide the count in OSS because it will always show 0 in this case when it really is the full count of all namespaces.

src/lib/pages/nexus-endpoints.svelte
```diff
@@ -1,121 +1,179 @@
...
+        <caption class="sr-only" slot="caption">
+          {translate('nexus.endpoints')}
+        </caption>
+        <TableHeaderRow slot="headers">
+          <th>Name</th>
+          <th>Used By</th>
+          {@render headerColumns?.()}
+          <th>Last Updated</th>
+          <th>Created On</th>
+        </TableHeaderRow>
```
Alex-Tideman: It might be a good idea to pass in the full TableHeaderRow in the headerColumns snippet instead of having it in the middle of hardcoded headers. That way you have more flexibility in the future and can keep OSS/Cloud separation clean.

src/lib/pages/nexus-endpoint.svelte
```diff
@@ -31,8 +31,9 @@
         {endpoint.spec?.name || ''}
       </h1>
       <Button
-        href={routeForNexusEndpointEdit(endpoint.id)}
-        disabled={editDisabled}>{translate('common.edit')}</Button
+        href={endpoint.id ? routeForNexusEndpointEdit(endpoint.id) : '#'}
```
Alex-Tideman: Curious when endpoint wouldn't have id if it's always required?

GraceGardner: That's a goooood point.

src/routes/(app)/nexus/+page.svelte
```diff
@@ -1,15 +1,55 @@
 <script lang="ts">
   import PageTitle from '$lib/components/page-title.svelte';
+  import Timestamp from '$lib/components/timestamp.svelte';
+  import Link from '$lib/holocene/link.svelte';
+  import TableHeaderRow from '$lib/holocene/table/table-header-row.svelte';
+  import TableRow from '$lib/holocene/table/table-row.svelte';
   import { translate } from '$lib/i18n/translate';
   import NexusEndpoints from '$lib/pages/nexus-endpoints.svelte';
   import { page } from '$lib/svelte-mocks/app/stores';
```
rossnelson: I know this isnt your change. but pull from mocks struck me as strange. is it valid?

src/lib/pages/nexus-edit-endpoint.svelte
rossnelson: i wonder if this is intentional: `items-centeritems-center`

rossnelson (approved):
All good on your end. just noticed some preexisting stuff that might be worth addressing.

src/routes/(app)/nexus/+page.svelte
Alex-Tideman: def shouldn't have a mock in here, probs a bad auto-import

PR #3079, opened 2026-01-07
Svelte 5 Migrate skill

Alex-Tideman:
## Description & motivation 💭 
Add a new Claude Skill to migrate .svelte files from 4 to 5. Use with the pattern of 

`migrate ...file to svelte5` when in Claude Code.

I updated the Badge and Chip components to test. It updated all the dependencies of Chip as well.

I updated the .gitignore to allow checking in skills but no other .claude files so others can use the skills

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/.claude/skills/svelte-migrate/SKILL.md
+++ b/.claude/skills/svelte-migrate/SKILL.md
@@ -425,7 +425,7 @@ The `class:` directive still works but consider using conditional expressions:
 <div class={`${isActive ? 'active' : ''} ${disabled ? 'disabled' : ''}`}>
 ```
 
-### 12. App State
+### 13. App State
 
 **Before:**
 


.claude/skills/svelte-migrate/SKILL.md
```diff
@@ -0,0 +1,452 @@
...
+   - Component events (dispatch → callback props)
+   - Slots (slot → snippets)
+4. Update imports (add Snippet type if needed, remove createEventDispatcher if no longer used)
+5. Ensure TypeScript types are correct
+6. Run `pnpm check` to verify no type errors
+
+## Important Notes
+
+- Keep `onMount` and `onDestroy` - they're still valid
+- `context="module"` scripts remain unchanged
```
rossedfort: Svelte5/Runes uses `<script module>` instead of `<script context="module">`, should we include that?

rossedfort (approved):

.claude/skills/svelte-migrate/SKILL.md
```diff
@@ -0,0 +1,452 @@
...
+
+{@render children?.()}
+```
+
+**Named slots:**
+
+Before:
+
+```svelte
+<slot name="header" />
```
laurakwhit, line 265: Might run into conflicts with slot names and other props, but maybe we just add something here when/if we cross that bridge 😅

.claude/skills/svelte-migrate/SKILL.md
```diff
@@ -0,0 +1,452 @@
...
+```
+
+**After (either works):**
+
+```svelte
+<div class:active={isActive} class:disabled>
+<!-- or -->
+<div class={`${isActive ? 'active' : ''} ${disabled ? 'disabled' : ''}`}>
+```
+
```
laurakwhit, line 427: Could also add a line for moving from `import { page } from '$app/stores'` to `import { page } from '$app/state'`.

Alex-Tideman: 👍 added App State section

.claude/skills/svelte-migrate/SKILL.md
Alex-Tideman: Yeah I'm not sure what to put for that yet... let's see how it goes!

PR #3090, opened 2026-01-13
Nav update icon projects/kt

KATIETOLER:
This PR is meant to add the underlying structure for the new nav. 
The effect of this PR adds more svelte5 migrations because the nav is central to a lot of things and updating it ended up being more hairy than I had hoped for. 
The LLM helped me do this, so there are probably errors or weirdness somewhere I am code blind to.
The application seemed to run fine though, so maybe I am paranoid. 

The icons are added and there are lots of guard rails to prevent any of these changes to prod until we are ready.


src/lib/holocene/navigation/navigation-container.svelte
```diff
@@ -34,28 +35,61 @@
...
-        {isCloud ? 'Cloud' : 'Self-Hosted'}
-      </p>
-    </a>
-    <button
-      title={$navOpen ? 'Collapse Navigation' : 'Expand Navigation'}
-      class="mx-2 flex items-center justify-center opacity-0 transition-[opacity,transform] focus-visible:outline-none focus-visible:ring-2 group-hover:opacity-100 group-focus:opacity-100 group-data-[nav=open]:rotate-180 group-data-[nav=closed]:p-2 {isCloud
-        ? 'focus-visible:ring-success'
-        : 'focus-visible:ring-primary/70'}"
-      onclick={toggle}
+  {#if isCloud}
```
Alex-Tideman, line 38: We should make this it's own component instead of an if statement. Start with a CloudNavBar and OSSNavBar and if we want to use CloudNavBar for OSS we can switch to that

KATIETOLER: yess thats a much better plan.

KATIETOLER:
See this PR -- without all the unintended changes -- here --> https://github.com/temporalio/ui/pull/3095

PR #3091, opened 2026-01-13
feat(auth): add configurable max session duration

rossnelson:
## Summary

- Add server-side enforcement of maximum session duration independent of token expiry
- When `maxSessionDuration` is configured, users must re-authenticate after the specified time even if tokens are still refreshable
- Add Claude Code skills for auth testing documentation

## Changes

- Add `MaxSessionDuration` field to Auth config struct
- Set `session_start` cookie on login callback to track session age
- Validate session duration in auth middleware, return 401 when exceeded
- Configure OIDC test server session TTL to match for local testing
- Document testing scenarios in config files

## Test plan

- [ ] Run `pnpm dev:with-auth` to start local auth environment
- [ ] Login and verify `session_start` cookie is set
- [ ] Wait for token expiry (~60s) and verify automatic refresh works
- [ ] Wait for session expiry (~2m) and verify redirect to SSO login
- [ ] Verify user must re-enter credentials at OIDC provider

### DIFF
--- a/.claude/skills/auth-testing/SKILL.md
+++ b/.claude/skills/auth-testing/SKILL.md
@@ -74,6 +74,39 @@ ttl: {
 }
 ```
 
+## Max Session Duration
+
+The `maxSessionDuration` config enforces a hard limit on how long a user can stay logged in, independent of token expiry.
+
+### How it works
+
+1. **On login**: Server sets `session_start` cookie with current timestamp
+2. **On each request**: Server validates session age against `maxSessionDuration`
+3. **On expiry**: Server returns 401, UI redirects to SSO login
+
+### Config location
+
+```yaml
+# server/config/with-auth.yaml
+auth:
+  enabled: true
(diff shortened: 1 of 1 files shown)


server/server/auth/auth.go
```diff
@@ -105,6 +106,49 @@ func SetUser(c echo.Context, user *User) error {
...
+// This should only be called on initial login, NOT on token refresh.
+func SetSessionStart(c echo.Context, maxSessionDuration time.Duration) {
+	if maxSessionDuration <= 0 {
+		return
+	}
+
+	cookie := &http.Cookie{
+		Name:     sessionStartCookie,
+		Value:    strconv.FormatInt(time.Now().Unix(), 10),
+		MaxAge:   int(maxSessionDuration.Seconds()),
```
asmclean, line 149: Both of these values can be changed, so this is really more than a recommendation than anything.  If we need to truly enforce this, then we'll need an assurance that the values are unchanged  (This could be handled by my suggestion in the PR for RT POC, but there are other options, too.)

PR #3093, opened 2026-01-14
feat(ui-server): add forward of static headers from env

wensiet:
## Description & motivation 💭 

This PR introduces static headers forwarding to gRPC temporal frontend. 

Our scenario is:
- custom authorizer & claim-mapper plugin for temporal server
- TLS is provided by the infrastructure and cannot be configured
- to authenticate request, we need custom header, like `x-service-auth=some-token-here`
- it works with temporal-cli or SDK's, with grpc-meta, but UI only has forwarding headers, which is not suitable

To solve this issue, we introduce new env var: TEMPORAL_SERVER_STATIC_HEADERS. It is later converted to config field staticHeaders.

Format is: TEMPORAL_SERVER_STATIC_HEADERS="some-header=someval,another-header=someval"

No ui components changed.

### Design Considerations 🎨 

No ui components changed.

## Testing 🧪 

Manually tested with temporal-frontend. Added unit tests to ensure logic is correct.

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

Add staticHeaders to configs, example:

```yaml
staticHeaders:
  authorization: some-tkn
```

Or use ENV var TEMPORAL_SERVER_STATIC_HEADERS="authorization=some-tkn"

### Any docs updates needed? 

Yes, UI startup [docs](https://github.com/temporalio/documentation/blob/992cb96f1427bd98620783aa590347846cd0e4d0/docs/references/web-ui-environment-variables.mdx).

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Once you have the prerequisites going, run the following:
 pnpm install
 ```
 
-Running `pnpm install` will attempt to download and install the most recent version of [Temporal CLI][] into `./bin/cli/temporal`. The development server will attempt to use use this version of this Temporal when starting up.
+Running `pnpm install` will attempt to download and install the most recent version of [Temporal CLI][] into `./bin/cli/temporal`. The development server will attempt to use this version of this Temporal when starting up.
 
 - If that port is already in use, the UI will fallback to trying to talk to whatever process is running on that port.
 - If you do not have a version of Temporal CLI at `./bin/cli/temporal`, the development server will look for a version of Temporal CLI in your path.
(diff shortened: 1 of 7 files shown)


laurakwhit:
Adding static credentials via env. vars has potential security implications and is not something we want to enable. Especially since it has the potential to accidentally be configured or composed in unsafe ways. 

A potential solution we might consider looking into is updating the forwardHeaders config to support server-side headers in addition to the existing client-forwarded headers. This would enable the UI server to attach auth tokens (or other credentials) to gRPC calls without requiring them to come from the browser.

PR #3100, opened 2026-01-15
Use CountSchedules for counting schedules

lina-temporal:
## Description & motivation 💭 
- Makes use of the incoming [`CountSchedules API`](https://github.com/temporalio/api/pull/696). Using `ListWorkflowExecutions` to query schedules has always been private API, and it's about to break with the release of CHASM schedules. This makes use of a public API for both V1 and V2 schedules.


laurakwhit (approved):

src/lib/services/workflow-counts.ts
```diff
@@ -66,30 +70,45 @@ export const fetchWorkflowCountByExecutionStatus = async ({
...
+}): Promise<string> => {
+  try {
+    const countRoute = routeForApi('schedules.count', { namespace });
+    const { count } = await requestFromAPI<CountSchedulesResponse>(countRoute, {
+      params: query ? { query } : {},
+      notifyOnError: false,
+    });
+    return count ?? '0';
+  } catch (error: unknown) {
+    if (isNotImplemented(error)) {
```
laurakwhit: Actually, I'm seeing a `404 Not Found` error when running `pnpm dev` locally.
We might need to account for that as well. We're currently using [v1.60.1](https://github.com/temporalio/api/releases/tag/v1.60.1) in the ui-server and it looks like CountSchedules operation was added in [v.1.62.0](https://github.com/temporalio/api/releases/tag/v1.62.0).

lina-temporal: thanks - fixed, added routing from 404 as well.

PR #3102, opened 2026-01-19
[DT-3505] add iso option to timestamp formats

andrewzamojc:
### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/timezone-select.svelte
+++ b/src/lib/components/timezone-select.svelte
@@ -123,6 +123,7 @@
     id="timezones-menu"
     {position}
     class="w-[10rem] sm:w-[20rem] md:w-[28rem]"
+    maxHeight=""
   >
     <Input
       label={translate('common.search')}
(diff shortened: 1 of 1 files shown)


src/lib/pages/workflows-with-new-search.svelte
```diff
@@ -98,7 +98,7 @@
   });
 
   $effect(() => {
-    $workflowsQuery = query;
+    $workflowsQuery = query ?? '';
```
andrewzamojc, line 101: Annotation: danger file made me do it

src/lib/pages/workflows-with-new-search.svelte
```diff
@@ -228,8 +228,8 @@
 
 <header class="flex flex-col gap-2">
   <div class="flex flex-col justify-between gap-2 md:flex-row">
-    <div class="flex flex-row flex-wrap items-start gap-2">
-      <div>
+    <div class="flex flex-col gap-2">
+      <div class="flex flex-row items-center gap-1">
```
andrewzamojc, line 232: Annotation: On workflows page, just under the main heading there is a timestamp. When the "long" format was selected it pushed the refresh button away. This rejigs the flexbox to fix that.

PR #3107, opened 2026-01-22
Add Portal component

laurakwhit:
## Description & motivation 💭 
This PR introduces a new `Portal` component to the Holocene design system and integrates it with the existing `Menu` component via an opt-in `usePortal` prop. 

 `Portal` renders children at a specific position relative to an anchor element with advanced positioning features:       
  - Renders content in `document.body` to escape overflow/z-index constraints                                                                                                                                                                                            
  - Has 8 position options (`top`, `bottom`, `left`, `right`, `top-left`, `top-right`, `bottom-left`, `bottom-right`)                                                                      
  - Automatically flips position when content would overflow viewport or scroll container if `flipOnCollision` is enabled                                                                           
  - Tracks scrolling in custom containers, not just window with the optional `scrollContainer` prop                                                                                                
  - Hides portal content when the anchor element scrolls out of view if `hideWhenAnchorHidden` is enabled     
  

```svelte
<script>
  import { Portal } from '$lib/holocene/portal';
  import Button from '$lib/holocene/button.svelte';

  let isOpen = $state(false);
</script>

<Button id="portal-trigger" on:click={() => (isOpen = !isOpen)}>Toggle Portal</Button>

<Portal anchor="portal-trigger" open={isOpen}>
  <div>Portal content</div>
</Portal>
```                                                                                                       

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added
- [x] Storybook stories

### Issue(s) closed 
Related to `DT-3325`


src/lib/holocene/menu/menu.svelte
```diff
@@ -51,9 +57,13 @@
...
     keepOpen?: boolean;
     position?: 'left' | 'right' | 'top-left' | 'top-right';
-    menuElement?: HTMLUListElement;
+    menuElement?: HTMLUListElement | null;
     maxHeight?: string;
     class?: ClassNameValue;
+    usePortal?: boolean;
+    scrollContainer?: string;
+    flipOnCollision?: boolean;
+    hideWhenAnchorHidden?: boolean;
```
andrewzamojc: We might not need the last two props. Menu probably always wants to flip and hide when anchor is hidden?

src/lib/holocene/menu/menu.svelte
```diff
@@ -101,20 +131,46 @@
...
+      bind:clientHeight={height}
+      onfocusout={handleFocusOut}
+      onclick={handleClick}
+      {...rest}
+    >
+      {@render children?.()}
+    </ul>
+  </Portal>
+{:else}
+  <ul
```
andrewzamojc, line 131: Could make a snippet of this ul, so it's not repeated. I see the style attribute is different, but that could be a param to the snippet.

src/lib/holocene/portal/portal.svelte
```diff
@@ -0,0 +1,187 @@
...
+  import { twMerge as merge } from 'tailwind-merge';
+
+  import { portal } from './portal-action';
+  import { calculatePosition, getElementRect } from './position-calculator';
+  import type { PortalProps } from './types';
+
+  let {
+    anchor,
+    open = true,
+    position = 'bottom',
```
andrewzamojc, line 14: Is this mixing concepts between portal and overlay? I haven't gone this deep on portals but I didn't think they had a position like a popover...

andrewzamojc:
This is great stuff! Pretty cool to see the guts of a Portal.

src/lib/holocene/menu/menu.svelte
laurakwhit: Yeah I went back and forth here. They default to `true` but thought it might be good to surface the option. I'll remove for now and we can add back if needed in the future!

src/lib/holocene/portal/portal.svelte
laurakwhit: Yeah this is a positioned portal. We could consider having just a portal component and then handling the positioning separately, but I figured for 99% of our use cases we're going to want to position it. 

In my experience an overlay wouldn't need to be positioned since it's usually the same (centered) in every instance. Is that not your experience? Curious if you have any more insights into patterns for portals you've seen and liked!

andrewzamojc: Ya I see. Basically this is practical and always how we're going to use it 👍

src/lib/holocene/portal/portal.svelte
```diff
@@ -0,0 +1,187 @@
...
+    if (rafId !== null) {
+      cancelAnimationFrame(rafId);
+    }
+    rafId = requestAnimationFrame(() => {
+      updatePosition();
+      rafId = null;
+    });
+  }
+
+  $effect(() => {
```
GiantRobots: Can we use a named function here instead of an arrow function so we know what the intent of this effect is? ex
```
$effect(function ayooo() {
```

laurakwhit: [Updated](https://github.com/temporalio/ui/pull/3107/commits/fa9cf7a7e4048b286be73a61aec64425c41e15c9) ✅

src/lib/holocene/menu/menu.svelte
```diff
@@ -83,6 +94,21 @@
     $menuElementCtx = menuElement;
   });
 
+  $effect(() => {
```
GiantRobots, line 97: Named function please

PR #3108, opened 2026-01-22
[DT-3501] reactive timestamp

andrewzamojc:
## Description & motivation 💭 

This PR adds a new way for a page to render a timestamp such that it always respects users' timezone and format preferences. Rather than the page needing multiple stores to ensure reactivity, that is now captured in the new `timestamp` function.

```
<div>{$timestamp(someTime)}</div>

// output in the format the user has selected in the timezone popover
<div>Jan 22, 2026 12:12:12</div>
```

This change is motivated by the task to add 12/24 hour option to the timezone popover. When this choice was added, it wasn't reactive like the short/default/long choice. To make it reactive would require touching each call to formatDate to pass a new reactive parameter.

Instead of doing that, this PR replaces most calls to `formatDate` with the reactive function `timestamp`. Since this function accesses the stores internally, new functionality can be added like the 12/24 hour option, without needing to update each call again. Since `timestamp` is actually a derived function, it's return value is reactive to the user's timezone and format preferences.

Screenshot 2026-01-22 at 5 36 09 PM

## Testing 🧪 

In progress...

### How was this tested 👻 

- [X] Manual testing
- [ ] E2E tests added
- [X] Unit tests added

### DIFF
--- a/src/lib/components/timestamp.svelte
+++ b/src/lib/components/timestamp.svelte
@@ -8,31 +8,23 @@
   } from '$lib/stores/time-format';
   import {
     formatDate,
-    type TimestampFormat,
+    type FormatDateOptions,
     type ValidTime,
   } from '$lib/utilities/format-date';
-
-  type TimestampOptions = {
-    format?: TimestampFormat | 'relative';
-    relativeLabel?: string;
-  };
 
   export const timestamp = derived(
     [timeFormat, relativeTime, timestampFormat],
     ([$timeFormat, $relativeTime, $timestampFormat]) => {
       return (
         date: ValidTime | undefined | null,
-        options?: TimestampOptions,
+        options: FormatDateOptions = {},
       ): string => {
-        const isRelativeOverride = options?.format === 'relative';
-
-        const format = isRelativeOverride
-          ? $timestampFormat
-          : (options?.format ?? $timestampFormat);
-        const relative = isRelativeOverride ? true : $relativeTime;
-        const relativeLabel = options?.relativeLabel ?? undefined;
+        const format = options?.format ?? $timestampFormat;
+        const relative = options?.relative ?? $relativeTime;
+        const relativeLabel = options?.relativeLabel;
 
         return formatDate(date, $timeFormat, {
+          ...options,
           relative,
           format,
           relativeLabel,
@@ -48,17 +40,15 @@
   import type { Snippet } from 'svelte';
 
   type T = $$Generic<keyof SvelteHTMLElements>;
-  type DateTime = ValidTime | null | undefined;
 
   export { timestampSnippet as timestamp };
 
   type Props = SvelteHTMLElements[T] & {
-    dateTime: DateTime;
+    dateTime: ValidTime | null | undefined;
     as?: T;
     fallback?: string;
     leading?: Snippet<[]>;
-    // only used to override the user's stored preference in certain cases
-    overrideTimestampFormat?: TimestampFormat;
+    options?: FormatDateOptions; // overrides the user's stored preference
   };
 
   let {
@@ -66,17 +56,14 @@
     as = undefined,
     class: className = undefined,
     fallback = undefined,
-    overrideTimestampFormat = undefined,
+    options = undefined,
     leading,
     ...rest
   }: Props = $props();
 </script>
 
-{#snippet timestampSnippet(dateTime: DateTime)}
-  {$timestamp(
-    dateTime,
-    overrideTimestampFormat ? { format: overrideTimestampFormat } : undefined,
-  )}
+{#snippet timestampSnippet(dateTime: ValidTime | null | undefined)}
+  {$timestamp(dateTime, options)}
 {/snippet}
 
 {#if as}
--- a/src/lib/components/timestamp.test.ts
+++ b/src/lib/components/timestamp.test.ts
@@ -96,7 +96,7 @@ describe('timestamp', () => {
   it('should accept relative override to force relative time', () => {
     timeFormat.set('local');
     relativeTime.set(false);
-    const result = get(timestamp)(date, { format: 'relative' });
+    const result = get(timestamp)(date, { relative: true });
     expect(result).toContain('ago');
   });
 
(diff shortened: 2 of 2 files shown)


src/lib/components/deployments/version-details.svelte
```diff
@@ -21,10 +16,7 @@
     />
     <WorkflowDetail
       title={translate('deployments.rollout-started')}
-      content={formatDate(version.createTime, $timeFormat, {
-        relative: $relativeTime,
-        format: $timestampFormat,
-      })}
+      content={$timestamp(version.createTime)}
```
andrewzamojc, line 19: This happens a lot in this PR. A call to `formatDate` that needs three stores is replaced with a call to `$timestamp`.

$timestamp is a derived store that derives to a function, so you can call it! The return is reactive to user preference changes.

```
let formattedTime = $timestamp(someTime)
```

src/lib/components/event/event-summary-row.svelte
```diff
@@ -165,18 +160,10 @@
       !currentEvent?.userMetadata?.summary,
   );
 
-  const eventTime = $derived(
-    formatDate(currentEvent?.eventTime, $timeFormat, {
-      relative: $relativeTime,
-      format: $timestampFormat,
-    }),
-  );
+  const eventTime = $derived($timestamp(currentEvent?.eventTime));
```
andrewzamojc, line 163: If we put the call to `$timestamp` in a `$derived`, then the return is reactive to user preferences like format _and_ the parameter passed in like `currentEvent?.eventTime`

src/lib/components/lines-and-dots/workflow-details.svelte
```diff
@@ -122,10 +120,7 @@
   {#if workflow?.startDelay}
     <DetailListLabel>{translate('workflows.execution-start')}</DetailListLabel>
     <DetailListTextValue
-      text={formatDate(workflow?.executionTime, $timeFormat, {
-        relative: $relativeTime,
-        format: $timestampFormat,
-      })}
+      text={$timestamp(workflow?.executionTime)}
       tooltipText={formatDate(workflow?.executionTime, $timeFormat, {
         relative: !$relativeTime,
```
andrewzamojc, line 125: This shows how a developer can fall back to manually setting up a `formatDate` if the logic is special, such as this !$relativeTime business.

src/lib/components/schedule/schedule-upcoming-runs.svelte
```diff
@@ -13,11 +13,7 @@
   <h2 class="mb-4">{translate('schedules.upcoming-runs')}</h2>
   {#each futureRuns.slice(0, 5) as run}
     <div class="row">
-      <Timestamp
-        as="p"
-        dateTime={run}
-        options={{ relativeLabel: translate('common.from-now') }}
```
andrewzamojc, line 19: from-now is handled inside `$timestamp`

src/lib/components/search-attribute-filter/datetime-filter.svelte
```diff
@@ -21,18 +21,17 @@
...
     startDate,
     startHour,
     startMinute,
     startSecond,
-    TIME_UNIT_OPTIONS,
     timeFormat,
     timeFormatType,
   } from '$lib/stores/time-format';
   import { getSelectedTimezone } from '$lib/utilities/format-date';
+  import { getTimezone, TIME_UNIT_OPTIONS } from '$lib/utilities/timezone';
```
andrewzamojc, line 34: There were some circular-ish dependencies between stores and utilities. To untangle it, I created the timezone utility. 

In general, stores import utilities, so this PR removes some utilities importing from stores (mostly for constants).

src/lib/stores/time-format.ts
```diff
@@ -1,14 +1,15 @@
...
+import { type TimestampFormat } from '$lib/utilities/format-date';
 import {
-  getLocalTimezone,
-  type TimestampFormat,
-} from '$lib/utilities/format-date';
+  BASE_TIME_FORMAT_OPTIONS,
+  getAdjustedTimeformat,
+  TIME_UNIT_OPTIONS,
+  Timezones,
+} from '$lib/utilities/timezone';
```
andrewzamojc, line 12: More unwinding of store/utility interdependency. The timezone file seems to do the trick.

src/lib/utilities/timezone.ts
```diff
@@ -0,0 +1,124 @@
+import { enUS } from 'date-fns/locale';
+import * as dateTz from 'date-fns-tz';
+
+export const TIME_UNIT_OPTIONS = ['minutes', 'hours', 'days'];
+
+export const BASE_TIME_FORMAT_OPTIONS = {
+  LOCAL: 'local',
+  UTC: 'UTC',
+};
```
andrewzamojc, line 9: This stuff isn't new. It's pulled into this file from other places like time-format store. Since they're just constants, they make more sense in the utility file.

src/lib/components/lines-and-dots/workflow-pending-task.svelte
```diff
@@ -25,30 +20,15 @@
...
-      <Badge
-        >{formatDate(pendingTask.scheduledTime, $timeFormat, {
-          relative: $relativeTime,
-          format: $timestampFormat,
-        })}</Badge
-      >
+      <Badge>{$timestamp(pendingTask.scheduledTime)}</Badge>
     </p>
     <p class="flex items-center gap-4">
       {translate('workflows.started-time')}
```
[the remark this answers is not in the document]
andrewzamojc: is this my error, or just seeing it now because I touched the file and danger file is new?

src/lib/components/timestamp.svelte
```diff
@@ -11,43 +9,61 @@
...
     class: className = undefined,
     fallback = undefined,
-    overrideTimestampFormat = undefined,
+    options = undefined,
     leading,
     ...rest
   }: Props = $props();
 </script>
 
-{#snippet timestamp(dateTime: DateTime, options: FormatDateOptions)}
```
rossedfort, line 45: We can probably just get rid of this snippet now since it just wraps the $timestamp store, but up to you.

PR #3113, opened 2026-01-23
[DT-3501][DT-3505] 12/24 hour and ISO formats added to timezone popover

andrewzamojc:
## Description & motivation 💭 

This PR adds 24 hour option to the time formatting popover, as well as an ISO format option. This lets the user see times in the app, formatted in different ways that they're used to.

This new functionality is reactive throughout the app thanks to the recent `$timestamp` function, which formats dates reactively based on the user's format preferences. 12/24 hour options are disabled for ISO format because it does not have these variants.

Screenshot 2026-01-26 at 12 04 45 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

### Issue(s) closed 
https://temporalio.atlassian.net/browse/DT-3501
https://temporalio.atlassian.net/browse/DT-3505

### DIFF
--- a/src/lib/components/timezone-select.svelte
+++ b/src/lib/components/timezone-select.svelte
@@ -176,6 +176,11 @@
             active={$timestampFormat === 'long'}
             on:click={() => setTimestampFormat('long')}>Long</ToggleButton
           >
+          <ToggleButton
+            size="xs"
+            active={$timestampFormat === 'iso'}
+            on:click={() => setTimestampFormat('iso')}>ISO</ToggleButton
+          >
         </ToggleButtons>
       </div>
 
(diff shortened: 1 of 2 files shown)


src/lib/components/timezone-select.svelte
```diff
@@ -177,8 +176,47 @@
             active={$timestampFormat === 'long'}
             on:click={() => setTimestampFormat('long')}>Long</ToggleButton
           >
+          <ToggleButton
+            size="xs"
+            active={$timestampFormat === 'iso'}
+            on:click={() => setTimestampFormat('iso')}>ISO</ToggleButton
+          >
```
rossedfort, line 183: I initially thought we might need a different sort of control for this as the list grew, however it seems to work fine. May just want to run by design.

andrewzamojc: For sure. I've got some UX tweaks I'd like to propose after this PR and I'll get UX looped in on those.

PR #3119, opened 2026-01-27
[DT-3580] Add tooltips to worker deployment version status badges

rossnelson:
## Summary
- Add descriptive tooltips to deployment status badges explaining what each state means (Current, Ramping, Draining, Drained, Inactive)
- Tooltips help users understand the lifecycle states of worker deployments

## Test plan
- [ ] Verify tooltips appear on hover for each status badge type
- [ ] Confirm tooltip text matches the status being displayed

### DIFF
--- a/src/lib/pages/deployment.svelte
+++ b/src/lib/pages/deployment.svelte
@@ -38,9 +38,7 @@
   ];
 </script>
 
-<header
-  class="flex flex-row flex-wrap justify-between gap-8 border-b border-subtle px-4 pb-4 pt-8 md:pt-20 xl:px-8"
->
+<header class="flex flex-row flex-wrap justify-between gap-8">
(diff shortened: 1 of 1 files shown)


src/lib/components/deployments/deployment-status.svelte
```diff
@@ -35,6 +45,8 @@
   );
 </script>
 
-<p class={deploymentStatus({ status })}>
-  <Icon name={icon[status]} />{label}
-</p>
+<Tooltip text={tooltip[status]} top width={250}>
```
Alex-Tideman: topLeft is a little easier to read, keeps it lined up better at start of sentence

rossnelson: done!

PR #3128, opened 2026-02-02
feat: add long-term caching for immutable assets

autocracy:
## Description & motivation 💭 
Add Cache-Control headers for /_app/immutable/* paths to enable aggressive browser caching of static content-addressable filenames. This has a strong performance impact when the UI is behind a proxy that is preventing multiplexing connections. Without it, I observe slow serial downloading of these static assets.

- Set Cache-Control: public, max-age=31536000, immutable (one year)
- Applied in hooks.server.ts

### How was this tested 👻 

- [ X ] Manual testing
- [ ] E2E tests added
- [ X ] Unit tests added

Add comprehensive test coverage for Cache-Control header middleware in hooks.server.ts:
- Verify headers are set correctly for /_app/immutable/* paths
- Verify headers are not set for other paths
- Test preservation of existing headers
- Cover various immutable asset subdirectories

Observe header application and local caching using developer tools in browser.

### DIFF
--- a/src/hooks.server.ts
+++ b/src/hooks.server.ts
@@ -1,12 +1,14 @@
 import type { Handle } from '@sveltejs/kit';
 
+const ONE_YEAR = 31536000; // 365 days in seconds
+
 export const handle: Handle = async ({ event, resolve }) => {
   const response = await resolve(event, {});
 
-  if (event.url.pathname.startsWith('/_app/immutable/')) {
+  if (response.ok && event.url.pathname.startsWith('/_app/immutable/')) {
     response.headers.set(
       'Cache-Control',
-      'public, max-age=31536000, immutable',
+      `public, max-age=${ONE_YEAR}, immutable`,
     );
   }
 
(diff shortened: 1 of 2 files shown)


src/hooks.server.ts
```diff
@@ -3,6 +3,13 @@ import type { Handle } from '@sveltejs/kit';
 export const handle: Handle = async ({ event, resolve }) => {
   const response = await resolve(event, {});
 
+  if (event.url.pathname.startsWith('/_app/immutable/')) {
+    response.headers.set(
+      'Cache-Control',
+      'public, max-age=31536000, immutable',
```
laurakwhit: NIT: Might be helpful for readability to make a constant for cache duration e.g. 
```suggestion
      'public, max-age=${ONE_YEAR}, immutable',
```

src/hooks.server.ts
```diff
@@ -3,6 +3,13 @@ import type { Handle } from '@sveltejs/kit';
 export const handle: Handle = async ({ event, resolve }) => {
   const response = await resolve(event, {});
 
+  if (event.url.pathname.startsWith('/_app/immutable/')) {
+    response.headers.set(
```
laurakwhit, line 9: Let's add a check for successful responses here so we're not caching errors as well.

PR #3131, opened 2026-02-03
fix: prevent 'No Workers Running' flash during workflow load

rossnelson:
## Summary

Fixes a race condition where the "No Workers Running" warning would briefly flash when opening a workflow, even when workers were actively running.

**Root cause:** The `workflow-call-stack-error` component checked `!workers?.pollers?.length` to show the warning, but the initial store state has `pollers: []`. When a workflow loaded, the warning would briefly appear before worker data finished loading.

**Fix:** Added a `workersLoaded` flag to the store to distinguish between:
- "Workers are loading" (`workersLoaded: false`, `pollers: []`)
- "Workers loaded with none found" (`workersLoaded: true`, `pollers: []`)

The warning now only shows in the second case.

## Changes

- `src/lib/stores/workflow-run.ts` - Added `workersLoaded: boolean` to store type
- `src/lib/layouts/workflow-run-layout.svelte` - Sets `workersLoaded: true` after fetching
- `src/lib/components/workflow/workflow-call-stack-error.svelte` - Checks `workersLoaded` before showing warning

## Test plan

- [ ] Open a workflow that has workers running - should NOT see the "No Workers Running" flash
- [ ] Open a workflow with no workers polling its task queue - should still see the warning (after load completes)
- [ ] Navigate between workflows - warning should not flash during transitions

### DIFF
--- a/src/lib/components/workflow/workflow-call-stack-error.svelte
+++ b/src/lib/components/workflow/workflow-call-stack-error.svelte
@@ -18,7 +18,7 @@
       title={translate('workflows.workflow-error-no-workers-title')}
     >
       {translate('workflows.workflow-error-no-workers-description', {
-        taskQueue: workflow?.taskQueue,
+        taskQueue: workflow?.taskQueue ?? '',
       })}
     </Alert>
   </div>
--- a/src/lib/layouts/workflow-run-layout.svelte
+++ b/src/lib/layouts/workflow-run-layout.svelte
@@ -39,9 +39,9 @@
   $: showJson = $page.url.searchParams.has('json');
   $: fullJson = { ...$workflowRun, eventHistory: $fullEventHistory };
 
-  let workflowError: NetworkError;
+  let workflowError: NetworkError | null = null;
   let workflowRunController: AbortController;
-  let refreshInterval;
+  let refreshInterval: ReturnType<typeof setInterval> | null = null;
 
   const { copy, copied } = copyToClipboard();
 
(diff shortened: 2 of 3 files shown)


src/lib/layouts/workflow-run-layout.svelte
```diff
@@ -97,7 +97,7 @@
     const { taskQueue } = workflow;
     const workers = await getPollers({ queue: taskQueue, namespace });
 
-    $workflowRun = { ...$workflowRun, workflow, workers };
+    $workflowRun = { ...$workflowRun, workflow, workers, workersLoaded: true };
```
Alex-Tideman, line 104: Do we ever set this back to false onDestroy/navigate away from workflow page?

src/lib/components/workflow/workflow-call-stack-error.svelte
```diff
@@ -5,8 +5,9 @@
   import { translate } from '$lib/i18n/translate';
   import { workflowRun } from '$lib/stores/workflow-run';
 
-  $: ({ workflow, workers } = $workflowRun);
-  $: runningWithNoWorkers = workflow?.isRunning && !workers?.pollers?.length;
+  $: ({ workflow, workers, workersLoaded } = $workflowRun);
+  $: runningWithNoWorkers =
+    workersLoaded && workflow?.isRunning && !workers?.pollers?.length;
```
andrewzamojc, line 10: Makes sense to me.

PR #3148, opened 2026-02-10
Fix subtitle size and padding on proj nav

KATIETOLER:
Slight fix that does two things 
- makes subtitle size smaller on project nav 
- adds padding to new subtitles to align subtitle when small with nav items (while leaving original style intact)

### DIFF
--- a/src/lib/holocene/navigation/cloud-nav-bar.svelte
+++ b/src/lib/holocene/navigation/cloud-nav-bar.svelte
@@ -38,7 +38,7 @@
         class={merge(
           'font-secondary mb-0 hidden whitespace-nowrap font-medium text-white group-data-[nav=open]:block',
           subtitle === 'Cloud' ? 'text-2xl' : 'text-base',
-          subtitle === 'Cloud' ? '' : 'px-2',
+          subtitle !== 'Cloud' && 'px-2',
         )}
       >
         {subtitle}


src/lib/holocene/navigation/cloud-nav-bar.svelte
```diff
@@ -66,5 +73,6 @@
   >
     <span class="sr-only">{translate('common.version')}</span>
     {version}
+    x
```
laurakwhit: Was this intentional?

src/lib/holocene/navigation/cloud-nav-bar.svelte
```diff
@@ -33,11 +33,18 @@
       width={24}
       class={merge('m-1', subtitle !== 'Cloud' && 'hidden')}
     />
-    {#if subtitle}
+    {#if subtitle === 'Cloud'}
```
laurakwhit: Curious why this change is needed? We can just add additional conditions (e.g. `subtitle !== 'Cloud' ? 'text-2xl' : 'text-lg'`) to `merge()` if needed, no?

src/lib/holocene/navigation/cloud-nav-bar.svelte
KATIETOLER: no :(

src/lib/holocene/navigation/cloud-nav-bar.svelte
laurakwhit: > We needed additional padding on the smaller 'new nav' subtitle so that it aligns with the nav items. If this padding is on the old nav it makes the 'Cloud' word off centered.

I believe [this change](https://github.com/temporalio/ui/pull/3144/changes#diff-a5cdbaa3b4a73e335902964044f0966455f86eac1acb655667aae109b7c6deb2R44) should have fixed the padding. Can you verify this is still happening on the latest?

KATIETOLER: Like the font size, the padding is only needed on the 'new nav' and very much not needed on the 'old nav.'
It's all a bit messy with all these conditionals 😅. 
I will switch to the tw Merge for both. 

Screenshot 2026-02-10 at 4 14 01 PM
Screenshot 2026-02-10 at 4 36 14 PM

src/lib/holocene/navigation/cloud-nav-bar.svelte
```diff
@@ -37,7 +37,8 @@
       <h2
         class={merge(
           'font-secondary mb-0 hidden whitespace-nowrap font-medium text-white group-data-[nav=open]:block',
-          subtitle !== 'Cloud' ? 'text-2xl' : 'text-lg',
+          subtitle === 'Cloud' ? 'text-2xl' : 'text-base',
+          subtitle === 'Cloud' ? '' : 'px-2',
```
laurakwhit: So to clarify, with `Cloud` subtitle (old nav) we want **no** extra padding and text-2xl?

```suggestion
          subtitle !== 'Cloud' && 'px-2',
```

KATIETOLER: yes, that's correct! Thanks!

PR #3154, opened 2026-02-11
Add fallbacks for activities routes

laurakwhit:
## Description & motivation 💭 

Falls back to `/activities/...` routes if `/activities-deprecated/...` routes are not found.

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/services/workflow-activities-service.ts
+++ b/src/lib/services/workflow-activities-service.ts
@@ -1,3 +1,5 @@
+import { page } from '$app/state';
+
 import type {
   ActivityPauseRequest,
   ActivityPauseResponse,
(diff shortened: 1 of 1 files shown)


src/lib/services/workflow-activities-service.ts
```diff
@@ -12,22 +14,29 @@ import { isNotFound, isNotImplemented } from '$lib/utilities/handle-error';
...
   route: string,
   init: Parameters<typeof requestFromAPI>[1],
 ): Promise<T> => {
+  const fallbackRoute = route.replace(
+    '/activities-deprecated/',
+    '/activities/',
+  );
+  const version = page.data?.settings?.version;
+
+  if (version && !minimumVersionRequired('2.45.3', version)) {
```
laurakwhit, line 29: > Can we know which route to used ahead of time based on a version number or do we have to make a request?

@rossedfort good point! Added a `minimumVersionRequiredCheck`.

PR #3177, opened 2026-02-23
Fix some minor bugs in the Standalone Activities UI

rossedfort:
## Description & motivation 💭 
Fix some minor bugs in the Standalone Activities UI
- Add custom search attributes to filter dropdown on list page
- Fix unnecessary error state in activity metadata component when activity has no metadata
- Move polling logic into the shared boxes and arrows component and unset data onDestroy. Solves a bug where UI would have stale activity execution info

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/standalone-activities/activities-summary-filter-bar/filter.svelte
+++ b/src/lib/components/standalone-activities/activities-summary-filter-bar/filter.svelte
@@ -20,7 +20,6 @@
 
   import type { SearchAttributeFilter } from '$lib/models/search-attribute-filters';
   import { activityFilters } from '$lib/stores/filters';
-  import { activitySearchAttributeOptions } from '$lib/stores/search-attributes';
   import { emptyFilter } from '$lib/utilities/query/to-list-workflow-filters';
 
   import ActivityDropdownFilterList from './dropdown-filter-list.svelte';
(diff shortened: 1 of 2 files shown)


src/lib/components/standalone-activities/activities-summary-filter-bar/filter.svelte
```diff
@@ -66,5 +64,5 @@
 
 <div class="flex shrink flex-wrap items-center justify-start gap-2">
   <ActivityDropdownFilterList />
-  <ActivitySearchAttributeMenu {options} />
+  <ActivitySearchAttributeMenu options={$activitySearchAttributeOptions} />
```
laurakwhit: NIT: Do we need to pass these options as props or can we just import them directly in the component?

src/lib/layouts/standalone-activity-layout.svelte
```diff
@@ -65,12 +77,26 @@
...
+      queue: $activityExecution.info.taskQueue,
+      namespace,
+    });
+  });
+
+  onMount(async () => {
+    poller.start();
+  });
+
+  onDestroy(() => {
```
laurakwhit, line 93: NIT: Can this just be the return function in `onMount()`?

rossedfort: Good thought, but TIL we can't because the callback fn to `onMount` is async

https://svelte.dev/docs/svelte/lifecycle-hooks#onMount

PR #3181, opened 2026-02-24
fix(workflow): remove expand/minimize button from Event History

rossnelson:
## Summary
- Removes the Expanded/Minimized toggle button from the Event History toolbar
- The expand/minimize toggle controls Timeline view behavior and is not relevant in the Event History tab
- Cleans up the unused `Tooltip` import

## Test plan
- [ ] Verify the Event History toolbar no longer shows the Expanded/Minimized button
- [ ] Confirm Descending, Filter, Auto Refresh, and Download buttons still render correctly
- [ ] Verify the Timeline tab's expand/minimize behavior is unaffected


src/lib/layouts/workflow-history-layout.svelte
```diff
@@ -176,23 +175,6 @@
               : translate('common.ascending')}
           </ToggleButton>
         {/if}
-        <Tooltip
-          text={$minimizeEventView
```
Alex-Tideman, line 180: We should be able to rip this out and delete it. Just double check its no longer used anywhere

PR #3184, opened 2026-02-24
fix(workflow): remove max height from event history table and default to timeline

rossnelson:
## Summary
- Removes the constrained `maxHeight` on the event history table (`calc(100vh - 555px)`) so the browser window scrolls instead of the table scrolling independently
- Updates links that navigate to other workflows to point to the Timeline tab instead of Event History
- Updates the workflow detail page redirect to default to `/timeline`

### Changes
- **`event-summary-table.svelte`** — `maxHeight` changed from conditional calc to `"none"`
- **`workflow-history-layout.svelte`** — removed `minimized` prop from `EventSummaryTable`
- **`[run]/+page.ts`** — redirect `/history` → `/timeline`
- **`event-details-link.svelte`** — execution and child-workflow links use `routeForTimeline` instead of `routeForEventHistory`
- **`workflow-header.svelte`** — reset workflow success link uses `routeForTimeline`

## Test plan
- [ ] Open Event History tab — table should not have its own scrollbar, window scrolls instead
- [ ] Verify pagination still works correctly
- [ ] Navigate to a workflow — should land on Timeline tab by default
- [ ] Click a child workflow link in event details — should land on Timeline tab
- [ ] Click an execution link in event details — should land on Timeline tab
- [ ] Reset a workflow and click the "here" link — should land on Timeline tab
- [ ] Timeline tab behavior unchanged (graph visibility still controlled by minimized prop)

### DIFF
--- a/src/lib/utilities/event-link-href.test.ts
+++ b/src/lib/utilities/event-link-href.test.ts
@@ -57,7 +57,7 @@ describe('getEventLinkHref', () => {
     );
   });
 
-  it('should return event history route as fallback', () => {
+  it('should return timeline route as fallback', () => {
     const link = {
       workflowEvent: {
         namespace: 'test-ns',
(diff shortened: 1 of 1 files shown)


Alex-Tideman:
I think we need to do an audit of routeForEventHistory use and switch to a new routeForTimeline if that should be the default

PR #3187, opened 2026-02-25
Add back event history legend and auto refresh

laurakwhit:
## Description & motivation 💭 

* Adds back the [event history legend ](https://github.com/temporalio/ui/pull/3160) that was removed with changes in https://github.com/temporalio/ui/pull/3109. This PR adds back the legend in `/timeline` view and adds an event type legend to the `Event Type` column in the table in `/history` view. It also improves accessibility for the category icons by adding a `title` prop that describes the event type.
* Adds the [auto refresh UI update](https://github.com/temporalio/ui/pull/3137) to the `/timeline` view.

| Timeline | Event History |
|-|-|
|Screenshot 2026-02-24 at 8 24 00 PM|Screenshot 2026-02-24 at 8 24 23 PM|

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/layouts/workflow-history-layout.svelte
+++ b/src/lib/layouts/workflow-history-layout.svelte
@@ -46,7 +46,7 @@
   $effect(() => {
     const urlParams = parseEventFilterParams(page.url);
     $eventFilterSort = urlParams.sort;
-    $pauseLiveUpdates = urlParams.frozen;
+    $pauseLiveUpdates = urlParams.refresh_off;
   });
 
   $effect(() => {
@@ -108,8 +108,12 @@
     updateEventFilterParams(page.url, { sort: newSort }, goto);
   };
 
-  const onFreezeToggle = () => {
-    updateEventFilterParams(page.url, { frozen: !$pauseLiveUpdates }, goto);
+  const onAutoRefreshToggle = () => {
+    updateEventFilterParams(
+      page.url,
+      { refresh_off: !$pauseLiveUpdates },
+      goto,
+    );
   };
 
   const onAllClick = () => {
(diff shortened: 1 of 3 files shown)


src/lib/components/lines-and-dots/svg/line.svelte
```diff
@@ -50,14 +50,17 @@
...
+    }
     if (classification) {
       color = getStatusStrokeColor(classification);
     }
     if (delayed && [classification, status].includes('Running')) {
       color = getStatusStrokeColor('Delayed');
     }
-
-    if (category) {
+    if (category && ['pending', 'retry'].includes(category)) {
```
Alex-Tideman, line 62: do we need to add paused here too?

src/lib/components/lines-and-dots/event-history-legend.svelte
```diff
@@ -113,7 +115,7 @@
       category,
     })}"
   >
-    <Icon name={CategoryIcon[category]} class="h-3.5 w-3.5 shrink-0" />
+    <Icon name={CategoryIcon[category].name} class="h-3.5 w-3.5 shrink-0" />
```
Alex-Tideman: Only worry here is if it's a new event we aren't handling and it blows up with undefined.name. So might want a fallback or ?. check on all these CategoryIcon

laurakwhit: I checked with `strict` mode enabled. These are all currently typed to be an `EventTypeCategory`.

src/lib/components/lines-and-dots/svg/line.svelte
laurakwhit: I just based this change of the order of css classes [here](https://github.com/temporalio/ui/pull/3160/changes#diff-199632ce987e7c3233f36c158ebd5082ddaaf672ebdf9c453d648451c621a2deL83-L145). We didn't have a color for `paused`, but we could look into adding one.

Alex-Tideman: 👍 sounds good

PR #3196, opened 2026-02-26
Workflow pause updates

laurakwhit:
## Description & motivation 💭 

  - Treats `Paused` workflows similar to `Running` across the UI — activity actions, termination, cancel-in-progress checks, "no workers" warnings, and "results pending" placeholder for input and results.
  - Enable "auto refresh" button for a paused workflow.
  - Adds `workflowExecutionPausedEventAttributes` and `workflowExecutionUnpausedEventAttributes` to the event attribute keys so that the attributes are displayed in the event details.

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

On a `Paused` Workflow with pending activities > activity pause, unpause, update, or reset 
   - [x] Should appear on pending activities
   - [x] Should work as expected and trigger a page refresh when selected

When selecting `Pause Workflow` and "auto refresh" is enabled
   - [x] "auto refresh" should remain on
   - [x] "auto refresh" can be toggled on/off

If a Workflow is `Running` or `Paused` 
   - [x] "results" should show "Results will appear upon completion."

The workflow callstack error should show if there are no workers
   - [x] the Workflow is `Running` or `Paused`

When batch terminating Workflows, they should include
   - [x] Workflows that are `Running` or `Paused`

If a cancel is in progress and a Workflow is `Running` or `Paused`
   - [x] an alert should be shown

The event details for `Workflow Execution Paused` and `Workflow Execution Unpaused` events
   - [x] should list the identity, reason, and request ID

### DIFF
--- a/src/lib/components/workflow/client-actions/pause-confirmation-modal.svelte
+++ b/src/lib/components/workflow/client-actions/pause-confirmation-modal.svelte
@@ -43,7 +43,6 @@
       open = false;
       reason = '';
       triggerRefresh(Action.Pause);
-      $pauseLiveUpdates = true;
       toaster.push({
         id: 'workflow-pause-success-toast',
         message: translate('workflows.pause-success'),
--- a/src/lib/components/lines-and-dots/end-time-interval.svelte
+++ b/src/lib/components/lines-and-dots/end-time-interval.svelte
@@ -35,7 +35,7 @@
     if (pauseLiveUpdates) {
       clearInterval(endTimeInterval);
       endTimeInterval = null;
-    } else if (!endTimeInterval && workflow.isRunning) {
+    } else if (!endTimeInterval && (workflow.isRunning || workflow.isPaused)) {
       endTimeInterval = setInterval(() => {
         endTime = rightNow();
       }, 1000);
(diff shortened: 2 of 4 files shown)


src/lib/components/workflow/client-actions/pause-confirmation-modal.svelte
```diff
@@ -42,6 +43,7 @@
       open = false;
       reason = '';
       triggerRefresh(Action.Pause);
+      $pauseLiveUpdates = true;
```
Alex-Tideman: Do we want to pause live updates? I think we want to keep it live since they would want to know if in-flight things finish

laurakwhit: Yeah I'm not 100% sure what we want here. I just added this because we were disabling the "auto refresh" button and I wanted it to be consistent. We could keep the button enabled and allow the user to turn off/on. However, once those in-flight activities are finished it seems like we wouldn't want to keep polling... Not sure how we'd easily determine that though.

laurakwhit: Updated to [keep the button enabled](https://github.com/temporalio/ui/pull/3196/commits/7bea1792fed2b9a05b3cf31b971f1ac7d030fde3) on paused workflows for now.

PR #3212, opened 2026-03-05
Add initial timestamp values into hour/minute/second inputs

Alex-Tideman:
## Description & motivation 💭 
When reopening a datetime filter chip (e.g., StartTime >= 3/5/26, 9:15:00 AM CST), the hour/minute/second inputs were empty instead of showing the previously set values.

  Root cause: The dropdown-filter-chip.svelte component initialized its local time state to empty strings on every mount, never parsing the existing filter's ISO timestamp value back into date/time components.

  Fix:
  - Parse the filter's stored UTC ISO value back into date, hour, minute, and second using utcToZonedTime to respect the user's selected timezone
  - Handles both single datetime (>=, <=) and BETWEEN range filters
  - Also fixed the "add new filter" flow (datetime-filter.svelte) to initialize from persisted stores so values carry over
  - Extracted parsing logic into a shared utility (datetime-filter-parse.ts) with 12 unit tests covering single datetime, BETWEEN ranges, timezone
  conversion, and edge cases

Screenshot 2026-03-05 at 2 19 09 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

### DIFF
--- a/src/lib/components/workflow/filter-bar/dropdown-filter-chip.svelte
+++ b/src/lib/components/workflow/filter-bar/dropdown-filter-chip.svelte
@@ -3,6 +3,7 @@
 
   import { addHours, addMinutes, addSeconds, startOfDay } from 'date-fns';
   import { zonedTimeToUtc } from 'date-fns-tz';
+  import { untrack } from 'svelte';
 
   import { timestamp } from '$lib/components/timestamp.svelte';
   import Button from '$lib/holocene/button.svelte';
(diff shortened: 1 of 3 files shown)


src/lib/components/search-attribute-filter/datetime-filter.svelte
```diff
@@ -39,16 +39,16 @@
...
   const { filter, handleSubmit } = getContext<FilterContext>(FILTER_CONTEXT);
 
-  // Local state for date/time values - defaults to today
-  let localStartDate = $state(startOfDay(new Date()));
-  let localStartHour = $state('');
-  let localStartMinute = $state('');
-  let localStartSecond = $state('');
+  // Local state for date/time values - initialize from stores to preserve previous values
+  let localStartDate = $state($startDate ?? startOfDay(new Date()));
+  let localStartHour = $state($startHour ?? '');
```
laurakwhit, line 44: Do we want to update the values in the inputs if the user changes their timezone? E.g. user selected filter with UTC, then changes to PST and opens filter again (sees UTC inputs). 

Image

laurakwhit: Updated [here](https://github.com/temporalio/ui/pull/3212/commits/ece80a5c58019d1ae32644904db78313e45a2d7e) ✅

PR #3215, opened 2026-03-09
Fix custom search attributes inputs for different attribute types

laurakwhit:
<!--- Note to EXTERNAL Contributors -->
<!-- Thanks for opening a PR!
If it is a significant code change, please **make sure there is an open issue** for this.
We work best with you when we have accepted the idea first before you code. -->

<!--- For ALL Contributors 👇 -->

## Description & motivation 💭 <!-- Describe what has changed in this PR and the motivation behind it -->
Custom search attribute input was not updating based on the attribute type in the "Start a Workflow" form. E.g when selecting a custom search attribute with a `Bool` type, the value input stayed as a text input instead of switching to a true/false select. 

  - Introduced local `$state` variables (type, value, label) instead of reading directly from attribute (which isn't deeply reactive through `$bindable()`)
  - All inputs now `bind:value` to local state, with an `$effect` syncing value back to `attribute.value` only when the local state value changes
  - `handleAttributeChange` updates local state first, then reassigns `attribute = { label, value, type }` to notify the parent
  - Changed `$derived(attributesToAdd)` to `$derived([...attributesToAdd])` so that when a child updates its attribute via binding, the derived value produces a new array reference, causing all rows  to receive updated attributes and re-evaluate `isDisabled`
  - Added integration tests for custom search attributes


### Screenshots (if applicable) 📸 <!-- Add screenshots or videos -->

### Design Considerations 🎨 <!-- Any questions, concerns, thoughts for Design? -->

## Testing 🧪 <!-- Describe what has changed in this PR and the motivation behind it -->

### How was this tested 👻 <!--- Please describe how you tested your changes and tests that were added -->

- [x] Manual testing
- [x] E2E tests added
- [ ] Unit tests added

### Steps for others to test: 🚶🏽‍♂️🚶🏽‍♀️ <!--- Please describe how we can test the changes in the PR -->

 **Setup**                                                                                                                                                                                            
                                                            
  1. Ensure you have custom search attributes configured (Bool, Keyword, Int, Double, Datetime, KeywordList types)                                                                                 
  2. Navigate to `/workflows/start-workflow` page                                                                                                                                                               
  3. Click "More options" to expand the search attributes section
---

  - Click "Add" to add a search attribute row
     - [ ] Select a `Bool` attribute → verify a true/false select appears and a value can be entered
     - [ ] Select a `Keyword` attribute → verify a text input appears and a value can be entered
     - [ ] Select an `Int` attribute → verify a number input appears and a value can be entered
     - [ ] Select a `Datetime` attribute → verify a datetime input appears and a value can be entered
     - [ ] Select a `KeywordList` attribute → verify a chip input appears and a value can be entered

 - Click the close button on the rows
   - [ ] Verify the attributes are removed

 - Select a `Keyword` attribute 
    - [ ] Change to `Bool` → verify text input is replaced with true/false Select
    - [ ] Change to `Int` → verify Select is replaced with number input
    - [ ] Change back to `Keyword` → verify number input is replaced with text input

 - Add a row and select a `Bool` attribute
     - [ ] Add a second row → open the "Search Attribute" select → verify the `Bool` option is disabled/greyed out
     - [ ] Verify the other attributes are still selectable
     - [ ] In the first row, change from the `Bool` attribute to the `Keyword` attribute → open the second row's "Search Attribute" select → verify the `Bool` attribute option is now enabled and the `Keyword` attribute is disabled

## Checklists

### Draft Checklist <!-- Add todos if not ready to review -->

### Merge Checklist <!-- Add todos if not ready to merge -->

### Issue(s) closed <!-- add issue number here -->
`DT-3653`

## Docs

### Any docs updates needed? <!--- Update README if applicable or point out where to update docs.temporal.io -->

### DIFF
--- a/src/lib/components/workflow/search-attribute-input/index.svelte
+++ b/src/lib/components/workflow/search-attribute-input/index.svelte
@@ -33,6 +33,8 @@
 
   $effect(() => {
     const v = value;
+    // effect should only re-run when the local value state changes
+    // changes to attribute should not re-trigger the effect
     untrack(() => {
       attribute.value = v;
     });


src/lib/components/workflow/search-attribute-input/index.svelte
```diff
@@ -26,26 +27,37 @@
...
-  let _label = $derived(attribute.label || (label && label.toString()));
+  let type = $state(attribute.type);
+  let value = $state(attribute.value);
+  let label = $state(attribute.label);
+
+  $effect(() => {
+    const v = value;
+    untrack(() => {
+      attribute.value = v;
+    });
```
rossedfort, line 40: Is this `untrack` here because we don't want the effect to run when `attribute.value` changes? Maybe add a comment explaining.

laurakwhit: Yeah we only want the effect to run when local `value` state changes, not when there are changes to the `attribute`. 

Added a comment [here](https://github.com/temporalio/ui/pull/3215/commits/14e41a2654c894f6ffd9404b2cd9d029e7455012) ✅

PR #3218, opened 2026-03-10
Add unique id to filter

laurakwhit:
## Description & motivation 💭 
Adds a unique id field to `SearchAttributeFilter` so Svelte can correctly track which component belongs to which filter and fixes a bug when removing a filter when there are multiple filters with the same attribute.

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

On the `/namespaces/{namespace}/workflows` page > `+ Add Filter` for `StartTime` > add another filter for `StartTime` > select the first `StartTime` filter chip and `Remove`
   - [ ] Verify the dropdown closes
   - [ ] Verify the remaining `StartTime` filter value is the second `StartTime` filter that was added

### DIFF
--- a/src/lib/utilities/query/to-list-workflow-filters.ts
+++ b/src/lib/utilities/query/to-list-workflow-filters.ts
@@ -54,13 +54,22 @@ export const getLargestDurationUnit = (duration: Duration): Duration => {
 const isDatetimeStatement = is(SEARCH_ATTRIBUTE_TYPE.DATETIME);
 const isBoolStatement = is(SEARCH_ATTRIBUTE_TYPE.BOOL);
 
-export const emptyFilter = (): SearchAttributeFilter => ({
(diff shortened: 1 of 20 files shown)


src/lib/utilities/query/to-list-workflow-filters.ts
```diff
@@ -54,13 +54,22 @@ export const getLargestDurationUnit = (duration: Duration): Duration => {
...
   operator: '',
   parenthesis: '',
   conditional: '',
+};
+
+export const createFilter = (
+  overrides: Partial<SearchAttributeFilter> = {},
+): SearchAttributeFilter => ({
+  ...defaultFilter,
+  id: crypto.randomUUID(),
```
Alex-Tideman, line 71: Good call

PR #3229, opened 2026-03-18
feat: add capability-aware feature flag system

rossnelson:
## Summary

- Adds `featureFlags` persistStore holding a `Record<string, boolean>` under the `featureFlags` localStorage key — the local override layer
- Extends the `Capabilities` type with `serverlessWorkers?: boolean | null` for forward compatibility with the proto
- Adds `withLocalFallback` and `isCapabilityEnabled` to `capability-enablement.ts` — server capability is authoritative when present, localStorage is the fallback when absent
- Adds `CapabilityGuard` component for declarative feature gating

https://github.com/user-attachments/assets/b1d055e1-d1bd-4357-9b4a-6f4f54ac90b5

## How it works

| Server capability | localStorage | Result |
|---|---|---|
| `true` | anything | ✅ enabled |
| `false` | `true` | ❌ disabled (server wins) |
| absent | `true` | ✅ enabled |
| absent | absent/false | ❌ disabled |

## Usage

```svelte
<CapabilityGuard capability="serverlessWorkers">
  <NewUI />
  {#snippet fallback()}
    <OldUI />
  {/snippet}
</CapabilityGuard>
```

Enable locally until the server reports the capability:
```js
localStorage.setItem('featureFlags.serverlessDeployments', 'true')
// then refresh
```

Adding a new capability with local override = one line in `CAPABILITY_FLAG_MAP`.

## Test plan

- [ ] `pnpm lint` passes with no new errors
- [ ] `pnpm check` passes with no new errors
- [ ] `pnpm test -- --run` passes all unit tests with no regressions
- [ ] `CapabilityGuard` renders children when capability/flag enabled, fallback when disabled, nothing when disabled with no fallback
- [ ] Server capability `false` overrides localStorage `true`
- [ ] localStorage fallback works when server capability is absent

### DIFF
--- a/src/lib/stores/capability-enablement.ts
+++ b/src/lib/stores/capability-enablement.ts
@@ -13,28 +13,20 @@ const LOCAL_OVERRIDE_CAPABILITIES = new Set<keyof Capabilities>([
   'serverlessDeployments',
 ]);
 
-function withLocalFallback(
-  capabilityKey: keyof Capabilities,
-): Readable<boolean> {
-  const localStore = getFlagStore(String(capabilityKey));
-  return derived([page, localStore], ([$page, $local]) => {
-    const serverValue = $page.data?.systemInfo?.capabilities?.[capabilityKey];
-    if (serverValue !== undefined && serverValue !== null) {
-      return Boolean(serverValue);
-    }
-    return $local;
-  });
-}
-
 export function isCapabilityEnabled(
   key: keyof Capabilities,
 ): Readable<boolean> {
-  if (LOCAL_OVERRIDE_CAPABILITIES.has(key)) {
-    return withLocalFallback(key);
+  if (!LOCAL_OVERRIDE_CAPABILITIES.has(key)) {
+    return derived(page, ($page) =>
+      Boolean($page.data?.systemInfo?.capabilities?.[key]),
+    );
   }
-  return derived(page, ($page) =>
-    Boolean($page.data?.systemInfo?.capabilities?.[key]),
-  );
+
+  const localStore = getFlagStore(String(key));
+  return derived([page, localStore], ([$page, $local]) => {
+    const serverValue = $page.data?.systemInfo?.capabilities?.[key];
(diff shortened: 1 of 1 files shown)


src/lib/stores/capability-enablement.ts
```diff
@@ -1,11 +1,42 @@
...
+
+function withLocalFallback(
+  capabilityKey: keyof Capabilities,
+): Readable<boolean> {
+  const localStore = getFlagStore(String(capabilityKey));
+  return derived([page, localStore], ([$page, $local]) => {
+    const serverValue = $page.data?.systemInfo?.capabilities?.[capabilityKey];
+    if (serverValue !== undefined && serverValue !== null) {
+      return Boolean(serverValue);
+    }
```
Alex-Tideman: Curious why we need this when below we do the same capability check in isCapabilityEnabled

rossnelson: It's not a redundant check. It ensures that the system-info value takes precedence. But i can see how the way its written makes it seem redundant. Ill update it to be more clear.

src/lib/stores/capability-enablement.ts
```diff
@@ -1,11 +1,34 @@
...
+const LOCAL_OVERRIDE_CAPABILITIES = new Set<keyof Capabilities>([
+  'serverlessDeployments',
+]);
+
+export function isCapabilityEnabled(
+  key: keyof Capabilities,
+): Readable<boolean> {
+  if (!LOCAL_OVERRIDE_CAPABILITIES.has(key)) {
+    return derived(page, ($page) =>
+      Boolean($page.data?.systemInfo?.capabilities?.[key]),
```
laurakwhit, line 21: We also have `page.data.namespace.namespaceInfo?.capabilities` which is where `workerHeartbeats` is currently enabled/disabled. Maybe we want a distinction between `isSystemCapabilityEnabled` and `isNamespaceCapabilityEnabled`?

PR #3230, opened 2026-03-18
Re-wire auth to use a provider pattern. Lots of tests remove cloud references

GiantRobots:
## Description 

This PR updates the auth code in request-from-api to use dependency injection and middleware instead of the current if based approach for injecting different authentication strategies for other providers specifically temporal cloud. 

This allows auth to be swapped wholesale without implementation leakage around the project and consolidating request/response wrappers also ensuring a single implementation is being used throughout (there were previously some edge cases where two sets of authnz code would be executed)


src/lib/utilities/core-provider.ts
```diff
@@ -0,0 +1,92 @@
...
+      preRequest: options.api?.preRequest ?? passthrough,
+      postResponse: options.api?.postResponse ?? passthroughResponse,
+    },
+    getDataEncoderEndpoint: options.getDataEncoderEndpoint ?? (async () => ''),
+    searchNamespaces: options.searchNamespaces ?? (async () => []),
+  };
+}
+
+export async function getAccessToken(): Promise<string> {
+  if (!BROWSER || !provider) return '';
```
andrewzamojc, line 58: silent, but seems fine.

src/lib/utilities/oss-provider.svelte.ts
```diff
@@ -0,0 +1,71 @@
...
+
+  return response;
+};
+
+export async function ossGetDataEncoderEndpoint(
+  _namespace: string,
+): Promise<string> {
+  const settings = page.data?.settings;
+  return getCodecEndpoint(settings);
+}
```
andrewzamojc, line 71: wizardry...

src/lib/utilities/request-from-api.ts
```diff
@@ -89,46 +90,52 @@ export const requestFromAPI = async <T>(
...
+    let context = { url, options: baseOptions };
+
+    if (isBrowser) {
+      context = await runPreRequest(context);
+    }
+
+    let response = await executeRequest(context);
+
+    if (isBrowser) {
+      response = await runPostResponse(response, {
```
andrewzamojc, line 122: Dope

src/lib/utilities/oss-provider.svelte.ts
GiantRobots: This is approximately how we do it now. I very much dislike how we use a store and push it down by setting it in the page it's VERY VERY side effecty so this is step1 in changing it.

PR #3231, opened 2026-03-18
Add underline, arrow, and subtitle href structure

KATIETOLER:
New style structure for Contextual Nav
This adds a subtitle that is also a link. 
This will allow the subtitles to link back to previous nav contexts. 

- Adds underline to subtitle if it is an href
- Adds arrow on the left of the subtitle to indicate navigation
- Adds some padding to account for new UI elements
- Adds new icons for the nav: [action, usage, slash-forward, workflows, support, settings, overview, house]

Screenshot 2026-03-18 at 4 03 20 PM

### DIFF
--- a/src/lib/holocene/navigation/cloud-nav-bar.svelte
+++ b/src/lib/holocene/navigation/cloud-nav-bar.svelte
@@ -46,7 +46,7 @@
           <Icon
             label="Project Namespaces"
             name="arrow-left"
-            class="m-1 text-indigo-100 group-data-[nav=closed]:hidden"
+            class="m-1.5 text-indigo-100 group-data-[nav=closed]:hidden"
           />
           <h2
             class={merge(


src/lib/holocene/navigation/navigation-item.svelte
```diff
@@ -38,7 +39,7 @@
     data-track-intent="navigate"
     data-track-text={label}
     class={merge(
-      'mb-1 flex items-center whitespace-nowrap px-2 py-1 text-sm ',
```
bilal-karim, line 41: The icon feels too close to the edge of the hover and active background.

Screenshot 2026-03-19 at 10 30 08 AM

Let's stick with `px-2 py-1` for padding, as we currently do in prod.

src/lib/holocene/navigation/navigation-container.svelte
bilal-karim: Let's try `px-2 px-4` here

src/lib/holocene/icon/paths.ts
```diff
@@ -138,10 +141,12 @@ import userGroup from './svg/user-group.svelte';
 import verticalEllipsis from './svg/vertical-ellipsis.svelte';
 import warning from './svg/warning.svelte';
 import workflow from './svg/workflow.svelte';
+import workflows from './svg/workflows.svelte';
```
GraceGardner, line 144: Is this replacing the workflow icon?

GraceGardner (approved):

src/lib/holocene/icon/paths.ts
KATIETOLER: yes. It is.

PR #3234, opened 2026-03-19
Fix infinite loop on root page if publicPath is set

laurakwhit:
## Description & motivation 💭 

Requesting the root page (e.g. `https://hostname/:port/`) when `publicPath` is set (e.g. `/custom-path`) without the `publicPath` caused an infinite loop because the server served `index.html` with base the `/custom-path` but the URL didn't match and caused SvelteKit to repeatedly fail to route.

This PR updates the catch-all index handler so it now checks `RequestURI` (which preserves the original path before the rewrite middleware). If the `publicPath` prefix is missing, it redirects to the prefixed URL (e.g. `/` → `/custom-path`). All API, health, and asset routes should be unaffected since they match their own specific handlers.

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

`cd server && go test ./server/route/ -v -run TestBuildUIIndexHandler`

Run the following (with a `publicPath` set) ⬇️ 
```
cd server
pnpm build:server 
docker build -t temporal-ui-local .
docker run -p 5252:8080 -e TEMPORAL_UI_PUBLIC_PATH=/custom-path -e TEMPORAL_ADDRESS=host.docker.internal:7233 temporal-ui-local

pnpm dev:ui-server
```

Then go to http://localhost:5252
 - [ ] Verify it redirects to  `/custom-path` and does not cause an infinite loop
 - [ ] Verify `/namespaces` redirects to `/custom-path/namespaces`

Run the following (**without** a `publicPath` set)  ⬇️ 
```
docker run -p 5252:8080 -e TEMPORAL_ADDRESS=host.docker.internal:7233 temporal-ui-local
```

Then go to http://localhost:5252
 - [ ] Verify the route works as expected
 
### Issue(s) closed 

Related [Slack thread](https://temporalio.slack.com/archives/C011AAP3GFP/p1773863664893019?thread_ts=1772227388.908679&cid=C011AAP3GFP).

### DIFF
--- a/server/server/route/ui.go
+++ b/server/server/route/ui.go
@@ -92,11 +92,12 @@ func buildUIIndexHandler(publicPath string, assets fs.FS) (echo.HandlerFunc, err
 		missingPublicPath := hasPublicPath && !hasPublicPathPrefix && !isExactPublicPath
 
 		if missingPublicPath {
-			target := publicPath + c.Request().URL.Path
+			cleanPath := path.Clean(c.Request().URL.Path)
+			target := publicPath + cleanPath
 			if c.Request().URL.RawQuery != "" {
 				target += "?" + c.Request().URL.RawQuery
 			}
-			return c.Redirect(http.StatusTemporaryRedirect, target)
+			return c.Redirect(http.StatusPermanentRedirect, target)
 		}
 		return c.Stream(200, "text/html", bytes.NewBuffer(indexHTMLBytes))
 	}, nil
--- a/server/server/route/ui_test.go
+++ b/server/server/route/ui_test.go
@@ -146,7 +146,7 @@ func TestBuildUIIndexHandler_WithPublicPath_RedirectsWhenPrefixMissing(t *testin
 
 			err := handler(c)
 			assert.NoError(t, err)
-			assert.Equal(t, http.StatusTemporaryRedirect, rec.Code)
+			assert.Equal(t, http.StatusPermanentRedirect, rec.Code)
 			assert.Equal(t, tt.expectedTarget, rec.Header().Get("Location"))
 		})
 	}


server/server/route/ui.go
```diff
@@ -86,6 +87,17 @@ func buildUIIndexHandler(publicPath string, assets fs.FS) (echo.HandlerFunc, err
 	}
 
 	return func(c echo.Context) (err error) {
+		hasPublicPathPrefix := strings.HasPrefix(c.Request().RequestURI, publicPath+"/")
```
rossnelson, line 90:
```suggestion
hasPublicPathPrefix := strings.HasPrefix(c.Request().URL.Path, publicPath+"/")
```

for consistency

server/server/route/ui.go
```diff
@@ -86,6 +87,17 @@ func buildUIIndexHandler(publicPath string, assets fs.FS) (echo.HandlerFunc, err
...
+		isExactPublicPath := c.Request().URL.Path == publicPath
+		missingPublicPath := hasPublicPath && !hasPublicPathPrefix && !isExactPublicPath
+
+		if missingPublicPath {
+			target := publicPath + c.Request().URL.Path
```
rossnelson:
```suggestion
			cleanPath := path.Clean(c.Request().URL.Path)
      target := publicPath + cleanPath
```

add sanitization to the built path

server/server/route/ui.go
```diff
@@ -86,6 +87,17 @@ func buildUIIndexHandler(publicPath string, assets fs.FS) (echo.HandlerFunc, err
 	}
 
 	return func(c echo.Context) (err error) {
+		hasPublicPathPrefix := strings.HasPrefix(c.Request().RequestURI, publicPath+"/")
+		isExactPublicPath := c.Request().URL.Path == publicPath
+		missingPublicPath := hasPublicPath && !hasPublicPathPrefix && !isExactPublicPath
```
rossnelson, line 92: It might be possible unrelated paths get incorrectly redirected?

```suggestion
  isRootPath := c.Request().URL.Path == "/"
  missingPublicPath := hasPublicPath && !hasPublicPathPrefix && !isExactPublicPath && isRootPath
```

server/server/route/ui.go
```diff
@@ -86,6 +87,17 @@ func buildUIIndexHandler(publicPath string, assets fs.FS) (echo.HandlerFunc, err
...
+
+		if missingPublicPath {
+			target := publicPath + c.Request().URL.Path
+			if c.Request().URL.RawQuery != "" {
+				target += "?" + c.Request().URL.RawQuery
+			}
+			return c.Redirect(http.StatusTemporaryRedirect, target)
```
rossnelson: do we want browsers to cache and avoid the round-trip on every cold load? or is this truly a temp redirect? we can avoid the round-trip if we use 308.

```suggestion
    return c.Redirect(http.StatusPermanentRedirect, target)
```

server/server/route/ui.go
laurakwhit: For `hasPublicPathPrefix` we need `RequestURI` because we're asking "did the original request include the publicPath prefix?". After rewrite, `URL.Path` has already been stripped, so it will never start with the prefix.

For `isExactPublicPath` we're using `URL.Path` because the rewrite only matches publicPath + "/*" (with a trailing slash). A request to exactly `/custom-path` (no trailing slash) doesn't match the rewrite pattern, so `URL.Path` stays `/custom-path`.

Definitely open to suggestions is there's a better way to go about this!

server/server/route/ui.go
laurakwhit: I think we want other paths to get redirected as well though? E.g. `/namespaces` --> `/custom-path/namespaces`.

server/server/route/ui.go
rossnelson: That was the context i needed, thanks!

PR #3285, opened 2026-04-09
Workflows Table Row Density

Alex-Tideman:
## Description & motivation 💭 
Reduce the whitespace in the Workflows List to allow users to more quickly scan information, especially if they have configured lots of columns.

### Summary

  - Add a table density toggle to the workflows summary table, allowing users to switch between Dense (dense rows with truncated values and short timestamps) and Comfortable (standard spacing with full values) views
  - Add table-dense and table-comfy icons to the Holocene icon system for the density toggle button
  - Extract a reusable truncateValue utility and add a persisted tableDensity store

### Test plan
  - Verify the density toggle button appears in the workflows table toolbar, with the first time showing the FeatureTag icon (indigo dot) to draw attention to it.
  - Click the toggle and confirm the table switches between compact and comfy views
  - In Dense mode: values are truncated, timestamps use short format, row actions are smaller
  - In comfy mode: full values are shown with filter/copy buttons, timestamps use long format
  - Refresh the page and confirm the density preference persists
  - Verify both table-dense and table-comfy icons render correctly at various sizes

https://github.com/user-attachments/assets/6cabce91-0e7e-43f6-89e7-c6cd21ffb758

### Design Considerations 🎨 

- [x] Do we want to keep Copy and Filter icon buttons on Compact density? If so, where would they go? Yes keep.

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


acav-temporalio:
The design team is currently doing our own table density work to make sure everything plays nice for all the various types of styles we have in our tables, as well as adjusting sizing of components within tables that will impact table row height and truncation. We definitely want to tackle this but we want to make sure we're covering all our bases so this can become a pattern across the board for all high density tables. We already have "All/Compact" and then adding a "Compact/Comfy" may start crossing some boundaries we want to care for preemptively.

Can we pause on this while the UX team wraps up our discovery and comparison and then we'll come back with a comprehensive plan for how a compact/comfy treatment will work across all tables in the UI?

This would be an excellent candidate for a UX/Eng collaborative pairing.

Noting some things that this would cover:

- row height
- row width
- truncation and inline button access
- timestamp format (? overrides user-chosen timestamp format?)

src/lib/components/workflow/workflows-summary-configurable-table/table-body-cell.svelte
```diff
@@ -113,6 +120,7 @@
           run: workflow.runId,
           archival,
         })}
+        truncate
```
laurakwhit: Was this intentional? Should Run ID always be truncated even in "comfortable" mode?

src/lib/components/workflow/workflows-summary-configurable-table/table-body-cell.svelte
```diff
@@ -173,13 +186,16 @@
...
+        {truncate}
       />
     {/if}
   </td>
 {:else}
   <td
     class="workflows-summary-table-body-cell"
     data-testid="workflows-summary-table-body-cell"
+    class:comfy={!truncate}
+    class:dense={truncate}
```
laurakwhit: Looks like these aren't being used?

```suggestion
```

laurakwhit: Also, noticed we have `class="dense"` in a few places, but I don't think it's being used anymore. Maybe we remove for clarity?

src/lib/i18n/locales/en/common.ts
```diff
@@ -220,4 +220,7 @@ export const Strings = {
   'slack-community': 'Slack Community',
   'community-forum': 'Community Forum',
   'change-log': 'Change Log',
+  compact: 'Compact',
```
laurakwhit: Could we just use this one instead of having `workflows.compact`? It also looks like this one is not currently being used.

src/lib/components/workflow/workflows-summary-configurable-table/table-body-cell.svelte
```diff
@@ -32,6 +35,7 @@
     archival?: boolean;
   };
   let { column, workflow, archival = false }: Props = $props();
+  let truncate = $derived($tableDensity === 'compact');
```
laurakwhit: Don't think this is an actual performance concern, but every cell component subscribes to `$tableDensity` independently with each triggering a re-derive. We could just pass `truncate` as a prop from the parent (`workflows-summary-configurable-table.svelte`) so it only reads the store once. Not sure if there's a pattern to "follow" here, but thought I'd bring up.

src/lib/holocene/feature-tag.svelte
```diff
@@ -1,15 +1,13 @@
 <script lang="ts">
   import { viewedFeatureTags } from '$lib/stores/new-feature-tags';
 
-  export let feature: string;
-  export let alpha = false;
+  type Props = { feature: string; alpha?: boolean };
```
laurakwhit: NIT
```suggestion
  interface Props {
    feature: string;
    alpha?: boolean;
  }
  
```

src/lib/components/workflow/workflows-summary-configurable-table/table-body-cell.svelte
```diff
@@ -69,7 +73,8 @@
 
 {#if filterableLabels.includes(label) || isCustomKeywordOrTextAttribute}
   <td
-    class="workflows-summary-table-body-cell filterable"
+    class="workflows-summary-table-body-cell"
+    class:filterable={!truncate}
```
laurakwhit, line 90: If there aren't a lot of columns in the table, the filter and copy buttons can get super far away from the actually cell data. Wondering if it would look at all better to make them stick close to the end of the text? 

Image

src/lib/components/workflow/workflows-summary-configurable-table/filterable-table-cell.svelte
```diff
@@ -15,31 +15,25 @@
...
     type?: SearchAttributeType;
+    truncate?: boolean;
   };
   let {
     attribute,
     filterOrCopyButtonsVisible = false,
     value,
     href,
     type = SEARCH_ATTRIBUTE_TYPE.KEYWORD,
+    truncate = true,
```
laurakwhit: Should this default to `false` since we default to `comfortable`?

```suggestion
    truncate = false,
```

src/lib/components/workflow/workflows-summary-configurable-table/filterable-table-cell.svelte
```diff
@@ -61,14 +55,14 @@
   };
 </script>
 
-{#if isRunId}
-  <Tooltip text={value} top class="min-w-0">
-    <Link {href} class="cursor-help">{truncateRunId(value)}</Link>
+{#if href}
+  <Tooltip text={value} top class="min-w-0" hide={!truncate}>
+    <Link {href}>{truncate ? truncateValue(value) : value}</Link>
```
laurakwhit, line 67: NIT: Not a blocker, but would like to figure out a way to not show the tooltip text if the value isn't actually being truncated (e.g. is less than 15 chars).

src/lib/holocene/feature-tag.svelte
```diff
@@ -1,15 +1,13 @@
...
+  let { feature, alpha }: Props = $props();
```
laurakwhit: Do we still want a default here?

```suggestion
  let { feature, alpha = false }: Props = $props();
```

src/lib/components/workflow/workflows-summary-configurable-table/table-body-cell.svelte
Alex-Tideman: I thought we'd probs still want to truncate that but upon second thought let's not.

src/lib/components/workflow/workflows-summary-configurable-table/table-body-cell.svelte
Alex-Tideman: I know we have "dense" in our core table which doesn't really have anything to do with "dense" mode so I'm going to leave those for now and we can do another pass in table.

src/lib/holocene/feature-tag.svelte
Alex-Tideman: I set it to true, for the most part feature tags are for alpha releases.

src/lib/components/workflow/workflows-summary-configurable-table/filterable-table-cell.svelte
Alex-Tideman: 👍

src/lib/components/workflow/workflows-summary-configurable-table/filterable-table-cell.svelte
Alex-Tideman: Updated it to check if truncateValue(value).length is smaller than TRUNCATE_LENGTH

src/lib/holocene/feature-tag.svelte
Alex-Tideman: 👍

src/lib/i18n/locales/en/common.ts
Alex-Tideman: 🗑️

src/lib/components/workflow/workflows-summary-configurable-table/table-body-cell.svelte
Alex-Tideman: Based in as prop and changed to "dense" instead of "compact"

src/lib/components/workflow/workflows-summary-configurable-table/table-body-cell.svelte
Alex-Tideman: Yeah was thinking about this, but if you don't have many columns, you aren't going to use Dense mode. I don't want to add a bunch of complexity here when its not really the point of the mode to use when you just have a couple of columns.

laurakwhit (approved):

PR #3292, opened 2026-04-10
Add relative path prefix support to routeFor utilities

Alex-Tideman:
## Summary

- Adds a `routePrefix` mechanism via `CoreProvider` that allows consuming projects (e.g., Temporal Cloud) to inject a path prefix into all UI route functions
- Introduces a `withPrefix` helper that inserts the prefix between the SvelteKit `base` path and the route path across the 12 root `resolve()` calls — leaf functions inherit the prefix automatically through existing string concatenation
- Auth routes (`routeForAuthentication`, `routeForLoginPage`) are explicitly excluded to avoid breaking SSO flows
- Moves `getRoutePrefix` to `CoreProvider` so the prefix is available globally and configured at initialization

## Changes

- **`core-provider.ts`**: Added `getRoutePrefix` to the `CoreProvider` type and `InitOptions`, with a default no-op returning `''`
- **`route-for.ts`**: Added `withPrefix()` helper wrapping all root route functions; auth routes left unwrapped
- **`hooks.client.ts`**: OSS init passes `getRoutePrefix: () => ''` to maintain current behavior
- **`route-for.test.ts`**: Added test suite verifying prefix propagation through root, namespace, workflow, nexus, and schedule routes, plus exclusion from auth routes
- **`route-for-base-path.test.ts`**: Added comprehensive parameterized tests covering all `routeFor` functions with prefix enabled, ensuring `base + prefix` appears exactly once and auth routes remain prefix-free

## Test plan

- [ ] `pnpm test -- --run` passes all new and existing route-for tests
- [ ] Verify OSS app behavior is unchanged (prefix defaults to empty string)
- [ ] Verify consuming project can set a prefix and all navigation routes include it
- [ ] Verify auth/login routes do not include the prefix

### DIFF
--- a/src/hooks.client.ts
+++ b/src/hooks.client.ts
@@ -24,6 +24,7 @@ consumeAuthCookies();
 initCoreProvider({
   getAccessToken: async () => getAuthUser().accessToken ?? '',
   getIdToken: async () => getAuthUser().idToken,
+  getRoutePrefix: () => '',
   api: {
     preRequest: ossPreRequest,
     postResponse: ossPostResponse,
--- a/src/lib/stores/route-prefix.ts
+++ b/src/lib/stores/route-prefix.ts
@@ -1,0 +1,5 @@
+import { get, writable } from 'svelte/store';
+
+export const routePrefix = writable<string>('');
+
+export const getRoutePrefix = (): string => get(routePrefix);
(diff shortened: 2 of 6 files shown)


KATIETOLER (approved):
This looks great, tested on the cloud-ui test PR and it's very robust and idiomatic, thanks for doing this! My only question is how getRoutePrefix() actually reads the value and whether it's safe to call outside of component initialization context?

PR #3299, opened 2026-04-13
refactor: replace PayloadDecoder and MetadataDecoder with unified Payload component

rossedfort:
## Description & motivation 💭 

The monolithic <Payload> component handled 3 rendering modes plus a headless decode mode through a single 14-prop interface, where most props were irrelevant to any given call site. This refactor replaces it with four focused components, each with a minimal prop surface.

  New components

  - PayloadCodeBlock — renders a <CodeBlock>, replaces <Payload mode="code-block"> (~24 call sites). Uses decodeEventAttributes + parsePayloadAttributes for full event attribute tree decoding.
  - PayloadSummary — renders a <Badge> or <span>, replaces <Payload mode="summary">. Uses decodeUserMetadata (correct path for single raw Payloads — avoids double-decoding with codec servers).
  - PayloadInline — renders a truncated inline code block, replaces <Payload mode="inline-truncated">.
  - PayloadDecoder — headless decode-only component that passes decoded value to a child snippet.

  Shared decode utility

  Extracted the repeated decodeEventAttributes → parsePayloadAttributes → stringifyWithBigInt pipeline from PayloadCodeBlock, PayloadDecoder, and PayloadInline into src/lib/utilities/decode-payload-value.ts. Exports decodePayloadValue and getInitialPayloadValue with a DecodableValue union type.

  Correctness fixes

  - PayloadSummary now uses decodeUserMetadata instead of decodeEventAttributes — the previous path would double-decode raw Payloads when a codec server is configured.
  - All four components use $effect instead of onMount, so decoded values update reactively when props change rather than only on mount.
  - Added onDecode callback to PayloadSummary (was missing, needed by timeline graph row).

### How was this tested 👻 

- [ ] Manual testing
- [x] E2E tests added - added in #3321
- [ ] Unit tests added

- check out this branch, run `pnpm dev`, `pnpm codec-server`, and `pnpm run-workflows`.
- check the payload-coverage-workflow and user-metadata-workflow to ensure all payloads are decoded

#### Optionally in addition to the above:
1. in a separate terminal in the ui repo, run `git worktree add /path/to/worktree main`, `cd /path/to/worktree`, `cp /path/to/ui/.env /path/to/worktree`, `pnpm i`.
2. change the `VITE_API` env variable in `.env` to `"http://localhost:8081"`
3. change `preview.port` in `vite.config.ts` to `3031`.
4. Then run `pnpm build:local` and `pnpm preview:local`
5. open `http://localhost:3001` to spot-check differences between `main` and this branch.


src/lib/components/event/event-card.svelte
```diff
@@ -52,6 +52,7 @@
         typeof value === 'object' && Object.keys(value).length > 0,
     ),
   );
+  $inspect(payloadFields);
```
Alex-Tideman: 🗑️

src/lib/services/workflow-service.ts
```diff
@@ -837,38 +836,53 @@ export const fetchInitialValuesForStartWorkflow = async ({
     const firstEvent = await fetchInitialEvent(params);
 
     const startEvent = firstEvent as WorkflowExecutionStartedEvent;
-    const convertedAttributes = (await decodeEventAttributesForExport(
-      startEvent?.attributes?.input,
-    )) as PotentiallyDecodable;
+    const decodedInput = await decodePayloadAndParseDataToJSON(
+      startEvent.attributes.input[0],
```
rossnelson: Looks like inputs has an additional level to get the actual payload. 

```suggestion
      startEvent.attributes.input?.payloads?.[0],
```

Before: 
Image

After:

Image

src/lib/utilities/decode-local-activity.ts
```diff
@@ -30,17 +31,18 @@ export type LocalActivityDecodeOptions = {
 export const decodeLocalActivity = async (
   event: IterableEvent,
 ): Promise<SummaryAttribute | undefined> => {
+  console.log(event);
```
rossnelson: X ?

src/lib/components/payload/payload-inline.svelte
```diff
@@ -0,0 +1,25 @@
...
+  let { value, truncateAt = 60, class: className = '' }: Props = $props();
+</script>
+
+<PayloadDecoder {value}>
+  {#snippet children(decodedValue)}
+    <div
+      class="overflow-hidden border border-subtle bg-code-block px-1 py-0.5 font-mono text-xs text-primary {className}"
+    >
+      <code>
+        <pre class="truncate">{decodedValue.slice(0, truncateAt)}</pre>
```
rossnelson: looks like this is expecting to truncate text but decoded value is actually an array?

```suggestion
        <pre class="truncate">{(decodedValue[0] ?? '').slice(0, truncateAt)}</pre>
```

Before:
Image

After:
Image

src/lib/utilities/decode-payload.ts
```diff
@@ -198,38 +229,64 @@ const keyIs = (key: string, ...validKeys: string[]) => {
...
+    return null;
+  }
+
+  return parseRawPayloadToJSON(decoded[0], returnDataOnly);
+}
+
+export const decodePayloadsAndParseDataToJSON = async (
+  payloads: Payloads | null | undefined,
+): Promise<unknown[]> => {
+  const decoded = await decodePayloadsWithRemoteCodec(payloads.payloads);
```
rossnelson: The function signature implies that payloads could be null or undefined. 

```suggestion
  const decoded = await decodePayloadsWithRemoteCodec(payloads?.payloads);
```

src/lib/components/payload/payload-summary.svelte
```diff
@@ -0,0 +1,63 @@
...
+          decodedValue = applyPrefix(result);
+          onDecode?.(decodedValue);
+        } else {
+          decodedValue = fallback;
+        }
+      })
+      .catch(() => {
+        decodedValue = fallback;
+      });
+  });
```
rossnelson, line 60: Edge case, but there could be a race condition here if the component is re-rendered while still decoding. Maybe cancel on effect cleanup?

rossnelson (commented):
Great work! I left a couple minor comments.

src/lib/utilities/export-history.ts
```diff
@@ -18,7 +18,8 @@ const decodePayloads = async (
       decodeSetting,
     );
 
-    return parsePayloadAttributes(convertedAttributes, false);
+    return convertedAttributes;
+    // return parsePayloadAttributes(convertedAttributes, false);
```
laurakwhit: Should this be commented out?

src/lib/components/event/event-details-row.svelte
```diff
@@ -29,21 +30,13 @@
         {format(key)}
       </p>
     {/if}
-    {#if typeof value === 'object'}
+    {#if isRawPayload(value) || isRawPayloads(value)}
```
laurakwhit, line 33: Looks like `value` still can be the type `Record<string, unknown>`. Should there be a fallback for an object that isn't a payload here or should the type be removed?

src/lib/components/payload/payload-decoder.svelte
```diff
@@ -0,0 +1,52 @@
...
+    decodePayloadAndParseDataToJSON,
+    decodePayloadsAndParseDataToJSON,
+    isRawPayload,
+    isRawPayloads,
+    type PayloadContainingObject,
+    type PotentiallyDecodable,
+  } from '$lib/utilities/decode-payload';
+  import { stringifyWithBigInt } from '$lib/utilities/parse-with-big-int';
+
+  export const decodePayloadValue = async (
```
laurakwhit: Do we need to export this? If so, maybe it should be moved to the `decode-payload-value.ts` file instead.

```suggestion
  const decodePayloadValue = async (
```

src/lib/components/payload/payload-code-block.svelte
```diff
@@ -0,0 +1,35 @@
...
+    testId?: string;
+  }
+
+  let { value, maxHeight, testId }: Props = $props();
+</script>
+
+<PayloadDecoder {value}>
+  {#snippet children(decodedValue)}
+    <div class="space-y-2">
+      {#each decodedValue as data (data)}
```
laurakwhit: Should we key on `index` instead of `data` here in case the data might be duplicated?

src/lib/components/workflow/client-actions/update-confirmation-modal.svelte
```diff
@@ -167,14 +176,10 @@
...
-    {#if success}
+    {#if success && typeof success === 'object'}
       <Alert intent="success" title="Success">
         {#if success?.payloads?.[0] && success.payloads[0].data}
-          <PayloadDecoder value={success.payloads[0]}>
-            {#snippet children(decodedValue)}
-              <CodeBlock class="mt-4" content={decodedValue} language="text" />
-            {/snippet}
-          </PayloadDecoder>
+          <PayloadCodeBlock value={success} />
```
laurakwhit, line 182: Should we pass `language` to `PayloadCodeBlock` as well? Looks like this was previously `language="text"` but `PayloadCodeBlock` uses `language="json"`.

src/lib/components/payload/decode-payload-value.ts
```diff
@@ -0,0 +1,44 @@
...
+export type DecodableValue =
+  | PotentiallyDecodable
+  | EventAttribute
+  | WorkflowEvent
+  | Memo
+  | RawPayload
+  | null
+  | undefined;
+
+export const getInitialPayloadValue = (
```
laurakwhit: Are the functions in this file currently being used anywhere?

src/lib/components/event/event-card.svelte
```diff
@@ -184,55 +182,17 @@
...
         {translate('workflows.call-stack-tab')}
       </p>
-      <CodeBlock
-        content={stackTrace}
-        language="text"
-        maxHeight={384}
-        copyIconTitle={translate('common.copy-icon-title')}
-        copySuccessIconTitle={translate('common.copy-success-icon-title')}
-      />
+      <CodeBlock content={stackTrace} language="text" maxHeight={384} />
```
laurakwhit: Looks like we're no longer passing in `copyIconTitle` or `copySuccessIconTitle`. Should we just default them to some values in the `CodeBlock` component for a11y support?

src/lib/utilities/decode-payload.ts
```diff
@@ -198,38 +229,64 @@ const keyIs = (key: string, ...validKeys: string[]) => {
...
+): Promise<unknown>;
+export async function decodePayloadAndParseDataToJSON(
+  payload: Payload,
+  returnDataOnly: false,
+): Promise<ParsedPayload>;
+export async function decodePayloadAndParseDataToJSON(
+  payload: Payload | null | undefined,
+  returnDataOnly: boolean = true,
+): Promise<unknown | ParsedPayload> {
+  const decoded = await decodePayloadsWithRemoteCodec(toArray(payload));
```
laurakwhit, line 270: Should we be short-circuiting nullish payloads in the decode path?

```suggestion
  if (!payload) return '';
  const decoded = await decodePayloadsWithRemoteCodec(toArray(payload));
```

src/lib/utilities/decode-payload.ts
```diff
@@ -198,38 +229,64 @@ const keyIs = (key: string, ...validKeys: string[]) => {
...
+    return null;
+  }
+
+  return parseRawPayloadToJSON(decoded[0], returnDataOnly);
+}
+
+export const decodePayloadsAndParseDataToJSON = async (
+  payloads: Payloads | null | undefined,
+): Promise<unknown[]> => {
+  const decoded = await decodePayloadsWithRemoteCodec(payloads?.payloads);
```
laurakwhit, line 283: Same question here.

src/lib/components/payload/IMPROVEMENTS.md
```diff
@@ -0,0 +1,413 @@
+# Payload Component Improvements
```
laurakwhit: NIT: If we want to preserve this file maybe we should start a docs folder? Feels a bit weird to ship it under `/src`.

src/lib/components/payload/payload-summary.svelte
laurakwhit: +1 here. We should have a way to cancel this async operation. Looks like we also removed the `{#key}` in `workflow-json-navigator.svelte`, so very wary that this won't handle reactivity correctly if the value changes while decode is in flight. 

Could we update `decodePayloadAndParseDataToJSON` to accept and honor an `AbortSignal`?

src/lib/components/payload/payload-decoder.svelte
```diff
@@ -0,0 +1,52 @@
...
+    } else if (isRawPayloads(value)) {
+      const parsedPayloadsData = await decodePayloadsAndParseDataToJSON(value);
+      const stringified = parsedPayloadsData.map((data) =>
+        stringifyWithBigInt(data),
+      );
+      onDecode?.(stringified);
+      return stringified;
+    } else {
+      const decoded = await decodeEventAttributes(value);
+      const stringified = stringifyWithBigInt(decoded);
```
laurakwhit, line 32: Do we need `parsePayloadAttributes` as well here? Or even better, can we just use the function in `decode-payload-value.ts`? 

```suggestion
      const decoded = await decodeEventAttributes(value);
      const parsed = parsePayloadAttributes(decoded);
      const stringified = stringifyWithBigInt(parsed);
```

rossedfort: > Do we need parsePayloadAttributes as well here?

We don't. The `decodeEventAttributes` function also does the base64 parsing back to human-readable values

> Or even better, can we just use the function in decode-payload-value.ts?

That file ended up being unused, so I deleted it. I had forgot that I refactored that util into this component, as it wasn't being used anywhere else.

src/lib/components/payload/IMPROVEMENTS.md
rossedfort: Removed 🗑️

src/lib/components/payload/decode-payload-value.ts
rossedfort: Nope - I nuked it 🗑️

laurakwhit (approved):

PR #3300, opened 2026-04-13
feat: add onItemsChange callback to PaginatedTable

andrewzamojc:
## Summary

- Adds an optional `onItemsChange` callback prop to the `api-paginated` `PaginatedTable` component
- Fires whenever `$store.visibleItems` changes (initial fetch, page navigation, cached page, page size change)
- Follows the existing callback pattern used by `onError`, `onShiftUp`, `onShiftDown`, `onSpace`
- Enables consumers to react to visible table data changes without slot prop workarounds

**Motivation:** Cloud UI's billing invoices page ([DT-3724](https://temporalio.atlassian.net/browse/DT-3724)) needs to sync a chart component with the table's visible data. The current approach uses a comma expression in the slot to extract `visibleItems`, which is fragile. This callback provides a clean, first-class way to observe item changes.

This table needs the prop, for this chart.

Screenshot 2026-04-13 at 3 28 30 PM

## Test plan

I built ui and tested the new prop in cloud-ui locally. Works well.

- [x] Verify existing PaginatedTable behavior is unchanged when `onItemsChange` is not provided
- [x] Pass `onItemsChange` callback and verify it fires on initial load
- [x] Verify callback fires on forward/backward page navigation
- [x] Verify callback fires on cached page navigation (no new fetch)
- [x] Verify callback fires on page size change

[DT-3724]: https://temporalio.atlassian.net/browse/DT-3724?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/holocene/table/paginated-table/api-paginated.svelte
+++ b/src/lib/holocene/table/paginated-table/api-paginated.svelte
@@ -32,6 +32,7 @@
     maxHeight?: string;
     onError?: ((error: Error | unknown) => void) | undefined;
     onFetch: () => Promise<PaginatedRequest<T>>;
+    onItemsChange?: (items: T[]) => void;
     onShiftUp?: KeyboardHandler;
     onShiftDown?: KeyboardHandler;
     onSpace?: KeyboardHandler;
@@ -51,6 +52,7 @@
   export let maxHeight = '';
   export let onError: ((error: Error) => void) | undefined = undefined;
   export let onFetch: () => Promise<PaginatedRequest<T>>;
+  export let onItemsChange: ((items: T[]) => void) | undefined = undefined;
   export let onShiftUp: KeyboardHandler = undefined;
   export let onShiftDown: KeyboardHandler = undefined;
   export let onSpace: KeyboardHandler = undefined;
@@ -180,6 +182,10 @@
         }
         break;
     }
+  }
+
+  $: if (onItemsChange && $store.visibleItems) {
+    onItemsChange($store.visibleItems);
   }
 
   $: adjustedTotal =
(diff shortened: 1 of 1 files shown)


src/lib/holocene/table/paginated-table/api-paginated.svelte
```diff
@@ -182,6 +184,10 @@
     }
   }
 
+  $: if (onItemsChange && $store.visibleItems) {
+    onItemsChange($store.visibleItems);
+  }
```
laurakwhit: Seems like ideally this should be migrated to Svelte 5. That way we could have a `$bindable` `visibleItems` prop and let consumers react in their own `$effect`. 

In the meantime, this currently depends on the full `$store` derived store and will get triggered every time any pagination state changes. Maybe we want to guard with a reference check here?

```suggestion
  let previousItems;                                                                                                                                                                              
  $: if (onItemsChange && $store.visibleItems !== previousItems) {                                                                                                                                
      previousItems = $store.visibleItems;
      onItemsChange($store.visibleItems);                                                                                                                                                         
  }  
```

Also, thought it might be worth noting that this fires on the initial load. Not sure if that's expected behavior or not. If we wanted to prevent that we could add a check like so ⬇️ 

```js
if (previousItems !== undefined) {                                                                                                                                                          
    onItemsChange($store.visibleItems);                                                                                                                                                     
}       
```

andrewzamojc: I think the initial load one might be necessary to get the first set.

I'll add previous items. It doesn't seem right to fire when the reference hasn't changed 👍

laurakwhit (approved):

PR #3306, opened 2026-04-14
Common Errors for Event History

Alex-Tideman:
## Summary

  - Adds a common errors system that detects 35 workflow/activity misconfigurations and surfaces them as warnings, errors, and info messages on the Workflow Details page.
  - Detection covers workflow timeouts, retry policies, continue-as-new patterns, activity timeouts, heartbeat issues, local activities, delayed start, ID reuse policies, and more
  - Includes a standalone `/common-errors` reference page with tab-based severity filtering
  - Refactors workflow-run-layout to Svelte 5 and adds a headerSnippet slot for injecting the common errors component

For this initial release, the Common Errors are not included by default to the Workflow Header. We will opt in to internally test first by including in the headerSnippet:

```
<WorkflowRunLayout>
  {#snippet headerSnippet()}
    <WorkflowCommonErrors />
  {/snippet}
  {@render children()}
</WorkflowRunLayout>
```

Test plan

- There are two significant changes outside of Common Errors to test:
1. Worklow Run Layout was converted to Svelte5, specifically with $effect and untrack to do data fetching. This is an important file that fetches our workflow data. High risk change.
2. Alert component was converted to Svelte5. Low risk change.

To test common errors, you can include the above snippet and run workflows with any of the 35 scenarios such as a short workflowExecutionTimeout to view the error.

Screenshot 2026-04-01 at 3 58 17 PM
Screenshot 2026-04-01 at 3 59 11 PM
Screenshot 2026-04-07 at 12 07 39 PM
Screenshot 2026-04-07 at 8 44 37 AM


src/lib/components/workflow/workflow-common-errors.svelte
```diff
@@ -0,0 +1,22 @@
...
+  import CommonErrorList from '$lib/components/common-errors/common-error-list.svelte';
+  import { fullEventHistory } from '$lib/stores/events';
+  import { workflowRun } from '$lib/stores/workflow-run';
+  import { getApplicableCommonErrors } from '$lib/utilities/common-error-detection';
+
+  const { workflow } = $derived($workflowRun);
+  const firstEvent = $derived($fullEventHistory[0]);
+  const errors = $derived(
+    workflow
+      ? getApplicableCommonErrors(workflow, firstEvent, $fullEventHistory)
```
laurakwhit: This recomputes on every `$fullEventHistory` change. Wondering if there's a way we can improve this for a "Running" workflow. Maybe some way of only running this for events that haven't been checked for errors yet?

src/lib/components/common-errors/common-errors-data.ts
```diff
@@ -0,0 +1,378 @@
...
+      'Memo fields are stored as plaintext and are visible in the UI and API responses. If you are passing sensitive information such as PII, consider using headers instead. Header values pass through the Data Converter, which supports encryption, and can be propagated using SDK interceptors.',
+    link: 'https://docs.temporal.io/dataconversion',
+    action: 'Headers and Context Propagation docs',
+    category: 'memo-headers',
+  },
+];
+
+export function getCommonErrorById(id: number): CommonError | undefined {
+  return COMMON_ERRORS.find((error) => error.id === id);
+}
```
laurakwhit: Instead of using `Array.find()` on every call. Could we do a Map<number, CommonError> lookup here instead?

```suggestion
const COMMON_ERRORS_BY_ID = new Map<number, CommonError>(                                                                                        
  COMMON_ERRORS.map((error) => [error.id, error]),                                                                                               
);                                                                                                                                               
                                                                                                                                  
export function getCommonErrorById(id: number): CommonError | undefined {
  return return COMMON_ERRORS_BY_ID.get(id);
}
```

src/lib/utilities/common-error-detection.ts
```diff
@@ -0,0 +1,494 @@
...
+      }
+    }
+
+    // #23: Heartbeat timeout set but no heartbeats detected
+    if (heartbeat > 0 && !act.lastHeartbeatTime) {
+      errorIds.add(23);
+    }
+  }
+
+  return [...errorIds].map(getCommonErrorById).filter(Boolean);
```
laurakwhit, line 219: Might need a "type predicate filter" here to fix type narrowing here. Something like 

```js
function isDefined(error: CommonError | undefined): error is CommonError {                                                                        
  return error !== undefined;                                                                                                              
} 
[...]
return [...errorIds].map(getCommonErrorById).filter(isDefined);  
```

src/lib/components/common-errors/common-error-list.svelte
```diff
@@ -0,0 +1,23 @@
...
+    'data-testid'?: string;
+  }
+
+  let { errors, class: className = '', ...rest }: Props = $props();
+</script>
+
+{#if errors.length > 0}
+  <div
+    class="flex flex-col gap-4 {className}"
+    data-testid={rest['data-testid']}
```
laurakwhit: NIT
```suggestion
    {...rest}
```

src/lib/components/common-errors/common-error.svelte
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  let { error, class: className = '', ...rest }: Props = $props();
+</script>
+
+<Alert
+  intent={error.severity}
+  title={error.title}
+  class={className}
+  data-testid={rest['data-testid']}
```
laurakwhit: NIT
```suggestion
  {...rest}
```

src/lib/utilities/common-error-detection.ts
```diff
@@ -0,0 +1,494 @@
...
+    'seconds' in duration
+  ) {
+    const seconds = Number((duration as { seconds: string | number }).seconds);
+    const nanos = Number((duration as { nanos?: number }).nanos);
+    return (isNaN(seconds) ? 0 : seconds) + (isNaN(nanos) ? 0 : nanos / 1e9);
+  }
+
+  if (typeof duration !== 'string') return 0;
+
+  const nsMatch = duration.match(/^(\d+\.?\d*)s$/);
```
laurakwhit: NIT

```suggestion
  const secondsMatch = duration.match(/^(\d+\.?\d*)s$/);
```

src/lib/components/workflow/workflow-common-errors.svelte
laurakwhit: Actually, it looks like this component currently isn't being used. Is it needed?

src/lib/components/common-errors/common-errors-data.ts
```diff
@@ -0,0 +1,378 @@
...
+
+export function getCommonErrorsByCategory(
```
laurakwhit: Is this being used anywhere?

src/lib/types/common-errors.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export interface CommonError {
+  id: number;
+  severity: CommonErrorSeverity;
+  title: string;
+  description: string;
+  link: string;
+  action: string;
+  category: CommonErrorCategory;
+  cloudOnly?: boolean;
```
laurakwhit, line 24: Looks like this is defined on some errors but has no effect. Do we need to update the UI somewhere to not show some errors based on this?

src/lib/components/workflow/workflow-common-errors.svelte
Alex-Tideman: Some of this was guessing what we'd want to do for a full page after feedback, but no we don't need this. Will 🗑️ it

src/lib/components/common-errors/common-errors-data.ts
Alex-Tideman: Nope, didn't know if we'd want that yet. Will 🗑️ and we can always add it back

src/lib/types/common-errors.ts
Alex-Tideman: Yeah I'm predicting I need it for cloud

src/lib/utilities/common-error-detection.ts
Alex-Tideman: 👍

src/lib/components/workflow/workflow-common-errors.svelte
laurakwhit: Was this intentionally [added back](https://github.com/temporalio/ui/pull/3306/commits/987cec36958ff4363973d1f17f7cc67266de2271)?

laurakwhit (approved):

src/lib/components/workflow/workflow-common-errors.svelte
Alex-Tideman: Yeah I added it back, this is the component we will use in the layout. We just aren't using it yet in ui.

```
<WorkflowRunLayout>
  {#snippet headerSnippet()}
    <WorkflowCommonErrors />
  {/snippet}
  {@render children()}
</WorkflowRunLayout>
```

PR #3307, opened 2026-04-14
Keep Pending Activities/Operations at the top of Timeline and Compact view

Alex-Tideman:
Make it easier for our users to find pending activities and operations without needing to scroll to find them.

### Summary

  - Pending activities and nexus operations now consistently appear at the top of Compact and Timeline views (descending) or bottom (ascending), matching the existing All view behavior
  - Compact view pending groups display "Pending Activity" / "Pending Nexus Operation" as the event type with a pulsing icon, matching the All view presentation
  - Fixed a bug in the timeline's activeGroupsHeightAboveGroup that used numeric ID comparison to determine row ordering — replaced with index-based comparison that works correctly regardless of group reordering

### Test plan

  - Open a workflow with pending activities in descending order
  - Verify All view shows pending activities at the top (existing behavior, unchanged)
  - Switch to Compact view — pending groups should appear at the top with "Pending Activity" label and pulsing icon
  - Switch to Timeline view — pending groups should appear as the topmost rows
  - Toggle to ascending sort — pending groups should move to the bottom in all three views
  - Verify expanding a group row in Timeline view does not overlap adjacent rows
  - Test with the failed/pending status filter enabled in Timeline view
  - Verify workflows with no pending activities render normally in all views
  - Unit tests pass: pnpm test -- --run src/lib/utilities/order-groups-by-pending.test.ts
 

Screenshot 2026-04-14 at 2 08 30 PM
Screenshot 2026-04-14 at 2 12 30 PM
Screenshot 2026-04-14 at 3 07 38 PM

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

### DIFF
--- a/src/lib/components/event/event-summary-row.svelte
+++ b/src/lib/components/event/event-summary-row.svelte
@@ -280,9 +280,10 @@
       <Icon
         name={CategoryIcon[event.category].name}
         title={CategoryIcon[event.category].title}
-        class="mr-1 inline {isEventGroup(event) && event.isPending
-          ? ' animate-pulse'
-          : ''}"
+        class={merge(
+          'mr-1 inline',
+          isEventGroup(event) && event.isPending && 'animate-pulse',
+        )}
       />
       {displayName}
     </p>
--- a/src/lib/components/event/event-summary-row.svelte
+++ b/src/lib/components/event/event-summary-row.svelte
@@ -280,8 +280,7 @@
       <Icon
         name={CategoryIcon[event.category].name}
         title={CategoryIcon[event.category].title}
-        class="mr-1 inline{isEventGroup(event) &&
-        (event.pendingActivity || event.pendingNexusOperation)
+        class="mr-1 inline {isEventGroup(event) && event.isPending
           ? ' animate-pulse'
           : ''}"
       />
--- a/src/lib/components/event/event-summary-row.svelte
+++ b/src/lib/components/event/event-summary-row.svelte
@@ -111,7 +111,11 @@
 
   const displayName = $derived(
     isEventGroup(event)
-      ? event.label
+      ? event.pendingActivity
+        ? translate('workflows.pending-activity')
+        : event.pendingNexusOperation
+          ? translate('workflows.pending-nexus-operation')
+          : event.label
       : isLocalActivityMarkerEvent(event)
         ? translate('events.category.local-activity')
         : spaceBetweenCapitalLetters(event.name),
@@ -276,7 +280,10 @@
       <Icon
         name={CategoryIcon[event.category].name}
         title={CategoryIcon[event.category].title}
-        class="mr-1 inline"
+        class="mr-1 inline{isEventGroup(event) &&
+        (event.pendingActivity || event.pendingNexusOperation)
+          ? ' animate-pulse'
+          : ''}"
       />
       {displayName}
     </p>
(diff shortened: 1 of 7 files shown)


src/lib/components/event/event-summary-row.svelte
```diff
@@ -276,7 +280,9 @@
       <Icon
         name={CategoryIcon[event.category].name}
         title={CategoryIcon[event.category].title}
-        class="mr-1 inline"
+        class="mr-1 inline {isEventGroup(event) && event.isPending
+          ? ' animate-pulse'
+          : ''}"
```
laurakwhit: NIT
```suggestion
        class={merge(
          'mr-1 inline',
          isEventGroup(event) && event.isPending && 'animate-pulse' 
        )}
```

src/lib/components/lines-and-dots/svg/timeline-graph.svelte
```diff
@@ -51,10 +50,10 @@
   };
 
   $: activeGroupsHeightAboveGroup = (group: EventGroup) => {
+    const groupIndex = filteredGroups.findIndex((g) => g.id === group.id);
     const activeGroupIsAbove = $activeGroups?.filter((id) => {
-      if ($eventFilterSort === 'ascending')
-        return parseInt(id) < parseInt(group.id);
-      return parseInt(id) > parseInt(group.id);
+      const activeIndex = filteredGroups.findIndex((g) => g.id === id);
```
laurakwhit: `filteredGroups.findIndex()` is called once for the group itself, then again for every active group.  Concerned about the performance of this for large histories. Could we build and id-to-index Map from filteredGroups and reuse it here instead? 

```js
  $: groupIndexMap = new Map(filteredGroups.map((g, i) => [g.id, i]));                                                                                      
                                                                                                                                                            
  $: activeGroupsHeightAboveGroup = (group: EventGroup) => {                                                                                                
    const groupIndex = groupIndexMap.get(group.id) ?? -1;                                                                                                   
    const activeGroupIsAbove = $activeGroups?.filter((id) => {                                                                                              
      const activeIndex = groupIndexMap.get(id) ?? -1;
      return activeIndex !== -1 && activeIndex < groupIndex;                                                                                                
    });                      
                                                                                                                                                           
```

src/lib/utilities/order-groups-by-pending.ts
```diff
@@ -0,0 +1,13 @@
...
+export const orderGroupsByPending = (
+  groups: EventGroups,
+  reverseSort: boolean,
+): EventGroups => {
+  const pendingGroups = groups.filter((g) => g.isPending);
+  const nonPendingGroups = groups.filter((g) => !g.isPending);
+
+  return reverseSort
+    ? [...pendingGroups, ...nonPendingGroups]
+    : [...nonPendingGroups, ...pendingGroups];
```
laurakwhit: NIT: We can do this in a single pass with one allocation using `toSorted`.

```suggestion
  return groups.toSorted((a, b) => {
    if (a.isPending === b.isPending) return 0;
    if (a.isPending) return reverseSort ? -1 : 1;
    return reverseSort ? 1 : -1;
  });
```

Alex-Tideman: TIL toSorted!

laurakwhit (approved):

PR #3320, opened 2026-04-17
Add usePortal support to Tooltip component

rossnelson:
## Summary

- Adds `usePortal` and `scrollContainer` props to `<Tooltip>` following the same pattern Laura used for the `<Menu>` component
- When `usePortal` is true, the tooltip renders via `<Portal>` anchored to the wrapper element, escaping any `overflow:hidden` ancestor
- Hover state is tracked in JS (`on:mouseenter`/`on:mouseleave`) since the portal renders outside the DOM subtree
- Falls back to the original CSS `group-hover` absolute positioning when `usePortal` is false (no breaking changes)

## Test plan

- [ ] Hover tooltips in a container with `overflow:hidden` — with `usePortal` the tooltip should appear unclipped
- [ ] Verify all position variants (`top`, `bottom`, `left`, `right`, `topLeft`, `topRight`, `bottomLeft`, `bottomRight`) work correctly with `usePortal`
- [ ] Verify existing tooltips without `usePortal` are unaffected
- [ ] Test on the Deployments page where the original clipping was observed

### DIFF
--- a/src/lib/holocene/tooltip.svelte
+++ b/src/lib/holocene/tooltip.svelte
@@ -3,6 +3,8 @@
 
   import type { IconName } from '$lib/holocene/icon';
   import Icon from '$lib/holocene/icon/icon.svelte';
+  import Portal from '$lib/holocene/portal/portal.svelte';
+  import type { PortalPosition } from '$lib/holocene/portal/types';
   import type { Only } from '$lib/types/global';
 
   type BaseProps = {
@@ -13,6 +15,8 @@
     class?: string;
     tooltipClass?: string;
     show?: boolean;
+    usePortal?: boolean;
+    scrollContainer?: string;
   };
 
   type BasePositionProps = {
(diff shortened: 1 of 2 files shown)


Alex-Tideman:
Might be the right time to convert Tooltip to Svelte5

src/lib/holocene/tooltip.svelte
```diff
@@ -71,42 +75,90 @@
   export let width: number | null = null;
   export let tooltipClass = '';
   export let show = false;
+  export let usePortal = false;
+  export let scrollContainer: string | undefined = undefined;
+
+  let wrapperElement: HTMLElement | null = null;
+  let isHovered = false;
+
+  $: portalPosition = ((): PortalPosition => {
```
andrewzamojc, line 84: I know this is a pre-existing pattern so no change needed, but I wanted to put on record I think its weird to use boolean style props that can't be combined. Not the end of the world, and reads nice on the usage, but a `position` prop seems simpler.

andrewzamojc (approved):
LGTM

PR #3321, opened 2026-04-17
test(e2e): add end-to-end payload decoder coverage

rossedfort:
## Description & motivation 💭 
Adds a payload-coverage-workflow that exercises every Temporal payload touchpoint — workflow input/output, memo, search attributes, activity input/output, activity failure, child workflow, signals, and queries — with all payloads AES-GCM encrypted via the codec server.

Uses this workflow as the foundation for a new e2e test suite (payload-decoder.desktop.spec.ts) that verifies the UI correctly decodes payloads at every touchpoint when the codec endpoint is configured.

What changed

Temporal infrastructure (temporal/)
  - Adds payload-coverage-workflow with activities, child workflows, signals (add-data), and queries (get-status, get-field)
  - Replaces payload-codec.ts with a full encryption-codec.ts implementing AES-GCM encrypt/decrypt
  - Adds crypto.ts for key generation and data-converter.ts wiring
  - Extends codec-server.ts with encryption support and client.ts with startPayloadCoverageWorkflow

E2e tests (tests/)
  - global-setup.ts: starts payload-coverage-workflow fire-and-forget so the workflow stays running during tests (enabling signal/query interactions)
  - payload-decoder.desktop.spec.ts: 11 tests covering workflow input, memo, search attributes, 5 event history event types, 2 query types, signal sending + history verification, and call stack
  - 4 new page object classes: WorkflowDetailPage, EventHistoryPage, WorkflowQueryPage, SignalModalPage
### How was this tested 👻 

- [ ] Manual testing
- [x] E2E tests added
- [ ] Unit tests added


Alex-Tideman:
Definitely don't need to add to this PR, but would be nice to also cover Updates, Nexus Operations, Schedules, SAAs, Start Workflows

utilities/temporal-server.ts
```diff
@@ -94,13 +94,18 @@ export const createTemporalServer = async ({
...
+    { flag: 'history.externalPayloadsEnabled', value: 'true' },
+    { flag: 'activity.enableStandalone', value: 'true' },
   ].forEach(({ flag, value }) => {
     flags.push('--dynamic-config-value');
     flags.push(`${flag}=${value}`);
   });
 
   const temporal = $`${cliPath} server start-dev ${flags}`.quiet();
 
+  await $`${cliPath} operator search-attribute create --name CustomKeywordField --type Keyword --name CustomIntField --type Int`;
```
Alex-Tideman: Do you need to pass in a namespace?

rossedfort: Nope, it just adds them to the `default` namespace. For some reason this was causing a failure in CI so I had to inline it with the main `temporal server start-dev` command above

rossedfort: Oh, maybe it was failing because I didn't supply a namespace. Either way, should be fixed now.

Alex-Tideman (approved):

PR #3340, opened 2026-04-22
refactor: move fetchDeployment into Deployment component

rossnelson:
## Summary

Moves deployment data fetching into the `Deployment` component rather than SvelteKit's load layer, and cleans up the related page infrastructure that was no longer doing anything useful.

## Why

`fetchDeployment` requires a server-relative base URL that isn't available at load time for package consumers. Running it inside the component ensures the transport is fully initialized before the call is made.

## Test plan

- [ ] Deployment detail page loads and displays version table
- [ ] Delete version flow refreshes the page correctly
- [ ] Delete deployment flow still works
- [ ] Existing integration tests pass

### DIFF
--- a/src/lib/pages/deployment-page.ts
+++ b/src/lib/pages/deployment-page.ts
@@ -1,19 +1,8 @@
-import { fetchDeployment } from '$lib/services/deployments-service';
-import { decodeURIForSvelte } from '$lib/utilities/encode-uri';
-
 export function loadDeploymentPage({
-  params,
   depends,
 }: {
   params: { namespace: string; deployment: string };
   depends: (...deps: `${string}:${string}`[]) => void;
 }) {
   depends('data:deployment');
-  const deploymentName = decodeURIForSvelte(params.deployment);
-  return {
-    deploymentPromise: fetchDeployment({
-      namespace: params.namespace,
-      deploymentName,
-    }),
-  };
 }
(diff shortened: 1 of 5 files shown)


src/lib/pages/deployment-page.ts
```diff
@@ -1,19 +1,8 @@
-import { fetchDeployment } from '$lib/services/deployments-service';
```
Alex-Tideman, line 1: Just 🗑️ this whole file, don't need it

rossnelson: done.

src/lib/pages/deployment.svelte
```diff
@@ -91,6 +95,7 @@
           {namespace}
           {deploymentName}
           conflictToken={deployment.conflictToken}
+          onVersionDeleted={() => reloadCount++}
```
Alex-Tideman: I'd use a timestamp instead of a count

let lastInvalidatedAt = $state(Date.now());

  const effectiveDeploymentPromise = $derived.by(() => {
    lastInvalidatedAt; // tracked so incrementing it re-fetches
    return fetchDeployment({ namespace, deploymentName });
  });

function reload() {
  lastInvalidatedAt = Date.now()
}

onVersionDeleted={reload}

rossnelson: done :)

Alex-Tideman (approved):

PR #3344, opened 2026-04-24
feat(DT-3657): Support shift click for bulk selection in workflow table

tegan-temporal:
## Description & motivation 💭 

The workflow table supports selecting multiple workflows in order to perform batch actions on the selections. Right now, you can either select all workflows, the entire page of workflows, or individual workflows (one at a time).

A common pattern is to allow multi select by selecting an item, pressing (and holding shift key), and selecting another item. All items between the two (including the two) will either be selected or deselected based on the selection state of the item.

https://github.com/user-attachments/assets/0f5ba403-fc39-4df4-b42e-0d15e81f4c35

Holding shift key after having selected or deselected a workflow, all workflows between the previously clicked workflow and the next clicked workflow will be selected (if the next clicked  workflow is currently unselected) or deselected (if the next clicked workflow is currently selected).

https://github.com/user-attachments/assets/12089333-4f87-4642-a945-ba7a17153dd5

Clear memory of prev clicked workflow when selected or deselecting all or page. (It's not intuitive to maintain the index after making a different bulk selection change).

https://github.com/user-attachments/assets/c9405c22-35ec-4045-ac27-45557877b172

There is a single shift+click scope.
  - Children that are not visible are ignored by the shift+click scope.
  - Children that are collapsed are deselected.
  - All visible children are in the same shift+click scope as the parents (selection range does not have parent/child boundaries).

### Design Considerations 🎨 

N/A (no visual change)

### How was this tested 👻 

- [x] Manual testing
- [x] E2E tests added
- [ ] Unit tests added

- Go to page with workflow summary table
- Select a workflow
- Press and hold shift key
- Select a different workflow
- See that all workflows within the range are now selected.

### Draft Checklist 

- [x] Figure out if child workflows need to be supported and if so, how.

### Issue(s) closed 

DT-3657


src/lib/holocene/checkbox.svelte
```diff
@@ -41,7 +41,9 @@
   export { className as class };
 
   let inputElement: HTMLInputElement;
-  $: inputElement !== undefined && (inputElement.indeterminate = indeterminate);
+  $: if (inputElement !== undefined) {
+    inputElement.indeterminate = indeterminate;
+  }
```
tegan-temporal, line 46: Fixes a warning about unused expression.

src/lib/holocene/checkbox.svelte
```diff
@@ -67,15 +69,27 @@
   };
 
   $: checked = group !== undefined ? group.includes(value) : checked;
+
+  $: checkIconName = indeterminate
+    ? ('hyphen' as const)
+    : checked
+      ? ('checkmark' as const)
+      : null;
```
tegan-temporal, line 77: Ensures `checkIconName` is `'hyphen' | 'checkmark' | null` which fixes a type error about the name prop for the Icon component not accepting `null`

src/lib/holocene/checkbox.svelte
```diff
@@ -67,15 +69,27 @@
...
 </script>
 
 <div
-  data-testid={$$restProps['data-testid'] ?? null}
-  on:click|stopPropagation
+  data-testid={$$restProps['data-testid']}
+  on:click|stopPropagation={() => {
```
tegan-temporal, line 82: Fixes issue where `on:click` handler passed into checkbox component was being called multiple times with different currentTargets (the checkbox input and this div).

src/lib/pages/workflows-with-new-search.svelte
```diff
@@ -96,10 +100,10 @@
...
 
   $effect(() => {
-    namespace;
-    query;
-    perPage;
-    $refresh;
+    void namespace;
+    void query;
+    void perPage;
+    void $refresh;
```
tegan-temporal, line 101: Fixes warnings about unused expressions.

src/lib/pages/workflows-with-new-search.svelte
```diff
@@ -135,15 +139,19 @@
...
 
   const openBatchTerminateConfirmationModal = () => {
-    $selectedWorkflows.length > 1
-      ? (batchTerminateConfirmationModalOpen = true)
-      : (terminateConfirmationModalOpen = true);
+    if ($selectedWorkflows.length > 1) {
+      batchTerminateConfirmationModalOpen = true;
+    } else {
+      terminateConfirmationModalOpen = true;
+    }
```
tegan-temporal, line 147: Fixes warnings about unused expressions

src/lib/pages/workflows-with-new-search.svelte
```diff
@@ -266,9 +301,8 @@
   >
     <WorkflowsSummaryConfigurableTable
       onClickConfigure={openCustomizationDrawer}
-    >
-      <slot name="cloud" slot="cloud" />
-    </WorkflowsSummaryConfigurableTable>
+      {cloud}
```
tegan-temporal, line 280: This change requires a small update in cloud-ui to pass in a snippet instead of a slot

src/lib/components/workflow/workflows-summary-configurable-table.svelte
```diff
@@ -105,37 +190,67 @@
...
-    {#each visibleItems as workflow (`${workflow.id}:${workflow.runId}`)}
+    {#each visibleRows as row, visibleRowIndex (row.value.runId)}
+      {@const isChildRow = row.rowType === 'child'}
       <TableRow
-        {workflow}
-        {viewChildren}
-        childCount={childrenActive(workflow)?.children.length}
+        workflow={row.value}
+        toggleChildrenVisibility={(workflow) => {
+          toggleChildrenVisibility(workflow);
```
tegan-temporal: This is a bit of a hack to get rid of a type warning. `toggleChildrenVisibility` has a return type of `Promise<void>` since it's `async` however the prop expects a return type of just `void`.

src/lib/components/workflow/workflows-summary-configurable-table/table-row.svelte
```diff
@@ -80,11 +90,11 @@
           <Button
             size="xs"
             variant={childrenShown ? 'primary' : 'ghost'}
-            on:click={() => viewChildren(workflow)}
+            on:click={() => toggleChildrenVisibility(workflow)}
             class={$tableDensity === 'dense' ? 'mt-1 h-5 w-5' : ''}
           >
             <Tooltip
-              text={childrenShown
+              text={childrenShown && childCount != null
```
tegan-temporal: Fix type error about `{count: undefined} | {count: null}` not being an accepted type for I18n

src/lib/holocene/checkbox.svelte
laurakwhit: Svelte supports `bind:indeterminate` [directly on checkbox inputs](https://svelte.dev/docs/svelte/bind#input-bind:indeterminate). I _think_ we can drop `inputElement` ref and the reactive block entirely.

src/lib/pages/workflows-with-new-search.svelte
laurakwhit: We'll just want to make sure we have a commit ready to add to the generated Cloud release PR that updates this.

src/lib/components/workflow/workflows-summary-configurable-table.svelte
```diff
@@ -105,37 +219,69 @@
...
+        }}
+        childCount={!isChildRow && row.childCount > 0
+          ? row.childCount
+          : undefined}
+        child={isChildRow}
+        onClickBatchSelect={(event) => {
+          // this is required due to how the underlying Checkbox component
+          // get's it's onclick type from svelte event forwarding. It does not
```
laurakwhit: NIT
```suggestion
          // gets its onclick type from svelte event forwarding. It does not
```

src/lib/components/workflow/workflows-summary-configurable-table.svelte
```diff
@@ -24,73 +32,179 @@
...
 
-  const viewChildren = async (workflow: WorkflowExecution) => {
-    if (childrenActive(workflow)) {
-      childrenIds = childrenIds.filter(
-        (id) => id.workflowId !== workflow.id && id.runId !== workflow.runId,
-      );
-    } else {
+    if (visibleChildren?.length) {
+      // we are collapsing the children so if there is an inflight request
+      // we don't want it's resolution to reopen the children.
```
laurakwhit: NIT
```suggestion
      // we don't want its resolution to reopen the children.
```

src/lib/components/workflow/workflows-summary-configurable-table.svelte
```diff
@@ -24,73 +32,179 @@
...
-  }[] = [];
+  const visibleChildrenMap = new SvelteMap<string, WorkflowExecution[]>();
 
-  const clearChildren = () => {
-    childrenIds = [];
-  };
+  $effect(() => {
+    void $refresh;
+    void query;
+    visibleChildrenMap.clear();
```
laurakwhit, line 68: Might we want `inFlightChildRequests` here too so that an in-flight fetch from before the change can't write to the new map?

```suggestion
    visibleChildrenMap.clear();
    inFlightChildRequests.clear();
```

src/lib/components/workflow/workflows-summary-configurable-table.svelte
laurakwhit: Could we just make it a required prop instead?

```js
 export let toggleChildrenVisibility: (
    workflow: WorkflowExecution,
  ) => void;
```

src/lib/components/workflow/workflows-summary-configurable-table/table-row.svelte
laurakwhit: We could also default the value (e.g. `count: childCount ?? 0`)

src/lib/holocene/checkbox.svelte
laurakwhit, line 117: Should this no longer be `bind:checked` (and instead just be `{checked}`) since it's computed from `$selectedWorkflows`?

src/lib/components/workflow/workflows-summary-configurable-table.svelte
```diff
@@ -24,73 +32,179 @@
...
 
-  $: dense = $tableDensity === 'dense';
+  const dense = $derived($tableDensity === 'dense');
 
   const setTableDensity = () => {
     $tableDensity = dense ? 'comfortable' : 'dense';
     viewFeature('tableDensity');
   };
+
+  let visibleItems: WorkflowExecution[] = $state([]);
```
laurakwhit: Wondering if we might want to rename this. Might get a bit confusing with the use of the slot scoped `visibleItems` on line 286 ⬇️ 

```js
<svelte:fragment slot="actions-end-additional" let:visibleItems let:page>
```

src/lib/components/workflow/workflows-summary-configurable-table/table-header-row.svelte
```diff
@@ -42,8 +41,9 @@
         labelHidden
         id="select-visible-workflows"
         data-testid="batch-actions-checkbox"
-        bind:checked={$pageSelected}
-        {indeterminate}
+        checked={pageSelectionStatus === 'checked' ||
+          pageSelectionStatus === 'partial'}
```
laurakwhit: Was including `|| pageSelectionStatus === 'partial'` intentional here? 

```suggestion
        checked={pageSelectionStatus === 'checked'}
```

laurakwhit (commented):
This is looking great ⌨️ 🎉  Nothing major, but a few things to clarify.

src/lib/holocene/checkbox.svelte
tegan-temporal: This was a change to just remove the warning, I wasn't trying to make a larger change... I could still do this if we want tho

src/lib/holocene/checkbox.svelte
tegan-temporal: Other uses of holocene checkbox might want to bind to checked

src/lib/components/workflow/workflows-summary-configurable-table.svelte
tegan-temporal: good call!

src/lib/components/workflow/workflows-summary-configurable-table/table-header-row.svelte
tegan-temporal: It _was_ intentional because for some reason I was under the impression that checked had to be true for the indeterminate symbol to show up, but it doesn't! So I'll remove it.

src/lib/components/workflow/workflows-summary-configurable-table/table-row.svelte
tegan-temporal: yeah good call

src/lib/components/workflow/workflows-summary-configurable-table.svelte
tegan-temporal: I'll rename to `visiblePaginatedItems`

src/lib/components/workflow/workflows-summary-configurable-table.svelte
tegan-temporal: child rows won't have a toggle function to call so I don't think we should make it required.

For this though, I updated it so the function can return a Promise or nothing at all

src/lib/components/workflow/workflows-summary-configurable-table.svelte
```diff
@@ -103,39 +218,69 @@
...
+      empty={visiblePaginatedItems.length === 0}
       slot="headers"
-      let:visibleItems
-      workflows={visibleItems}
+      workflows={visiblePaginatedItems}
+      {pageSelectionStatus}
+      onSelectPage={handleSelectPage}
     >
-      {#each columns as column}
+      {#each columns as column (column)}
```
laurakwhit: `columns` is $derived and therefore the object is re-created each time the derived runs. Probably want to key by `column.label` instead.

```suggestion
      {#each columns as column (column.label)}
```

Although wondering if there might be a scenario where the column label could be duplicated. E.g. a user creates a search attribute with the name "Status" 🤔 Maybe we want to create a unique Id for each column and use that to be on the safe side.

src/lib/components/workflow/workflows-summary-configurable-table.svelte
```diff
@@ -103,39 +218,69 @@
...
+          }
+
+          selectWorkflows(isChecked, targetedWorkflows);
+
+          prevClickedRow = row;
+        }}
       >
-        {#each columns as column}
-          <TableBodyCell {workflow} {column} truncate={dense} />
+        {#each columns as column (column)}
```
laurakwhit: Same here.

```suggestion
        {#each columns as column (column.label)}
```

src/lib/components/workflow/workflows-summary-configurable-table.svelte
```diff
@@ -24,73 +32,180 @@
...
+      }
+
+      return;
+    }
+
+    if (inFlightChildRequests.has(workflow.runId)) return;
+
+    inFlightChildRequests.add(workflow.runId);
+    try {
       const children = await fetchAllChildWorkflows(
```
laurakwhit, line 100: Perhaps out of scope for this PR and probably not causing any bugs atm, but wondering what you think about having some kind of AbortSignal for fetchAllChildWorkflows. Seems like it might be possible to expand, collapse, and then expand child workflows again before the first fetch has resolved.

laurakwhit (approved):
Two small comments r.e. keyed by `(column)`, but otherwise 🚀

src/lib/components/workflow/workflows-summary-configurable-table.svelte
tegan-temporal: yeah an abort signal would be nice. I didn't want to go down that rabbit hole for this PR though.

PR #3345, opened 2026-04-24
DT-3751 - download external payloads

rossedfort:
## Description & motivation 💭 
Adds download functionality for externally stored payloads (note: only for payloads that display in a Code Block). Other payload touch points will be addressed in a following PR.

**WIP**
This PR is a work in progress. Currently the download button shows, but the download functionality doesn't work.

#### Generic error (when codec server not configured or download fails)

Screenshot 2026-05-05 at 11 22 48 AM

#### Error when codec server is configured, but no `/download` endpoint

Screenshot 2026-05-05 at 11 23 21 AM

#### Disable download button and show tooltip when codec server not configured

Screenshot 2026-05-05 at 11 23 43 AM

#### Sample download w/ semantic filename

Screenshot 2026-05-12 at 10 54 00 AM

#### Event History Row w/ External Payload

Screenshot 2026-05-12 at 12 30 52 PM

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

1. Check out this branch and run `pnpm i` and `pnpm build:server` and `pnpm dev`
2. Clone Justin's fork of [sdk-go](https://github.com/jmaeagle99/temporalio-sdk-go/tree/codec-server-demo) which contains an external payload workflow and codec server
3. Follow the instructions in the notion doc titled: `External Storage Bug Bash 04/17`
4. Set the `AWS_PROFILE` and `S3_BUCKET` environment variables
5. In your local version of Justin's fork of temporalio/sdk-go, run `go run demos/worker/main.go`, and in a separate shell session, run `go run demos/codecserver/main.go`
7. Run `temporal workflow start --task-queue "echo-task-queue" --type "GenerateDocument" --workflow-id "generate-doc" --input '1000000'` to run a workflow with an externally stored payload.


src/lib/components/payload/payload-code-block.svelte
```diff
@@ -1,35 +1,160 @@
...
+      }
+    } finally {
+      downloadLoading = false;
+    }
+  };
 </script>
 
 <PayloadDecoder {value}>
-  {#snippet children(decodedValue)}
+  {#snippet children(results)}
```
GiantRobots, line 99: Does this results have a type?

Could we have this snippet be configurable so we could re-use the decoding logic. But swap with our own implementation?

GiantRobots: NVM you already did that

src/lib/components/schedule/schedule-form/schedule-input-payload.svelte
```diff
@@ -35,28 +40,32 @@
...
+        const parsedMetadata = base64ParsePayloadMetadata(payloads);
 
-    if (isPayloadInputEncodingType(currentEncoding)) {
-      $encoding = currentEncoding;
-      initialEncoding = $encoding;
-      if (currentEncoding === 'json/protobuf' && currentMessageType) {
-        messageType = currentMessageType;
-        initialMessageType = currentMessageType;
+        currentEncoding =
+          (parsedMetadata[0]?.encoding as PayloadInputEncoding) ?? 'json/plain';
```
GiantRobots: Should this be `parsedMetadata?.[0]` for safe access if that's not an array?

src/lib/components/payload/payload-decoder.svelte
```diff
@@ -1,52 +1,102 @@
...
 </script>
 
-{#await decodePayloadValue(value)}
+{#await decodeValue(value)}
   {@render loading?.()}
-{:then decodedValue}
-  {@render children(decodedValue)}
+{:then decodeResult}
+  {@render children(decodeResult)}
 {/await}
```
GiantRobots, line 102: I feel like we could really easily put the error state in here and pass it a retry snippet as well so we can get per request errors instead of the "last used" errors

src/lib/components/payload/payload-code-block.svelte
```diff
@@ -1,35 +1,160 @@
...
+        {:else}
+          <CodeBlock
+            content={stringifyWithBigInt(result.decodedValue)}
+            {maxHeight}
+            copyIconTitle={translate('common.copy-icon-title')}
+            copySuccessIconTitle={translate('common.copy-success-icon-title')}
+            {testId}
+            language="json"
+          />
+        {/if}
```
GiantRobots, line 167: I feel like there's a lot of stuff going on with these files that could be weird or error, maybe we should throw down an error boundary somewhere just in case to prevent app crashes

src/lib/utilities/decode-payload.ts
```diff
@@ -237,8 +244,23 @@ const keyIs = (key: string, ...validKeys: string[]) => {
...
+  payload: unknown,
+): payload is ParsedExternalPayload => {
   return (
-    keys.length === 2 && keys.includes('metadata') && keys.includes('data')
+    isParsedPayload(payload) &&
+    has(payload.metadata, 'messageType') &&
+    payload.metadata.messageType ===
+      'temporal.api.sdk.v1.ExternalStorageReference'
   );
 };
```
GiantRobots, line 265: Just spitballing on these, I wonder if instead of using typeGuards (which are a typesystem escape hatch) we could use a discriminated union on like a data container so you could easily identify which phase we're in for decoding. 

Just thinking out loud don't need to act on it.

GiantRobots (approved):

PR #3346, opened 2026-04-28
Show current duration for pending timeline events

Alex-Tideman:
## Description & motivation 💭 
Show the current duration of a pending activity or operation to quickly show how long it has been running. The duration is from scheduled event to now.

Screenshot 2026-04-28 at 9 18 52 AM

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/lines-and-dots/svg/timeline-graph-row.svelte
+++ b/src/lib/components/lines-and-dots/svg/timeline-graph-row.svelte
@@ -48,7 +48,6 @@
     readOnly = false,
   }: Props = $props();
 
-  $inspect('TIMELINE GRAPH ROW RENDER');
   const { height, gutter, radius } = TimelineConfig;
 
   let hovering = $state(false);
--- a/src/lib/components/lines-and-dots/svg/timeline-graph-row.svelte
+++ b/src/lib/components/lines-and-dots/svg/timeline-graph-row.svelte
@@ -34,7 +34,7 @@
     y: number;
     group: EventGroup;
     startTime: string | Timestamp;
-    endTime: string | Date;
+    endTime: string | Date | number;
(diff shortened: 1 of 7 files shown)


src/lib/components/lines-and-dots/svg/timeline-graph.svelte
```diff
@@ -120,7 +126,12 @@
           {/key}
         {/if}
         {#if !readOnly && $activeGroups.includes(group.id)}
-          <GroupDetailsRow y={y + 1.33 * radius} {group} {canvasWidth} />
+          <GroupDetailsRow
+            y={y + 1.33 * radius}
+            {group}
+            {canvasWidth}
+            currentTime={workflow?.endTime ? endTime : currentTime}
```
tegan-temporal: I wonder if there is a better name for this prop than `currentTime` since it isn't doesn't seem to always match up with the real `currentTime` 🤔 (e.g., when a workflow has ended).

Alex-Tideman: Good call, switched it to endTime because that's really want we are checking for. currentTime still makes sense for the EndTimeInterval bind prop.

tegan-temporal (approved):
:shipit:

src/lib/components/lines-and-dots/svg/timeline-graph-row.svelte
```diff
@@ -48,6 +48,7 @@
     readOnly = false,
   }: Props = $props();
 
+  $inspect('TIMELINE GRAPH ROW RENDER');
```
laurakwhit: 🗑️ 

```suggestion
```

PR #3236, opened 2026-03-19
feat: serverless worker deployment CRUD

rossnelson:
## Summary

Implements full CRUD UI for serverless worker deployments under \`/workers/deployments/\`:

https://github.com/user-attachments/assets/9833e5af-5e6d-4ce8-b825-fa258e1454f9

Also see worker code at: https://github.com/rossnelson/temporal-serverless-worker/blob/main/SETUP.md

**Deployments**
- Deployment list page with empty state and capability guard
- Deployment detail page with compute badge, version table, and actions menu
- Create deployment form: name + compute provider picker (AWS Lambda / GCP Cloud Run), Lambda ARN, IAM Role ARN, scaling and limits (min/max instances)
- Delete deployment (available when no versions exist)
- Three-call orchestration on create: validate compute config → create deployment → create initial version, with rollback on failure

**Versions**
- Version table rows with lazy-loaded detail expansion, compute details, and actions menu
- Create version form wired to \`CreateWorkerDeploymentVersion\` API
- Edit/update version compute config
- Version actions menu rendered via portal to avoid table overflow clipping
- Filter version instances by deployment

**Compute Provider**
- Compute provider picker UI with AWS Lambda and GCP Cloud Run options
- AWS/GCP provider icons added to Holocene icon library
- Compute badge component for list/detail views
- Setup guide toggle for AWS Lambda setup instructions

**Infrastructure**
- \`deployments-service.ts\` with typed API calls for all CRUD operations
- \`serverless-worker-service.ts\` for legacy serverless worker calls
- \`CapabilityGuard\` component for feature-flagging UI sections
- Deployment sub-components extracted for reuse
- Deployments service unit tests
- Route helpers for deployment/version paths with test coverage
- Bump \`go.temporal.io/api\` to \`v1.62.8\`

## Test plan

- [ ] Deployment list renders with empty state when no deployments exist
- [ ] Create form validates ARN format on submit; errors appear inline
- [ ] Successful create calls API with correct payload (region extracted from Lambda ARN)
- [ ] Delete button appears only when deployment has zero versions
- [ ] Version table expands row to show compute details
- [ ] Create version form validates and submits correctly
- [ ] Actions menu in version table renders above table overflow boundary
- [ ] Capability guard hides sections when feature is disabled
- [ ] All deployments service unit tests pass

PR #3351, opened 2026-04-29
Fix null conditionals in search attribute filter

laurakwhit:
## Description & motivation 💭 

When selecting `Is null` / `Is not null` in the Workflow search attribute filter it should now create the correct query. The filter was either being dropped (because `value === ''`) or emitting a malformed statement with the empty value interpolated in it.

Behavior was introduced when value was set to empty string [here](https://github.com/temporalio/ui/pull/3236/changes#diff-340e6397a2f250727f71de58746d9ef34d63774f0736fb02d133b24982cdb22aR329).
   
### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

Go to `/workflows` > `+ Add Filter` > select a `Keyword` search attribute
   - [ ] Select `Is null` and verify it updates the query correctly
   - [ ] Select `Is not null` and verify it updates the query correctly

### DIFF
--- a/src/lib/components/workflow/filter-bar/dropdown-filter-chip.svelte
+++ b/src/lib/components/workflow/filter-bar/dropdown-filter-chip.svelte
@@ -326,7 +326,7 @@
         active={localFilter.conditional === option.value}
         on:click={() => {
           if (isNullConditional(option.value)) {
-            localFilter.value = '';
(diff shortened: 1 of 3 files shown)


GraceGardner (approved):
Thanks for getting this in

PR #3356, opened 2026-04-29
feat(DT-3944): SignalWithStart system Nexus obfuscation — decode infrastructure + UI

rossnelson:
## Summary

Temporal exposes a built-in Nexus endpoint, `__temporal_system`. A workflow calls it to signal-with-start another workflow, and that arrives in event history as generic plumbing — `NexusOperationScheduled` / `NexusOperationCompleted`, carrying an endpoint, a service, an operation name and a request ID. The transport, not what the caller did. The payload compounds it: a `SignalWithStartWorkflowExecutionRequest` encoded as `binary/protobuf`, which the UI had no way to read.

### Caller

| timeline | history |
|--------|--------|
| CleanShot 2026-08-25 at 13 41 26 CleanShot 2026-08-25 at 13 39 13  | CleanShot 2026-08-25 at 13 39 23  |

### Target

| timeline | history |
|--------|--------|
| CleanShot 2026-08-25 at 13 40 43 | CleanShot 2026-08-25 at 13 40 31 | 

### History Cards 

CleanShot 2026-08-25 at 13 43 13

This decodes the payload in the browser and renders the operation instead — on both sides of the call — and puts all operation knowledge behind one registry so the next operation is additive.

Closes DT-3945, DT-3946, DT-3947, DT-3948, DT-3949, DT-4385. Part of DT-3944.

## What a reviewer sees

**Caller history** — events 5 and 6 read **Signal With Start Workflow Execution Initiated** and **Delivered**, in signal pink with the signal icon rather than the Nexus star. The Nexus wrapper fields (`endpoint`, `service`, `operation`, `requestId`) are hidden.

| Card | Shows |
| --- | --- |
| Initiated | Target Execution (link), Signal Name, and the input split into **Signal Input** and **Workflow Input** |
| Delivered | **Initiated Event ID** linking back to event 5, Target Execution, and the decoded result |

The two Target Execution links resolve differently, by necessity. Initiated has only a workflow ID — the run does not exist yet, which is the point of signal-with-start — so it opens the workflow list filtered by that ID. Delivered has the run ID from the response, so it links to that run.

**Target history** — `Workflow Execution Signaled` leads with **Caller Execution**, and expanded also shows **Caller Event**, deep-linking to `NexusOperationScheduled (5)` in the caller. The two histories connect in both directions.

**Expansion** — signal-with-start is synchronous and its two events carry nearly identical fields, so opening one event opens one card rather than the whole group. Opening the collapsed **Compact** row, which represents the group, still opens both.

## How it is built

All operation knowledge lives in `src/lib/system-nexus-endpoints/`. Host components never branch on an operation.

```
src/lib/system-nexus-endpoints/
├── index.ts                        registry + the public API
├── types.ts                        SystemNexusOperationDefinition contract
├── shared.ts                       link / target / state helpers
└── signal-with-start-workflow/
    ├── schemas.ts                  proto message types + bufbuild schemas
    ├── definition.ts               everything about this operation
    └── input-renderer.svelte       how its payload renders
```

Six consumers each ask one question and render the answer:

| Consumer | Asks |
| --- | --- |
| `event-card`, `event-summary-row` | `resolveSystemNexusEvent(event, ctx)` |
| `event-details-full` | `…?.expandsIndividually` |
| `timeline-graph-row` | `…?.timelineCategory` |
| `get-group-name` | `systemNexusGroupLabel(event)` |
| `decode-payload` | `schemaForMessageType(messageType)` |
| `event-card`, `input-and-results-payload` | `systemNexusInputRenderer(payload)` |

`resolveSystemNexusEvent` returns `null` for anything unregistered, so every other event falls through to existing rendering untouched.

Payload decoding is separate and general: `decode-binary-protobuf.ts` is a pure decoder taking a schema, wired into `decode-payload.ts` so any `binary/protobuf` payload with a registered `messageType` renders as JSON through every existing surface — event cards, compact rows, the JSON view, copy and download.

## Adding the next operation

```bash
pnpm system-nexus-operation:new StartWorkflowExecution --kind start-workflow --category workflow
```

Writes the three files and registers the operation in the kind union and the registry array. The generated code compiles and lints as-is; two TODOs mark what to fill in. `.claude/skills/system-nexus-operation/SKILL.md` documents the contract for whoever — or whatever — does it next.

## Security fixes from review

- **Query escaping.** The target search interpolated a decoded workflow ID straight into a visibility query literal. That ID is authored by the caller, so a quote closed the literal and changed the query. Backslashes and quotes are now escaped.
- **Endpoint verification.** A terminal Nexus event carries no endpoint, so the Completed branch trusted the payload's own `messageType` to decide an event was a system operation — meaning a handler on a tenant's own endpoint could stamp that type and be rendered as trusted. It now requires the operation's initiating event, which does carry the endpoint.

## Server support

This needs a Temporal server built from `main` (1.32 development, not 1.31.x), with `history.enableChasm` and `history.enableSignalWithStartFromWorkflow` enabled — the latter is namespace-scoped and defaults to off. Until 1.32 ships, these events arrive as ordinary Nexus operations and the UI renders them exactly as it does today, so merging early is safe.

## Known gaps

- **`NexusOperationFailed`, `TimedOut` and `Canceled` are not remapped.** Their attributes carry only `scheduled_event_id`, `failure` and `request_id` — no endpoint, no operation — so the event cannot identify itself. `stateVerbs` already holds the labels; the initiating-event plumbing added for the endpoint check is what the wiring needs.
- **`Control` and `Identity` render nothing.** Both exist in the proto, but no caller populates them; verified against live histories on upstream `main`. A design decision rather than a fix.
- **Caller-side links on the signaled event** key off the event type rather than the registry, because the target event carries nothing identifying the operation.
- **`WorkflowId="…"` interpolation appears in four other places** in the codebase, unrelated to this change. Only the one in this diff is fixed here.

## Test plan

- [x] `pnpm check` — 0 errors
- [x] `pnpm test -- --run` — 189 files, 2649 passing
- [x] `pnpm lint` — clean for tracked files
- [x] Caller history: both events relabelled, both link tiers, decoded Signal Input and Workflow Input
- [x] Target history: Caller Execution and Caller Event deep-link back to the caller
- [x] Expanding one event in **All** opens one card; the **Compact** group row opens both
- [x] Non-system Nexus operations and all other events render unchanged


Alex-Tideman:
How does this show in the Timeline? As a signal?

rossnelson:
> How does this show in the Timeline? As a signal?

@Alex-Tideman yes, ive added screenshots to the desc. The operation is green on completion and when you hover it shows the pink signal color.

Alex-Tideman (approved):

rossnelson:
🎉

PR #3359, opened 2026-04-30
refactor(DT-3906): Simple Svelte 5 migrations

tegan-temporal:
## Description & motivation 💭 

We want to (eventually) have all our code in Svelte 5 syntax.

This PR migrates components where the migration is very straightforward.
  - No named slots
  - No event handlers

## Testing 🧪 

Manual smoke testing all changes should be 1:1 (no behavior changes).
### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/components/lines-and-dots/svg/history-graph.svelte
+++ b/src/lib/components/lines-and-dots/svg/history-graph.svelte
@@ -11,27 +11,26 @@
   import HistoryGraphRowVisual from './history-graph-row-visual.svelte';
   import Line from './line.svelte';
 
-  export let groups: EventGroups;
-  export let history: WorkflowEventWithPending[];
+  interface Props {
+    groups: EventGroups;
+    history: WorkflowEventWithPending[];
+  }
 
-  $: workflowTaskGroups = groupWorkflowTaskEvents(
-    $filteredEventHistory,
-    $eventFilterSort,
+  let { groups, history }: Props = $props();
+
+  const workflowTaskGroups = $derived(
+    groupWorkflowTaskEvents($filteredEventHistory, $eventFilterSort),
   );
-  $: allGroups = [...workflowTaskGroups, ...groups];
+  const allGroups = $derived([...workflowTaskGroups, ...groups]);
 
   const { height, radius } = HistoryConfig;
 
   const nodeBuffer = 4 * radius;
   const maxWidth = 600;
 
-  let canvasWidth = nodeBuffer;
-  let visualWidth = 0;
-
-  $: canvasHeight = history.length * height;
-
-  $: getMaxWidth = (items: WorkflowEventWithPending[]) => {
-    items.forEach((event) => {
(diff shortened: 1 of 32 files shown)


src/lib/components/lines-and-dots/svg/history-graph.svelte
```diff
@@ -11,40 +11,40 @@
...
         $eventFilterSort,
       );
-      canvasWidth = Math.max(canvasWidth, offset * 1.75 * radius + nodeBuffer);
-      visualWidth = Math.min(canvasWidth - 2 * radius, maxWidth);
+      width = Math.max(width, offset * 1.75 * radius + nodeBuffer);
     });
-  };
+    return width;
+  });
+  const visualWidth = $derived(Math.min(canvasWidth - 2 * radius, maxWidth));
```
andrewzamojc, line 45: I like the untangling here. I find the svelte5 version so much easier to understand.

src/lib/components/lines-and-dots/workflow-error-stack-trace.svelte
```diff
@@ -42,5 +48,5 @@
   </Accordion>
 {/if}
 {#if failure?.cause}
-  <svelte:self failure={failure.cause} />
+  <Self failure={failure.cause} />
```
andrewzamojc, line 51: neat! TIL both old and new way here.

src/lib/components/payload-input.svelte
```diff
@@ -27,6 +37,12 @@
     }
   };
 
+  const computedError = $derived(!isValidInput(input));
+
+  $effect(() => {
+    error = computedError;
+  });
```
andrewzamojc, line 44: Bindable pattern makes sense to me.

src/lib/holocene/combobox/async-test.svelte
```diff
@@ -3,16 +3,20 @@
...
-  let loading = false;
+  const syncOptions = ['one', 'two', 'three'];
+  const asyncOptions = ['asyncone', 'asynctwo', 'asyncthree'];
+  let value = $state('');
+  let options = $state<string[]>(syncOptions);
+  let loading = $state(false);
   let abortController: AbortController | null = null;
 
-  $: i = 0;
+  let i = $derived(0);
```
andrewzamojc: Should this be state instead of derived? Or if it doesn't need to be reactive, then const?

andrewzamojc (approved):
This LGTM. Nice choice of changes to keep it easy to review and not too many changes needed to consumers in `cloud-ui`.

src/lib/holocene/combobox/async-test.svelte
tegan-temporal: Yeah it can be made into a `$state(0)` instead. $derived was a more direct translation of what was previously there but since the line had no dependencies it wasn't ever reseting back to 0.

PR #3360, opened 2026-04-30
refactor(DT-3906): Update components/event components to Svelte 5 syntax

tegan-temporal:
## Description & motivation 💭 

Update components/event components to Svelte 5 syntax

### DIFF
--- a/src/lib/components/auto-refresh-workflow.svelte
+++ b/src/lib/components/auto-refresh-workflow.svelte
@@ -4,9 +4,9 @@
   import { translate } from '$lib/i18n/translate';
   import { autoRefreshWorkflow } from '$lib/stores/event-view';
 
-  export let onChange: () => void;
+  let { onChange }: { onChange: () => void } = $props();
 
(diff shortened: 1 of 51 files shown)


src/lib/components/event/event-link.svelte
```diff
@@ -4,26 +4,41 @@
...
+    label = translate('nexus.link'),
+    href,
+    class: className = '',
+    linkClass = '',
+  }: Props = $props();
+
+  const resolvedHref = $derived(href ?? getEventLinkHref(link));
+  const resolvedValue = $derived(
+    value ??
+      (resolvedHref.split('workflows/')?.[1] || resolvedHref) ??
```
GraceGardner: Did you already look into this? I can see why you would have removed the null/undefined stuff. Not sure why we would be using this component in a case where this is undefined? But might want to just make sure that removing that is safe? Which you may have already done?

GraceGardner (commented):
Some weird flags that popped up about undefined and such that might be worth a look before merging? Happy to approve when you're ready though

PR #3363, opened 2026-05-01
refactor(DT-3906): Enable Svelte 5 runes on files not using legacy features

tegan-temporal:
## Description & motivation 💭 

Turn on runes mode for files not currently relying on legacy features.


KATIETOLER (approved):
This looks great, thanks!

PR #3364, opened 2026-05-01
refactor(DT-3906): More Svelte 5 Migrations (trivial ones)

tegan-temporal:
Migrates a bunch of components and routes from Svelte 4 legacy syntax to Svelte 5 runes. This batch is scoped to "trivial" components -- ones with straightforward props/event handler conversions and no complex reactivity, slot forwarding, or two-way bindings worth calling out.

### DIFF
--- a/src/lib/pages/workflow-query-modal.svelte
+++ b/src/lib/pages/workflow-query-modal.svelte
@@ -1,1 +1,0 @@
-<svelte:options runes />
--- a/src/lib/pages/workflow-query-modal.svelte
+++ b/src/lib/pages/workflow-query-modal.svelte
@@ -1,0 +1,1 @@
+<svelte:options runes />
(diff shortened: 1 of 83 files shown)


src/lib/pages/workflow-query-modal.svelte
```diff
@@ -0,0 +1 @@
+<svelte:options runes />
```
laurakwhit: We can probably just delete this file 🗑️

tegan-temporal: How confident are you? I thought it was weird to have a blank file here but wasn't sure if it was potentially still needed for a route

laurakwhit: Pretty confident. Checked both repos and doesn't look like it's being used anywhere.

tegan-temporal: Deleted

laurakwhit (approved):

PR #3368, opened 2026-05-04
refactor(DT-3906): UI Svelte 5 Migrations medium

tegan-temporal:
## Description & motivation 💭 

A collection of svelte 5 migrations


src/lib/components/is-legacy-cloud-guard.svelte
```diff
@@ -1,15 +0,0 @@
-<script lang="ts">
-  import { page } from '$app/stores';
-
-  import { isCloudMatch } from '$lib/services/settings-service';
```
laurakwhit, line 4: Can remove `isCloudMatch` if it's no longer being used anywhere.

src/lib/components/search-attribute-filter/duration-filter.svelte
```diff
@@ -10,10 +10,8 @@
 
   const { filter, handleSubmit } = getContext<FilterContext>(FILTER_CONTEXT);
 
-  $: ({ value } = $filter);
-  $: _value = value;
-
-  let isValid = true;
+  let _value = $derived($filter.value);
```
laurakwhit, line 13: I don't think we can bind a derived value. I think we might need something like this 🤔 

```suggestion
  let _value = $state($filter.value);
  $effect(() => {
    _value = $filter.value;
  });
```

src/lib/components/search-attribute-filter/number-filter.svelte
```diff
@@ -11,8 +11,7 @@
     getContext<FilterContext>(FILTER_CONTEXT);
   const min = 0;
 
-  $: ({ value } = $filter);
-  $: _value = value ? Number(value) : null;
+  let _value = $derived($filter.value ? Number($filter.value) : null);
```
laurakwhit, line 14: Same here.

src/lib/components/search-attribute-filter/text-filter.svelte
```diff
@@ -11,8 +11,7 @@
 
   const { filter, handleSubmit } = getContext<FilterContext>(FILTER_CONTEXT);
 
-  $: ({ value } = $filter);
-  $: _value = value;
+  let _value = $derived($filter.value);
```
laurakwhit, line 14: And here.

src/lib/components/search-attribute-filter/duration-filter.svelte
laurakwhit: Just seeing your comment r.e. [overriding derived values](https://svelte.dev/docs/svelte/$derived#Overriding-derived-values) 👀 And testing it out it seems to work ✅

laurakwhit (approved):

src/lib/components/is-legacy-cloud-guard.svelte
tegan-temporal: It's still used within the settings-service, it needs to still be exported in order to test it.

PR #3371, opened 2026-05-06
chore: upgrade TypeScript to v6.0.3

rossnelson:
## Summary

- Bumps `typescript` from `^5.2.2` to `6.0.3` (latest stable, released March 2026)
- Adds explicit `"strict": false` to `tsconfig.json` — TS 6 now defaults `strict` to `true`, so this preserves existing behavior
- Fixes two event handler type errors in `duration-filter.svelte` and `batch-operation-confirmation-form.svelte` caused by the TS 6 DOM lib now typing `on:input` as `InputEvent` instead of `Event`

## Why no other changes?

The project was already well-positioned for this upgrade:
- `moduleResolution: "bundler"` was already explicit (TS 6 changed the default)
- `module: "es2020"` was already explicit
- `esModuleInterop: true` was already explicit
- `.svelte-kit/tsconfig.json` already sets `"types": ["node"]` (TS 6 now defaults `types` to `[]`)
- `target: "es2019"` is still valid (minimum is `es2015`)
- No deprecated `import assert {}` syntax in the codebase

## Ecosystem compatibility

| Package | Peer dep | Status |
|---|---|---|
| `svelte-check` 4.4.6 | `typescript >=5.0.0` | ✅ |
| `typescript-eslint` 8.58.1 | `typescript >=4.8.4 <6.1.0` | ✅ |

## Test plan

- [x] `pnpm check` — 0 errors, 92 pre-existing warnings (all Svelte-related, not TS 6)
- [x] `pnpm install` — clean install with TS 6.0.3


tegan-temporal (approved):
LGTM

tegan-temporal (approved):
LGTM

PR #3379, opened 2026-05-06
Scope group-hover to Tooltip component

laurakwhit:
## Description & motivation 💭 

Tailwind's `group-hover:` matches any ancestor with group so                                         
  1. If a tooltip's slot content uses its own `group-hover:` utilities, hovering the tooltip wrapper triggers them.         
  2. If the tooltip is placed inside another group (a card, row, button), the tooltip wrapper inherits that hover state too.

This PR uses Tailwind's named-group syntax (`group/tooltip`) so the hover relationship is scoped to the `Tooltip` component only.                                             

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/holocene/tooltip.svelte
+++ b/src/lib/holocene/tooltip.svelte
@@ -100,7 +100,7 @@
   <!-- svelte-ignore a11y-no-static-element-interactions -->
   <div
     bind:this={wrapperElement}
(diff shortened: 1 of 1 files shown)


andrewzamojc (approved):
Great!

PR #3385, opened 2026-05-11
fix: portal maximizable to body to escape stacking context

rossnelson:
## Summary

- The `Maximizable` component's full-screen mode (`fixed z-100`) was rendering behind the side nav and header because `content-wrapper` (`relative overflow-auto`) creates a browser stacking context, trapping `z-100` within it — below the side nav's `z-30` at the root level
- Fixed by using the existing `portal` action from `portal-action.ts` to move the wrapper element to `document.body` when maximized, so `z-100` competes at the root stacking context level
- On minimize, the element is inserted back to its original DOM position using the saved parent/sibling references
- Uses the same `portalElement` naming convention as `drawer.svelte` and `portal.svelte`

## Test plan

- [ ] Navigate to `/namespaces/default/workers/deployments/create`
- [ ] Click the expand icon on the CloudFormation template code block — should go full screen covering nav and header
- [ ] Click minimize (or press Escape) — should return to the inline panel position
- [ ] Verify other `Maximizable` code blocks elsewhere (event history, workflow payloads) expand and minimize correctly

### DIFF
--- a/src/lib/holocene/maximizable/maximizable.svelte
+++ b/src/lib/holocene/maximizable/maximizable.svelte
@@ -1,6 +1,8 @@
 <script lang="ts">
-  import { type Snippet } from 'svelte';
+  import { onDestroy, type Snippet } from 'svelte';
   import { twMerge as merge } from 'tailwind-merge';
+
+  import { portal } from '$lib/holocene/portal/portal-action';
 
   import MaximizeButton from './button.svelte';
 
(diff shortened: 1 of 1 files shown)


src/lib/holocene/maximizable/maximizable.svelte
```diff
@@ -40,11 +42,45 @@
...
+  });
+
+  onDestroy(() => {
+    if (portalElement && wrapperEl) {
+      if (originalParent) {
+        originalParent.insertBefore(wrapperEl, originalNextSibling);
+      }
+      portalElement.destroy();
+      portalElement = null;
+    }
```
laurakwhit: Is the `insertBefore` necessary here? The component is unmounting anyways. 

```suggestion
      portalElement?.destroy();
      portalElement = null;
```

src/lib/holocene/maximizable/maximizable.svelte
```diff
@@ -40,11 +42,45 @@
...
+  });
 </script>
 
 <svelte:window onkeydown={escapeListener} />
 
 <div
```
laurakwhit, line 69: Might consider the use of `focusTrap` (similar to `drawer.svelte`) when `maximized`.

src/lib/holocene/maximizable/maximizable.svelte
```diff
@@ -40,11 +42,45 @@
       maximized = false;
     }
   };
+
+  let wrapperEl: HTMLElement | undefined = $state();
+  let originalParent: ParentNode | null = null;
+  let originalNextSibling: ChildNode | null = null;
+  let portalElement: ReturnType<typeof portal> | null = null;
+
+  $effect(() => {
```
laurakwhit, line 50: Actually, could we rewrite this $effect using it's cleanup return? Then we wouldn't need `onDestroy` and possible unwanted re-runs.

src/lib/holocene/maximizable/maximizable.svelte
```diff
@@ -44,36 +45,21 @@
...
+    const nextSibling = el.nextSibling;
+    const pe = portal(el, document.body);
 
-  onDestroy(() => {
-    if (portalElement && wrapperEl) {
-      if (originalParent) {
-        originalParent.insertBefore(wrapperEl, originalNextSibling);
+    return () => {
+      pe?.destroy();
+      if (parent) {
```
laurakwhit: NIT: We might want to guard against a detached parent here, seems like the CodeBlock might be likely to re-render frequently. Maybe with `parent?.isConnected`?

laurakwhit (approved):

PR #3386, opened 2026-05-11
Remove capability guard from Set Current Version menu item

rossnelson:
## Summary

- Removes the `serverScaledDeployments` capability guard from the **Set Current Version** menu item in the version actions menu, so it renders unconditionally regardless of server capability
- Adds a temporary dev override in `+layout.ts` that forces `serverScaledDeployments = false` to test the UI without the capability enabled

CleanShot 2026-05-11 at 15 47 30

## Test plan

- [ ] Verify **Set Current Version** appears in the version actions menu when `serverScaledDeployments` is disabled (via the layout override)
- [ ] Verify other guarded menu items (Edit, Validate Connection, Delete) are still hidden when the capability is disabled
- [ ] Remove the `+layout.ts` override before merging if this is not intended to ship

### DIFF
--- a/src/lib/components/deployments/version-actions-menu.svelte
+++ b/src/lib/components/deployments/version-actions-menu.svelte
@@ -45,11 +45,9 @@
           {translate('deployments.edit')}
         </MenuItem>
       </CapabilityGuard>
-      <CapabilityGuard capability="serverScaledDeployments">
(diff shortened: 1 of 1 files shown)


jaypipes (approved):
thanks! :)

PR #3393, opened 2026-05-13
Fix publicPath URL duplication when prefix appears as substring

### LINKED ISSUE #3325: Broken UI when using custom path since 2.48.1
**Describe the bug**
This bug has resurfaced: https://github.com/temporalio/ui/issues/3183 

The UI breaks when using a custom publicPath when using 2.48.1 and up.
Was fixed in 2.47.3

**To Reproduce**

1. Update config and set publicPath (env TEMPORAL_UI_PUBLIC_PATH) to /custom
2. Deploy temporal/ui 2.48.1
3. Navigate to the UI and click around.
4. The URL includes path twice. e.g., http://my-temporal-ui/custom/custom

**Expected behavior**
The URL should only contains publicPath once.

**Screenshots**

Image

**Desktop** (please complete the following information):
OS: MacOS 26.4.1
Browser: Chrome 147.0.7727.56 

**Additional context**

laurakwhit:
## Description & motivation 💭 

The substring `/custom` is short and generic enough to appear inside a URL path other than at position 0 (e.g. asset chunks like `/_app/immutable/chunks/custom-XXXXX.js`). Any time the middleware saw a request whose `RequestURI` contained `/custom` somewhere past position 0, it rewrote the `req.URL` to just the suffix and threw away the prefix. That mangled `URL.Path` then drove the redirect logic in `buildUIIndexHandler`.

This PRs fixes this by
* Always reading `RequestURI`(which middleware doesn't touch) instead of a mix of `RequestURI` and `URL.Path`
* Consolidating prefix check via `hasPathPrefix`(handles exact paths and sub-paths)
* Anchoring the rewrite regex at the start by changing `route.PublicPath` to use `middleware.RewriteWithConfig{ RegexRules: ... }` with an explicitly-anchored pattern (`^/custom/(.*?)$`) to prevent the substring matching

It also inlines `resolve()` into `withPrefix()` so deprecated `base` path is no longer needed.

**Note:** 308 is cached aggressively by browsers. Once a wrong redirect target is cached, the doubled URL sticks around until cache is cleared.

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

`pnpm test src/lib/utilities/route-for-base-path.test.ts`

`pnpm test src/lib/utilities/route-for.test.ts`

` cd server && go test ./server/route/...`

Run the following ⬇️ 
```
cd server

pnpm build:server 
pnpm dev:ui-server

docker build -t temporal-ui-local .
docker run -p 5252:8080 -e TEMPORAL_UI_PUBLIC_PATH=/custom -e TEMPORAL_ADDRESS=host.docker.internal:7233 temporal-ui-local

```

Then go to http://localhost:5252
 - [ ] Verify it redirects to  `/custom` and does not cause an infinite loop
 - [ ] Verify `/namespaces` redirects to `/custom/namespaces`

Clear browser cache and then run the following (**without** a `publicPath` set)  ⬇️ 

```
docker run -p 5252:8080 -e TEMPORAL_ADDRESS=host.docker.internal:7233 temporal-ui-local
```

Then go to http://localhost:5252
 - [ ] Verify the route works as expected

### DIFF
--- a/server/server/route/ui.go
+++ b/server/server/route/ui.go
@@ -87,20 +87,27 @@ func buildUIIndexHandler(publicPath string, assets fs.FS) (echo.HandlerFunc, err
 	}
 
 	return func(c echo.Context) (err error) {
-		hasPublicPathPrefix := strings.HasPrefix(c.Request().RequestURI, publicPath+"/")
-		isExactPublicPath := c.Request().URL.Path == publicPath
-		missingPublicPath := hasPublicPath && !hasPublicPathPrefix && !isExactPublicPath
-
-		if missingPublicPath {
-			cleanPath := path.Clean(c.Request().URL.Path)
-			target := publicPath + cleanPath
-			if c.Request().URL.RawQuery != "" {
-				target += "?" + c.Request().URL.RawQuery
+		reqPath, rawQuery := splitRequestURI(c.Request().RequestURI)
+		if hasPublicPath && !hasPathPrefix(reqPath, publicPath) {
+			target := publicPath + path.Clean(reqPath)
+			if rawQuery != "" {
+				target += "?" + rawQuery
 			}
 			return c.Redirect(http.StatusPermanentRedirect, target)
 		}
 		return c.Stream(200, "text/html", bytes.NewBuffer(indexHTMLBytes))
 	}, nil
+}
+
+func hasPathPrefix(p, prefix string) bool {
+	return p == prefix || strings.HasPrefix(p, prefix+"/")
+}
+
+func splitRequestURI(uri string) (reqPath, rawQuery string) {
+	if i := strings.Index(uri, "?"); i >= 0 {
+		return uri[:i], uri[i+1:]
+	}
+	return uri, ""
 }
 
 func buildUIAssetsHandler(assets fs.FS) echo.HandlerFunc {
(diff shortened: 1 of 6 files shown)


server/server/route/ui.go
```diff
@@ -87,22 +87,29 @@ func buildUIIndexHandler(publicPath string, assets fs.FS) (echo.HandlerFunc, err
...
-			target := publicPath + cleanPath
-			if c.Request().URL.RawQuery != "" {
-				target += "?" + c.Request().URL.RawQuery
+		reqPath, rawQuery := splitRequestURI(c.Request().RequestURI)
+		if hasPublicPath && !hasPathPrefix(reqPath, publicPath) {
+			target := publicPath + path.Clean(reqPath)
+			if rawQuery != "" {
+				target += "?" + rawQuery
 			}
 			return c.Redirect(http.StatusPermanentRedirect, target)
```
laurakwhit, line 96: Wondering if we should consider changing this to a 307 vs. a 308. Currently, any changes to the public path after rollout will require users to clear cached redirects.

Alex-Tideman (approved):

rossnelson (approved):

PR #3394, opened 2026-05-14
feat(history): show Nexus operation name in compact view 

KATIETOLER:
## Show Nexus operation name in compact view alongside Activity and Update DT-3903

# Testing
Used nexusScheduledOperation
Screenshot 2026-05-14 at 11 37 01 AM
 ## Summary

  Show the Nexus operation name in the workflow history compact view, the same way
  Activity name and Update name appear today.

  ## Why

  Internal ask: the compact history view shows the operation/type name for
  Activity and Update rows, but Nexus operation rows fell back to `endpointId` instead
  — making it hard to tell which Nexus operation a row represents without expanding it.
   This adds the operation name so you can read the full picture at a glance.

  Before: `Nexus Operation` row showed `EndpointId: <endpoint-id>`
  After: `Nexus Operation` row shows `Operation: <operation-name>`

  Mirrors the existing pattern — Activity shows `ActivityType`, Update shows `Name`,
  Nexus now shows `Operation`.

  ## What changed

  - `src/lib/utilities/get-single-attribute-for-event.ts` — added a
  `NexusOperationScheduled` branch in `getEventSummaryAttribute` that returns `{ key:
  'operation', value: event.attributes.operation }`, placed right after the existing
  Update branch so it wins over the `endpointId` fallback in `preferredSummaryKeys`.
  - `src/lib/utilities/get-single-attribute-for-event.test.ts` — updated the existing
  Nexus test to assert the new contract.

  No changes to the row component itself — `event-summary-row.svelte` already routes
  everything through `getPrimaryAttributeForEvent` and reads from `event.initialEvent`
  for grouped events, so the operation name surfaces automatically for both the
  standalone scheduled event and the grouped Nexus operation row.

  Pending Nexus operation rows (`pending-nexus-summary-row.svelte`) are intentionally
  **out of scope** for this PR — they can be a separate change if desired.

  ## Testing

  - `pnpm lint` — clean (no new errors/warnings)
  - `pnpm check` — clean
  - `pnpm test -- --run` — 1964 tests pass, including the updated Nexus assertion
  - Manual: ran the UI against a local workflow with Nexus operation events, toggled to
   Compact view, confirmed the operation name renders on the Nexus row

### DIFF
--- a/src/lib/utilities/get-single-attribute-for-event.ts
+++ b/src/lib/utilities/get-single-attribute-for-event.ts
@@ -17,6 +17,7 @@ import { has } from './has';
 import { isObject } from './is';
 import {
   isLocalActivityMarkerEvent,
+  isNexusOperationScheduledEvent,
   isWorkflowExecutionUpdateAcceptedEvent,
 } from './is-event-type';
 import {
@@ -321,6 +322,15 @@ export const getEventSummaryAttribute = (
       return {
         key: 'name',
         value: event.attributes.acceptedRequest.input.name,
+      };
+    }
+  }
+
+  if (isNexusOperationScheduledEvent(event)) {
+    if (event.attributes?.operation) {
+      return {
+        key: 'operation',
+        value: event.attributes.operation,
       };
     }
   }
--- a/src/lib/utilities/get-single-attribute-for-event.ts
+++ b/src/lib/utilities/get-single-attribute-for-event.ts
@@ -17,7 +17,6 @@ import { has } from './has';
 import { isObject } from './is';
 import {
   isLocalActivityMarkerEvent,
-  isNexusOperationScheduledEvent,
   isWorkflowExecutionUpdateAcceptedEvent,
 } from './is-event-type';
 import {
@@ -265,6 +264,7 @@ const preferredSummaryKeys = [
   'identity',
   'parentInitiatedEventId',
   'endpointId',
+  'operation',
 ] as const;
 
 /**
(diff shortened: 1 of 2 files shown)


andrewzamojc (approved):
LGTM

src/lib/utilities/get-single-attribute-for-event.ts
```diff
@@ -325,6 +323,15 @@ export const getEventSummaryAttribute = (
     }
   }
 
+  if (isNexusOperationScheduledEvent(event)) {
+    if (event.attributes?.operation) {
+      return {
+        key: 'operation',
```
laurakwhit: Could we just add `operation` to the list of `preferredSummaryKeys` instead?

KATIETOLER: oo probably

src/lib/utilities/get-single-attribute-for-event.ts
```diff
@@ -264,6 +264,7 @@ const preferredSummaryKeys = [
   'identity',
   'parentInitiatedEventId',
   'endpointId',
+  'operation',
```
laurakwhit, line 267: @KATIETOLER I think this will never end up being used because `input` also exists and since it comes earlier in the array will take precedence. The test in `src/lib/utilities/get-single-attribute-for-event.test.ts` is still returning the input (not the operation, which is what we wanted in this PR).

PR #3400, opened 2026-05-15
Fix codec server request URL

### LINKED ISSUE #3399: Codec server "could not connect" — new URL(type, endpoint) drops endpoint's last path segment (regression from #3345)
**Describe the bug**

Since #3345 merged on 2026-05-15, codec server requests fail to reach the configured endpoint when the endpoint URL has path segments and no trailing slash.

The new code in `src/lib/services/data-encoder.ts`:

```ts
const url = new URL(type, endpoint);
url.searchParams.set('preserveStorageRefs', 'true');
```

…relies on URL resolution, which treats the last path segment of `endpoint` as a "filename" and discards it. The previous code (``endpoint + `/${type}` ``) did not have this behavior. v2.47.0 (last tagged) is fine; the regression only affects deployments tracking `main`, including `cloud.temporal.io`.

Result: a configured endpoint of `https://example.com/v1/api/codec` is requested as `https://example.com/v1/api/decode`, returning 404 (no CORS headers), surfaced in the UI as "Codec Server could not connect."

Verified in the production `cloud.temporal.io` bundle today — the chunk containing `preserveStorageRefs` uses the `new URL(...)` form. Same browser, four URL-resolution variants:

| endpoint                             | `new URL('decode', endpoint)`               |
| ------------------------------------ | ------------------------------------------- |
| `https://example.com/v1/api/codec`   | `https://example.com/v1/api/decode` ❌       |
| `https://example.com/v1/api/codec/`  | `https://example.com/v1/api/codec/decode` ✅ |
| `https://example.com/v1/api/codec/x` | `https://example.com/v1/api/codec/decode` ✅ |
| (old) `endpoint + '/' + 'decode'`    | `https://example.com/v1/api/codec/decode` ✅ |

**To Reproduce**

1. Run a codec server with `/decode` and `/encode` mounted under a non-root path, e.g. `https://example.com/v1/api/codec/decode`.
2. In Temporal Cloud → Settings → Codec Server, set the endpoint to `https://example.com/v1/api/codec` (no trailing slash). The settings page strips any trailing slash you add, so this is the only saveable form.
3. Open any encrypted workflow.

**Expected behavior**

The UI requests `https://example.com/v1/api/codec/decode` and decodes payloads (behavior in v2.47.0 and earlier).

**Actual behavior**

The UI requests `https://example.com/v1/api/decode` → 404 → CORS preflight failure → "Codec Server could not connect" banner.

**Desktop**

- OS: macOS 15
- Browser: Chrome 148, Brave 1.x (Chromium 148)
- Temporal Cloud UI: production `cloud.temporal.io`, build post-2.47.0 (bundle contains `preserveStorageRefs` from #3345).

**Additional context**

Workaround for users today: configure the endpoint with a meaningless trailing segment, e.g. `https://example.com/v1/api/codec/x`. `new URL('decode', '…/codec/x')` resolves to `…/codec/decode` correctly. This isn't discoverable and is a temporary hack.

Suggested fix — either preserve the previous semantics with string concat:

```ts
const url = new URL(`${endpoint.replace(/\/$/, '')}/${type}`);
url.searchParams.set('preserveStorageRefs', 'true');
```

…or normalize to a trailing slash before resolving:

```ts
const base = endpoint.endsWith('/') ? endpoint : endpoint + '/';
const url = new URL(type, base);
url.searchParams.set('preserveStorageRefs', 'true');
```

The first form is closer to the docs' mental model ("endpoint + /decode") and to the v2.47.0 behavior. Either fix should also be applied to wherever the namespace-settings page (and the per-browser codec popover) strips trailing slashes on save — if those didn't strip the slash, the bug would have a self-service workaround.

cc @rossedfort (author of #3345).

rossedfort:
## Description & motivation 💭 
Constructing a new `URL()` with `type` and `endpoint` drops any route prefix the user has configured.

Screenshot 2026-05-15 at 2 23 08 PM

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

### DIFF
--- a/src/lib/services/data-encoder.ts
+++ b/src/lib/services/data-encoder.ts
@@ -75,8 +75,9 @@ export async function codeServerRequest({
         body: stringifyWithBigInt(payloads),
       };
 
-  const url = new URL(type, endpoint);
-  url.searchParams.set('preserveStorageRefs', 'true');
+  // explicitly not constructing a new URL here because it
+  // drops any route prefix the user has configured, eg localhost:8080/codec-server
+  const url = `${endpoint}/${type}?preserveStorageRefs=true`;
 
   const decoderResponse: Promise<PotentialPayloads> = fetch(url, requestOptions)
     .then((response) => {
--- a/src/lib/services/data-encoder.test.ts
+++ b/src/lib/services/data-encoder.test.ts
@@ -27,6 +27,25 @@ describe('Codec Server Requests for Decode and Encode', () => {
     vi.clearAllMocks();
   });
 
+  it('should preserve a route prefix if the user has one configured', async () => {
+    const mockFetch = vi.fn(async () => {
+      return {
+        json: () => Promise.resolve(payloads),
+      };
+    });
+    vi.stubGlobal('fetch', mockFetch);
+
+    codecEndpoint.set('http://localcodecserver.com/prefix');
+    await codeServerRequest({
+      type: 'decode',
+      payloads,
+    });
+    expect(mockFetch).toBeCalledWith(
+      'http://localcodecserver.com/prefix/decode?preserveStorageRefs=true',
+      expect.any(Object),
+    );
+  });
+
   it('should send a request and return decoded payloads', async () => {
     global.fetch = vi.fn(() =>
       Promise.resolve({
--- a/src/lib/utilities/decode-payload.test.ts
+++ b/src/lib/utilities/decode-payload.test.ts
@@ -341,7 +341,7 @@ describe('decodeEventAttributes', () => {
     );
 
     expect(mockFetch).toBeCalledWith(
-      expect.any(URL),
+      expect.any(String),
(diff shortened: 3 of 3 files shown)


GiantRobots (approved):

src/lib/services/data-encoder.ts
```diff
@@ -75,8 +75,9 @@ export async function codeServerRequest({
         body: stringifyWithBigInt(payloads),
       };
 
-  const url = new URL(type, endpoint);
-  url.searchParams.set('preserveStorageRefs', 'true');
+  // explicitly not constructing a new URL here because it
+  // drops any route prefix the user has configured, eg localhost:8080/codec-server
+  const url = `${endpoint}/${type}?preserveStorageRefs=true`;
```
tegan-temporal, line 80: @rossedfort I think simple concatenation fails if the endpoint already ends with a `/` character. Do we strip it out before hand somewhere?

PR #3402, opened 2026-05-15
Refactor status counts and refresh button

laurakwhit:
## Description & motivation 💭 

  - Swaps `ActivityCountRefresh` for the shared `CountRefreshButton` on standalone activities page
  - Adds formatted refresh time to the workers page header
  - Adjusts workers and workflows header spacing for the refresh button
  - Unifies `WorkflowCounts` and `ActivityCounts` into one `StatusCounts` component

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


src/lib/utilities/get-activity-status-and-count.ts
```diff
@@ -37,11 +39,13 @@ export const toActivityStatus = (
...
 
-export const getActivityStatusAndCountOfGroup = (groups = []) => {
+export const getActivityStatusAndCountOfGroup = (
+  groups: CountWorkflowExecutionsResponse['groups'] = [],
+): { status: ActivityStatus; count: number }[] => {
   return groups
     .map((group) => {
       const rawStatus = parseRawPayloadToJSON(
-        group?.groupValues[0],
+        (group?.groupValues as unknown as Payload[])[0],
```
rossedfort: What type is `group.groupValues` ? I would like to avoid type casting payload related stuff as it previously was done fairly frequently and caused a huge headache when refactoring the payload decoder code in preparation for supporting external payloads.

src/lib/pages/workflows-with-new-search.svelte
```diff
@@ -71,6 +73,19 @@
 
   const refreshTimeFormatted = $derived($timestamp(refreshTime));
 
+  const hasTaskFailureAttribute = $derived(
+    !!page.data.namespace?.namespaceInfo?.capabilities
+      ?.reportedProblemsSearchAttribute,
+  );
+
+  $effect(() => {
+    void refreshTime;
```
rossedfort, line 82: Is this so that this effect will re-run when refreshTime changes?

src/lib/utilities/get-activity-status-and-count.ts
laurakwhit: It was typed to `Payloads`, but updated the response to be `Payload[]` instead in https://github.com/temporalio/ui/pull/3402/changes/dd1b3c833a2d12233e0802214854a5bf2fbb8cb6 ✅

src/lib/pages/workflows-with-new-search.svelte
laurakwhit: Correct, we want to update the task failure counts on refresh similar to other status counts.

rossedfort (approved):

PR #3404, opened 2026-05-18
Fix pauseLiveUpdate logic and remove currentEventHistory store

Alex-Tideman:
## Description & motivation 💭 

This PR updates workflow run loading so live polling respects pauseLiveUpdates.

When live updates are paused, workflow history and metadata requests are made without the live polling abort signal, which prevents the event history endpoint from using `waitNewEvent=true`. The layout also aborts any existing live polling controller when pause is enabled so an already-running long poll does not continue in the background. 

It also removes currentEventHistory store: currentEventHistory was a workaround for the older behavior: keep the rendered event list frozen while fullEventHistory continued to update in the background. Now the fix is at the source: when pauseLiveUpdates is true, live polling is not started, and any in-flight polling controller is aborted. With the guard:

if (signal?.aborted) return;

The aborted fetch no longer writes [] back into $fullEventHistory.

## Why

The history page correctly stopped updating when `pauseLiveUpdates` was enabled, but an in-flight workflow run poll could still continue after the user paused live updates. This caused background history refetch behavior even though the UI indicated auto refresh was off. Also, if a user refreshed the page with refresh_off=true, the history would not load, showing only pending events.

This change makes the pause state apply consistently to workflow details, metadata, and event history loading.

  - Verified `/history?refresh_off=true` still loads initial event history.
  - Verified pausing live updates stops live polling/refetch behavior.
  - Verified workflow details and history remain usable while live updates are
  paused.

## Testing 🧪 
Pause Live Updates during a running workflow and ensure no new events render and the workflow doesn't update. 

Pause Live Updates during a running workflow and then unpause and ensur new events render and the workflow does update. 

Pause Live Updates during a running workflow, wait, and refresh the page. Ensure new events are rendered from the pause to the refresh, but ensure no new events render.

### How was this tested 👻 

- [x] Manual testing
- [x] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/layouts/workflow-run-layout.svelte
+++ b/src/lib/layouts/workflow-run-layout.svelte
@@ -18,11 +18,7 @@
   import { fetchWorkflow } from '$lib/services/workflow-service';
   import { resetLastDataEncoderSuccess } from '$lib/stores/data-encoder-config';
   import { eventFilterSort, type EventSortOrder } from '$lib/stores/event-view';
-  import {
-    fullEventHistory,
-    pauseLiveUpdates,
-    timelineEvents,
-  } from '$lib/stores/events';
+  import { fullEventHistory, timelineEvents } from '$lib/stores/events';
   import {
     initialWorkflowRun,
     refresh,
@@ -47,6 +43,7 @@
   let workflowId = $derived(page.params.workflow);
   let runId = $derived(page.params.run);
   let showJson = $derived(page.url.searchParams.has('json'));
+  let pauseLiveUpdates = $derived(parseEventFilterParams(page.url).refresh_off);
   let fullJson = $derived({ ...$workflowRun, eventHistory: $fullEventHistory });
 
   let workflowError: NetworkError | null = $state(null);
@@ -90,7 +87,6 @@
     workflowId: string,
     runId: string,
   ) => {
-    const { settings } = page.data;
     const { workflow, error } = await fetchWorkflow({
       namespace,
       workflowId,
@@ -114,6 +110,7 @@
     $workflowRun = { ...$workflowRun, workflow, workers, workersLoaded: true };
 
     workflowRunController = new AbortController();
+    const signal = pauseLiveUpdates ? undefined : workflowRunController.signal;
 
     if (workflow.isRunning && workers?.pollers?.length) {
       getWorkflowMetadata(
@@ -124,20 +121,24 @@
             runId,
           },
         },
-        $pauseLiveUpdates ? undefined : workflowRunController.signal,
+        signal,
       ).then((metadata) => {
         $workflowRun.metadata = metadata;
       });
     }
 
-    $fullEventHistory = await fetchAllEvents({
+    const events = await fetchAllEvents({
       namespace,
       workflowId,
       runId,
       sort: 'ascending',
-      signal: $pauseLiveUpdates ? undefined : workflowRunController.signal,
+      signal,
       historySize: workflow.historyEvents,
     });
+
+    if (signal?.aborted) return;
+
+    $fullEventHistory = events;
   };
 
   const getOnlyWorkflowWithPendingActivities = async (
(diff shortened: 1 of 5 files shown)


src/lib/layouts/workflow-run-layout.svelte
```diff
@@ -124,20 +121,24 @@
...
         },
-        workflowRunController.signal,
+        signal,
       ).then((metadata) => {
         $workflowRun.metadata = metadata;
       });
     }
 
-    $fullEventHistory = await fetchAllEvents({
+    const events = await fetchAllEvents({
```
laurakwhit, line 130: Do we maybe want a try/catch here now? Same with `getWorkflowMetadata` above? Something that ignores the abort errors but rethrows anything else?

src/lib/layouts/workflow-run-layout.svelte
```diff
@@ -47,6 +43,7 @@
   let workflowId = $derived(page.params.workflow);
   let runId = $derived(page.params.run);
   let showJson = $derived(page.url.searchParams.has('json'));
+  let pauseLiveUpdates = $derived(parseEventFilterParams(page.url).refresh_off);
```
laurakwhit, line 46: Looks like we're still just setting `$pauseLiveUpdates = false` in `workflow-hsitory-layout.svelte`. Do we now need to update to something like this ⬇️ ?

```js
      if (isNotPending && $pauseLiveUpdates) {
        updateEventFilterParams(page.url, { refresh_off: false }, goto);
      }
```

src/lib/layouts/workflow-run-layout.svelte
```diff
@@ -197,18 +197,15 @@
 
   $effect(() => {
     const refreshValue = $refresh;
-    const pause = $pauseLiveUpdates;
+    const pause = pauseLiveUpdates;
+    if (pause && workflowRunController) {
+      workflowRunController.abort();
+    }
     untrack(() => {
       getOnlyWorkflowWithPendingActivities(refreshValue, pause);
```
laurakwhit, line 205: After pause (and abort) and then unpausing are we actually starting polling again? Or do we need to re-trigger `getWorkflowAndEventHistory` after it was paused and then unpaused? 

In the network tab, it doesn't look the `/history` is being called again (just the workflow summary endpoint).

src/lib/layouts/workflow-run-layout.svelte
```diff
@@ -114,6 +110,7 @@
     $workflowRun = { ...$workflowRun, workflow, workers, workersLoaded: true };
 
     workflowRunController = new AbortController();
```
laurakwhit, line 112: Not sure if this scenario is any real cause for concern, but probably best practice to abort any existing controller before constructing a new one
```suggestion
    workflowRunController?.abort();
    workflowRunController = new AbortController();
```

src/lib/layouts/workflow-run-layout.svelte
laurakwhit, line 169: NIT: Abort and _then_ clear. 
```suggestion
   if (workflowRunController) {
     workflowRunController.abort();
   }
   $fullEventHistory = [];
```

tests/integration/workflow-event-history.spec.ts
```diff
@@ -43,6 +43,21 @@ test.describe('Workflow History', () => {
...
+    await mockWorkflowApis(page);
+    await page.goto(`${workflowUrl}?refresh_off=true`);
+
+    await expect(page.getByTestId('pause')).toContainText('Auto Refresh Off');
+    await expect(
+      page
+        .getByTestId('event-summary-row')
+        .filter({ hasText: 'Workflow Execution Started' })
+        .first(),
+    ).toBeVisible();
```
laurakwhit, line 58: NIT: Could we also test for the mid-flight toggle path? Maybe something like this ⬇️?

```suggestion
    await mockWorkflowApis(page);
  
    const longPollPromise = page.waitForRequest(
      (req) => req.url().includes('waitNewEvent=true'),
      { timeout: 2000 },
    ).catch(() => null);
  
    await page.goto(`${workflowUrl}?refresh_off=true`);
  
    await expect(page.getByTestId('pause')).toContainText('Auto Refresh Off');
    await expect(
      page
        .getByTestId('event-summary-row')
        .filter({ hasText: 'Workflow Execution Started' })
        .first(),
    ).toBeVisible();

    const eventRows = await page.getByTestId('event-summary-row').count();
    expect(eventRows).toBeGreaterThan(1);
  
    expect(await longPollPromise).toBeNull();
```

PR #3420, opened 2026-05-21
fix: enable external payload download button with namespace level codec endpoint

### LINKED ISSUE #3419: External storage payload download button disabled when codec server is configured at namespace level
**Describe the bug**
When a codec server is configured at the namespace level, the download button for externally-stored payloads is permanently disabled. Hovering shows the tooltip "Add a codec server with a /download endpoint to download this payload." even though the codec server is reachable and actively decoding payloads on the same page.

**Steps To Reproduce**

1. Configure a codec server at the namespace level. Do not configure it via the per-browser "Configure Codec Server" dialog (top-right of any workflow page). If you've previously configured at the browser level, you'll need to clear the endpoint from local storage as well.
2. Run a workflow that produces a payload large enough to be offloaded to external storage via the SDK's ExternalStorage feature.
3. Navigate to that workflow execution in the Cloud UI.
4. Observe that workflow payloads are decoded correctly (the codec server `/decode` endpoint is being called successfully), but any payload that was offloaded to external storage shows a disabled download button.

**Expected behavior**
The download button should be enabled whenever a codec server endpoint is resolvable, whether set at the namespace level or as a browser-level override. The codec server is clearly reachable (it handles /decode for every page load), so the button should not be gated on a separate browser-local store.

**Screenshots**
Image

  Desktop:
  - OS: macOS 15.7.4
  - Browser: Chrome 146
  - Version: (Temporal Cloud UI, current production)

**Additional context**

The root cause seems to be in `src/lib/components/payload/payload-code-block.svelte`. The button's disabled and tooltip hide attributes check `$codecEndpoint`, the browser localStorage store, rather than getCodecEndpoint(page.data.settings), which is the function already used  by `codeServerRequest` in `data-encoder.ts` for all codec requests. `getCodecEndpoint` correctly falls back to `settings.codec.endpoint` when the browser store is empty, but the button check does not.

If desired, I can make a PR for this!

jtschulz:
## Description & motivation 💭
The download button for externally-stored payloads was gated on `$codecEndpoint` (the browser localStorage store), which is null when the codec server is configured only at the namespace/settings level. This left the button permanently disabled even though the codec server was reachable and actively serving /decode requests on the same page.

Replace the `$codecEndpoint` checks with `getCodecEndpoint(page.data.settings)`, which is already the authoritative resolver used by `codeServerRequest` for all encode/decode/download requests. It correctly falls back to `settings.codec.endpoint` when the browser-level override is not set.

image

### Design Considerations 🎨
I just tried to follow the pattern used elsewhere that namespace-level codec endpoints are respected, using `getCodecEndpoint`

### How was this tested 👻

- [ ] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

* Run the UI with a namespace-level codec endpoint configured, supporting the external storage /download endpoint
* Run a workflow with a large payload
* Observe that you can use the download button

### Issue(s) closed
#3419 

### Any docs updates needed?
n/a

### DIFF
--- a/src/lib/components/payload/payload-code-block.svelte
+++ b/src/lib/components/payload/payload-code-block.svelte
@@ -9,6 +9,8 @@
 </script>
 
 <script lang="ts">
+  import { page } from '$app/state';
+
   import Button from '$lib/holocene/button.svelte';
   import CodeBlock from '$lib/holocene/code-block.svelte';
   import Icon from '$lib/holocene/icon/icon.svelte';
@@ -17,7 +19,6 @@
   import Tooltip from '$lib/holocene/tooltip.svelte';
   import { translate } from '$lib/i18n/translate';
   import { downloadExternalPayloadWithCodec } from '$lib/services/data-encoder';
-  import { codecEndpoint } from '$lib/stores/data-encoder-config';
   import type { Payload, Payloads } from '$lib/types';
   import {
     isExternallyStoredRawPayload,
@@ -26,6 +27,7 @@
     type PayloadContainingObject,
   } from '$lib/utilities/decode-payload';
   import { formatBytes } from '$lib/utilities/format-bytes';
+  import { getCodecEndpoint } from '$lib/utilities/get-codec';
   import { isNetworkError } from '$lib/utilities/is-network-error';
   import { stringifyWithBigInt } from '$lib/utilities/parse-with-big-int';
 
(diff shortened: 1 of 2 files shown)


Alex-Tideman:
Appreciate it @jtschulz! Will get this in today.

Alex-Tideman (approved):

PR #3435, opened 2026-05-22
[DT-4039] Workflow query builder doesn't work with numeric search attributes

andrewzamojc:
## Summary

The workflow list query builder wrapped `Int` and `Double` search attribute values in double quotes (e.g. `Foo="1"`), can cause problems in open source with SQL db. The builder now emits `Foo=1`.

While fixing the serializer, the tokenizer also needed adjustment: it merged unquoted values with their conditional (e.g. `Foo=1` → tokens `Foo`, `=1`) and split 2-char operators like `>=` across two tokens with the second piece glued to the value. Boolean parsing had a local workaround for this; for numeric, the 2-char operators (`>=`, `<=`, `!=`) made that workaround insufficient. The tokenizer now emits the conditional as its own token and accepts operator-style 2-char conditionals without a trailing space. The boolean parser is simplified to match the new token shape.

Slack thread: https://temporaltechnologies.slack.com/archives/C048BE2NVRB/p1779408194209509

### BEFORE

Screenshot 2026-05-22 at 3 08 20 PM

### AFTER

Screenshot 2026-05-22 at 3 09 05 PM

### To Test

Use the preview link on this PR https://github.com/temporalio/cloud-ui/pull/2645 which combines this change with cloud-ui for easier previewing.

Notice no more double quotes on ints and doubles
Screenshot 2026-05-22 at 3 50 55 PM

## Jira

https://temporalio.atlassian.net/browse/DT-4039

## Test plan

- [x] Create or use an existing numeric custom search attribute (`Int` and/or `Double`)
- [x] Start a workflow with that attribute set
- [x] On the workflow list page, filter via the query builder using `=`, `>`, `>=`, `<=`, `<`, `!=` — confirm the URL shows `Foo=1` (not `Foo="1"`) and the list returns matching workflows
- [x] Refresh the page; confirm the filter chip rehydrates correctly
- [x] Confirm boolean and keyword filters still work unchanged

### DIFF
--- a/src/lib/utilities/query/tokenize.ts
+++ b/src/lib/utilities/query/tokenize.ts
@@ -11,6 +11,10 @@ import {
 
 type Tokens = string[];
 
+const OPERATOR_CONDITIONALS = new Set(['>=', '<=', '!=', '==']);
+const isOperatorConditional = (value: string): boolean =>
+  OPERATOR_CONDITIONALS.has(value);
+
 export const tokenize = (string: string): Tokens => {
   const tokens: Tokens = [];
   const addBufferToTokens = (): void => {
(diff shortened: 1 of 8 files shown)


src/lib/utilities/query/tokenize.ts
```diff
@@ -91,16 +95,19 @@ export const tokenize = (string: string): Tokens => {
...
+      // To prevent false positives like "inspect" being a "in" conditional, check for space, quote, or parenthesis after the midConditional.
+      // Operator-style conditionals like >=, <=, != never collide with identifiers, so accept them even without a trailing space.
       buffer += minConditional;
       addBufferToTokens();
       cursor += 2;
       continue;
     } else if (isConditional(character)) {
       addBufferToTokens();
       buffer += character;
+      addBufferToTokens();
```
Alex-Tideman, line 110: Just double checking, this was meant to be added twice?

andrewzamojc: Ya, this is intentional. For unquoted values, the old code glued the conditional to the value, and later we'd strip it with `filter.value = nextToken.replace('=', '');` . That got complicated when int/double came into play, so instead we separate the conditional token from the value token up front.

Alex-Tideman (approved):

PR #3443, opened 2026-05-22
fix(a11y): announce workflow / event status on timeline graph nodes (WCAG 1.4.1)

bilal-karim:
## Summary

Adds `aria-label` to each SVG `<g role="button">` wrapper on the timeline-graph nodes so screen readers announce the workflow's id and status (or the event's type and classification) instead of bare "button".

```diff
 <g
   role="button"
   tabindex="0"
+  aria-label={translate('workflows.row-accessible-name', {
+    workflowId: workflow.id,
+    status: statusLabel,
+  })}
   class="relative cursor-pointer"
   {height}
 >
```

Previously, terminal statuses on the timeline graph (Completed, Failed, Canceled, TimedOut, Fired, Signaled) were distinguishable **only by node color**. The legend that decodes those colors lives inside a `<Tooltip>` — keyboard- and AT-reachable since PR #3429, but still requires the user to leave the graph, decode the color in the legend, and come back to identify each node. This change makes the graph self-describing at the node level.

## Audit context

- WCAG 2.2 SC **1.4.1 Use of Color (Level A)** — Serious. Affects every workflow timeline view, which is the most-trafficked product surface for workflow investigation.
- Issue file: `audit-output/issues/1.4.1-timeline-graph-status.md` (this is Option A — the recommended fix).
- The sibling 1.4.1 fix (label required indicator) shipped in PR #3439.

## Cross-cutting wins

- Closes the SC **4.1.2 Name, Role, Value** missing-accessible-name defect on the same `<g role="button">` widgets.
- Closes the SC **1.3.1 Info and Relationships** matrix-row item d (deferred to the Robust wave).

## New i18n keys

- `workflows.row-accessible-name`: `"Workflow {{workflowId}}: {{status}}"`
- `events.row-accessible-name`: `"Event {{eventType}}: {{classification}}"`

## Test plan

- [x] Open a workflow with failed/completed children. Tab into the timeline graph; with VoiceOver/NVDA, confirm each node announces `"Workflow {id}: {status}, button"` instead of just `"button"`.
- [x] On the event history graph: tab to each dot; AT announces `"Event {eventType}: {classification}, button"`.
- [x] Right-click → Inspect a node to confirm the `<g>` has the expected `aria-label` attribute.
- [ ] axe-core `svg-img-alt` rule should no longer flag the timeline graph rows.
- [x] Zero visual change for sighted users.

A11y-Audit-Ref: 1.4.1-timeline-graph-status

### DIFF
--- a/src/lib/components/lines-and-dots/svg/history-graph-row-visual.svelte
+++ b/src/lib/components/lines-and-dots/svg/history-graph-row-visual.svelte
@@ -1,4 +1,5 @@
 <script lang="ts">
+  import { translate } from '$lib/i18n/translate';
   import type {
     EventGroup,
     EventGroups,
@@ -69,9 +70,27 @@
         : (undefined as EventTypeCategory | 'pending' | undefined),
   );
   const reverseSort = $derived($eventFilterSort === 'descending');
+
+  const classificationLabel = $derived(
+    classification
+      ? translate(`events.event-classification.${classification.toLowerCase()}`)
+      : translate('common.unknown'),
+  );
+
+  const accessibleName = $derived(
+    translate('events.row-accessible-name', {
+      eventType: event?.eventType ?? translate('common.unknown'),
+      classification: classificationLabel,
+    }),
+  );
 </script>
 
-<g role="button" tabindex="0" class="relative cursor-pointer">
+<g
+  role="button"
+  tabindex="0"
+  aria-label={accessibleName}
+  class="relative cursor-pointer"
+>
   {#if connectLine}
     <Line
       startPoint={[canvasWidth, y]}
--- a/src/lib/components/lines-and-dots/svg/history-graph-row-visual.svelte
+++ b/src/lib/components/lines-and-dots/svg/history-graph-row-visual.svelte
@@ -71,15 +71,45 @@
   );
   const reverseSort = $derived($eventFilterSort === 'descending');
 
+  const CLASSIFICATION_LABEL_KEY = {
+    Unspecified: 'events.event-classification.unspecified',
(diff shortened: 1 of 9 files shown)


ardiewen:
### `getStatusLabel` split + workflow/event status overlap

Extracting the shared label helper surfaced that **6 status names already overlap** between `WorkflowStatus` and `EventClassification`: `Running`, `Completed`, `Failed`, `Canceled`, `Terminated`, `TimedOut`. They're identical English today but live in separate i18n namespaces (`workflows.*` vs `events.event-classification.*`), so a future change to one could silently mislabel the other domain.

Handled in `d5c7186d`:
- **Domain resolvers** (exhaustive over their own union): `getWorkflowStatusLabel` → `workflows.*` (workflow-row), `getEventClassificationLabel` → `events.event-classification.*` (graph event nodes).
- **`getStatusLabel`** stays polymorphic for `WorkflowStatus.svelte` only — it renders one badge for any status (workflow/schedule/classification) and sees a bare string at runtime. On the 6 overlaps it prefers the workflow label, preserving existing behavior.

Covered by `get-status-label.test.ts` + `timeline-graph-accessible-names.spec.ts`.

@laurakwhit or @Alex-Tideman , do either one of you know why [these lines](https://github.com/temporalio/ui/pull/3443/changes#diff-255e057f86c169f60931a2e65eeaec570c6834ad9549c13baf757ffbdbfa0d2aL48-L68) represent some Workflow Statuses and Some EventClassification with overlap and aren't separated into their own translation domains?

ardiewen (approved):

PR #3446, opened 2026-05-22
Add event time filter to timeline

laurakwhit:
## Description & motivation 💭 

Adds the ability to filter timeline events based on a period of time. 

| End Time Disabled | End Time Error | Auto Refresh Disabled |
| - | - | - |
|Screenshot 2026-05-22 at 12 56 55 PM|Screenshot 2026-05-22 at 12 57 43 PM|Screenshot 2026-05-22 at 12 59 56 PM|

### How was this tested 👻 

- [x] Manual testing
- [x] E2E tests added
- [x] Unit tests added

* Go to `/workflows/[workflow]/timeline` > select the "Time" filter
   - [ ] Verify "End Time" is disabled by default
   - [ ] Verify applying just a "Start" time works as expected
   - Add an "End" time
      - [ ] Verify adding an end time after the start time errors and disables "Apply"
      - [ ] Verify it works as expected
      - [ ] Verify "Auto Refresh" is disabled w/ a tooltip if the workflow is still running
   - [ ] Verify "Clear all" works as expected

### Issue(s) closed 
`DT-3927`

### DIFF
--- a/src/lib/components/lines-and-dots/end-time-interval.svelte
+++ b/src/lib/components/lines-and-dots/end-time-interval.svelte
@@ -8,6 +8,7 @@
 
   export let workflow: WorkflowExecution;
   export let startTime: string | Timestamp;
+  export let overrideEndTime: string | undefined = undefined;
 
   let currentTime = Date.now();
 
@@ -16,7 +17,7 @@
     return currentTime + 1000;
   };
 
-  $: endTime = workflow?.endTime || rightNow();
+  $: endTime = overrideEndTime || workflow?.endTime || rightNow();
   $: duration = getMillisecondDuration({
     start: startTime,
     end: endTime,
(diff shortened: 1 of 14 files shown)


tigernaut:
**End time disabled:**
- the toggle for "end" + the lock icon in the field + disabled input nice and reinforce each other.
- the toggle's current state is a little unclear as i'm reading it as "off" - just want to confirm that's intended. does this default "end" to off? if so, it could mean less or zero extra input, lowering cognitive load vs presenting two fully active date/time pickers up front.

**End time error state**
- the error under the input is clear. 
- looks like the error is triggered by the 11 vs 12 time on the same date and the validation is comparing the full timestamp not just dates. wondering if users will understand why it's invalid as both show "Fri May 22, 2026" and the only differing value is the hour, i.e. could it be 11pm?

**Side questions**
- can users apply a start only filter? (in image 1 there is no "apply" button)
- for design team consideration: the segmented hrs/mins/sec are visually bulky and reads a bit heavy. although it is an existing pattern, wondering if we should revisit this and simplify.

PR #3467, opened 2026-05-27
feat(DT-4017): Schedules List UI Update

tegan-temporal:
## Description & motivation 💭 

Updates Schedules List View:
  - Add refresh button and count polling for schedules
  - Add workflow type link and filter/copy buttons

As part of the above changes, there were two refactors:
  - The count polling logic was abstracted to be reusable.
  - The Filter/Copy button overlays on table cells was abstracted to it's own component. Previously, it was more cumbersome, having to manually track the hover/focus with many event listeners and state.
  
  ### Minor Changes
  - Replace ad-hoc `id="filter-button"` / `id="copy-button"` focusout checks with a `composeEventHandlers` helper plus a `currentTarget.contains(next)` subtree check, and cover the helper with unit tests.
- Add `common.filter-items` translation as the default filter icon title.
- Key the schedules `recentActions` / `futureActionTimes` `{#each}` loops.
- Scope `.holocene-table-body` with `:where()` so consumers can override
  borders/hover without specificity bumps.
  

Screenshot 2026-05-27 at 2 54 16 PM

Screenshot 2026-05-27 at 2 54 38 PM

https://github.com/user-attachments/assets/efee37a0-d36b-4596-8025-53bd81d9a808

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

### Issue(s) closed 
https://temporalio.atlassian.net/browse/DT-4017


src/lib/holocene/filter-or-copy-buttons.svelte
```diff
@@ -44,7 +44,6 @@
         }}
         class="copy-or-filter-button"
         class:filtered
-        id="filter-button"
```
ardiewen, line 47: I think `src/lib/components/workers/workers-table/workers-table-cell.svelte` (L78) was out of scope of your PR, but it contains:

```typescript
if (nextTarget && !['filter-button', 'copy-button'].includes(nextTarget.id))
    hideFilterOrCopy();
```

May want to migrate this component as well due to the refactor.

tegan-temporal: Good catch!

src/lib/components/status-counts.svelte
```diff
@@ -84,11 +81,31 @@
...
+      void perPage;
+      void $refresh;
+    },
+  });
 
-  const initialIntervalSeconds = 60;
-  const maxAttempts = 20;
+  $effect(function propagateRefreshTimeChange() {
+    refreshTime = countPoller.refreshTime;
+  });
```
ardiewen, line 108: There's something funky going on here. I tested two consumer components that should be receiving propagated time, and neither of them are updating on polling:

```
src/lib/pages/standalone-activities.svelte
```

and

```
src/lib/pages/workflows-with-new-search.svelte
```

tegan-temporal: Thanks, I've updated to fix this. BTW there is a small behavior change. count polling will not update the refresh time. the refresh time will reflect the timestamp for the currently visible data, not including the "newCount".

ardiewen (approved):
LGTM!

PR #3498, opened 2026-06-04
feat: improve local-temporal skill score (90% → 94%)

yogesh-tessl:
Hey @Alex-Tideman 👋

truly impressive work. Four skills for auth testing, Svelte migration, worktree setup, and local Temporal dev. The fact that you already have a claude.yml in your workflows tells me you're serious about AI-assisted development, and the Svelte migration skill especially is the kind of thing that saves real time.

ran your skills through `tessl skill review` at work and found some targeted improvements. Here's the before/after:

| Skill | Before | After | Change |
|-------|--------|-------|--------|
| local-temporal | 90% | 94% | +4% |

<details>
<summary>Changes made</summary>

**Description improvement (main impact):**
- **local-temporal**: Expanded the frontmatter description from a general summary to list specific concrete actions - `make start`, `pnpm dev:local-temporal`, and syncing `go.temporal.io/api` proto versions. Added "proto version mismatches" as an additional trigger term so the skill activates when developers hit the common "unknown field" error.

**Content tightening:**
- Removed the redundant `(aliased to pnpm dev:ui-server)` clause from the "Why not pnpm dev?" section
- Condensed the proto sync explanation - collapsed the grpc-gateway architecture paragraph into a single sentence while preserving the critical "unknown field" diagnostic insight
- Renamed the section header from "Keeping the ui-server in sync with the Temporal server branch" to "Keeping the ui-server proto versions in sync" for scannability

</details>

---

also stress-tested your `local-temporal` skill against [a few real-world task evals](https://tessl.io/registry/skills/github/temporalio/ui/setup-worktree/evals) and it held up really well on the proto version mismatch debugging scenario. The 4-step sync workflow is exactly what a developer needs when they hit "unknown field" errors after switching branches. Kudos for that.

quick honest disclosure. I work at https://github.com/tesslio where we build tooling around skills like these. Not a pitch, just saw room for improvement and wanted to contribute.

if you want to self-improve your skills, or define your own scenarios to pressure test, just ask your agent (Claude Code, Codex, etc.) to evaluate and optimize your skill with Tessl. Ping me [@yogesh-tessl](https://github.com/yogesh-tessl), if you hit any snags.

### DIFF
--- a/.claude/skills/local-temporal/SKILL.md
+++ b/.claude/skills/local-temporal/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: local-temporal
-description: Instructions for running the UI against a local Temporal server build instead of the built-in CLI dev server. Use when asked how to start the dev environment, run the UI locally, or connect to a local Temporal repo.
+description: "Set up and connect the Temporal UI to a locally built Temporal server by starting the server via make start, running pnpm dev:local-temporal, and syncing go.temporal.io/api proto versions between the server branch and server/go.mod. Use when asked how to start the dev environment, run the UI locally, connect to a local Temporal repo, or troubleshoot proto version mismatches."
 ---
 
 # Local Temporal Dev Setup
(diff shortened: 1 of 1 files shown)


yogesh-tessl:
hey @Alex-Tideman, just bumping this in case it got lost in the queue, happy to address any feedback!

No pressure if reviews run on a schedule here.

PR #3508, opened 2026-06-08
feat(DT-4078): Add Set Ramping Version to Worker Deployment Versions

rossnelson:
## Summary

- Adds **Set Ramping Version** action to the Worker Deployment Version actions menu, matching CLI parity
- New `SetRampingVersionModal` with percentage input (0–100%), a no-active-pollers guard modal, a remove-ramping destructive button, and a warning when replacing an existing ramping version
- New `setRampingDeploymentVersion` and `removeRampingDeploymentVersion` service functions posting to the `set-ramping-version` API endpoint
- Extends `VersionActionsMenu` with a context-sensitive menu item ("Set Ramping Version" or "Edit Ramping Percentage")

CleanShot 2026-06-08 at 13 41 29

## Test plan

- [ ] Version with no active pollers → menu item opens info modal explaining pollers are required
- [ ] Version with active pollers → modal shows percentage input and "Set Ramping Version" confirm
- [ ] Setting a ramping percentage shows success toast and refreshes the table
- [ ] When a different version is already ramping, modal warns about the replacement
- [ ] "Remove Ramping" button appears only for the currently-ramping version and removes it on click
- [ ] "Edit Ramping Percentage" label shown for the currently-ramping version
- [ ] `confirmDisabled` when percentage is NaN, < 0, or > 100
- [ ] Current version does not show the ramping menu item

### DIFF
--- a/src/lib/components/deployments/set-ramping-version-modal.svelte
+++ b/src/lib/components/deployments/set-ramping-version-modal.svelte
@@ -1,0 +1,101 @@
+<script lang="ts">
+  import Button from '$lib/holocene/button.svelte';
+  import NumberInput from '$lib/holocene/input/number-input.svelte';
+  import Modal from '$lib/holocene/modal.svelte';
+  import { translate } from '$lib/i18n/translate';
+
+  interface Props {
+    buildId: string;
+    deploymentName: string;
+    open: boolean;
+    hasActivePollers: boolean;
+    isRamping: boolean;
+    existingRampingBuildId?: string;
+    existingRampingPercentage?: number;
+    percentage: number;
+    error?: string;
+    onConfirm: () => void;
+    onRemove: () => void;
+    onCancel: () => void;
+  }
+
+  let {
+    buildId,
+    deploymentName,
+    open,
+    hasActivePollers,
+    isRamping,
+    existingRampingBuildId,
+    existingRampingPercentage,
+    percentage = $bindable(),
+    error = '',
+    onConfirm,
+    onRemove,
+    onCancel,
+  }: Props = $props();
+</script>
+
+{#if !hasActivePollers}
+  <Modal
+    id="set-ramping-version-modal-{buildId}"
+    {open}
+    confirmText=""
+    cancelText={translate('common.close')}
+    hideConfirm
+    on:cancelModal={onCancel}
+  >
+    <h3 slot="title">
+      {translate('deployments.set-ramping-no-pollers-title')}
+    </h3>
+    <div slot="content">
+      <p class="text-sm">
+        {translate('deployments.set-ramping-no-pollers-body')}
+      </p>
+    </div>
+  </Modal>
+{:else}
+  <Modal
+    id="set-ramping-version-modal-{buildId}"
+    {open}
+    {error}
+    confirmText={translate('deployments.set-ramping-version')}
+    cancelText={translate('common.cancel')}
+    confirmDisabled={Number.isNaN(percentage) ||
+      percentage < 0 ||
+      percentage > 100}
+    on:confirmModal={onConfirm}
+    on:cancelModal={onCancel}
+  >
+    <h3 slot="title">{translate('deployments.set-ramping-version')}</h3>
+    <div slot="content" class="flex flex-col gap-4">
+      <p class="text-sm">
+        This will route a percentage of new traffic to Build ID
+        <span class="font-mono font-medium">{buildId}</span>
+        for <strong>{deploymentName}</strong>. The Current Version will continue
+        to receive the remaining traffic.
+      </p>
+      {#if existingRampingBuildId && existingRampingBuildId !== buildId}
+        <p class="text-sm">
+          Worker Deployment Version
+          <span class="font-mono font-medium">{existingRampingBuildId}</span>
+          is currently ramping at {existingRampingPercentage ?? 0}%.
+        </p>
+      {/if}
+      <NumberInput
+        id="ramping-percentage-{buildId}"
+        bind:value={percentage}
+        min={0}
+        max={100}
+        units="%"
+        label={translate('deployments.ramping-percentage-label')}
+      />
+    </div>
+    <svelte:fragment slot="footer">
+      {#if isRamping}
+        <Button variant="destructive" on:click={onRemove}>
+          {translate('deployments.remove-ramping')}
+        </Button>
+      {/if}
+    </svelte:fragment>
+  </Modal>
+{/if}
--- a/src/lib/components/deployments/set-ramping-version-modal.svelte
+++ b/src/lib/components/deployments/set-ramping-version-modal.svelte
@@ -10,9 +10,11 @@
     open: boolean;
     hasActivePollers: boolean;
     isRamping: boolean;
-    existingRampingBuildId?: string;
+    existingRampingVersion?: string;
     existingRampingPercentage?: number;
+    currentPercentage?: number;
     percentage: number;
+    loading?: boolean;
     error?: string;
     onConfirm: () => void;
     onRemove: () => void;
@@ -25,9 +27,11 @@
     open,
     hasActivePollers,
     isRamping,
-    existingRampingBuildId,
+    existingRampingVersion,
     existingRampingPercentage,
+    currentPercentage,
     percentage = $bindable(),
+    loading = false,
     error = '',
     onConfirm,
     onRemove,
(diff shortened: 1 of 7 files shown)


src/lib/components/deployments/set-ramping-version-modal.svelte
```diff
@@ -0,0 +1,101 @@
...
+    <div slot="content" class="flex flex-col gap-4">
+      <p class="text-sm">
+        This will route a percentage of new traffic to Build ID
+        <span class="font-mono font-medium">{buildId}</span>
+        for <strong>{deploymentName}</strong>. The Current Version will continue
+        to receive the remaining traffic.
+      </p>
+      {#if existingRampingBuildId && existingRampingBuildId !== buildId}
+        <p class="text-sm">
+          Worker Deployment Version
```
laurakwhit: NIT: Can we get the copy in this file in i18n?

src/lib/services/deployments-service.ts
```diff
@@ -218,6 +218,40 @@ export const setCurrentDeploymentVersion = async (
...
+): Promise<void> => {
+  const route = routeForApi('worker-deployment-set-ramping-version', {
+    namespace: request.namespace,
+    deploymentName: request.deploymentName,
+  });
+  await requestFromAPI<unknown>(route, {
+    options: {
+      method: 'POST',
+      body: stringifyWithBigInt({
+        build_id: request.buildId,
```
laurakwhit: Should this be camelCase instead? 

```suggestion
        buildId: request.buildId,
```

src/lib/services/deployments-service.ts
```diff
@@ -218,6 +218,40 @@ export const setCurrentDeploymentVersion = async (
...
+
+export const removeRampingDeploymentVersion = async (
+  request: DeploymentParameters,
+  onError?: ErrorCallback,
+): Promise<void> => {
+  const route = routeForApi('worker-deployment-set-ramping-version', request);
+  await requestFromAPI<unknown>(route, {
+    options: {
+      method: 'POST',
+      body: stringifyWithBigInt({ build_id: '' }),
```
laurakwhit: Same here. Or maybe we can just remove it entirely?

```suggestion
      body: stringifyWithBigInt({ buildId: '' }),
```

src/lib/services/deployments-service.ts
```diff
@@ -218,6 +218,40 @@ export const setCurrentDeploymentVersion = async (
...
+  const route = routeForApi('worker-deployment-set-ramping-version', {
+    namespace: request.namespace,
+    deploymentName: request.deploymentName,
+  });
+  await requestFromAPI<unknown>(route, {
+    options: {
+      method: 'POST',
+      body: stringifyWithBigInt({
+        build_id: request.buildId,
+        percentage: request.rampingVersionPercentage,
```
laurakwhit, line 245: Should we also be passing the `conflictToken` (similar to `deleteWorkerDeploymentVersion`) here and in `removeRampingDeploymentVersion`?
```suggestion
        percentage: request.rampingVersionPercentage,
         ...(conflictToken && { conflictToken }),
```

src/lib/components/deployments/version-table-row.svelte
```diff
@@ -145,6 +148,13 @@
     }),
   );
 
+  const hasActivePollers = $derived(
+    parseVersionStatus(drainageStatus).status !== 'Created',
```
laurakwhit, line 152: Isn't there an "Inactive" status? Would `hasActivePollers` also return `true` for that and is that expected behavior?

src/lib/components/deployments/version-table-row.svelte
```diff
@@ -203,6 +216,62 @@
...
+    rampingPercentage = isRamping
+      ? (routingConfig.rampingVersionPercentage ?? 0)
+      : 0;
+    setRampingError = '';
+    showSetRampingModal = true;
+  }
+
+  async function handleSetRamping() {
+    setRampingError = '';
+    await setRampingDeploymentVersion(
```
laurakwhit: Is there a loading state we could set so that a user can't submit this twice while the first request is pending?

src/lib/components/deployments/version-actions-menu.svelte
```diff
@@ -48,6 +52,13 @@
       <MenuItem onclick={onSetCurrent} disabled={isCurrent}>
         {translate('deployments.set-as-current')}
       </MenuItem>
+      {#if !isCurrent}
+        <MenuItem onclick={onSetRamping}>
```
laurakwhit: Instead of hiding, should we just disable this option? 
```suggestion
        <MenuItem onclick={onSetRamping} disabled={isCurrent || !hasActivePollers}>
```

src/lib/components/deployments/set-ramping-version-modal.svelte
```diff
@@ -0,0 +1,101 @@
...
+    </div>
+  </Modal>
+{:else}
+  <Modal
+    id="set-ramping-version-modal-{buildId}"
+    {open}
+    {error}
+    confirmText={translate('deployments.set-ramping-version')}
+    cancelText={translate('common.cancel')}
+    confirmDisabled={Number.isNaN(percentage) ||
```
laurakwhit, line 67: Should we also disable if the value is 0 or if it is the same as the existing value?

src/lib/components/deployments/version-actions-menu.svelte
laurakwhit: Also, does this need a `CapabilityGuard` as well?

src/lib/components/deployments/set-ramping-version-modal.svelte
```diff
@@ -0,0 +1,101 @@
...
+          is currently ramping at {existingRampingPercentage ?? 0}%.
+        </p>
+      {/if}
+      <NumberInput
+        id="ramping-percentage-{buildId}"
+        bind:value={percentage}
+        min={0}
+        max={100}
+        units="%"
+        label={translate('deployments.ramping-percentage-label')}
```
laurakwhit, line 98: Should we support decimals here? 
```suggestion
        label={translate('deployments.ramping-percentage-label')}
        step={0.1}
```

src/lib/components/deployments/version-table-row.svelte
```diff
@@ -145,6 +148,13 @@
     }),
   );
 
+  const hasActivePollers = $derived(
+    parseVersionStatus(drainageStatus).status !== 'Created',
+  );
+  const otherVersionRamping = $derived(
+    rampingBuildId && !isRamping ? rampingBuildId : undefined,
```
laurakwhit: Would we ever want the name as well the build Id here? Maybe we should pass `routingConfig.rampingDeploymentVersion` instead and display both in the modal title?

src/lib/components/deployments/set-ramping-version-modal.svelte
rossnelson: Fixed! Both strings already existed in the i18n file, the modal just wasn't using them.

src/lib/services/deployments-service.ts
rossnelson: Fixed, changed to `buildId`.

src/lib/services/deployments-service.ts
rossnelson: Fixed! changed to `buildId: ''`.

src/lib/services/deployments-service.ts
rossnelson: Fixed! added `conflictToken?` to both `setRampingDeploymentVersion` and `removeRampingDeploymentVersion` using the same conditional spread pattern as the delete functions.

src/lib/components/deployments/version-table-row.svelte
rossnelson: Yes, `Inactive` returns `true` and that's intentional. `Inactive` means the version has polled the server before (it was previously current or ramping) so workers exist and ramping makes sense. Only `Created` (never polled) should block ramping since no workers have connected yet.

src/lib/components/deployments/version-table-row.svelte
rossnelson: Fixed! added `setRampingLoading` state that is set before the request and cleared after, passed as `loading` to the modal to disable the confirm button while pending.

src/lib/components/deployments/version-actions-menu.svelte
rossnelson: Fixed! removed the `{#if !isCurrent}` block and replaced with `disabled={isCurrent}` so the item stays visible but is grayed out.

src/lib/components/deployments/set-ramping-version-modal.svelte
rossnelson: Fixed! `confirmDisabled` now also blocks on `percentage <= 0` and when `percentage === currentPercentage` (passed down from the parent as `routingConfig.rampingVersionPercentage` when `isRamping`).

src/lib/components/deployments/version-actions-menu.svelte
rossnelson: Intentionally no `CapabilityGuard` here, Set Ramping is a standard deployment feature, not Cloud/serverless-only. This matches Set Current Version which also has no guard.

src/lib/components/deployments/set-ramping-version-modal.svelte
rossnelson: Fixed! added `step={0.1}`.

src/lib/components/deployments/version-table-row.svelte
rossnelson: Fixed! `otherVersionRamping` now builds the canonical `${rampingDeploymentName}.${rampingBuildId}` string using the existing `rampingDeploymentName` derived. The modal prop was renamed from `existingRampingBuildId` to `existingRampingVersion` and the i18n string updated to `{{ version }}` accordingly.

src/lib/components/deployments/version-table-row.svelte
```diff
@@ -145,6 +148,15 @@
     }),
   );
 
+  const hasActivePollers = $derived(
```
laurakwhit: NIT: Would something like "canRampToVersion" be a little clearer here?

src/lib/components/deployments/version-table-row.svelte
```diff
@@ -203,6 +219,67 @@
...
+        rampingVersionPercentage: rampingPercentage,
+        conflictToken,
+      },
+      (err) => {
+        setRampingError =
+          (err as { body?: { message?: string } })?.body?.message ??
+          translate('deployments.set-ramping-error');
+      },
+    );
+    setRampingLoading = false;
```
laurakwhit: NIT: 

```js
    try {
      await setRampingDeploymentVersion(...);
    } finally {
      setRampingLoading = false;
    }
```

laurakwhit (approved):

src/lib/components/deployments/version-table-row.svelte
rossnelson: Done — renamed to `canRampToVersion`.

src/lib/components/deployments/version-table-row.svelte
rossnelson: Done — wrapped in `try/finally` for both `handleSetRamping` and `handleRemoveRamping`.

PR #3511, opened 2026-06-08
feat(DT-4079): miscellaneous WD UI fixes

rossnelson:
## Summary
- Hide "Set Current Version" in version actions menu when the version is already Current
- Re-arrange WD detail page header: Create New Version (primary), More Actions dropdown (View Workflows, Ramp to Unversioned Workers, Delete)
- Add Ramp to Unversioned Workers modal with a percentage input; calls `set-ramping-version` API with empty `build_id`
- Show alert banner on the deployment detail page when unversioned workers are actively ramping, with inline Edit and Delete actions
- Remove border box from ComputeBadge — render as plain `[icon] label` matching the Namespace Region style
- **[DT-4077 gap]** Add "Unset Current" to the current version's actions menu — calls `set-current-version` with an empty version string, which transitions the version to Draining and eventually Drained, allowing it to be deleted. Without this, a deployment could never be fully emptied (and therefore never deleted) because one version is always pinned as current.

## Test plan
- [ ] Current version row no longer shows "Set Current Version" in its actions menu; instead shows "Unset Current"
- [ ] Clicking "Unset Current" on the current version transitions it to Draining then Drained
- [ ] Deployment detail header shows "Create New Version" (primary) + "More Actions" dropdown
- [ ] More Actions contains View Workflows, Ramp to Unversioned Workers, and Delete
- [ ] Ramp to Unversioned Workers opens a modal with percentage input; confirms and refreshes
- [ ] When unversioned workers are ramping, a yellow banner appears with Edit/Delete inline actions
- [ ] Compute provider shows as icon + label without a border box in the version list

### DIFF
--- a/src/lib/components/deployments/delete-deployment-modal.svelte
+++ b/src/lib/components/deployments/delete-deployment-modal.svelte
@@ -44,7 +44,9 @@
     {open}
     {error}
     title={translate('deployments.delete-deployment')}
-    description={translate('deployments.delete-deployment-description')}
+    description={translate('deployments.delete-deployment-description', {
+      name: deploymentName,
+    })}
     entityName={deploymentName}
     {onConfirm}
     {onCancel}
--- a/src/lib/components/deployments/delete-deployment-modal.svelte
+++ b/src/lib/components/deployments/delete-deployment-modal.svelte
@@ -1,23 +1,52 @@
 <script lang="ts">
   import DeleteConfirmModal from '$lib/components/delete-confirmation-modal.svelte';
+  import Modal from '$lib/holocene/modal.svelte';
   import { translate } from '$lib/i18n/translate';
 
   interface Props {
     open: boolean;
     deploymentName: string;
+    hasVersions: boolean;
+    error?: string;
     onConfirm?: () => void;
     onCancel?: () => void;
   }
 
-  let { open, deploymentName, onConfirm, onCancel }: Props = $props();
+  let {
+    open,
+    deploymentName,
+    hasVersions,
+    error = '',
+    onConfirm,
+    onCancel,
+  }: Props = $props();
 </script>
 
-<DeleteConfirmModal
-  id="delete-deployment-modal"
-  {open}
-  title={translate('deployments.delete-deployment')}
-  description={translate('deployments.delete-deployment-description')}
-  entityName={deploymentName}
-  {onConfirm}
-  {onCancel}
-/>
+{#if hasVersions}
+  <Modal
+    id="delete-deployment-modal"
+    {open}
+    confirmText=""
+    cancelText={translate('common.close')}
+    hideConfirm
+    on:cancelModal={onCancel}
+  >
+    <h3 slot="title">{translate('deployments.cannot-delete-deployment')}</h3>
+    <div slot="content">
+      <p class="text-sm">
+        {translate('deployments.cannot-delete-deployment-body')}
+      </p>
+    </div>
+  </Modal>
+{:else}
+  <DeleteConfirmModal
+    id="delete-deployment-modal"
+    {open}
+    {error}
+    title={translate('deployments.delete-deployment')}
+    description={translate('deployments.delete-deployment-description')}
+    entityName={deploymentName}
+    {onConfirm}
+    {onCancel}
+  />
+{/if}
--- a/src/lib/components/deployments/deployment-table-row.svelte
+++ b/src/lib/components/deployments/deployment-table-row.svelte
@@ -11,7 +11,10 @@
   import MenuItem from '$lib/holocene/menu/menu-item.svelte';
   import Menu from '$lib/holocene/menu/menu.svelte';
   import { translate } from '$lib/i18n/translate';
-  import { deleteWorkerDeployment } from '$lib/services/deployments-service';
+  import {
+    deleteWorkerDeployment,
+    fetchDeployment,
+  } from '$lib/services/deployments-service';
   import type { ConfigurableTableHeader } from '$lib/stores/configurable-table-columns';
   import type { WorkerDeploymentSummary } from '$lib/types/deployments';
   import { parseVersionStatus } from '$lib/utilities/deployments';
@@ -38,11 +41,21 @@
 
   let showDeleteModal = $state(false);
   let deleteError = $state('');
+  let conflictToken = $state<string | undefined>(undefined);
+
+  async function openDeleteModal() {
+    showDeleteModal = true;
+    const result = await fetchDeployment({
+      namespace,
+      deploymentName: deployment.name,
+    });
+    conflictToken = result.conflictToken;
+  }
 
   async function handleDelete() {
     deleteError = '';
     await deleteWorkerDeployment(
-      { namespace, deploymentName: deployment.name },
+      { namespace, deploymentName: deployment.name, conflictToken },
       (err) => {
         deleteError =
           (err as { body?: { message?: string } })?.body?.message ??
@@ -151,7 +164,7 @@
         position="right"
         usePortal
       >
-        <MenuItem onclick={() => (showDeleteModal = true)} destructive>
+        <MenuItem onclick={openDeleteModal} destructive>
           {translate('common.delete')}
         </MenuItem>
       </Menu>
--- a/src/lib/components/deployments/deployment-table-row.svelte
+++ b/src/lib/components/deployments/deployment-table-row.svelte
@@ -4,8 +4,14 @@
   import CapabilityGuard from '$lib/components/capability-guard.svelte';
   import Timestamp from '$lib/components/timestamp.svelte';
   import Copyable from '$lib/holocene/copyable/index.svelte';
+  import Icon from '$lib/holocene/icon/icon.svelte';
   import Link from '$lib/holocene/link.svelte';
+  import MenuButton from '$lib/holocene/menu/menu-button.svelte';
+  import MenuContainer from '$lib/holocene/menu/menu-container.svelte';
+  import MenuItem from '$lib/holocene/menu/menu-item.svelte';
+  import Menu from '$lib/holocene/menu/menu.svelte';
   import { translate } from '$lib/i18n/translate';
+  import { deleteWorkerDeployment } from '$lib/services/deployments-service';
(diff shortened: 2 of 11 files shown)


bilal-karim:
> Remove border box from ComputeBadge — render as plain [icon] label matching the Namespace Region style

I noticed in https://github.com/temporalio/ui/pull/3518 PR screenshots that the border is around the icon + label:

image

<br >

Apologies if this wasn't clear, but it should be like Namespace Regions in the list table. I border should be around the cloud provider icon only.

Screenshot 2026-06-11 at 11 37 55 AM

src/lib/services/deployments-service.ts
```diff
@@ -218,6 +218,63 @@ export const setCurrentDeploymentVersion = async (
...
+): Promise<void> => {
+  const route = routeForApi(
+    'worker-deployment-set-ramping-version',
+    parameters,
+  );
+  await requestFromAPI<unknown>(route, {
+    options: {
+      method: 'POST',
+      body: stringifyWithBigInt({
+        build_id: '',
```
laurakwhit: Should this be camelCase?

```suggestion
        buildId: '',
```

src/lib/services/deployments-service.ts
```diff
@@ -218,6 +218,63 @@ export const setCurrentDeploymentVersion = async (
...
+  onError?: ErrorCallback,
+): Promise<void> => {
+  const route = routeForApi(
+    'worker-deployment-set-ramping-version',
+    parameters,
+  );
+  await requestFromAPI<unknown>(route, {
+    options: {
+      method: 'POST',
+      body: stringifyWithBigInt({ build_id: '' }),
```
laurakwhit: Same here. Also wondering if it should include `conflictToken`.

```suggestion
      body: stringifyWithBigInt({ buildId: '' }),
      ...(conflictToken && { conflictToken }),
```

src/lib/components/deployments/version-table-row.svelte
```diff
@@ -203,6 +204,11 @@
     onChange?.();
   }
 
+  async function handleUnsetCurrentVersion() {
+    await unsetCurrentDeploymentVersion({ namespace, deploymentName });
```
laurakwhit: Do we want any kind of error handling here? Maybe a toast notification at least?

laurakwhit: Also, do we want this behind a confirm modal as well?

src/lib/components/deployments/deployment-table-row.svelte
```diff
@@ -15,13 +21,38 @@
...
+    !!deployment.latestVersionSummary?.deploymentVersion,
+  );
+
+  let showDeleteModal = $state(false);
+  let deleteError = $state('');
+
+  async function handleDelete() {
+    deleteError = '';
+    await deleteWorkerDeployment(
+      { namespace, deploymentName: deployment.name },
```
laurakwhit: Do we want a `conflictToken` here?

src/lib/components/deployments/deployment-table-row.svelte
```diff
@@ -104,4 +135,38 @@
...
+        class="flex h-8 w-8 items-center justify-center"
+      >
+        <Icon name="vertical-ellipsis" class="h-4 w-4" />
+      </MenuButton>
+      <Menu
+        id="deployment-actions-{deployment.name}"
+        position="right"
+        usePortal
+      >
+        <MenuItem onclick={() => (showDeleteModal = true)} destructive>
```
laurakwhit: Does this need a `CapabilityGuard`?

src/lib/pages/deployment.svelte
```diff
@@ -54,21 +59,66 @@
...
+  }
 </script>
 
 {#await effectiveDeploymentPromise}
   <SkeletonTable rows={15} />
 {:then deployment}
   {@const info = deployment.workerDeploymentInfo}
+  {@const unversionedRampingPercentage =
+    !info.routingConfig?.rampingDeploymentVersion &&
+    info.routingConfig?.rampingVersionPercentage
```
laurakwhit: Could there be zero percent ramp? And if so, is that the same as no ramp?

src/lib/pages/deployment.svelte
```diff
@@ -54,21 +59,66 @@
...
+          translate('deployments.ramp-to-unversioned-error');
+      },
+    );
+    if (rampUnversionedError) return;
+    showRampUnversionedModal = false;
+    reload();
+  }
+
+  async function handleRemoveRampUnversioned() {
+    await removeRampingUnversionedWorkers({ namespace, deploymentName });
```
laurakwhit: Do we want any kind of error handling here as well?

src/lib/components/deployments/ramp-unversioned-modal.svelte
```diff
@@ -0,0 +1,62 @@
...
+  confirmText={translate('common.confirm')}
+  cancelText={translate('common.cancel')}
+  on:confirmModal={() => onConfirm?.(percentage)}
+  on:cancelModal={onCancel}
+>
+  <Button
+    slot="footer"
+    variant="destructive"
+    size="sm"
+    class={!onRemove ? 'invisible' : ''}
```
laurakwhit: Would we prefer conditional rendering (e.g. ` {#if onRemove}`) instead of setting `invisible`? Seems like that's the pattern most places follow.

src/lib/pages/deployment.svelte
```diff
@@ -5,14 +5,19 @@
   import CapabilityGuard from '$lib/components/capability-guard.svelte';
   import DeleteDeploymentModal from '$lib/components/deployments/delete-deployment-modal.svelte';
   import DeploymentHeader from '$lib/components/deployments/deployment-header.svelte';
+  import RampUnversionedModal from '$lib/components/deployments/ramp-unversioned-modal.svelte';
   import VersionTableRow from '$lib/components/deployments/version-table-row.svelte';
+  import Alert from '$lib/holocene/alert.svelte';
+  import Button from '$lib/holocene/button.svelte';
```
laurakwhit:
```suggestion
```

src/lib/components/deployments/delete-deployment-modal.svelte
```diff
@@ -1,23 +1,52 @@
...
+    {open}
+    confirmText=""
+    cancelText={translate('common.close')}
+    hideConfirm
+    on:cancelModal={onCancel}
+  >
+    <h3 slot="title">{translate('deployments.cannot-delete-deployment')}</h3>
+    <div slot="content">
+      <p class="text-sm">
+        {translate('deployments.cannot-delete-deployment-body')}
```
laurakwhit, line 37: Should we include the deployment name somewhere in the copy here?

src/lib/services/deployments-service.ts
rossnelson: Fixed, good catch.

src/lib/services/deployments-service.ts
rossnelson: Fixed -- switched to camelCase and added conflictToken.

src/lib/components/deployments/version-table-row.svelte
rossnelson: Added both -- error handling with inline error display and a confirm modal before the action fires.

src/lib/components/deployments/deployment-table-row.svelte
rossnelson: The list response doesn't include a conflictToken, only the full deployment fetch does. Left it out for now since adding it would require an extra fetch per row.

src/lib/components/deployments/deployment-table-row.svelte
rossnelson: Delete applies to all deployments regardless of compute type, so no guard needed here.

src/lib/pages/deployment.svelte
rossnelson: Good point -- fixed with != null so 0% is treated as an active ramp.

src/lib/pages/deployment.svelte
rossnelson: Added -- surfaces the API error message in the modal.

src/lib/components/deployments/ramp-unversioned-modal.svelte
rossnelson: Agreed, switched to {#if onRemove}.

src/lib/pages/deployment.svelte
rossnelson: Removed.

src/lib/components/deployments/delete-deployment-modal.svelte
rossnelson: Added the deployment name to the description.

src/lib/services/deployments-service.ts
```diff
@@ -218,6 +218,68 @@ export const setCurrentDeploymentVersion = async (
...
+  onError?: ErrorCallback,
+): Promise<void> => {
+  const route = routeForApi(
+    'worker-deployment-set-current-version',
+    parameters,
+  );
+  await requestFromAPI<unknown>(route, {
+    options: {
+      method: 'POST',
+      body: stringifyWithBigInt({ version: '' }),
```
laurakwhit: Should we use `buildId`? Looks like maybe `version` is deprecated. Also wondering about passing the `conflictToken` here as well. 

```suggestion
      body: JSON.stringify({
          buildId: '',
          ...(conflictToken && { conflictToken }),
        }),
```

src/lib/components/deployments/deployment-table-row.svelte
laurakwhit: Could we fetch the `conflictToken` when the user hits the confirm modal? Or is it just not needed?

rossnelson: Yes -- fetching the full deployment when the modal opens to get the conflictToken before the user confirms.

src/lib/services/deployments-service.ts
rossnelson: Fixed -- switched to buildId and added conflictToken.

laurakwhit (approved):

PR #3512, opened 2026-06-08
[DT-4100] Holocene → Table: allow pinned bottom "total"-style row

andrewzamojc:
## Summary

Screenshot 2026-06-08 at 4 55 40 PM

Adds an optional `footer` slot to the Holocene `Table` component, rendered in a sticky `<tfoot>` that mirrors the existing sticky header so a "Total"-style row stays pinned at the bottom of the scroll viewport — always visible just above the pagination controls instead of requiring a scroll to the end of the rows.

- New `footer` named slot, only rendered when content is provided (existing tables unchanged).
- Sticky footer styled like a normal striped body row (`surface-background`), opaque so rows don't bleed through while scrolling.
- `bottom: var(--table-footer-bottom, 0)` lets consumers lift the row above a fixed bottom bar (e.g. pagination controls).
- Storybook stories: standalone pinned total row, and one pinned above a pagination-style bar (also demonstrating consumer-side `font-mono tabular-nums` numerals).

Component-only change in this repo; consuming it on the cloud-ui usage page is a separate follow-up.

## Jira

https://temporalio.atlassian.net/browse/DT-4100

## Test plan

- [ ] Storybook: **Table → Pinned Total Row** — total row stays pinned while scrolling
- [ ] Storybook: **Table → Pinned Total Row, Above Footer Bar** — total row pins above the bar
- [ ] Tables without a `footer` slot render unchanged
- [ ] No regressions in adjacent table flows

### DIFF
--- a/src/lib/holocene/table/paginated-table/index.svelte
+++ b/src/lib/holocene/table/paginated-table/index.svelte
@@ -27,7 +27,7 @@
   export { className as class };
 
   let tableContainer: HTMLDivElement;
-  let footerBarHeight = 0;
+  let actionBarHeight = 0;
 
   $: tableOffset = tableContainer?.offsetTop
     ? tableContainer.offsetTop + 32
@@ -47,7 +47,7 @@
   bind:this={tableContainer}
   style="max-height: {maxHeight || `calc(100vh - ${tableOffset}px)`};
 
-  --table-header-h: 2.25rem; --table-footer-bottom: {footerBarHeight}px;"
+  --table-header-h: 2.25rem; --table-footer-bottom: {actionBarHeight}px;"
 >
   {#if loading}
     {#if $$slots.loading}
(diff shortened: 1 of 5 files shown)


bilal-karim:
Love this!

src/lib/holocene/table/paginated-table/index.svelte
```diff
@@ -46,7 +47,7 @@
   bind:this={tableContainer}
   style="max-height: {maxHeight || `calc(100vh - ${tableOffset}px)`};
 
-  --table-header-h: 2.25rem;"
+  --table-header-h: 2.25rem; --table-footer-bottom: {footerBarHeight}px;"
```
andrewzamojc: In the paginated table that there is already a sticky pagination bar at the bottom of the scroll region, but outside the table. This means my sticky table footer overlaps the sticky pagination. Two options, I went with #2

1) In this file, move the pagination control outside the scroll region so the footer can be the only sticky thing at the bottom. Only changes this component, but could impact every caller of PaginatedTable. More testing required

2) Give the tfoot a vertical offset so it sits above the pagination row. It has to be recalculated on resize because the height isn't static, but the perf impact should be minimal.

andrewzamojc:
doing it in cloud-ui instead https://github.com/temporalio/cloud-ui/pull/2794

PR #3515, opened 2026-06-09
fix(archival): use WorkflowTypeName attribute when filtering

royreznik:
archived workflows by type

## Description & motivation 💭 
The archival page now searching for "WorkflowType" but the actual name of the field inside the archival is "WorkflowTypeName"

### How was this tested 👻 

- [x] Manual testing

### DIFF
--- a/src/lib/components/workflow/workflows-summary-configurable-table/table-body-cell.svelte
+++ b/src/lib/components/workflow/workflows-summary-configurable-table/table-body-cell.svelte
@@ -93,7 +93,7 @@
 {#if filterableLabels.includes(label) || isCustomKeywordOrTextAttribute}
   {#if label === 'Type'}
     {@render renderFilterableTableCell({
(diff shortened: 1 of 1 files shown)


royreznik:
@laurakwhit Not sure who to tag here, just saw you are active contributer

laurakwhit:
Hey @royreznik, thanks for raising this! Looking at the docs it looks like [google cloud](https://github.com/temporalio/temporal/tree/main/common/archiver/gcloud#visibility-query-syntax) expects `WorkflowType` while [s3](https://github.com/temporalio/temporal/tree/main/common/archiver/s3store#visibility-query-syntax) expects `WorkflowTypeName`. 

Checked in with the server team and they're going to look into supporting both `WorkflowType` and `WorkflowTypeName` for s3, which means we don't need to update our logic here. Keep an eye out for support in the `1.32` [release](https://github.com/temporalio/temporal/releases)!

PR #3516, opened 2026-06-09
fix(DT-4080): surface error when validate connection has no compute config, warn on set current

rossnelson:
## Summary
- **Validate Connection empty pop-up**: When a version has no compute config (created outside the UI without IAM details), the modal now shows an actionable error message instead of a blank pop-up
- **Set Current warning**: When attempting to set a version with no compute config as current, a warning alert is shown in the confirmation modal

## Test plan
- [ ] Validate Connection on a version with no compute config shows "This version has no compute configuration..." error instead of empty modal
- [ ] Validate Connection on a valid version still shows pass/fail result as before
- [ ] Set Current Version modal shows a warning alert when the version lacks a compute configuration
- [ ] Set Current Version modal shows no warning for versions with valid compute config

### DIFF
--- a/src/lib/components/deployments/set-current-version-modal.svelte
+++ b/src/lib/components/deployments/set-current-version-modal.svelte
@@ -1,5 +1,4 @@
 <script lang="ts">
-  import Alert from '$lib/holocene/alert.svelte';
   import Icon from '$lib/holocene/icon/icon.svelte';
   import Modal from '$lib/holocene/modal.svelte';
   import { translate } from '$lib/i18n/translate';
@@ -10,7 +9,6 @@
     deploymentName: string;
     open: boolean;
     error?: string;
-    hasComputeConfig?: boolean;
     onConfirm: () => void;
     onCancel: () => void;
   }
@@ -21,7 +19,6 @@
     deploymentName,
     open,
     error = '',
-    hasComputeConfig = true,
     onConfirm,
     onCancel,
   }: Props = $props();
@@ -38,12 +35,6 @@
 >
   <h3 slot="title">{translate('deployments.set-as-current')}</h3>
   <div slot="content" class="flex flex-col gap-4">
-    {#if !hasComputeConfig}
-      <Alert
-        intent="warning"
-        title={translate('deployments.set-current-no-config-warning')}
-      />
-    {/if}
     <p class="text-sm">
       This will set Build ID <span class="font-mono font-medium">{buildId}</span
       >
--- a/src/lib/components/deployments/set-current-version-modal.svelte
+++ b/src/lib/components/deployments/set-current-version-modal.svelte
@@ -1,4 +1,5 @@
 <script lang="ts">
+  import Alert from '$lib/holocene/alert.svelte';
   import Icon from '$lib/holocene/icon/icon.svelte';
   import Modal from '$lib/holocene/modal.svelte';
   import { translate } from '$lib/i18n/translate';
@@ -9,6 +10,7 @@
     deploymentName: string;
     open: boolean;
     error?: string;
+    hasComputeConfig?: boolean;
     onConfirm: () => void;
     onCancel: () => void;
   }
@@ -19,6 +21,7 @@
     deploymentName,
     open,
     error = '',
+    hasComputeConfig = true,
     onConfirm,
     onCancel,
   }: Props = $props();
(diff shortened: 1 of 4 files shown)


src/lib/components/deployments/version-table-row.svelte
```diff
@@ -268,6 +271,9 @@
     {editHref}
     {workflowHref}
     {isCurrent}
+    hasComputeConfig={isVersionSummaryNew(version)
+      ? !!version.computeConfig
+      : true}
```
laurakwhit, line 276: To verify, for the old version summary we **don't** want to show this warning?

src/lib/components/deployments/version-actions-menu.svelte
```diff
@@ -45,14 +47,16 @@
           {translate('deployments.edit')}
         </MenuItem>
       </CapabilityGuard>
-      <MenuItem onclick={onSetCurrent} disabled={isCurrent}>
-        {translate('deployments.set-as-current')}
-      </MenuItem>
-      <CapabilityGuard capability="serverScaledDeployments">
-        <MenuItem onclick={onValidate}>
-          {translate('deployments.validate-connection')}
+      {#if hasComputeConfig}
```
laurakwhit: Do we want this check if we also have the check in the modal with an alert?

src/lib/components/deployments/set-current-version-modal.svelte
```diff
@@ -19,6 +21,7 @@
     deploymentName,
     open,
     error = '',
+    hasComputeConfig = true,
```
laurakwhit: Feels like this should default to `false` or we should make `hasComputeConfig` a required prop.

```suggestion
    hasComputeConfig = false,
```

src/lib/i18n/locales/en/deployments.ts
```diff
@@ -85,6 +85,10 @@ export const Strings = {
   'validate-connection-error': 'Failed to validate connection',
   'validate-connection-valid': 'Connection is valid',
   'validate-connection-invalid': 'Connection is invalid',
+  'validate-connection-no-config':
+    'This version has no compute configuration. It may have been created outside the UI without the required IAM details.',
```
laurakwhit: NIT: Does this sound any better? Feel free to ignore if copy has already been reviewed.
```suggestion
    'This version is missing compute configuration. It may have been created outside the UI without the required IAM details.',
```

src/lib/i18n/locales/en/deployments.ts
```diff
@@ -85,6 +85,10 @@ export const Strings = {
...
+  'set-current-no-config-warning':
+    'This version has no compute configuration and may not function correctly as the current version.',
```
laurakwhit: NIT
```suggestion
    'This version is missing compute configuration. Setting it as current may not function as expected.',
```

src/lib/components/deployments/version-table-row.svelte
rossnelson: Right — old-shape summaries predate compute config so we can't determine it; treating them as configured avoids false warnings. This is moot now anyway since the modal warning was removed in favor of hiding the menu items (see other thread).

src/lib/components/deployments/version-actions-menu.svelte
rossnelson: Good catch! the modal alert became unreachable once the menu item was hidden, so I removed it and kept the menu check as the single source of truth. Also gated the hiding on the serverScaledDeployments capability: without it (OSS), Set as Current always shows via the guard fallback, since those versions never have computeConfig.

src/lib/components/deployments/set-current-version-modal.svelte
rossnelson: Removed the prop from the modal entirely along with the alert, it's required (no default) on VersionActionsMenu.

src/lib/i18n/locales/en/deployments.ts
rossnelson: Applied, thanks!

src/lib/i18n/locales/en/deployments.ts
rossnelson: This string was only used by the removed modal alert, so it's deleted now.

laurakwhit (approved):

PR #3517, opened 2026-06-09
fix(DT-4081): surface task queue not registered error in validate connection modal

rossnelson:
## Summary

- When a Lambda invokes successfully but errors before registering its Task Queue, Validate Connection previously showed nothing (silent failure)
- Now checks `taskQueueInfos` on the fetched version details and surfaces a specific message when empty: _"This Worker Deployment Version's Task Queue is not registered. The Lambda may be failing after invocation. Check the function for configuration or runtime errors, then re-run it from the AWS Console. Alternatively, create a new Version to retry."_

## Test plan

- [ ] Create a Worker Deployment Version that has compute config but no task queues registered
- [ ] Click "Validate Connection" → should show the "Task Queue not registered" message in the modal
- [ ] Verify versions with registered task queues still proceed to the normal validation flow

### DIFF
--- a/src/lib/components/deployments/version-table-row.svelte
+++ b/src/lib/components/deployments/version-table-row.svelte
@@ -169,6 +169,15 @@
       validateLoading = false;
       return;
     }
+    const taskQueueInfos =
+      versionDetails.workerDeploymentVersionInfo.taskQueueInfos;
+    if (!taskQueueInfos?.length) {
+      validateResult = {
+        message: translate('deployments.validate-connection-no-task-queue'),
+      };
+      validateLoading = false;
+      return;
+    }
     let errorMessage: string | undefined;
     await validateWorkerDeploymentVersionComputeConfig(
       { namespace, deploymentName, buildId: versionBuildId, computeConfig },
--- a/src/lib/components/deployments/version-table-row.svelte
+++ b/src/lib/components/deployments/version-table-row.svelte
@@ -19,6 +19,7 @@
     type VersionSummary,
   } from '$lib/types/deployments';
   import { parseVersionStatus } from '$lib/utilities/deployments';
+  import { getMilliseconds } from '$lib/utilities/format-time';
   import {
     getBuildIdFromVersion,
     getDeploymentFromVersion,
@@ -154,6 +155,8 @@
   let validateLoading = $state(false);
   let validateResult = $state<{ message?: string } | null>(null);
 
+  const newVersionGracePeriodMs = 2 * 60 * 1000;
+
   async function handleValidateConnection() {
     validateResult = null;
     validateLoading = true;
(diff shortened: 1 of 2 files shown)


src/lib/i18n/locales/en/deployments.ts
```diff
@@ -85,6 +85,8 @@ export const Strings = {
   'validate-connection-error': 'Failed to validate connection',
   'validate-connection-valid': 'Connection is valid',
   'validate-connection-invalid': 'Connection is invalid',
+  'validate-connection-no-task-queue':
+    "This Worker Deployment Version's Task Queue is not registered. The Lambda may be failing after invocation. Check the function for configuration or runtime errors, then re-run it from the AWS Console. Alternatively, create a new Version to retry.",
```
laurakwhit: Should the copy here be provider-neutral instead of using "AWS Console" in anticipation of GCP?

src/lib/components/deployments/version-table-row.svelte
```diff
@@ -169,6 +169,15 @@
       validateLoading = false;
       return;
     }
+    const taskQueueInfos =
+      versionDetails.workerDeploymentVersionInfo.taskQueueInfos;
+    if (!taskQueueInfos?.length) {
```
laurakwhit, line 179: Can newly created versions get this error? Should we verify it's not just newly created before showing this error?

src/lib/i18n/locales/en/deployments.ts
rossnelson: Good call. Updated the copy to be provider-neutral in 99df5310. "The Lambda may be failing" is now "The serverless function may be failing" and "AWS Console" is now "your cloud provider's console".

src/lib/components/deployments/version-table-row.svelte
rossnelson: Yes, they could have. Added a 2 minute grace period based on the version's createTime in 99df5310. Versions created within that window get a softer "the worker may still be starting up, wait a moment and retry" message instead of the failure guidance.

src/lib/components/deployments/version-table-row.svelte
```diff
@@ -172,8 +175,16 @@
     const taskQueueInfos =
       versionDetails.workerDeploymentVersionInfo.taskQueueInfos;
     if (!taskQueueInfos?.length) {
+      const createdMs = getMilliseconds(
+        versionDetails.workerDeploymentVersionInfo.createTime,
+      );
+      const isNewlyCreated = Date.now() - createdMs < newVersionGracePeriodMs;
```
laurakwhit: I think we actually want something like this ⬇️ here instead of `getMilliseconds`

```js
 const isNewlyCreated = Date.now() - new Date(createTime).getTime() < newVersionGracePeriodMs;
```

rossnelson: Good catch, getMilliseconds returns the milliseconds component of the date, not epoch time. Fixed in a0117904 using your suggestion.

laurakwhit (approved):

PR #3522, opened 2026-06-10
Date picker input field bug/kt

KATIETOLER:
Bug: 

The Holocene date-picker.svelte component has two user-facing bugs in its onInput handler. The displayed format (MM/DD/YYYY, 4-digit year) disagrees with the parser, which only fires when input is exactly 8 chars and assumes a 2-digit year (2000 + YY).

Impact:

Editing the date inline does nothing. Changing the prefilled year (e.g. 2026→2027) produces a >8-char string, so date change never fires. The displayed value updates but the bound value silently stays at the default or last calendar-clicked date — no validation feedback. On API key creation this means the key gets an expiration the user didn't choose.

Typing a 4-digit year produces a past date. Typing 06/05/2026 triggers parsing mid-type at 06/05/20 → year 2020; the final 26 makes it 10 chars so it's never re-parsed. The key silently expires in 2020.

Fix direction: Align displayed and accepted input on MM/DD/YYYY; parse only on a complete, valid date (reject partials and invalid dates like 02/30/2026); surface invalid input instead of silently keeping a stale value; apply isAllowed to typed input.

### DIFF
--- a/src/lib/holocene/date-picker.svelte
+++ b/src/lib/holocene/date-picker.svelte
@@ -52,7 +52,7 @@
     showDatePicker = true;
   };
 
-  const onInput = (e: Event) => {
+  const onBlur = (e: FocusEvent) => {
     const target = e.target as HTMLInputElement;
     const { date, error } = evaluateDatePickerInput(target.value, isAllowed);
     inputError = error;
@@ -97,7 +97,7 @@
     icon="calendar-plus"
     type="text"
     onfocus={onFocus}
-    oninput={onInput}
+    onblur={onBlur}
     placeholder={DATE_PICKER_INPUT_FORMAT}
     value={formatDatePickerInput(selected)}
     error={inputError}
--- a/src/lib/holocene/date-picker.svelte
+++ b/src/lib/holocene/date-picker.svelte
@@ -52,21 +52,20 @@
   });
 
   // handlers
-  const onFocus = () => {
-    showDatePicker = true;
+  const commitDate = (date: Date) => {
+    selected = date;
+    onDateChange?.(date);
   };
 
-  const onInput = (e: Event) => {
-    const target = e.target as HTMLInputElement;
-    const { date } = evaluateDatePickerInput(target.value, isAllowed);
-    if (date) onDateChange?.(date);
+  const onFocus = () => {
+    showDatePicker = true;
   };
 
   const onBlur = (e: FocusEvent) => {
     const target = e.target as HTMLInputElement;
     const { date, error } = evaluateDatePickerInput(target.value, isAllowed);
     inputError = error;
-    if (date) selected = date;
+    if (date) commitDate(date);
   };
 
   const next = () => {
@@ -89,8 +88,7 @@
 
   const handleDateChange = (d: Date) => {
     showDatePicker = false;
-    selected = d;
-    onDateChange?.(d);
+    commitDate(d);
   };
 
   const previousMonth = translate('date-picker.previous-month');
@@ -107,7 +105,6 @@
     icon="calendar-plus"
     type="text"
     onfocus={onFocus}
-    oninput={onInput}
     onblur={onBlur}
     placeholder={DATE_PICKER_INPUT_FORMAT}
     value={formatDatePickerInput(selected)}
@@ -152,10 +149,7 @@
         <button
           type="button"
           class="cursor-pointer text-[12px]"
-          onclick={() => {
-            selected = new Date();
-            onDateChange?.(selected);
-          }}
+          onclick={() => commitDate(new Date())}
         >
           {todayLabel}
         </button>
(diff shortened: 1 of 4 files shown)


src/lib/utilities/date-picker-input.ts
tegan-temporal, line 1: Can we rely on date-fns parse and format instead of rolling our own, I think the format string would be `MM/dd/yyyy`

- https://date-fns.org/v2.30.0/docs/parse
- https://date-fns.org/v2.30.0/docs/format

src/lib/holocene/date-picker.svelte
```diff
@@ -48,16 +54,9 @@
...
-    if (inputDate.length === 8) {
-      const inputDateSplit = inputDate?.split('/');
-      const yearEnd = parseInt(inputDateSplit[2]);
-      const year = 2000 + yearEnd;
-      const month = parseInt(inputDateSplit[0]) - 1;
-      const date = parseInt(inputDateSplit[1]);
-      const newDate = new Date(year, month, date);
-      onDateChange?.(newDate);
-    }
+    const { date, error } = evaluateDatePickerInput(target.value, isAllowed);
```
tegan-temporal, line 66: TBH It feels a bit off to me to validate while the user is typing. maybe we should validate onBlur instead?

KATIETOLER: Great point, thank you!

src/lib/utilities/date-picker-input.ts
KATIETOLER: oh god yeah. I forgot about fns once again.

src/lib/utilities/date-picker-input.test.ts
```diff
@@ -0,0 +1,122 @@
...
+
+  test('returns no date and no error for empty input', () => {
+    expect(evaluateDatePickerInput('')).toEqual({ date: null, error: false });
+    expect(evaluateDatePickerInput('   ')).toEqual({
+      date: null,
+      error: false,
+    });
+  });
+
+  test('returns no date and no error for incomplete entry (no dispatch)', () => {
```
tegan-temporal: I think the description for this test is wrong. We assert that there **is** an error in the test.

`'flags incomplete entry as an error without a date'` might be a better description

src/lib/holocene/date-picker.svelte
```diff
@@ -40,24 +45,18 @@
   let month = $derived(selected.getMonth());
   let year = $derived(selected.getFullYear());
   let showDatePicker = $state(false);
+  let inputError = $state(false);
```
tegan-temporal: I think we want `inputError` to clear whenever `selected` changes so making it a mutable derived might be better than state.

```suggestion
  // resets whenever selected changes. can override until then
  let inputError = $derived.by(() => {
    void selected;
    return false;
  });
```

src/lib/holocene/date-picker.svelte
```diff
@@ -96,9 +97,11 @@
...
     onfocus={onFocus}
-    oninput={onInput}
-    placeholder="MM/DD/YY"
-    value={selected.toDateString()}
+    onblur={onBlur}
+    placeholder={DATE_PICKER_INPUT_FORMAT}
+    value={formatDatePickerInput(selected)}
+    error={inputError}
+    hintText={inputError ? invalidDate : ''}
     clearable
```
tegan-temporal, line 114: I think we want to clear `inputError` if the user "clears" the input

```
onClear={() => (inputError = false)}
```

src/lib/holocene/date-picker.svelte
```diff
@@ -137,7 +140,10 @@
         <button
           type="button"
           class="cursor-pointer text-[12px]"
-          onclick={() => (selected = new Date())}
+          onclick={() => {
+            inputError = false;
```
tegan-temporal: I think this needs to call `onDateChange?.(selected)` because if the consumer relies on the handler instead of binding to selected the change won't propagate.

Also, if we do https://github.com/temporalio/ui/pull/3522/changes#r3405211704, that should clear the input error automatically since selected is changing

src/lib/holocene/date-picker.svelte
```diff
@@ -40,24 +45,18 @@
...
   let showDatePicker = $state(false);
+  let inputError = $state(false);
 
   // handlers
   const onFocus = () => {
     showDatePicker = true;
   };
 
-  const onInput = (e: Event) => {
+  const onBlur = (e: FocusEvent) => {
```
tegan-temporal, line 64: I had claude review this change and it caught a problem with going from onInput to onBlur. If someone is in a form and types in a date and then presses enter without blurring first, the form will submit the stale value instead of the current value.

So I think maybe my suggestion to go to onBlur was bad. We can either keep it onInput or split showing the validation error (would be on blur) and committing the change (would be onInput)

tegan-temporal (approved):

PR #3523, opened 2026-06-10
Add background to text labels in event history timeline

laurakwhit:
## Description & motivation 💭 

Text labels on the event-history timeline graph overlap the dotted vertical gridlines, making them hard to read.

This PR renders a small surface-colored SVG `<rect>` behind each non-backdrop timeline label, sized to the text's bounding box, so the text reads cleanly against the gridlines.

| Before | After |
|-|-|
|Screenshot 2026-06-10 at 10 11 03 AM|Screenshot 2026-06-10 at 8 57 58 AM|
|Screenshot 2026-06-10 at 10 10 37 AM|Screenshot 2026-06-10 at 10 02 20 AM|

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

  - [ ] Verify text labels are readable against the dotted gridlines on the timeline graph.
  - [ ] Expand a group row and confirm the background rect for labels in shifted rows tracks the text position. Background rectangles shouldn't appear on top of the expanded details.
  - [ ] Toggle dark mode and confirm the background fill matches the timeline background in both themes.
  - [ ] On events that have a `userMetadata.summary` confirm the background resizes to fit the decoded summary text after it loads.
  - [ ] Confirm hover/click on labeled timeline events still works as before (rect should not intercept events).
  - [ ] Verify existing `backdrop=true` case renders unchanged.

### Issue(s) closed 
`DT-4153`

### DIFF
--- a/src/lib/components/lines-and-dots/svg/text.svelte
+++ b/src/lib/components/lines-and-dots/svg/text.svelte
@@ -22,6 +22,12 @@
     config?: GraphConfig;
     label?: boolean;
     children?: Snippet;
+    /**
+     * Pass any value that captures the rendered content (e.g. the decoded text).
+     * Whenever it changes, the text bbox is re-measured so the background rect
+     * stays sized to the actual text.
+     */
+    contentKey?: unknown;
   };
 
   let {
@@ -37,6 +43,7 @@
     config = undefined,
     label = false,
     children,
+    contentKey,
   }: Props = $props();
 
   const [x, y] = $derived(point);
@@ -44,7 +51,11 @@
   let textElement: SVGTextElement = $state();
 
   const showIcon = $derived(icon && config);
-  const textBBox = $derived(textElement?.getBBox());
+  let textBBox = $state<DOMRect | undefined>();
+  $effect(() => {
+    void contentKey;
+    if (textElement) textBBox = textElement.getBBox();
+  });
   const textWidth = $derived(textBBox?.width || 0);
   const textHeight = $derived(textBBox?.height || 0);
   const backdropWidth = $derived(showIcon ? textWidth + 36 : textWidth + 12);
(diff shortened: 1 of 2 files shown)


src/lib/components/lines-and-dots/svg/text.svelte
```diff
@@ -37,17 +43,24 @@
...
-  const textWidth = $derived(textElement?.getBBox()?.width || 0);
-  const backdropWidth = $derived(showIcon ? textWidth + 36 : textWidth + 12);
+  let textElement: SVGTextElement | undefined = $state();
+
+  const iconInfo = $derived(icon && config ? { icon, config } : undefined);
+  let textBBox = $state<DOMRect | undefined>();
+  $effect(() => {
+    void contentKey;
+    if (textElement) textBBox = textElement.getBBox();
+  });
```
Alex-Tideman, line 58: This gonna cause performance concerns on large histories? Seems pretty innocent but just double checking.

Alex-Tideman:
Screenshot 2026-06-11 at 10 26 37 AM

Should we do the same for bars? Use a lighter color instead of opacity?

src/lib/components/lines-and-dots/svg/text.svelte
laurakwhit: The `getBBox()` was in a `$derived` previously, so no drastic change there. Only moved it to the `$effect` so that decoded content would re-trigger it and get the correct values based on the actual text.

src/lib/components/lines-and-dots/svg/text.svelte
```diff
@@ -93,6 +116,10 @@
     alignment-baseline: baseline;
   }
 
+  rect.text-background {
+    fill: rgb(var(--color-surface-primary));
```
laurakwhit, line 120: > Should we do the same for bars? Use a lighter color instead of opacity?

I think the main reason to add it for text is legibility, so I'm not sure it's needed for the bars. If we do a color audit here though might be worth updating.

Alex-Tideman (approved):

PR #3524, opened 2026-06-10
fix(workers): soften no-workers-polling alert for serverless worker deployments

rossnelson:
## What was changed

When a running workflow's task queue has no pollers but is served by a worker deployment with a compute config (serverless workers), the "No workers polling" warning is replaced with an informational alert instead of the alarming "check Kubernetes/ECS, restart workers" guidance.

CleanShot 2026-06-10 at 13 49 42

- `no-workers-polling-alert.svelte`: when the no-workers condition triggers and the workflow has a `TemporalWorkerDeployment` search attribute, fetch the deployment and check for a compute config. If found, render an info-intent alert explaining the deployment scales to zero when idle and workers start automatically, linking to the deployment page. The warning variant is held hidden until the deployment check resolves so the harsh copy never flashes first.
- `deployment-has-compute-config.ts` (+ tests): checks deployment-level compute config plus current/ramping/latest and per-version summaries for non-empty scaling groups, so a pinned older build with a compute config still counts.
- `deployments-service.ts`: `fetchDeployment` accepts an optional `onError` that suppresses the global error toast — a failed background check just falls back to the existing warning.
- New i18n keys for the serverless title, description, and deployment link.

## Why?

Serverless worker deployments scale to zero when idle, so an empty poller list from `DescribeTaskQueue` is an expected state, not a sign that workers crashed. The existing warning told users to check their orchestration system and restart workers, which is misleading for deployments Temporal manages.

Reported by Brandon: the alert appeared on a running workflow pinned to a serverless deployment (`loan-underwriting-app` / `build-1`) even though the deployment was handling tasks.

## Checklist

1. Closes [DT-4158]

2. How was this tested:

- Unit tests for `deploymentHasComputeConfig` covering deployment-level, current/latest/per-version summaries, and old-shape summaries
- `pnpm check`, `pnpm lint`, and full unit suite pass

3. Any docs updates needed?

No

[DT-4158]: https://temporalio.atlassian.net/browse/DT-4158?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/lib/components/workers/no-workers-polling-alert.svelte
+++ b/src/lib/components/workers/no-workers-polling-alert.svelte
@@ -20,37 +20,38 @@
 
   let serverlessDeployment = $state(false);
   let deploymentChecked = $state(false);
-  let lastCheckedDeployment: string | undefined = undefined;
 
   $effect(() => {
     if (!runningWithNoWorkers || !deployment) {
       serverlessDeployment = false;
       deploymentChecked = true;
-      lastCheckedDeployment = undefined;
       return;
     }
 
-    if (lastCheckedDeployment === deployment) return;
-    lastCheckedDeployment = deployment;
-
     deploymentChecked = false;
+    const controller = new AbortController();
+
     fetchDeployment(
       { namespace, deploymentName: deployment },
       fetch,
       () => {},
       false,
+      controller.signal,
     )
       .then((response) => {
+        if (controller.signal.aborted) return;
         serverlessDeployment = deploymentHasComputeConfig(
           response?.workerDeploymentInfo,
         );
+        deploymentChecked = true;
       })
       .catch(() => {
+        if (controller.signal.aborted) return;
         serverlessDeployment = false;
-      })
-      .finally(() => {
         deploymentChecked = true;
       });
+
+    return () => controller.abort();
   });
 </script>
 
--- a/src/lib/components/workers/no-workers-polling-alert.svelte
+++ b/src/lib/components/workers/no-workers-polling-alert.svelte
@@ -20,16 +20,26 @@
 
   let serverlessDeployment = $state(false);
   let deploymentChecked = $state(false);
+  let lastCheckedDeployment: string | undefined = undefined;
 
   $effect(() => {
     if (!runningWithNoWorkers || !deployment) {
       serverlessDeployment = false;
       deploymentChecked = true;
+      lastCheckedDeployment = undefined;
       return;
     }
 
+    if (lastCheckedDeployment === deployment) return;
+    lastCheckedDeployment = deployment;
+
     deploymentChecked = false;
-    fetchDeployment({ namespace, deploymentName: deployment }, fetch, () => {})
+    fetchDeployment(
+      { namespace, deploymentName: deployment },
+      fetch,
+      () => {},
+      false,
+    )
       .then((response) => {
         serverlessDeployment = deploymentHasComputeConfig(
           response?.workerDeploymentInfo,
--- a/src/lib/components/workers/no-workers-polling-alert.svelte
+++ b/src/lib/components/workers/no-workers-polling-alert.svelte
@@ -1,32 +1,88 @@
 <script lang="ts">
+  import { page } from '$app/state';
+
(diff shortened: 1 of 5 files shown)


src/lib/utilities/deployment-has-compute-config.ts
```diff
@@ -0,0 +1,25 @@
...
+  const versionConfigs = (deployment.versionSummaries ?? []).map((summary) =>
+    'computeConfig' in summary ? summary.computeConfig : undefined,
+  );
+
+  return [
+    deployment.computeConfig,
+    deployment.latestVersionSummary?.computeConfig,
+    deployment.currentVersionSummary?.computeConfig,
+    deployment.rampingVersionSummary?.computeConfig,
+    ...versionConfigs,
```
laurakwhit: Should we be checking every historical version or should we only be checking for current?

src/lib/services/deployments-service.ts
```diff
@@ -69,9 +69,13 @@ export const createWorkerDeployment = async (
 export const fetchDeployment = async (
   parameters: DeploymentParameters,
   request = fetch,
+  onError?: ErrorCallback,
 ): Promise<WorkerDeploymentResponse> => {
   const route = routeForApi('worker-deployment', parameters);
-  return requestFromAPI(route, { request });
+  return requestFromAPI(route, {
+    request,
+    ...(onError ? { onError, notifyOnError: false } : {}),
```
laurakwhit: Could we instead just make `notifyOnError` a parameter for `fetchDeployment` as well and pass it in where needed?

src/lib/components/workers/no-workers-polling-alert.svelte
```diff
@@ -1,32 +1,88 @@
...
+
+  $effect(() => {
+    if (!runningWithNoWorkers || !deployment) {
+      serverlessDeployment = false;
+      deploymentChecked = true;
+      return;
+    }
+
+    deploymentChecked = false;
+    fetchDeployment({ namespace, deploymentName: deployment }, fetch, () => {})
```
laurakwhit: Since `$workflowRun` is a polling store wondering if this could be triggered more often than we want/need 🤔

src/lib/utilities/deployment-has-compute-config.ts
rossnelson: Good call — narrowed it to just the routed versions (deployment-level config + current + ramping summaries) in 19029bbe. Dropped the latest/historical sweep since only current/ramping receive tasks.

src/lib/services/deployments-service.ts
rossnelson: Done in 19029bbe — `notifyOnError` is now a plain parameter (default `true` so existing callers are unchanged) and the alert passes `false`.

src/lib/components/workers/no-workers-polling-alert.svelte
rossnelson: The effect's deps are all primitives (boolean + strings), so store churn with unchanged values shouldn't re-run it — but to make it explicit I added a guard in 19029bbe that tracks the last-checked deployment name, so the fetch fires at most once per deployment regardless of polling.

src/lib/components/workers/no-workers-polling-alert.svelte
```diff
@@ -1,32 +1,98 @@
...
+      deploymentChecked = true;
+      lastCheckedDeployment = undefined;
+      return;
+    }
+
+    if (lastCheckedDeployment === deployment) return;
+    lastCheckedDeployment = deployment;
+
+    deploymentChecked = false;
+    fetchDeployment(
```
laurakwhit, line 34: Should we pass an abort here if the namespace or deployment changes so an older fetch isn't accidentally applied to alert state?

rossnelson: done!

laurakwhit (approved):

PR #3544, opened 2026-06-11
a11y(2.5.8): chip — pad remove button to 24×24 CSS px minimum

rosanusi:
## Description

The chip remove button (`<button><Icon name="close" /></button>`) had no padding and no explicit size, so its hit area collapsed to ~16×16 CSS px — below the WCAG 2.5.8 (Target Size Minimum, Level AA) requirement of 24×24.

**Fix:** add `inline-flex items-center justify-center p-1` to the button's class. `p-1` is 4 px each side, expanding the hit area to exactly 24×24 (`4 + 16 + 4`). The visible close icon is unchanged; `merge()` is already imported in the file. The button's `disabled ? 'hidden' : ''` hide behaviour is preserved.

The `IconButton` approach (recommended in the audit issue) was considered but deferred: `chip.svelte` is already in Svelte 5 runes mode while `icon-button.svelte` is Svelte 4 — `Button` explicitly types `onclick` as `never`, and using `on:click` syntax in a runes component triggers deprecation warnings. The padding approach satisfies 2.5.8, has zero interop risk, and keeps the diff minimal. A follow-on migration to `IconButton` can happen as part of the broader Svelte 5 component migration.

**Changed file:** `src/lib/holocene/chip.svelte` — 4 lines added to the remove button's `class`.

This fix cascades to all Chip consumers in both `ui-main` and `cloud-ui-main` (via the `@temporalio/ui` tarball).

## Screenshots

No visual change — close icon stays 16×16 px; padding is transparent. Hit area grows from ~16×16 to 24×24.

## Design

No design changes. The chip's outer `min-h-7` (28 px) container accommodates the 24-px button height with 2 px headroom each side.

## Testing

- [ ] Visit the workflows list with active filter chips visible.
- [ ] Inspect the chip remove button in DevTools: confirm computed size is 24×24 px.
- [ ] Tap-test on a touch device: confirm reliable activation without accidentally hitting adjacent chip content.
- [ ] Keyboard Tab to the remove button, press Enter/Space — chip is removed.
- [ ] Focus-visible ring is visible on the remove button.
- [ ] axe-core: no target-size violations on chip remove buttons.
- [ ] Visual regression: chip layout unchanged across filter-bar, search-attribute filter list, chip-input, and combobox chip variant.

## Checklist

- [x] No new i18n keys
- [x] No visual change to icon or chip container
- [x] Cascades to cloud-ui-main via `@temporalio/ui` tarball

## Docs

A11y-Audit-Ref: 2.5.8-chip-remove-button

### DIFF
--- a/scripts/a11y/manifest.yml
+++ b/scripts/a11y/manifest.yml
@@ -314,6 +314,13 @@
   bucket: 3
   severity: serious
   scope: ui-main
+- slug: 2.5.8-chip-remove-button
+  sc: '2.5.8'
+  bucket: 1
+  severity: moderate
+  scope: ui-main
+  files-touched:
+    - src/lib/holocene/chip.svelte
 - slug: 2.5.8-pagination-caret-buttons
   sc: '2.5.8'
   bucket: 2
--- a/src/lib/holocene/chip.svelte
+++ b/src/lib/holocene/chip.svelte
@@ -49,7 +49,10 @@
   {/if}
   <button
     aria-label={removeButtonLabel}
-    class={disabled ? 'hidden' : ''}
+    class={merge(
+      'inline-flex items-center justify-center p-1',
+      disabled ? 'hidden' : '',
+    )}
     data-track-name="chip"
     data-track-intent="remove"
     data-track-text={removeButtonLabel}
--- a/src/lib/holocene/chip.svelte
+++ b/src/lib/holocene/chip.svelte
@@ -64,7 +64,7 @@
 
 <style lang="postcss">
   .chip {
-    @apply surface-subtle flex min-h-7 w-fit min-w-fit flex-row items-center justify-between gap-1 whitespace-nowrap break-all rounded-sm p-1.5 text-sm leading-[1.5];
+    @apply surface-subtle flex min-h-7 w-fit min-w-fit flex-row items-center justify-between gap-1 whitespace-nowrap break-all rounded-sm p-1 pl-2 text-sm leading-[1.5];
 
     :global(.icon-button) {
       @apply ml-1 h-auto w-fit;


ardiewen:
@rosanusi just FYI the PR statement is incorrect about two things:

1) There _IS_ a visual change, though minor:

**Before:**
Screenshot 2026-06-17 at 11 31 37 AM

**After:**
Screenshot 2026-06-17 at 11 32 12 AM

**Explanation:**
> The icon itself never changed (16×16). Everything below flows from p-1 (4px padding on every side) making the button bigger, which then ripples out to the chip because the chip is sized to its contents.
>
> Width: 85 → 93px (+8px)
>
> - The remove button was just the bare 16px icon (padding: 0). p-1 adds 4px left + 4px right → button width becomes 4 + 16 + 4 = 24px. That's +8px of button width.
> - The chip is w-fit (shrink-to-fit) and lays its children out in a row, so the chip's width is the sum of its children. An 8px-wider button means an 8px-wider chip. The label and the inter-item gap-1 are unchanged, so the increase passes straight through: +8px.
>
> Height: 33 → 36px (+3px)
>
> The chip's height = (tallest child) + its own vertical padding. The chip has p-1.5 = 6px top + 6px bottom = 12px.
>
> - Before: the two children were the label text and the button. The label's line box is text-sm (14px) × leading-[1.5] ≈ 21px; the button was 16px. Tallest = 21px (the text). Chip = 21 + 12 = 33px.
> - After: p-1 makes the button 4 + 16 + 4 = 24px tall. Now the button (24px) is taller than the text line (21px), so it becomes the tallest child. Chip = 24 + 12 = 36px → +3px.
>
> So the height grew not because the chip's padding changed, but because the padded button overtook the text as the tallest item in the row.
>
> Why min-h-7 didn't absorb it
>
> The PR assumed the chip sits at min-h-7 (28px) with room to spare for a 24px button. But min-h-7 is a minimum, not a maximum — the chip's natural height (33px before, 36px after) already exceeds 28px, so that rule never engaged and couldn't cap the growth.
>
> (inline-flex items-center justify-center is a secondary effect: it centers the 16px icon inside the now-24px box and removes any inline whitespace, but the padding is what drives the size change.)

2) Testing instructions are wrong.

> "Visit the workflows list with active filter chips visible. Inspect the chip remove button: confirm 24×24."

But the "filter chips" on the workflows list are not that component. Tracing the render path, the workflows filter bar (filter-list.svelte:104) builds its active-filter pills from DropdownFilterChip and StatusFilterChip, which are Button-based dropdown pills (chevron ▾, click-to-edit) — they have no × remove button and no .chip markup at all.

`holocene/chip.svelte` only renders through ChipInput and combobox (multiselect). The most direct route:

- `/namespaces/default/workflows/start-workflow` → Search Attributes → Add a search attribute, set its type to "Keyword List" → the value field becomes a chip input; each entered value renders a real Chip with the enlarged × remove button.

ardiewen (approved):

ardiewen:
Rebalanced results here:

Screenshot 2026-06-17 at 2 29 31 PM

Screenshot 2026-06-17 at 2 29 36 PM

PR #3545, opened 2026-06-11
a11y(2.5.8): checkbox — add min-h-6 min-w-6 to Label for 24×24 hit area

rosanusi:
## Description

When `labelHidden={true}` the `<Checkbox>` Label's visible width collapses to ~16 px (the surrogate `h-4 w-4` box), giving a clickable region of ~16×18 CSS px — below the WCAG 2.5.8 (Target Size Minimum, Level AA) 24×24 requirement.

**Fix:** add `min-h-6` and `min-w-6` (`min-height: 24px; min-width: 24px`) to the Label's class array in `checkbox.svelte`. This ensures the `<label>` element — which is the actual click target wrapping the `sr-only` native input — is at least 24×24 for all consumers. With a visible label the Label continues to grow with its text content as before; only the floor changes.

The alternative of enlarging the visual surrogate from `h-4 w-4` to `h-6 w-6` was not chosen — it would change the checkbox's appearance across every consumer and require updating the inner icon positioning.

**Changed file:** `src/lib/holocene/checkbox.svelte` — 2 lines added to the Label class array.

This fix cascades to all Checkbox consumers in both `ui-main` and `cloud-ui-main` (via the `@temporalio/ui` tarball).

## Screenshots

No visual change for `labelHidden=false` (label text keeps the Label wider than 24 px). For `labelHidden=true` the invisible click region grows from ~16×18 to 24×24 around the same 16×16 visual surrogate.

## Design

No design changes. The 24-px floor is transparent to sighted users; the 16×16 checkbox appearance is unchanged.

## Testing

- [ ] Find a `<Checkbox labelHidden>` consumer (e.g. row-select checkboxes on the workflows list).
- [ ] Inspect the `<label>` element: `getBoundingClientRect()` returns `height ≥ 24` and `width ≥ 24`.
- [ ] Tap-test on a touch device: confirm reliable activation.
- [ ] Tab to a `labelHidden` checkbox: focus-visible ring appears on the visual surrogate.
- [ ] `labelHidden=false` consumer (e.g. namespace settings auth form): checkbox rows visually unchanged.
- [ ] axe-core: no target-size violations on `labelHidden` checkboxes.
- [ ] Storybook checkbox stories: spot-check both `labelHidden` and default variants.

## Checklist

- [x] No new i18n keys
- [x] No visual change to checkbox appearance
- [x] Cascades to cloud-ui-main via `@temporalio/ui` tarball

## Docs

A11y-Audit-Ref: 2.5.8-labelhidden-checkbox

---

## Correction (post-review)

Measured the actual `<label>` (the click target) before/after; the "no visual change" statements above need amending:

- **Every checkbox row grows ~6px taller — not only `labelHidden`.** `min-h-6` raises the height floor to 24px, but the `<label>` is naturally only ~18px tall (`leading-[18px]` + the 16px surrogate + its 1px border = **18×18**). So **visible-label** checkboxes go 153×18 → **153×24** as well; the `labelHidden=false` row is *not* unchanged.
- The visual surrogate is **18×18** (not 16×16) — `h-4 w-4` (16px) plus `border` with `box-content`. The glyph itself is unchanged.
- **Centering added:** with `min-w-6`, a `labelHidden` label is 24px wide and the 18px surrogate was flush-left (6px of dead space). Added `justify-center` **only when `labelHidden`**, so the checkbox is centered in the 24×24 target (3px all sides). Visible-label checkboxes keep the surrogate left, next to the text.

**Decision (with reviewer):** keep the 24px floor on **all** checkboxes — it guarantees a 24px-tall target everywhere for 2.5.8; the ~6px height growth on labelled rows is accepted.

Also merged latest `main` (resolved a `scripts/a11y/manifest.yml` append conflict with #3544).


ardiewen:
@rosanusi you may want to examine this one too:

**Example 1:**
Screenshot 2026-06-17 at 2 20 51 PM

**Example 2:**
Screenshot 2026-06-17 at 2 20 37 PM

It appears the checkbox is left aligned inside of its tap target ("label").

ardiewen (approved):

PR #3552, opened 2026-06-11
Schedules redesign 5/5: view page, list, and action modals

tegan-temporal:
## Description & motivation 💭

Final PR of the stack. After this merges, `feature/schedules` is byte-identical to `tegan/schedules-update` (verified with an empty `git diff`).

- New `schedule-view/` card-based page: spec card with full-spec toggle, advanced settings, recent/upcoming workflow runs, workflow input, custom search attributes
- `schedule-action-modals/`: trigger, backfill, and pause modals rendered through a conditionally-mounted host keyed on the confirmation-modal store, so every open starts with fresh state (no stale selections, reasons, or page-load timestamps). Trigger pre-selects Allow All (matching server behavior for manual triggers) and both overlap pickers mark the schedule's current policy
- Backfill modal interprets entered times in the schedule's timezone (caption shows it) and disables confirm when end ≤ start
- List components move to `schedules-list/`
- Final cleanup: deletes the superseded view components, `schedule-spec-label`, legacy schedule types (`ScheduleParameters` family, `Unspecified` overlap policy), and legacy i18n keys

### How was this tested 👻

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

`pnpm check` 0 errors, `pnpm lint` 0 errors, 2,080 unit tests passing, full integration suite 233 passed / 0 failed. New `schedule-actions.spec.ts` covers trigger patch bodies, backfill request + post-success modal usability, modal state reset on reopen, timezone caption, end-before-start validation, and delete. The e2e spec now does a full create → verify summary → delete loop but needs a run against a real Temporal server (not runnable locally).

On a schedule view page: trigger, backfill (try an end time before the start time, and reopen modals to confirm state resets), pause/unpause with a reason, and delete.

### Merge Checklist

- [ ] #3551 merges first
- [x] Run e2e suite against a real server before merging `feature/schedules` to `main`
- [ ] Merges into `feature/schedules` (not `main`)


src/lib/components/schedule/schedule-view/workflow-runs-recent.svelte
```diff
@@ -0,0 +1,86 @@
...
+    openBackfillConfirmationModal,
+    openTriggerConfirmationModal,
+  }: Props = $props();
+
+  const sortedRecentRuns = $derived.by(() => {
+    const runs = schedule?.info?.recentActions ?? [];
+    return runs
+      .filter(Boolean)
+      .sort(
+        (a, b) => getMilliseconds(a.actualTime) - getMilliseconds(b.actualTime),
```
laurakwhit: Do we want the newest (not the oldest) to show here?

```suggestion
        (a, b) => getMilliseconds(b.actualTime) - getMilliseconds(a.actualTime),
```

src/lib/components/schedule/schedule-view/advanced-settings-card.svelte
```diff
@@ -0,0 +1,141 @@
...
+              <li>{translate('common.none')}</li>
+            {/each}
+          </ul>
+        {:else}
+          {translate('common.none')}
+        {/if}
+      </dd>
+    </div>
+
+    {#if state.limitedActions}
```
laurakwhit:
```suggestion
    {#if state?.limitedActions}
```

src/lib/components/schedule/schedule-view/schedule-view.svelte
```diff
@@ -0,0 +1,233 @@
...
+    {translate('schedules.back-to-schedules')}
+  </Link>
+  <div class="flex items-start justify-between">
+    <div class="flex flex-col gap-2">
+      <h1
+        class="flex select-all flex-wrap items-center gap-2 text-3xl"
+        data-testid="schedule-name"
+      >
+        <WorkflowStatus
+          status={schedule?.schedule.state.paused ? 'Paused' : 'Running'}
```
laurakwhit:
```suggestion
          status={schedule?.schedule?.state?.paused ? 'Paused' : 'Running'}
```

src/lib/components/schedule/schedule-view/schedule-view.svelte
```diff
@@ -0,0 +1,233 @@
...
+        </div>
+
+        <div class="flex items-center gap-2">
+          <dt class="text-sm font-medium text-secondary">Created</dt>
+          <dd>
+            {$timestamp(schedule?.info?.createTime)}
+          </dd>
+        </div>
+
+        {#if schedule?.info.updateTime}
```
laurakwhit:
```suggestion
        {#if schedule?.info?.updateTime}
```

src/lib/components/schedule/schedule-view/workflow-input-card.svelte
```diff
@@ -4,18 +4,23 @@
   import { translate } from '$lib/i18n/translate';
   import type { Payloads } from '$lib/types';
 
-  let { input, scheduleId }: { input: Payloads; scheduleId: string } = $props();
+  interface Props {
+    input: Payloads;
```
laurakwhit: NIT: Can the input be `null` here?

src/lib/components/schedule/schedule-view/workflow-runs-upcoming.svelte
```diff
@@ -0,0 +1,52 @@
...
+    title={translate('schedules.workflow-runs-empty-state-upcoming-title')}
+    description={translate(
+      'schedules.workflow-runs-empty-state-upcoming-description',
+    )}
+    {openBackfillConfirmationModal}
+    {openTriggerConfirmationModal}
+  />
+{:else}
+  <ul class={twMerge('flex flex-col gap-2', className)}>
+    {#each sortedUpcomingRuns as run (run)}
```
laurakwhit: Can we key this #each block either by the value or the index?

src/lib/components/schedule/schedule-view/workflow-runs-upcoming.svelte
```diff
@@ -0,0 +1,52 @@
...
+  import { translate } from '$lib/i18n/translate';
+  import type { DescribeFullSchedule } from '$lib/types/schedule';
+  import { getMilliseconds } from '$lib/utilities/format-time';
+
+  import WorkflowRunsEmpty from './workflow-runs-empty.svelte';
+
+  interface Props {
+    class?: string;
+    schedule: DescribeFullSchedule;
+    workflowQuery: string;
```
laurakwhit:
```suggestion
```

src/lib/components/schedule/schedule-view/workflow-runs-card.svelte
```diff
@@ -0,0 +1,80 @@
...
+    <WorkflowRunsRecent
+      {schedule}
+      {namespace}
+      {openBackfillConfirmationModal}
+      {openTriggerConfirmationModal}
+    />
+  {:else}
+    <WorkflowRunsUpcoming
+      {schedule}
+      {workflowQuery}
```
laurakwhit: Looks like this isn't being used here.

```suggestion
```

src/lib/components/schedule/schedule-view/schedule-spec-card.svelte
```diff
@@ -0,0 +1,58 @@
...
+  const specs = $derived(getFormSpecFromSpec(schedule?.schedule?.spec));
+</script>
+
+<Panel class="flex w-full flex-col gap-4 border-subtle p-6" as="section">
+  <header class="mb-1 flex items-center justify-between">
+    <h2 class="text-2xl font-medium">
+      {translate('schedules.schedule-specs')}
+    </h2>
+    <Button
+      variant="ghost"
```
laurakwhit: NIT: Wondering if this should be a "secondary" button or maybe just have an up/down chevron icon? Or Maybe a toggle would be more intuitive here?

src/lib/components/schedule/schedule-view/schedule-view-loading.svelte
```diff
@@ -0,0 +1,25 @@
...
+  <div class="relative flex flex-col gap-4">
+    <Link href={routeForSchedules({ namespace })} icon="chevron-left">
+      {translate('schedules.back-to-schedules')}
+    </Link>
+    <h1 class="select-all" data-testid="schedule-name">
+      {scheduleId}
+    </h1>
+  </div>
+</header>
+<Loading />
```
laurakwhit: NIT: Instead of a loading spinner, wondering if we should follow the pattern of a skeleton loader we have elsewhere when loading a full page.

src/lib/components/schedule/schedule-view/schedule-view.svelte
```diff
@@ -0,0 +1,233 @@
...
+        {scheduleId}
+      </h1>
+
+      <dl class="contents">
+        <div class="flex items-center gap-2">
+          <dt class="text-sm font-medium text-secondary">Workflow Type</dt>
+          <dd class="inline-flex items-center gap-1">
+            {schedule?.schedule?.action?.startWorkflow?.workflowType?.name}
+            <div class="flex items-center gap-2">
+              <Link
```
laurakwhit: Could we use the `DetailList` components here? E.g. `DetailListLinkValue`.

src/lib/components/schedule/schedule-view/workflow-runs-card.svelte
```diff
@@ -0,0 +1,80 @@
...
+        namespace,
+        query: workflowQuery,
+      })}
+    >
+      {translate('common.view-all-runs')}
+    </Link>
+  </header>
+
+  <PillContainer class="mr-auto flex flex-row p-1">
+    <Pill
```
laurakwhit: Curious why pills instead of our [ToggleButtons](https://holocene.preview.thundergun.io/?path=/story/toggle-button--default) or [Tabs](https://holocene.preview.thundergun.io/?path=/story/tabs--default)?

src/lib/components/schedule/schedule-view/schedule-view.svelte
```diff
@@ -0,0 +1,233 @@
...
+    </SplitButton>
+  </div>
+
+  <dl class="flex flex-col gap-2">
+    <dt class="text-sm font-medium text-secondary">Total Workflows</dt>
+    <dd class="flex flex-wrap items-center gap-2">
+      <span class="font-mono" data-testid="workflow-count"
+        >{$workflowCount.count.toLocaleString()}
+      </span>
+      <Button
```
laurakwhit: Can we use the existing `CountRefreshButton` here?

laurakwhit (commented):
Looking good! Noticed a few small things, but no major blockers.

src/lib/components/schedule/schedule-view/schedule-spec-card.svelte
tegan-temporal: FWIW the design calls for this https://www.figma.com/design/tzggy7x7JIM0payi0D4Xr4/Cloud-UI-Redesign?node-id=3265-27833&m=dev

src/lib/components/schedule/schedule-view/schedule-view.svelte
tegan-temporal: TBH `CountRefreshButton` is more-roundabout. It would require an effect to trigger `refreshCurrentScheduleFetch({ namespace, scheduleId })` instead of the IMO much simpler version of just calling it in the handler when you click it.

src/lib/components/schedule/schedule-view/workflow-runs-card.svelte
tegan-temporal: Because the design called for pills 😅 https://www.figma.com/design/tzggy7x7JIM0payi0D4Xr4/Cloud-UI-Redesign?node-id=3265-28495&m=dev ... at least visually. The a11y semantics should probably be improved though

src/lib/components/schedule/schedule-view/workflow-runs-card.svelte
tegan-temporal: oh wow 🕵️‍♀️

src/lib/components/schedule/schedule-view/workflow-runs-upcoming.svelte
tegan-temporal: run is an ITimestamp which should be a string, I'll update to key by the index though, I guess theoretically sortedUpcomingRuns might have multiples of the same timestamp

src/lib/components/schedule/schedule-view/schedule-view.svelte
tegan-temporal: How strongly do you feel about that? DetailList seems like it adds a bunch of stuff without really adding any value for this case. I don't think we need all dl to be coupled.

src/lib/components/schedule/schedule-view/schedule-spec-card.svelte
laurakwhit: cc @acav-temporalio

src/lib/components/schedule/schedule-view/schedule-view.svelte
laurakwhit: Not strongly at all. Just figured it might be nice to have all the styling handled. And if we ever change the styling for `DetailList` in other places it would auto-update here.

src/lib/components/schedule/schedule-view/schedule-view.svelte
laurakwhit: That's fair. Could we make a version of the `CountRefreshButton` based on a boolean prop that isn't polling and is just clickable? Only thinking we might want to use it so we keep the UX/UI the same across the multiple places we have it.

src/lib/components/schedule/schedule-view/workflow-runs-card.svelte
laurakwhit: cc @acav-temporalio wondering if we can use existing Holocene components here.

src/lib/components/schedule/schedule-view/schedule-view.svelte
tegan-temporal: k, I just needed to add an optional onRefresh handler to the CountRefreshButton it works for me now.

src/lib/components/schedule/schedule-view/schedule-view.svelte
tegan-temporal: my dl seems to be different from what the DetailList renders

Screenshot 2026-06-18 at 8 17 07 AM

vs 

Screenshot 2026-06-18 at 8 20 02 AM

I think it would be an overall bigger change to get DetailList it seems like the layouts most similar to mine are (like the cards in the workers detail page) also inline the dl / dt / dd elements:

image

I would prefer not to use it here and maybe create a new task to create a more flexible DetailList component if want to tie everything together.

src/lib/components/schedule/schedule-view/workflow-runs-card.svelte
tegan-temporal: I created https://github.com/temporalio/ui/pull/3574 to make it so we can implement the pill-style tabs with the correct html semantics / aria-roles

src/lib/components/schedule/schedule-view/schedule-view-loading.svelte
tegan-temporal: Good idea! I've updated the loading view to be a skeleton

CleanShot 2026-06-18 at 08 40 58@2x
CleanShot 2026-06-18 at 08 41 07@2x

src/lib/components/schedule/schedule-view/schedule-spec-card.svelte
tegan-temporal: Updated to use secondary button -- also I realized it was missing disclosure semantics so I've added those as well.

https://github.com/user-attachments/assets/0f88be00-1a84-45b4-ad96-791be0c59508

src/lib/components/schedule/schedule-view/workflow-runs-card.svelte
tegan-temporal: Updated to use the tabs primitive + added back the pill container bg. The tabs primitive needs to merge before this one now.

https://github.com/user-attachments/assets/3350bad8-5e93-4607-93a0-d042a8ac1fd3

src/lib/components/schedule/schedule-view/schedule-view.svelte
laurakwhit: 💯% valid, we should leave the "Advanced Settings" for now. I was looking at these ("Workflow Type", "Created", etc.) ⬇️ and thinking they could look similar to the other `DetailList` components. 

Screenshot 2026-06-18 at 12 48 17 PM

src/lib/components/schedule/schedule-view/workflow-runs-card.svelte
laurakwhit: Amazing 😍 I think we're only using the `PillContainer` in one other spot and it could definitely benefit from your `Tabs` primitive 🙏

src/lib/components/schedule/schedule-view/schedule-view.svelte
tegan-temporal: omgggggg sorry I misunderstood the whole thing. I'll update that part

src/lib/components/schedule/schedule-view/schedule-view.svelte
```diff
@@ -0,0 +1,217 @@
...
+          {translate('common.created-label')}
+        </DetailListLabel>
+        <DetailListTimestampValue
+          class="font-sans text-sm"
+          timestamp={schedule?.info?.createTime}
+        />
+
+        {#if schedule?.info?.updateTime}
+          <DetailListLabel class="text-sm font-medium">
+            {translate('common.last-updated')}
```
laurakwhit: This has `Last Updated: `. Might want to just make it `Last Updated` here.

src/lib/components/schedule/schedule-view/schedule-view-loading.svelte
```diff
@@ -0,0 +1,59 @@
...
+        <Skeleton class="h-6 w-48" />
+        <Skeleton class="h-6 w-40" />
+      </div>
+    </div>
+
+    <Skeleton class="h-8 w-28 rounded-sm" />
+  </div>
+
+  <dl class="flex flex-col gap-2">
+    <dt class="text-sm font-medium text-secondary">Total Workflows</dt>
```
laurakwhit: NIT: Use translation here and in `src/lib/components/schedule/schedule-view/schedule-view.svelte`.

src/lib/components/schedule/schedule-view/schedule-view.svelte
```diff
@@ -0,0 +1,217 @@
...
+                `WorkflowType="${
+                  schedule?.schedule?.action?.startWorkflow?.workflowType?.name
+                }"`,
+                `TemporalScheduledById="${scheduleId}"`,
+              ].join(' AND '),
+            })}
+          >
+            <Icon
+              name="filter"
+              title="Filter scheduled workflows by this type"
```
laurakwhit: NIT: Add translation.

src/lib/components/schedule/schedule-action-modals/schedule-action-modals.svelte
```diff
@@ -0,0 +1,43 @@
...
+
+  interface Props {
+    schedule: DescribeFullSchedule;
+    namespace: string;
+    scheduleId: string;
+  }
+
+  let { schedule, namespace, scheduleId }: Props = $props();
+
+  // We are conditionally remndering each modal in order to reset modal state on open/close
```
laurakwhit:
```suggestion
  // We are conditionally rendering each modal in order to reset modal state on open/close
```

src/lib/components/schedule/schedule-view/workflow-runs-recent.svelte
```diff
@@ -0,0 +1,86 @@
...
+          class="col-span-2 row-start-2 flex justify-center sm:col-span-1 sm:col-start-2 sm:row-start-1 sm:justify-start"
+        >
+          <Link
+            href={routeForWorkflow({
+              workflow: run.startWorkflowResult.workflowId,
+              run: run.startWorkflowResult.runId,
+              namespace,
+            })}
+          >
+            {run.startWorkflowResult.workflowId}
```
laurakwhit:
```suggestion
              workflow: run.startWorkflowResult?.workflowId,
              run: run.startWorkflowResult?.runId,
              namespace,
            })}
          >
            {run.startWorkflowResult?.workflowId}
```

src/lib/components/schedule/schedule-view/schedule-view.svelte
```diff
@@ -0,0 +1,215 @@
...
+      >
+        <Icon
+          name="filter"
+          title="Filter scheduled workflows by this type"
+          class="h-4 w-4"
+        />
+      </Link>
+    </DetailListValue>
+
+    <DetailListLabel class="items-center text-sm font-medium">
```
laurakwhit: NIT: Do we need the `text-sm` or `font-medium` on the dt or dd elements here? I think it would be fine to leave it as the default styling and I think it should already be `text-sm` by default anyways.

laurakwhit (approved):
Few super small comments, but otherwise :shipit:

PR #3554, opened 2026-06-12
a11y(3.3.1): Holocene input primitives — add aria-invalid, aria-describedby, and uniform live-region error announcement

rosanusi:
## Summary

Adds uniform error-association to all seven Holocene input primitives (`Input`, `Textarea`, `NumberInput`, `ChipInput`, `Combobox`, `Select`, `Checkbox`).

Each primitive now:
- Sets `aria-invalid="true"` on the underlying control when an error is active
- Generates a unique `errorId` (`${id}-error`) per instance and sets `aria-describedby` pointing at the error element (only when an error is present — no dangling references)
- Renders the error text element with `id={errorId}` and `role="alert"` for consistent live-region announcement

This ensures screen-reader users hear the error text both on first render (live region) and when focus returns to an invalid field (via `aria-describedby` association).

**Before:** AT reads only the field label when focus returns to an invalid field; user cannot detect the error state programmatically.  
**After:** AT announces field label + "invalid entry, [error text]" via `aria-invalid` + `aria-describedby`.

## Files changed

- `src/lib/holocene/input/input.svelte`
- `src/lib/holocene/textarea.svelte`
- `src/lib/holocene/input/number-input.svelte`
- `src/lib/holocene/input/chip-input.svelte`
- `src/lib/holocene/combobox/combobox.svelte`
- `src/lib/holocene/select/select.svelte`
- `src/lib/holocene/checkbox.svelte`

## Test plan

- [ ] Inspect DOM for each primitive in an erroring state: confirm `aria-invalid="true"` on the control, `id="{id}-error"` on the error element, `aria-describedby="{id}-error"` on the control, and `role="alert"` on the error element
- [ ] Verify no `aria-describedby` is set when no error is present (no dangling references)
- [ ] With VoiceOver/NVDA: confirm error is announced on first render and re-announced when focus returns to the invalid field
- [ ] Regression: error text continues to render visually in the same position; non-error `hintText` still displays correctly when `error=false`/`valid=true`
- [ ] axe-core: zero new violations on a representative form page

## References

- WCAG 2.2 SC 3.3.1 Error Identification (Level A)
- Cross-walks: SC 1.3.1 Info and Relationships, SC 4.1.2 Name, Role, Value
- A11y-Audit-Ref: `3.3.1-holocene-input-primitives-error-association`

A11y-Audit-Ref: 3.3.1-holocene-input-primitives-error-association

### DIFF
--- a/src/lib/holocene/checkbox.svelte
+++ b/src/lib/holocene/checkbox.svelte
@@ -117,7 +117,7 @@
       data-track-name="checkbox"
       data-track-intent="toggle"
       data-track-text={label}
-      aria-invalid={showError ? 'true' : undefined}
+      aria-invalid={!valid ? 'true' : undefined}
       aria-describedby={showError ? errorId : undefined}
       bind:checked
       {disabled}
--- a/src/lib/holocene/checkbox.svelte
+++ b/src/lib/holocene/checkbox.svelte
@@ -180,7 +180,7 @@
       </span>
     </slot>
   </Label>
-  {#if showError}
-    <span id={errorId} role="alert" class="text-xs text-danger">{error}</span>
-  {/if}
+  <span id={errorId} role="alert" class="text-xs text-danger">
+    {#if showError}{error}{/if}
+  </span>
 </div>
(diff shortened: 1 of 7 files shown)


ardiewen (approved):

ardiewen:
Added a follow-up ticket: DT-4184 to unify the error/valid contract across the 7 primitives affected here.

PR #3556, opened 2026-06-12
a11y(4.1.2): add label prop to Button primitive and enforce accessible names

rosanusi:
## Summary

- Adds a `label?: string` prop to `BaseProps` in the Button primitive (`src/lib/holocene/button.svelte`) rendered as `aria-label` on both the `<a>` anchor branch and `<button>` branch
- Adds a dev-time `onMount` check that throws (dev) or `console.error`s (prod) when a button is rendered without any accessible name (no `label`, `aria-label`, `aria-labelledby`, `title`, slot text, leading/trailing icon, or count badge)
- Updates 7 consumer call-sites that were triggering the defect: Workflows List (per-row Start button + 3 toolbar buttons), Activities List (per-row Start button), Schedules List (toolbar configure-columns button), and Schedule Detail page (refresh button)
- Adds `configure-columns` i18n key to `src/lib/i18n/locales/en/common.ts`

## Affected files

- `src/lib/holocene/button.svelte`
- `src/lib/i18n/locales/en/common.ts`
- `src/lib/components/workflow/start-workflow-button.svelte`
- `src/lib/components/standalone-activities/start-activity-button.svelte`
- `src/lib/components/workflow/workflows-summary-configurable-table.svelte`
- `src/lib/pages/schedule-view.svelte`
- `src/lib/pages/schedules.svelte`

## Test plan

- [ ] Visual regression: all 7 consumer buttons render and function identically; no visible change
- [ ] Screen reader (NVDA/VoiceOver): each button announces its label before activation
- [ ] Dev-mode enforcement: render a `<Button>` with no name source — confirm it throws; add `label="…"` — confirm no throw
- [ ] Anchor branch: render `<Button href="…">` with no name source — confirm same enforcement fires
- [ ] Count badge: render `<Button count={3}>` — confirm enforcement does not fire (count provides a numeric name)
- [ ] `pnpm check` passes (no new type errors beyond pre-existing `$$_$$` slot interop errors on main)

A11y-Audit-Ref: 4.1.2-button-anchor-empty


ardiewen:
Pushed a revision reworking the approach based on review:

- **Dropped the `label` prop.** `aria-label` is already a typed attribute on `Button` (its props intersect `HTMLButtonAttributes`/`HTMLAnchorAttributes`), so `label` was redundant — and it reads like it should render *visible* text. Call sites now pass `aria-label` directly.
- **Dropped the `onMount` enforcement.** It throws in dev for any unlabeled Button (including Storybook/tests), and its `hasName` check treated a *decorative* `leadingIcon` as a name (`icon/svg.svelte` renders `<svg aria-hidden={!title}>`, and no `title` is passed) — so it simultaneously missed real violations (~22 `leadingIcon`-only buttons) and would have thrown on the equivalent icon-in-slot pattern.
- **Fixed 2 icon-only buttons this PR missed** (same defect, would have tripped the removed check): `activities-summary-configurable-table` configure-columns and `workflow-current-details` refresh.

On enforcement: the repo already has an axe suite (`tests/accessibility/`) that computes the real accessible name and would catch this correctly — but it isn't gated in CI and has rotted (the page fails to load before axe runs). Repairing + CI-gating it is the right systematic guard, tracked as a follow-up in [DT-4255](https://temporalio.atlassian.net/browse/DT-4255) rather than a bespoke runtime check.

[DT-4255]: https://temporalio.atlassian.net/browse/DT-4255?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

ardiewen (approved):

PR #3557, opened 2026-06-12
a11y(4.1.2): require label on CodeBlock primitive; add aria-readonly for read-only mode

rosanusi:
## Summary

- Makes `label: string` required on the `CodeBlock` primitive (`src/lib/holocene/code-block.svelte`), applied to CodeMirror's `.cm-content` via `EditorView.contentAttributes.of({ 'aria-label': label })` — every CodeMirror instance now has an accessible name
- Adds `aria-readonly="true"` to `contentAttributes` when `editable=false` so AT (especially JAWS forms-mode) correctly conveys non-editability instead of announcing the element as an editable text field
- Adds a dev-time `onMount` guard that throws (dev) / `console.error`s (prod) when `label` is empty at mount, catching dynamic-spread cases that TypeScript can't reach
- Fixes the orphaned `<Label for={id}>` association in `PayloadInput` (§NEW1): replaces `<Label for>` with `<span>` and forwards `label` to the inner `<CodeBlock>` so the accessible name lands on the labelable `.cm-content` element
- Migrates all ~28 affected files: primitive + 4 domain wrappers + ~22 direct consumers

## Affected files

**Primitive + wrappers (5 files):**
- `src/lib/holocene/code-block.svelte`
- `src/lib/components/payload/payload-code-block.svelte`
- `src/lib/components/payload-input.svelte`
- `src/lib/components/payload-input-with-encoding.svelte`
- `src/lib/components/workflow/input-and-results-payload.svelte`

**Locale additions (5 files):**
- `src/lib/i18n/locales/en/common.ts` — `json`
- `src/lib/i18n/locales/en/events.ts` — `import-format-1`, `import-format-2`
- `src/lib/i18n/locales/en/standalone-activities.ts` — `activity-input`, `activity-outcome`, `cluster-config`, `namespace-config`
- `src/lib/i18n/locales/en/workers.ts` — `heartbeat-config`, `setup-guide-command`
- `src/lib/i18n/locales/en/workflows.ts` — `query-result`, `callback-metadata`, `update-input`, `update-result`

**Direct consumers (~23 files):** event-card, workflow-error-stack-trace, schedules-calendar-view, schedule-frequency-panel, activity-input-and-outcome, standalone-activities-disabled, serverless-worker-setup-guide, worker-heartbeats-disabled, update-confirmation-modal, pending-activities, pending-activity-card, pending-nexus-operation-card, workflow-callback, workflow-json-navigator, import-events, standalone-activity-details, standalone-activity-search-attributes, workflow-call-stack, workflow-memo, workflow-metadata, workflow-query, workflow-search-attributes, archival/+page

## Test plan

- [ ] Screen reader: navigate to Workflow History Input/Result panels — `.cm-content` announces correct label; JAWS does not auto-switch to forms mode on read-only blocks
- [ ] Screen reader: navigate to Workflow Queries query-arg input (editable) — announces label and allows typing; `aria-readonly` absent
- [ ] Dev-mode enforcement: render `<CodeBlock content="…" label="" />` — confirm it throws; add non-empty `label` — confirm no throw
- [ ] Visual regression: all updated surfaces render identically to before; no visible change
- [ ] `pnpm check` passes (no new type errors beyond pre-existing `$$_$$` slot interop errors on main)
- [ ] `PayloadInput` standalone (e.g. update-confirmation-modal): visible `<span>` label text renders correctly; CodeBlock has `aria-label`

A11y-Audit-Ref: 4.1.2-codemirror-no-name


ardiewen:
Rebased onto main (was 91 behind; conflicts resolved — note two schedule files were deleted upstream, so their label edits dropped).

Changes from review:
- **Removed the `onMount` throw; kept `label` optional** — per #3556, enforcement belongs in the axe suite (DT-4255). A required prop / throw would break cloud-ui's ~74 label-less `CodeBlock` sites on repack.
- **Made read-only blocks focusable (`tabindex="0"`)** — the name/`aria-readonly` were in the AX tree, but `.cm-content` (contenteditable=false) wasn't focusable, so SRs skipped it (only the copy button was reachable). Each read-only block is now a tab stop announcing "…, read only". Note an accessible-name axe check passes without this — it doesn't test reachability.

`pnpm check` green; verified in-browser.

ardiewen (approved):

PR #3560, opened 2026-06-12
a11y(4.1.3): Toaster role="log" → per-toast role="status"/"alert" for correct AT announcement semantics

rosanusi:
## Summary

Fixes WCAG 2.2 SC **4.1.3 (Status Messages)** for the toast notification surface by introducing a small, **placeable live-region primitive** and decoupling the screen-reader announcement from the visual toast.

The earlier approach in this PR put `role="status"`/`role="alert"` + `aria-atomic` on **each per-toast `<div>`**. That inverts the one rule a live region must obey: **the region has to exist in the DOM, empty, _before_ its content is inserted.** A node that is simultaneously the region *and* its content (inserted in a single DOM mutation) does not reliably announce — and the polite case (`status`: success/info/warning/primary, the majority of toasts) is exactly the unreliable one. So the original change likely **regressed** the SC it set out to fix. This rework replaces it.

## Approach

A persistent sr-only live region, fed by the toaster store — one announcement per toast, independent of the animated visual node:

| File | Change |
|---|---|
| `src/lib/stores/announcer.ts` (new) | `createAnnouncer()` — a buffer of `{ id, message, politeness }`. `announce()` appends a uniquely-keyed node (append, **not** clear-and-reset); each is removed after `max(toast duration, 7s)`. `clear()` empties the buffer and cancels pending timers. |
| `src/lib/holocene/live-region.svelte` (new) | Two **always-rendered** sr-only regions — `aria-live="polite"` and `aria-live="assertive"`, both `aria-relevant="additions"`; messages routed by politeness, keyed by `id`. Optional `data-testid` applied to both regions (`-polite`/`-assertive`). |
| `src/lib/stores/toaster.ts` | `push()` calls `announce(message, variant === 'error' ? 'assertive' : 'polite', duration)`; exposes `announcements`; `clear()` also clears the announcer. |
| `src/lib/holocene/toaster.svelte` | Mounts `<LiveRegion>` (optional self-subscribing `announcements` prop, defaults to the singleton store); container keeps **no** `role`. |
| `src/lib/holocene/toast.svelte` | Reverted to purely visual — per-toast `role`/`aria-atomic` removed. |

### Why this shape
- **Region pre-exists empty** → reliable announcement (MDN Live Regions; WCAG ARIA22).
- `aria-live` on the wrapper (not `role`) avoids iOS-VoiceOver double-speak; `aria-relevant="additions"` keeps the timed removals silent.
- Unique-keyed append → **identical consecutive messages re-announce** (fixes the stale-announcement gap).
- A **placeable component** (not a body-level singleton) is deliberate: a body-level region is inerted by `aria-modal` under VoiceOver+Safari (W3C ARIA #1854). Modal-scoped messages (future, e.g. #3531) must mount their own `<LiveRegion>` inside the dialog; DOM-resident status (e.g. #3561) announces in place. **This PR scopes to toasts only.**

## Test plan

- [ ] **Screen-reader smoke test (VoiceOver + NVDA)** — success/info/warning/primary toast announces **politely** (no interrupt); error toast announces **assertively** (interrupts). *(Note: under a rapid burst, screen readers coalesce same-region polite updates to latest-wins; this matches react-aria/Radix and is expected.)*
- [ ] DOM: two `.sr-only[aria-live="polite"]` / `[aria-live="assertive"]` regions exist **before** any toast fires; the visible container `<div>` and each toast `<div>` carry **no** `role`.
- [ ] Workflow client-action confirmations (Terminate / Cancel / Reset / Signal) fire a toast that renders and announces.
- [ ] axe-core: no new violations on a page with a live toast.

**Automated:** unit tests for `createAnnouncer` (politeness mapping, lifetime/timeout, distinct ids, clear) and toaster store wiring; an integration assertion that the live regions pre-exist and a toast message routes to the polite region.

## Downstream (cloud-ui)
`announcements` is an **optional** prop defaulting to the singleton, so cloud-ui's two `<Toaster>` mounts compile and work unchanged after the `@temporalio/ui` pack bump — **no cloud-ui edits required.**

A11y-Audit-Ref: 4.1.3-toaster-role-status-alert

### DIFF
--- a/src/lib/holocene/live-region.svelte
+++ b/src/lib/holocene/live-region.svelte
@@ -1,12 +1,20 @@
 <script lang="ts">
-  import type { Announcement } from '$lib/stores/announcer';
+  import type { Announcement, Politeness } from '$lib/stores/announcer';
 
   interface Props {
     messages: Announcement[];
+    politeness?: Politeness | 'both';
     'data-testid'?: string;
   }
 
-  let { messages, 'data-testid': testid }: Props = $props();
+  let {
+    messages,
+    politeness = 'both',
+    'data-testid': testid,
+  }: Props = $props();
+
+  const showPolite = $derived(politeness !== 'assertive');
+  const showAssertive = $derived(politeness !== 'polite');
 
   const polite = $derived(
     messages.filter((message) => message.politeness === 'polite'),
@@ -16,23 +24,27 @@
   );
 </script>
 
-<div
-  class="sr-only"
-  aria-live="polite"
-  aria-relevant="additions"
-  data-testid={testid ? `${testid}-polite` : undefined}
->
-  {#each polite as { id, message } (id)}
-    <div>{message}</div>
-  {/each}
-</div>
-<div
-  class="sr-only"
-  aria-live="assertive"
-  aria-relevant="additions"
-  data-testid={testid ? `${testid}-assertive` : undefined}
->
-  {#each assertive as { id, message } (id)}
-    <div>{message}</div>
-  {/each}
-</div>
+{#if showPolite}
+  <div
+    class="sr-only"
+    aria-live="polite"
+    aria-relevant="additions"
+    data-testid={testid ? `${testid}-polite` : undefined}
+  >
+    {#each polite as { id, message } (id)}
+      <div>{message}</div>
+    {/each}
+  </div>
+{/if}
+{#if showAssertive}
+  <div
+    class="sr-only"
+    aria-live="assertive"
+    aria-relevant="additions"
+    data-testid={testid ? `${testid}-assertive` : undefined}
+  >
+    {#each assertive as { id, message } (id)}
+      <div>{message}</div>
+    {/each}
+  </div>
+{/if}
--- a/src/lib/holocene/live-region.svelte
+++ b/src/lib/holocene/live-region.svelte
@@ -1,20 +1,12 @@
 <script lang="ts">
-  import type { Announcement, Politeness } from '$lib/stores/announcer';
+  import type { Announcement } from '$lib/stores/announcer';
 
   interface Props {
     messages: Announcement[];
-    politeness?: Politeness | 'both';
     'data-testid'?: string;
   }
 
-  let {
-    messages,
-    politeness = 'both',
-    'data-testid': testid,
-  }: Props = $props();
-
-  const showPolite = $derived(politeness !== 'assertive');
-  const showAssertive = $derived(politeness !== 'polite');
+  let { messages, 'data-testid': testid }: Props = $props();
 
   const polite = $derived(
     messages.filter((message) => message.politeness === 'polite'),
@@ -24,27 +16,23 @@
   );
 </script>
 
-{#if showPolite}
-  <div
-    class="sr-only"
-    aria-live="polite"
-    aria-relevant="additions"
-    data-testid={testid ? `${testid}-polite` : undefined}
-  >
-    {#each polite as { id, message } (id)}
-      <div>{message}</div>
-    {/each}
-  </div>
-{/if}
-{#if showAssertive}
-  <div
-    class="sr-only"
-    aria-live="assertive"
-    aria-relevant="additions"
-    data-testid={testid ? `${testid}-assertive` : undefined}
-  >
-    {#each assertive as { id, message } (id)}
-      <div>{message}</div>
-    {/each}
-  </div>
-{/if}
+<div
+  class="sr-only"
+  aria-live="polite"
+  aria-relevant="additions"
+  data-testid={testid ? `${testid}-polite` : undefined}
+>
+  {#each polite as { id, message } (id)}
+    <div>{message}</div>
+  {/each}
+</div>
+<div
+  class="sr-only"
+  aria-live="assertive"
+  aria-relevant="additions"
+  data-testid={testid ? `${testid}-assertive` : undefined}
+>
+  {#each assertive as { id, message } (id)}
+    <div>{message}</div>
+  {/each}
+</div>
--- a/src/lib/holocene/live-region.svelte
+++ b/src/lib/holocene/live-region.svelte
@@ -6,7 +6,7 @@
     'data-testid'?: string;
   }
 
-  let { messages, ...rest }: Props = $props();
+  let { messages, 'data-testid': testid }: Props = $props();
 
   const polite = $derived(
     messages.filter((message) => message.politeness === 'polite'),
@@ -16,12 +16,22 @@
   );
 </script>
 
-<div class="sr-only" aria-live="polite" aria-relevant="additions" {...rest}>
+<div
+  class="sr-only"
+  aria-live="polite"
+  aria-relevant="additions"
+  data-testid={testid ? `${testid}-polite` : undefined}
+>
   {#each polite as { id, message } (id)}
     <div>{message}</div>
   {/each}
 </div>
-<div class="sr-only" aria-live="assertive" aria-relevant="additions">
+<div
+  class="sr-only"
+  aria-live="assertive"
+  aria-relevant="additions"
+  data-testid={testid ? `${testid}-assertive` : undefined}
+>
   {#each assertive as { id, message } (id)}
     <div>{message}</div>
   {/each}
--- a/src/lib/holocene/live-region.svelte
+++ b/src/lib/holocene/live-region.svelte
@@ -1,0 +1,28 @@
+<script lang="ts">
+  import type { Announcement } from '$lib/stores/announcer';
+
+  interface Props {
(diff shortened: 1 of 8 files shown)


ardiewen:
## Reworked: per-toast roles → placeable `LiveRegion` + `createAnnouncer()`

Pushed a rework of this PR's approach. The original commit put `role="status"`/`role="alert"` + `aria-atomic` on **each per-toast `<div>`**. That inverts the one structurally-correct property of a live region: **the region must already exist in the DOM, empty, before its content is inserted.** A node that is *both* the region *and* its content (inserted in one mutation) does not reliably announce — and the polite case (`status`, i.e. success/info/warning/primary — the majority of toasts) is exactly the unreliable one. `role="alert"` is a documented special case that usually announces on insert, so error toasts mostly survived, but the common path likely **regressed** the SC this PR claims to fix.

### New approach
A small, placeable live-region primitive — the announcement is decoupled from the visual toast:

| File | Change |
|---|---|
| `src/lib/stores/announcer.ts` (new) | `createAnnouncer()` — buffer of `{id, message, politeness}`; `announce()` appends a uniquely-keyed node (not clear-and-reset); each removed after 7s. |
| `src/lib/holocene/live-region.svelte` (new) | Two **always-rendered** sr-only regions — `aria-live="polite"` + `aria-live="assertive"`, both `aria-relevant="additions"`; messages routed by politeness, keyed by id. |
| `src/lib/stores/toaster.ts` | `push()` calls `announce(message, variant === 'error' ? 'assertive' : 'polite')` — one announcement per toast, from the store (not the rendered node). Exposes `announcements`; `clear()` also clears the announcer. |
| `src/lib/holocene/toaster.svelte` | Mounts `<LiveRegion>` (optional self-subscribing `announcements` prop defaulting to the singleton store); container keeps **no** `role`. |
| `src/lib/holocene/toast.svelte` | Reverted to purely visual — per-toast `role`/`aria-atomic`/`getRole` removed. |

### Why this shape (research-backed)
- Regions **pre-exist empty** → reliable announcement (MDN Live Regions; WCAG ARIA22).
- `aria-live` on the wrapper (not `role`) avoids iOS-VoiceOver double-speak; `aria-relevant="additions"` keeps the 7s removals silent.
- Unique-keyed append → **identical consecutive messages re-announce** (fixes the stale-announcement failure mode).
- A **placeable component** (vs a body-level singleton) is deliberate: a body-level region is **inerted by `aria-modal`** under VoiceOver+Safari (W3C ARIA #1854), so it can't serve modal-scoped messages. The component can be mounted *inside* a dialog where needed.

### Notes / follow-ups
- **Manual AT is a hard gate before this leaves draft.** The repo has no component-render harness, so the load-bearing property — that the regions actually *announce* — is only confirmable on real AT. Please verify on **NVDA + VoiceOver**: polite toast does not interrupt; error toast interrupts; and the two `.sr-only[aria-live]` regions exist before any toast fires. (The test-plan boxes above are still unchecked.)
- **cloud-ui cascade:** `announcements` is an *optional* prop defaulting to the singleton, so cloud-ui's two `<Toaster>` mounts compile and work unchanged after the pack bump — **no cloud-ui edits required**.
- **Future adopters of `LiveRegion` (not in this PR):** the session-expiry modal (#3531) must mount its own `<LiveRegion>` **inside the dialog** (the `aria-modal` inert-background reason above); status chips/counts (#3561) are DOM-resident and announce in place. This PR scopes to toasts only.
- Unit tests: `createAnnouncer` (politeness mapping, 7s timeout removal, distinct ids) + toaster store wiring. Full suite green (2050 passed).

ardiewen:
## Update — reworked onto a shared LiveRegion/announcer primitive

### What changed since the original PR
The original approach (per-toast `role="status"`/`"alert"` set on mount) regressed SC 4.1.3 — a polite role announced *on insertion* is unreliable. Reworked to the always-rendered-live-region pattern.

| Before | After |
|---|---|
| Role applied per-toast at mount | Toast is purely visual (no ARIA) |
| Region created with its content | Persistent empty `aria-live` regions pre-exist; toasts announced as dynamic content |
| Toaster-only | Reusable `createAnnouncer()` store + placeable `LiveRegion` component |

### Final architecture
- `stores/announcer.ts` — `createAnnouncer()`: `announce(message, politeness?, duration?)` + `clear()`; persists each message `max(duration, 7000ms)` then removes.
- `holocene/live-region.svelte` — two persistent sr-only regions (polite + assertive), keyed nodes.
- `stores/toaster.ts` — holds a private announcer; `push()` routes `error → assertive`, else `polite`.
- `holocene/toaster.svelte` — mounts `LiveRegion`; container/toast carry no `role`.

### Decisions
- ✅ Announcer kept **private to the toaster** — it's a standalone module, so layering is already clean and a future shared announcer forecloses nothing here.
- ↩️ A `politeness` zone-selection prop was added then **reverted** — speculative; no current consumer needs it.

### Out of scope (follow-up: "ambient SR announcements")
Audited `ui` + `cloud-ui` for the same transient-no-visible-host gap. Not part of this PR:

| Class | Surfaces | Fix |
|---|---|---|
| Copy-to-clipboard (silent icon swap) | `ui` ×4, `cloud-ui` ×3 — all via shared `copyToClipboard()` | Announce inside the shared util → covers both repos |
| Inline action confirms ("Success!"/"Saved") | `cloud-ui` ×3 (sink-verify ×2, internal-roles) | Per-surface `announce()` |

Banners/modals/status-chips/counts excluded (persistent or in-place → element-level live regions).

### Remaining gate
Manual NVDA + VoiceOver verification (polite no-interrupt, error interrupts, both regions pre-exist) before leaving draft. Note: VoiceOver coalesces rapid *polite* updates (latest-wins) — expected SR behavior, not a regression.

src/lib/stores/announcer.ts
```diff
@@ -0,0 +1,56 @@
...
+  messages: Readable<Announcement[]>;
+  announce: (
+    message: string,
+    politeness?: Politeness,
+    duration?: number,
+  ) => void;
+  clear: () => void;
+}
+
+const DEFAULT_TIMEOUT = 7000;
```
tegan-temporal, line 21: Am I understanding correctly that we are defaulting to a 7 second delay before starting the announcement? Is that normal? IMO it feels like a long time, if something is 'polite' does it need a manual delay at all?

src/lib/stores/announcer.ts
```diff
@@ -0,0 +1,56 @@
...
+  const clear = () => {
+    for (const handle of pendingTimers) {
+      clearTimeout(handle);
+    }
+    pendingTimers.clear();
+    messages.set([]);
+  };
+
+  return {
+    messages: { subscribe: messages.subscribe },
```
tegan-temporal: I think it might be more idiomatic to use https://svelte.dev/docs/svelte/stores#svelte-store-readonly to convert the messages writable to readonly here.

```suggestion
    messages: readonly(messages)
```

src/lib/holocene/live-region.svelte
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  let { messages, 'data-testid': testid }: Props = $props();
+
+  const polite = $derived(
+    messages.filter((message) => message.politeness === 'polite'),
+  );
+  const assertive = $derived(
+    messages.filter((message) => message.politeness === 'assertive'),
+  );
```
tegan-temporal, line 16: Feel free to ignore this suggestion, but it seems like a good use case for the newish [Object.groupBy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/groupBy).

```suggestion
  const { polite, assertive } = $derived(
    Object.groupBy(messages, (message) => message.politeness)
  );
```

tegan-temporal (approved):
couple comments, no blockers

src/lib/stores/announcer.ts
ardiewen: It's not a delay on the announcement itself, but rather how _long_ the announcement text stays in the DOM. This gives the screen reader a chance to actually pick up the announcement, while also ensuring it gets automatically cleared out of the dom after a certain amount of time.

The default of `7000ms` is just directly ripped from other popular implementations, e.g. [react-aria](https://github.com/adobe/react-spectrum/blob/main/packages/react-aria/src/live-announcer/LiveAnnouncer.tsx)

src/lib/holocene/live-region.svelte
ardiewen: `pnpm check` fails when we try to implement this:
```
1782322397445 ERROR "src/lib/holocene/live-region.svelte" 19:12 "Property 'groupBy' does not exist on type 'ObjectConstructor'. Do you need to change your target library? Try changing the 'lib' compiler option to 'es2024' or later."
```

Looks like it's an ES2024 and beyond feature 😞

PR #3561, opened 2026-06-12
a11y(4.1.3): Add live-region wrappers for workflow status, event count, filter counts, and Banner

rosanusi:
## Summary

Closes WCAG 2.2 SC 4.1.3 (Status Messages) gaps across ui-main — dynamic surfaces that changed without announcing to screen readers. Verified manually with VoiceOver + Safari/Chrome (see AT notes below).

| # | Surface | Fix |
|---|---|---|
| 1 | WorkflowStatus chip | Opt-in `announce` prop adds `role="status" aria-atomic="true"`; enabled **only** on the detail-header chip. Per-row table cells and the status-filter dropdown stay inert — avoids mounting dozens of live regions per list page (no proliferation). |
| 2 | Event history polling | sr-only `role="status"` in `event-summary-table`; **debounced (250 ms)** so a burst rendered across multiple reactive passes announces a single accurate total ("N new events loaded") instead of per-pass partial counts. |
| 3 | Filter result counts | Count **number + noun** wrapped together in one `role="status" aria-atomic="true"` region on the workflows / activities / schedules pages, so it announces "N Workflows" rather than a context-free "N". |
| 4 | Holocene Banner | `getRole()` maps `danger` → `role="alert"` (assertive), `default` → `role="status"` (polite) + `aria-atomic`. Banner's type space is only `default`/`danger`, so this mapping is complete (not an `Alert` mismatch). |
| 5 | HeartBeat indicator | `aria-hidden` on the decorative EKG wrapper so the animated graphic isn't surfaced as an unlabeled `img` inside the status chip's `aria-atomic` live region. |

## Files changed

- `src/lib/components/workflow-status.svelte`, `src/lib/layouts/workflow-header.svelte` — Defect 1
- `src/lib/components/event/event-summary-table.svelte`, `src/lib/i18n/locales/en/workflows.ts` — Defect 2
- `src/lib/pages/workflows-with-new-search.svelte`, `src/lib/pages/standalone-activities.svelte`, `src/lib/pages/schedules.svelte` — Defect 3
- `src/lib/holocene/banner/banner.svelte` — Defect 4
- `src/lib/components/heart-beat-indicator.svelte` — Defect 5

## Manual AT verification (VoiceOver + Safari/Chrome)

- ✅ **Defect 1** status chip announces on transition; correctly silent on initial page load (initial render, not a change).
- ✅ **Defect 2** event count announces an accurate burst total — debounce fix confirmed via DOM instrumentation (one signal = 4 raw events rendered as two +2 passes → single "4 new events loaded").
- ✅ **Defect 3** count region verified correct: persistent, inside `<main>`, never `aria-hidden`/`inert`, text mutates in place.
- ➖ **Defect 4** Banner role mapping is correct, but **dynamic-insertion announcement is not AT-tested**: holocene `Banner` has no consumer in OSS `ui` (and isn't exported from `@temporalio/ui`, so it doesn't cascade to cloud-ui). cloud-ui's own banner already uses the same `warning → alert, else status` pattern. Note: the `role` sits on a `<section>` mounted via `{#if show}` (created with its content), so dynamic appearance may not announce on all SR/browser combos — acceptable for banners shown on load; flagged for awareness.
- ⚠️ **VoiceOver caveat:** VO only reliably announces a polite region when the VO cursor is within that region's landmark, and suppresses background polite regions while a menu/dropdown is open. So Defect 3 may be silent on VoiceOver while filtering via an open dropdown — **NVDA/JAWS announce regardless**. The markup is correct; this is a VO platform limitation, not a defect here.

## Test plan

- [ ] **Defect 1**: trigger a status transition on a running workflow; confirm AT announces the change without navigation
- [ ] **Defect 2**: open event history for an active workflow, let it poll; confirm AT announces "N new events loaded" per batch (not per row), with accurate counts
- [ ] **Defect 3**: filter workflow/activity/schedule list; confirm AT announces "N Workflows/Activities/Schedules" on count change (best on NVDA/JAWS; see VO caveat)
- [ ] **Defect 4**: _(not AT-tested — no OSS consumer; role mapping verified by code review; matches cloud-ui's existing banner)_
- [ ] Visual regression: all surfaces unchanged visually
- [ ] axe-core: no new violations on workflow detail, workflow list, schedules, activities

## Related

- Follow-up: **DT-4199** — applying the first list filter drops focus to `<body>` (outside `<main>`), which also blocks the VoiceOver count announcement on that step. Focus-management fix tracked separately.

A11y-Audit-Ref: 4.1.3-ui-main-live-region-gaps


ardiewen:
@rosanusi **Changes made during review + AT testing** (`9351bfe9` → `d2236ac7` → `9638a85d`, 12 files):

1. **WorkflowStatus** — `role="status"` now behind an opt-in `announce` prop, header-only (no per-row/dropdown proliferation).
2. **Event count** — extracted to `utilities/count-announcer.ts` (debounced, coalesces bursts, re-announces identical counts) + `live-count-announcer.svelte`; 7 unit tests.
3. **Filter counts** — wrap number + noun in one `role="status"` region → "N Workflows", not a bare "N".
4. **Banner** — `getRole()`: `danger → alert`, `default → status`.
5. **HeartBeat** — `aria-hidden` so the decorative EKG isn't announced inside the chip's live region.

⚠️ VoiceOver won't announce polite regions while a dropdown is open (NVDA/JAWS do) — markup verified correct. Follow-up: **DT-4199** (first filter-add drops focus to `<body>`).

ardiewen (approved):

PR #3567, opened 2026-06-16
Fix required activity timeout inputs on Start a Standalone Activity form

laurakwhit:
## Description & motivation 💭 

On the Start a Standalone Activity form, either Start to Close Timeout or Schedule to Close Timeout is required. The validation worked when both fields were left blank, but broke as soon as a value was entered and then removed:
1. The `DurationInput` converted its raw numeric field to a seconds string with `${unit.convert(Number(durationValue))}s`. When the field was cleared, Number('') evaluated to 0, so the bound value became '0s' instead of empty.
  2. The form's required-check used plain truthiness (`!data.startToCloseTimeout`). Since '0s' is a truthy string, the check passed, the form considered itself valid and it submitted.
3. The same root cause ('0s' from an emptied field) also meant a deliberately-typed 0 slipped past validation.

The `DurationInput` now yields an empty value (`value = ''`) when the raw input is empty or non-numeric, instead of coercing to '0s'. The Start a Standalone Activity form now has an `isPositiveDuration` helper check for the required prop on the Start to Close Timeout and Schedule to Close Timeout fields.

### How was this tested 👻 

- [x] Manual testing
- [x] E2E tests added
- [ ] Unit tests added

* On the Start a Standalone Activity form > without entering a Start to Close Timeout or Schedule to Close Timeout hit submit
   - [ ] Verify the error ("Either "Start to Close Timeout" or "Schedule to Close Timeout" is required.") appears
* Enter a value for either Start to Close Timeout or Schedule to Close Timeout and then clear it and hit submit
   - [ ] Verify the error appears
* Enter "0" for either Start to Close Timeout or Schedule to Close Timeout and hit submit
   - [ ] Verify the error appears

### Issue(s) closed 

`DT-4181`

### DIFF
--- a/src/lib/holocene/duration-input/duration-input.svelte
+++ b/src/lib/holocene/duration-input/duration-input.svelte
@@ -93,10 +93,14 @@
 
   const convert = (durationValue: string, durationUnit: string) => {
     const unit = units.find((u) => u.label === durationUnit);
+    if (!unit) return;
 
-    if (unit) {
-      value = `${unit.convert(Number(durationValue))}s`;
+    if (durationValue === '' || isNaN(Number(durationValue))) {
+      value = '';
+      return;
     }
+
+    value = `${unit.convert(Number(durationValue))}s`;
   };
 
   const handleNumberInput: ChangeEventHandler<HTMLInputElement> = (e) => {
--- a/src/lib/holocene/duration-input/duration-input.svelte
+++ b/src/lib/holocene/duration-input/duration-input.svelte
@@ -95,7 +95,7 @@
     const unit = units.find((u) => u.label === durationUnit);
     if (!unit) return;
 
-    if (durationValue === '' || isNaN(Number(durationValue))) {
+    if (durationValue.trim() === '' || isNaN(Number(durationValue))) {
       value = '';
       return;
     }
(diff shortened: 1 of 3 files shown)


src/lib/holocene/duration-input/duration-input.svelte
```diff
@@ -93,10 +93,14 @@
 
   const convert = (durationValue: string, durationUnit: string) => {
     const unit = units.find((u) => u.label === durationUnit);
+    if (!unit) return;
 
-    if (unit) {
-      value = `${unit.convert(Number(durationValue))}s`;
+    if (durationValue === '' || isNaN(Number(durationValue))) {
```
tegan-temporal: We might want to trim durationValue too. `Number(' ')` is also apparently `0`

tegan-temporal (approved):
I ran across a similar issue with schedules and was too afraid to rework the logic of duration-input... hopefully nothing relies on this unexpected behavior 🤞

src/lib/holocene/duration-input/duration-input.svelte
laurakwhit: Good catch, updated [here](https://github.com/temporalio/ui/pull/3567/commits/8eb9fcbb075a2ce91f053f4f26f2d6a2827b376c) ✔️

PR #3572, opened 2026-06-17
Show single attempt in Standalone Activity details

laurakwhit:

| Before | After |
|-|-|
|Screenshot 2026-06-17 at 3 58 55 PM|Screenshot 2026-06-17 at 4 02 17 PM|

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added


dandavison (approved):
Great, I think this will be helpful.

Non-blocking: I wonder whether (and maybe this applies to Workflow UI also) when we're telling the user about a completed execution it would be better to use `Attempts` plural, since then they can read it as "how many attempts were there", rather than thinking "it's on attempt N".

PR #3574, opened 2026-06-17
feat(DT-4183): Add Tabs (Primitive)

tegan-temporal:
## Description & motivation 💭 

Adds a primitive tabs component. Our existing tabs component has built-in styles and expectations. This makes it difficult to use in situations outside it's particular use-case. I have a design I need to implement where the tab buttons are more pill-like than they are our existing tab component: https://www.figma.com/design/tzggy7x7JIM0payi0D4Xr4/Cloud-UI-Redesign?node-id=3265-28495&m=dev In order to implement that design with proper aria roles and focus management I've created this headless component.

I recommend looking at the story example code to see how it's intended to be used. [`DT-4183-primitive-tabs-panels?expand=1`#diff-40a886e3e6](https://github.com/temporalio/ui/compare/DT-4183-primitive-tabs-panels?expand=1#diff-40a886e3e6d3eb98c7359a4cd8f557b18f2c10609a906112bea8ec6befe977ccR27-R66)

The styles shown in the video are irrelevant, they are just basic styles added specifically to this story. The idea is you style it yourself for your particular situation.

https://github.com/user-attachments/assets/38522c86-27e3-4b51-8a19-337e68126215

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/holocene/tabs-primitive/context.ts
+++ b/src/lib/holocene/tabs-primitive/context.ts
@@ -1,0 +1,29 @@
+import { createContext } from 'svelte';
+
+export interface TabsContextType<T extends string = string> {
+  getIdForTab: (tab: T) => string;
+  getPanelIdForTab: (tab: T) => string;
+  selectedTab: T;
+  tabs: T[];
+  setSelectedTab: (selectedTab: T) => void;
+}
+
+const [getCtx, setCtx] = createContext<TabsContextType>();
+
+export function getTabsContext<
+  T extends string = string,
+>(): TabsContextType<T> {
+  const ctx = getCtx();
+
+  if (!ctx) {
+    throw new Error('Tabs context not found — did you call setTabsContext?');
+  }
+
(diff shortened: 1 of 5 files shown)


ardiewen:
@tegan-temporal I suppose there's a (valid) assumption that each tab button is in fact an html `<button>` element but it is not explicit in the type. This assumption allows us to delegate the `Enter` / `Space` semantics to the `button` element itself in manual mode.

In situations where the button is say, `<div role="tab">` instead, AND the setup is in `activation = manual`, then the user cannot actually activate a focused tab with a keyboard.

Do we want to explicitly capture the keydown `event.key` for `"Enter"` and `" "`?

tegan-temporal:
@ardiewen originally i didn't want to support anything besides buttons and anchor elements but adding the onkeydown logic wasn't too much more than already existed so I went ahead and added it.

ardiewen (approved):

PR #3577, opened 2026-06-22
Add  TEMPORAL_NAV_COLLAPSED_BY_DEFAULT env var to start UI with collapsed navs

Alex-Tideman:
## Description & motivation 💭 
Adds `TEMPORAL_NAV_COLLAPSED_BY_DEFAULT` for Docker/server config. When enabled, the left navigation and saved views navigation start collapsed by default, while preserving the existing click behavior and persisted user preferences. This is helpful for demo's that are run on startup of CLI, especially for split-screen demos. This allows for more horizontal real estate on the Web UI by default.

Existing localStorage values still win, so this acts as a first-load default rather than forcing nav state on every page load.

Screenshot 2026-06-22 at 9 40 18 AM

### Design Considerations 🎨 
No design changes. This only changes the configurable initial state of existing nav UI.

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

Tested with:

  - `pnpm exec vitest run src/lib/stores/nav-open.test.ts src/lib/services/settings-service.test.ts`
  - `pnpm exec svelte-check --tsconfig ./tsconfig.json --tsgo --incremental`
  - `GOCACHE=/private/tmp/temporal-ui-go-build go test ./server/config ./server/api`

  1. Start the UI with `TEMPORAL_NAV_COLLAPSED_BY_DEFAULT=true`.
  2. Clear localStorage for `navOpen` and `savedQueryNavOpen`.
  3. Load the app and verify the left nav starts collapsed.
  4. Open a Workflows or Standalone Activities page and verify saved views starts collapsed.
  5. Click either nav toggle and verify it expands/collapses and persists as before.
  6. Start without the env var, or with it set to `false`, and verify both navs default to expanded.

### DIFF
--- a/server/config/development.yaml
+++ b/server/config/development.yaml
@@ -10,6 +10,7 @@ refreshInterval: 1m
 defaultNamespace: default
 showTemporalSystemNamespace: false
 feedbackUrl:
+navCollapsedByDefault: false
 disableWriteActions: false
 workflowTerminateDisabled: false
 workflowCancelDisabled: false
--- a/server/config/docker.yaml
+++ b/server/config/docker.yaml
@@ -9,9 +9,9 @@ enableUi: {{ env "TEMPORAL_UI_ENABLED" | default "true" }}
 cloudUi: {{ env "TEMPORAL_CLOUD_UI" | default "false" }}
 defaultNamespace: {{ env "TEMPORAL_DEFAULT_NAMESPACE" | default "default" }}
 feedbackUrl: {{ env "TEMPORAL_FEEDBACK_URL" | default "" }}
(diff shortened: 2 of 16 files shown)


server/config/development.yaml
```diff
@@ -10,6 +10,7 @@ refreshInterval: 1m
 defaultNamespace: default
 showTemporalSystemNamespace: false
 feedbackUrl:
+navCollapsedByDefault: false
```
laurakwhit, line 13: Do we need to add this to `server/config/with-auth.yaml` as well?

laurakwhit (approved):

PR #3582, opened 2026-06-22
fix(auth): call /auth/logout endpoint on logout instead of client-side nav

### LINKED ISSUE #3579: Logout through ui navigates to /login via history.pushState
Currently, logging out through UI navigates to /login via history.PushState
It does not call /auth/logout, and without clearing the refresh cookie

There is already a /auth/logout endpoint that clears cookies, but SPA never use it

mudassiralladatkhan:
## What

Logging out through the UI never hit the backend, so the refresh cookie was never cleared.

The app-layout (`src/routes/(app)/+layout.svelte`) defined a local `logout` handler that only called `clearAuthUser()` and did a client-side `goto(routeForLoginPage())`. It never called `GET /auth/logout`, so the server-side refresh cookie survived the "logout" (#3579).

A correct `logout()` already exists in `src/lib/stores/auth-user.ts` — it calls `GET /auth/logout` with `credentials: 'include'`, clears local auth state, then redirects — but the SPA never used it.

## Change

- Wire the layout's `<UserMenu>` / `<UserMenuMobile>` to the shared `logout()` from `auth-user.ts`.
- Remove the now-unused local handler and its now-unused imports (`clearAuthUser`, `routeForLoginPage`). `goto` stays (still used for namespace navigation).

## ⚠️ One behavior change to confirm

The old local handler navigated to **`/login`** (soft `goto`). The shared `logout()` does `window.location.href = '/'` (hard navigation, which also fully resets app state). So after this change, logout lands on **`/`** instead of `/login`.

I think the hard nav to `/` is preferable (guarantees a clean state post-logout, and `/` will redirect unauthenticated users to login anyway), but if you'd rather preserve the exact `/login` destination, I can adjust `logout()` to redirect there instead — just let me know.

## Tests

`src/lib/stores/auth-user.ts` had no tests; added `auth-user.test.ts` covering `logout()`:
- calls `/auth/logout` with `credentials: 'include'`
- clears the local auth user
- redirects to `/`
- still clears state + redirects when the request fails (network error path)

## Verification

- `pnpm test -- --run src/lib/stores/auth-user.test.ts` → 4/4 pass
- `pnpm exec eslint` (changed files) → clean
- `pnpm prettier --check` (changed files) → clean
- `pnpm check` (svelte-check) → 0 errors

Closes #3579

### DIFF
--- a/src/lib/stores/auth-user.test.ts
+++ b/src/lib/stores/auth-user.test.ts
@@ -1,0 +1,59 @@
+import { get } from 'svelte/store';
+
+import { afterEach, beforeEach, describe, expect, test, vi } from 'vitest';
+
+import { authUser, logout, setAuthUser } from './auth-user';
+
+describe('auth-user store: logout', () => {
+  beforeEach(() => {
+    authUser.set({});
+    vi.restoreAllMocks();
+    vi.stubGlobal(
+      'fetch',
+      vi.fn().mockResolvedValue(new Response(null, { status: 200 })),
+    );
+    vi.stubGlobal('location', { href: '' } as Location);
+  });
+
+  afterEach(() => {
+    vi.unstubAllGlobals();
+  });
+
+  test('calls the /auth/logout endpoint with credentials so cookies are cleared', async () => {
+    await logout();
+
+    expect(fetch).toHaveBeenCalledWith('/auth/logout', {
+      method: 'GET',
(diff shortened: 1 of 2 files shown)


Alex-Tideman:
Appreciate the PR!

Closing this one to use https://github.com/temporalio/ui/pull/3590

PR #3590, opened 2026-06-23
Add direct SSO redirect and backend logout

### LINKED ISSUE #3578: Add passThroughProvider option to skip intermediate SSO page
Ask is to add a config flag so ui server can skip the "Continue to SSO" intermediate page 
and directs unauthenticated users direct to OIDC provider

Currently users pretty much have to deal with bigger workarounds like using nginx proxy to bypass this. 
Not expert in Svelte but think change would be to get SvelteKit layout to direct to /auth/sso instead of /login.

### LINKED ISSUE #3579: Logout through ui navigates to /login via history.pushState

Alex-Tideman:
## Description & motivation 💭 

This PR improves the OIDC authentication flow for self-hosted Temporal UI deployments.

It adds a new `auth.redirectToProvider` server configuration option that allows deployments to skip the intermediate Temporal UI `/login` page and send unauthenticated users directly to the configured OIDC provider through `/auth/sso`. This is useful for environments that want an IdP-first login flow and currently need proxy-level workarounds to bypass the "Continue to SSO" page.

It also updates UI logout behavior to call the existing `/auth/logout` backend endpoint instead of only clearing client state and navigating to `/login`. Calling the backend endpoint lets the server clear the refresh, session, and user cookies so the browser is actually logged out of the Temporal UI session.

Implementation details:

- Added `auth.redirectToProvider` to server config, Docker config, auth docs, and the `/api/v1/settings` auth response.
- Added `TEMPORAL_AUTH_REDIRECT_TO_PROVIDER` support for Docker config.
- Mapped `RedirectToProvider` into frontend settings as `settings.auth.redirectToProvider`, defaulting to `false` for compatibility with older settings responses.
- Updated unauthenticated app-route redirects to use `/auth/sso` when `redirectToProvider` is enabled.
- Updated `/login` route loading so direct visits also redirect to `/auth/sso` when the flag is enabled.
- Updated 401/403 handling to respect the same redirect setting.
- Updated the user-menu logout action to call `/auth/logout` with credentials before clearing local auth state and navigating home.

https://github.com/user-attachments/assets/7c5c68b8-2d4b-4c2f-a08e-d2a2f5610165

### Design Considerations 🎨 

No design review needed. The default behavior remains the existing `/login` page with the "Continue to SSO" step. The new direct provider redirect only applies when `auth.redirectToProvider` is explicitly enabled.

### How was this tested 👻 

- [x] Manual testing
- [x] E2E tests added
- [x] Unit tests added

Added/updated tests for:

- Server settings response includes `Auth.RedirectToProvider`.
- Frontend settings service maps `RedirectToProvider` and defaults it to `false`.
- Route helpers return `/login` by default and `/auth/sso` with `returnUrl` when direct provider redirects are enabled.
- Base-path behavior for the new auth redirect helper.
- 401/403 error handling respects `redirectToProvider`.
- UI logout calls `/auth/logout`, includes credentials, clears local auth state, and navigates home.
- Playwright OAuth flow coverage for default unauthenticated redirect, direct SSO redirect, and backend logout invocation.

Validated locally with:

```sh
pnpm test src/lib/utilities/route-for.test.ts src/lib/utilities/route-for-base-path.test.ts src/lib/services/settings-service.test.ts src/lib/utilities/handle-error.test.ts src/lib/stores/auth-user.test.ts
go test ./server/api
pnpm check
PW_MODE=integration pnpm exec playwright test tests/integration/oauth-flow.spec.ts --project "chromium desktop"
git diff --check
```

`pnpm check` reported 0 errors. Existing Svelte warnings remain.

1. Configure OIDC auth normally and leave `auth.redirectToProvider` unset or set to `false`.
2. Run ui with pnpm dev:with-auth
3. Visit a protected UI route while unauthenticated.
4. Confirm the UI still redirects to `/login` and shows the existing "Continue to SSO" flow.
5. Set `auth.redirectToProvider: true` or, for Docker config, `TEMPORAL_AUTH_REDIRECT_TO_PROVIDER=true`.
6. Restart the UI server and visit a protected UI route while unauthenticated.
7. Confirm the browser is redirected directly to `/auth/sso` and the original route is preserved in `returnUrl`.
8. Log in, then use the user-menu logout action.
9. Confirm the browser requests `/auth/logout`, the server clears auth cookies, and the UI navigates home.

### Draft Checklist 

Ready for review.

### Merge Checklist 

- [x] CI passes

### Issue(s) closed 

Closes #3578
Closes #3579

### Any docs updates needed? 

Updated `AUTHENTICATION.md` with the new `redirectToProvider` auth setting.

### DIFF
--- a/AUTHENTICATION.md
+++ b/AUTHENTICATION.md
@@ -38,6 +38,7 @@ auth:
 | Field                | Type     | Description                                                                                                                    |
 | -------------------- | -------- | ------------------------------------------------------------------------------------------------------------------------------ |
 | `enabled`            | boolean  | Enable or disable authentication                                                                                               |
+| `redirectToProvider` | boolean  | Skip the Temporal UI login page and redirect unauthenticated users directly to the configured OIDC provider                    |
 | `maxSessionDuration` | duration | Maximum session duration before forced re-login (e.g., `8h`, `24h`, `168h`). Set to `0` or omit for unlimited session duration |
 | `providers`          | array    | List of auth providers (currently only the first is used)                                                                      |
 
--- a/package.json
+++ b/package.json
@@ -154,6 +154,7 @@
     "cors": "^2.8.5",
     "cssnano": "^5.1.14",
     "desm": "^1.3.1",
+    "ejs": "^6.0.1",
     "eslint": "^9.39.2",
     "eslint-config-prettier": "^10.1.8",
     "eslint-plugin-import-x": "^4.16.1",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -277,6 +277,9 @@ importers:
       desm:
         specifier: ^1.3.1
         version: 1.3.1
+      ejs:
+        specifier: ^6.0.1
+        version: 6.0.1
       eslint:
         specifier: ^9.39.2
         version: 9.39.4(jiti@1.21.7)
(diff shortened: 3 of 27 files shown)


GraceGardner (approved):
thanks for getting on this so fast

enasikbst:
@Alex-Tideman 

`redirectToProvider` causes an infinite redirect loop for authenticated-but-unauthorized users (403)

We enabled `auth.redirectToProvider` (TEMPORAL_AUTH_REDIRECT_TO_PROVIDER=true) on a self-hosted setup with a custom authorizer/claim-mapper (OIDC via Google). It works great for expired sessions, but it loops forever for a user who authenticates successfully yet has no authorization.

Flow:
1. User hits the UI → redirected to /auth/sso → IdP.
2. IdP authenticates fine (valid SSO session) → callback → session cookies set. The user is authenticated.
3. UI loads and calls the frontend (e.g. cluster-info) → our authorizer returns PermissionDenied → 403 because the user has no role.
4. `redirectToProvider` treats the 403 the same as a 401 and redirects back to /auth/sso → IdP → step 2 → infinite loop.

Suggestion: only redirect to the provider on 401 (or when there's no session). On 403, stop and render an "access denied" / unauthorized state instead of re-triggering SSO.

PR #3592, opened 2026-06-23
Improve re-renders for large encoded event histories

laurakwhit:
## Description & motivation 💭 

1. Updates `PayloadSummary` to decode by content, not by reference. Similar to the guard for `PayloadDecoder` added in https://github.com/temporalio/ui/pull/3401 as a follow-up to https://github.com/temporalio/ui/pull/3299. A summary that's already decoded now stays decoded across polls instead of re-hitting the codec server.
2. The "Child Workflow" timeline rendered inside an expanded event card (`GraphWidget`) now refreshes in place instead of re-fetching the child's full workflow + history on every tick (caused by a $derived/{#await getWorkflowAndEventHistory()} whose dependencies were being invalidated on every parent re-render). 

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

**Prerequisites / setup**

* A codec server + encoded payloads (e.g. samples-typescript/encryption). Point the UI at it via Settings → Codec Server (set the endpoint).
* A long-running parent workflow that:
  - stays running (sleeps/loops so the UI keeps long-polling)
  - spawns many child workflows
  - sets user-metadata summaries on events — this is specifically what PayloadSummary in the timeline decodes (`group.userMetadata.summary`)

1. Open the running workflow's History / timeline view.
2. Click on a timeline row for a child
   - [ ] Verify the "Child Timeline" does not flash on every tick

### Issue(s) closed 

`DT-4192`

### DIFF
--- a/src/lib/components/lines-and-dots/svg/graph-widget.svelte
+++ b/src/lib/components/lines-and-dots/svg/graph-widget.svelte
@@ -1,16 +1,19 @@
 <script lang="ts">
+  import { untrack } from 'svelte';
+
   import { groupEvents } from '$lib/models/event-groups';
+  import type { EventGroups } from '$lib/models/event-groups/event-groups';
   import { fetchAllEvents } from '$lib/services/events-service';
   import { fetchWorkflow } from '$lib/services/workflow-service';
-  import type { WorkflowEvents } from '$lib/types/events';
   import type { WorkflowExecution } from '$lib/types/workflows';
 
   import TimelineGraph from './timeline-graph.svelte';
 
   interface Props {
     namespace: string;
-    workflowId: string;
-    runId?: string;
+    workflowId?: string | null;
+    runId?: string | null;
+    eventCount?: number;
     viewportHeight?: number;
     onLoad?: () => void;
     class?: string;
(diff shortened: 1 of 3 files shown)


src/lib/components/lines-and-dots/svg/graph-widget.svelte
```diff
@@ -20,36 +23,55 @@
...
+
+  const fetchKey = $derived(
+    `${namespace}|${workflowId}|${runId}|${eventCount}`,
+  );
+
+  $effect(() => {
+    void fetchKey;
+    let cancelled = false;
+    untrack(() => getWorkflowAndEventHistory()).then((next) => {
+      if (cancelled) return;
```
andrewzamojc, line 58: You could check for the !next case, but it seems super rare to me.

andrewzamojc (approved):
LGTM

src/lib/components/lines-and-dots/svg/graph-widget.svelte
laurakwhit: We want to intentionally set `snapshot = undefined` if `next` isn't defined so (based on the condition on line 68) it just won't render.

PR #3598, opened 2026-06-24
a11y(2.4.3): inert-based focus containment + restore; drop trap from native dialogs

ardiewen:
Alternative to #3536. Fixes WCAG 2.4.3 (Focus Order) using platform primitives instead of the hand-rolled keydown trap.

`focus-trap.ts` did two jobs — Tab containment + focus restoration — both of which have native equivalents:

- **`<dialog>` + `showModal()`** (`holocene/modal`, `command-palette/modal`) already traps *and* restores focus natively → **removed `use:focusTrap` entirely.**
- **Non-dialog** (`drawer` — no overlay, `maximizable`) needs help → the action now sets `inert` on background content (capturing/restoring focus around it). `inert` also blocks pointer + assistive-tech access, which the old keydown trap (first/last `Tab` only) did not.
- **`maximizable`** additionally captures focus in the component *before* toggling, because maximizing re-renders CodeMirror and blurs to `<body>` before the action runs (the action alone — and #3536 — would restore to `<body>`).

Net: −44 lines of trap machinery (keydown handlers, MutationObserver, first/last tracking).

## Testing
- Unit: +4 tests (inert on activate / clear on destroy, capture+restore, toggle lifecycle, preserve pre-existing `inert`). `pnpm check` clean, suite green.
- Browser-verified end-to-end:
  - **Modal** — outside-focus blocked natively, Esc restores focus to trigger (with `focusTrap` removed).
  - **Drawer** — real `inert` blocks background focus; close → inert cleared + focus restored to trigger.
  - **Maximizable** — real `inert` containment; focus restored to the maximize button via both Esc and the minimize button.
- [ ] NVDA/VoiceOver pass (background unreachable while a drawer/maximized surface is open).

`inert` is Baseline (Chrome 102+/FF 112+/Safari 15.5+), same era as `showModal()`.

A11y-Audit-Ref: 2.4.3-focus-restoration-on-close


src/lib/holocene/maximizable/maximizable.svelte
laurakwhit: Could `use:focusTrap` potentially run before the `portalElement` is set? Maybe we could make `portalElement` $state() and add a `null` check to `use:focusTrap` just in case.

src/lib/utilities/focus-trap.ts
```diff
@@ -9,64 +9,81 @@ export const getFocusableElements = (node: HTMLElement) =>
...
+        ? document.activeElement
+        : null;
 
-    if (!fromObserver) firstFocusable?.focus();
+    releaseBackground = inertBackground(node);
 
-    firstFocusable?.addEventListener('keydown', onKeydown);
-    lastFocusable?.addEventListener('keydown', onKeydown);
+    const [firstFocusable] = getFocusableElements(node);
+    (firstFocusable ?? node).focus();
```
laurakwhit, line 58: NIT: Might be nice to add `tabindex="-1"` to components like `maximizable.svelte` and `drawer.svelte` since by default they have none. Although, it shouldn't ever really be an issue because they will likely always contain a focusable child element.

src/lib/utilities/focus-trap.ts
```diff
@@ -9,64 +9,81 @@ export const getFocusableElements = (node: HTMLElement) =>
...
-    mutationRecords: MutationRecord[],
-    observer: MutationObserver,
-  ) => {
-    if (mutationRecords.length) {
-      cleanUp();
-      setFocus(true);
+    releaseBackground();
+    releaseBackground = () => {};
+
+    if (previouslyFocused && document.body.contains(previouslyFocused)) {
```
laurakwhit, line 68: Should we provide a way to pass a fallback or default here? That way if whatever opened the focus trap re-renders (e.g. a `CodeBlock`) while the user is in it they won't be kicked back to `<body>`.

src/lib/utilities/focus-trap.ts
```diff
@@ -9,64 +9,81 @@ export const getFocusableElements = (node: HTMLElement) =>
...
     }
+    current = current.parentElement;
+  }
+
+  return () => {
+    for (const element of inerted) element.inert = false;
   };
+};
+
+export const focusTrap = (node: HTMLElement, enabled: boolean) => {
```
laurakwhit, line 41: TIL Svelte runs action setup top-to-bottom, because if we change the order of `use:portal` and `use:focusTrap` in the `Drawer` component it silently breaks. Not sure we need to make any changes, just might be worth noting and/or adding some kind of "use portal before focus trap" lint warning.

src/lib/utilities/focus-trap.test.ts
```diff
@@ -78,3 +78,77 @@ describe('getFocusableElements', () => {
     expect(focusable.length).toBe(1);
   });
 });
+
+describe('focusTrap focus management (inert-based)', () => {
```
laurakwhit, line 82: Wouldn't hurt to add a test for overlapping traps (e.g. if there is a `Maximizable` component inside a `Drawer` component).

src/lib/holocene/maximizable/maximizable.svelte
ardiewen: Verified: yes.

[Went another direction here](https://github.com/temporalio/ui/pull/3598/changes/7a213942c9dec3038d56640d1d1cad84cf7ef990): removed the $effect entirely: `maximize()` now portals to `<body>` synchronously before setting `maximized = true`, so the node is already in place when `use:focusTrap` activates in the same flush. No more $effect phase ordering, so there's nothing to null-check.

src/lib/utilities/focus-trap.ts
```diff
@@ -9,64 +9,81 @@ export const getFocusableElements = (node: HTMLElement) =>
...
-        }
-      } else if (document.activeElement === lastFocusable) {
-        firstFocusable.focus();
-        event.preventDefault();
+  while (current !== document.body && current.parentElement) {
+    for (const sibling of Array.from(current.parentElement.children)) {
+      if (
+        sibling !== current &&
+        sibling instanceof HTMLElement &&
+        !sibling.inert
```
laurakwhit, line 27: Do we need to add a check for `aria-live` here as well? E.g. would a toast fired from within a drawer get inerted/unannounced?

src/lib/holocene/maximizable/maximizable.svelte
```diff
@@ -23,14 +23,69 @@
...
+
+    maximized = true;
+  };
+
+  const restorePosition = async () => {
+    await tick();
+    portalElement?.destroy();
+    portalElement = null;
+    if (wrapperEl && originalParent?.isConnected) {
+      originalParent.insertBefore(wrapperEl, originalNextSibling);
```
laurakwhit: Might not hurt to add a check that originalNextSibling is still a child of originalParent here since `Node.insertBefore` might throw an exception if it isn't.

laurakwhit (approved):
Two small comments that might be worth looking into, but otherwise 🚀

src/lib/holocene/maximizable/maximizable.svelte
ardiewen: great catch, fixed!

src/lib/utilities/focus-trap.ts
ardiewen: Yes, this is a real concern. I'm going to address this as a follow up ticket (DT-4252)

PR #3601, opened 2026-06-25
Add existing versions to worker create deployment version form

laurakwhit:
## Description & motivation 💭 

Displays a list of five+ past versions on the Create Worker Deployment Version page.

| Before | After |
|-|-|
|Screenshot 2026-06-24 at 6 11 38 PM|Screenshot 2026-06-26 at 1 52 52 PM|

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
`DT-4113`

### DIFF
--- a/src/lib/components/workers/serverless-worker-form/existing-versions.svelte
+++ b/src/lib/components/workers/serverless-worker-form/existing-versions.svelte
@@ -1,0 +1,56 @@
+<script lang="ts">
+  import DeploymentStatus from '$lib/components/deployments/deployment-status.svelte';
+  import Timestamp from '$lib/components/timestamp.svelte';
+  import { translate } from '$lib/i18n/translate';
+  import {
+    isVersionSummaryNew,
+    type VersionSummary,
+  } from '$lib/types/deployments';
+  import { parseVersionStatus } from '$lib/utilities/deployments';
+  import { getBuildIdFromVersion } from '$lib/utilities/get-deployment-build-id';
+
+  interface Props {
+    versions: VersionSummary[];
+  }
+
+  let { versions }: Props = $props();
+
+  const getBuildId = (version: VersionSummary): string =>
+    isVersionSummaryNew(version)
+      ? version.deploymentVersion.buildId
+      : getBuildIdFromVersion(version.version);
+
+  const getStatus = (version: VersionSummary) =>
+    parseVersionStatus(
+      isVersionSummaryNew(version) ? version.status : version.drainageStatus,
+    );
+</script>
+
+{#if versions?.length}
+  <div class="mt-4">
+    <h3 id="existing-versions-title" class="text-sm font-medium">
+      {translate('workers.existing-versions')}
+    </h3>
+    <ol class="flex flex-col" aria-labelledby="existing-versions-title">
+      {#each versions as version, index (index)}
+        {@const { status, label } = getStatus(version)}
+        <li
+          class="flex w-full items-start justify-between gap-2 border-b border-subtle py-2 last-of-type:border-b-0"
+        >
+          <div class="flex flex-wrap items-center gap-2 truncate">
+            <span class="select-all truncate font-mono text-sm"
+              >{getBuildId(version)}</span
+            >
+            <DeploymentStatus {status} {label} />
+          </div>
+          <Timestamp
+            class="shrink-0 text-sm text-secondary"
+            as="span"
+            dateTime={version.createTime}
+            options={{ relative: true }}
+          />
+        </li>
+      {/each}
+    </ol>
+  </div>
+{/if}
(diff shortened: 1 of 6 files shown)


src/lib/components/workers/serverless-worker-form/existing-versions.svelte
```diff
@@ -52,5 +64,10 @@
         </li>
       {/each}
     </ol>
+    {#if hiddenCount > 0}
+      <span class="mt-1 text-xs text-secondary">
+        {translate('workers.existing-versions-more', { count: hiddenCount })}
```
laurakwhit: @bilal-karim what do you think about adding a "+ [...] more" count here if there are more than five?

Image

bilal-karim: I guess I don't mind it, but there are two things I'm thinking about and would love your take on.

First, the main motivation for adding this list was to show prior versions during creation for easy reference, especially for users manually managing build IDs in the UI. That's why I thought showing the 5 most recent would be enough. What do you think? Do you think there's value in showing more?

Second, I'm questioning whether lifecycle status (current, ramping, draining, drained, inactive) adds much in this context. I don't think it's as useful here as the build ID and its timestamp on the right. Maybe we should remove the lifecycle status?

laurakwhit: > That's why I thought showing the 5 most recent would be enough. What do you think? Do you think there's value in showing more?

I think the 5 most recent is fine, just didn't want users to only see those 5 and wonder where the rest are. 

> Maybe we should remove the lifecycle status?

I don't think it hurts to have it! Can always remove it in the future.

bilal-karim: >I think the 5 most recent is fine, just didn't want users to only see those 5 and wonder where the rest are.

Fair point. We can rename "Existing Versions" to "Last 5 Versions".

laurakwhit: I think "Existing Versions" is fine, but let me know if you'd prefer I change it and remove the "+ [...] more"!

bilal-karim: Let's move forward with "Recent Versions". I'm ok with `+ [...] more` but can it be a secondary small button?

laurakwhit: Updated ⬇️ 

Screenshot 2026-06-26 at 1 52 52 PM

Alex-Tideman (approved):

src/lib/pages/worker-deployment-version-create.svelte
```diff
@@ -20,10 +22,30 @@
   let { namespace, deployment, onSuccess }: Props = $props();
 
   let error = $state<string | undefined>();
+  let versions = $state<VersionSummary[]>();
 
   const backHref = $derived(
     routeForWorkerDeployment({ namespace, deployment }),
   );
+
+  $effect(() => {
```
andrewzamojc, line 31: Just a thought on the effect for the promise. There's a pattern in the repo Thomsen is promoting and i've used a few times. It avoids the fetch in effect and leverages svelte await syntax. I'm starting to like it so I thought I'd promo it here.

```
// page.svelte

const deploymentPromise = $derived(fetchDeployment(...))

<CreateVersionForm {deploymentPromise}
```

```
// create-versions-form.svelte

{#await deploymentPromise}
   // maybe loading state if you need one
{:then deployment}
  {if deployemnt}whatever{/if}
{catch}
  error!
{/await}
```

src/lib/pages/worker-deployment-version-create.svelte
```diff
@@ -20,10 +22,30 @@
...
+      fetch,
+      undefined,
+      false,
+      controller.signal,
+    )
+      .then((response) => {
+        versions = response?.workerDeploymentInfo?.versionSummaries ?? [];
+      })
+      .catch(() => {
+        if (controller.signal.aborted) return;
```
andrewzamojc, line 45: Do you also need to check aborted in `.then`? I assume it's guarding against unmount?

andrewzamojc (approved):
LGTM. I left a couple comments that are optional / your call.

src/lib/pages/worker-deployment-version-create.svelte
laurakwhit: I opted not to go for this approach because there is currently no potential loading or error state for the recent versions.

src/lib/pages/worker-deployment-version-create.svelte
laurakwhit: A stale fetch would likely get overwritten immediately, but a check wouldn't hurt! 

Added [here](https://github.com/temporalio/ui/pull/3601/commits/7bc8fd626d3c923704103abc0b93dfa518e605a9) ✅

PR #3603, opened 2026-06-25
feature/schedules

tegan-temporal:
## Description & motivation 💭 

This is the feature branch for the schedules ui update. 5 separate PRs were merged into this one.


src/lib/components/schedule/schedule-form/spec-type-cron.svelte
```diff
@@ -0,0 +1,100 @@
...
+      (v) => ($form.specs[index] = { ...$form.specs[index], cronString: v })
+    }
+    placeholder="* * * * *"
+    required
+    error={!!$errors.specs?.[index]?.cronString?.[0]}
+    hintText={$errors.specs?.[index]?.cronString?.[0] ??
+      translate('schedules.cron-format-hint')}
+  />
+  <button
+    class="mr-auto underline hover:text-brand"
```
laurakwhit, line 93: Can we use a Holocene component here? If not, we should add the type so it doesn't default to submit. 
```suggestion
    type="button"
    class="mr-auto underline hover:text-brand"
```

src/lib/components/schedule/schedule-form/schedule-details-card.svelte
```diff
@@ -0,0 +1,248 @@
...
+          value="after"
+          label={translate('common.after')}
+        />
+        {#if $form.endKind === 'after'}
+          <div>
+            <Input
+              id="endAfterOccurrences"
+              type="number"
+              inputmode="numeric"
+              pattern="0-9+"
```
laurakwhit: Looks like maybe this is an invalid regex? 
```suggestion
              pattern="[0-9]+"
```

src/lib/components/schedule/schedule-form/schedule-details-card.svelte
```diff
@@ -0,0 +1,248 @@
...
+
+      <Input
+        id="jitter"
+        label={translate('schedules.jitter')}
+        type="number"
+        step={1}
+        min={0}
+        suffix={translate('common.seconds-abbreviated')}
+        bind:value={
+          () => $form.jitter.toString(), (v) => ($form.jitter = Number(v))
```
laurakwhit:
```suggestion
          () => $form.jitter?.toString() ?? '', (v) => ($form.jitter = Number(v))
```

src/lib/components/schedule/constants.ts
```diff
@@ -0,0 +1,87 @@
...
+  },
+  AllowAll: {
+    label: translate('schedules.overlap-allow-all-label'),
+    description: translate('schedules.overlap-allow-all-description'),
+  },
+});
+
+const MONTHS_UNIT: Unit<'month(s)'> = {
+  label: 'month(s)',
+  convert: (n: number) => n * 31 * 24 * 60 * 60,
```
laurakwhit, line 45: Any potential issues with hard coding this to 31 days 🤔 Maybe we can hint at this in the UI somewhere...

src/lib/holocene/duration-input/duration-input.svelte
```diff
@@ -79,16 +89,32 @@
...
+    if (!durationValue) return '';
+    const seconds = Number(parseDuration(durationValue));
+    if (!Number.isFinite(seconds)) return '';
+    const unitDef = units.find((u) => u.label === unitLabel);
+    const factor = unitDef ? unitDef.convert(1) : 1;
+    return factor ? String(seconds / factor) : String(seconds);
+  };
+
+  // svelte-ignore state_referenced_locally
+  let rawValue = $state(toUnitValue(value, initialUnit));
```
laurakwhit, line 116: Does the `rawValue` get updated if the `value` changes outside of the component?

src/lib/components/schedule/utilities/get-form-spec.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+export function getFormSpecFromSpec(
+  spec: ScheduleSpec | null | undefined,
+): FormSpecSchema[] {
+  const specs: FormSpecSchema[] = [];
+
+  for (const calendar of spec?.structuredCalendar ?? []) {
+    specs.push({
+      kind: 'frozen',
```
laurakwhit, line 22: Just want to confirm, once you create a calendar spec you can never edit it (even on create). Is that expected?
Image

src/lib/components/schedule/utilities/get-form-schedule-defaults.ts
```diff
@@ -0,0 +1,182 @@
...
+    // new schedules and off for existing ones, whose payloads aren't decoded
+    // back into the form.
+    editInput: !describeFullSchedule,
+    encoding: 'json/plain',
+    messageType: '',
+
+    specs: getSpecs(describeFullSchedule),
+    timezoneName: spec?.timezoneName || 'UTC',
+    startTime: isoStringToCalendarDateStr(
+      String(spec?.startTime || new Date().toISOString()),
```
laurakwhit, line 144: Is there potential for a bug here by setting the `startTime` when the user opens the form vs. submits it (e.g. if they submit a minute after midnight)? Seems like a very small likelihood, but thought it might be worth mention 😅

src/lib/components/schedule/schedule-form/schedule-policies-drawer.svelte
```diff
@@ -0,0 +1,341 @@
...
+
+  // svelte-ignore state_referenced_locally
+  const policiesForm = superForm(pickPolicies(get(form)), {
+    SPA: true,
+    dataType: 'json',
+    validators: zodClient(formSchedulePoliciesSchema),
+    resetForm: false,
+    onUpdate: ({ form: validated }) => {
+      if (!validated.valid) return;
+      form.update(($form) => ({ ...$form, ...validated.data }));
```
laurakwhit, line 71: When I change any of these values and hit "Update Policies" it looks like the values update for a second on the "Schedule Policies" card and then the whole form reloads and the values are lost.

src/lib/components/schedule/utilities/get-form-spec.ts
tegan-temporal: yeah, you have to delete it and create a different one

src/lib/components/schedule/utilities/get-form-schedule-defaults.ts
tegan-temporal: we set the startTime when the user opens the form so they see it in the form. I think it would be worse if we used a date different from what they see in the form. Also, `isoStringToCalendarDateStr` strips the time portion off, we're just using the date portion. Lastly, I don't think it necessarily matters at all if startDate is in the past. We don't force them to have startDate in the present or future. It just won't effect anything before the schedule is created.

src/lib/components/schedule/constants.ts
tegan-temporal: FWIW it is in the tooltip. I don't think there is a better solution besides just not supporting months (which I would be fine with, but design seemed to think it's still useful like this.
Image

laurakwhit: Ah missed that tooltip. Ignore my comment then 😅

src/lib/holocene/duration-input/duration-input.svelte
tegan-temporal: no, it doesn't. The reason is because it also previously didn't and changing that behavior seemed like opening a can of worms since duration-input is pretty widely used and it would potentially change values that other parts of the app were receiving. Schedules doesn't need value to be controlled

src/lib/components/schedule/utilities/get-form-schedule-defaults.ts
laurakwhit: That's fair. Is there a likely a scenario on edit where there **wouldn't** be a `startTime` defined? I see [your comments](https://github.com/temporalio/ui/pull/3603/changes#diff-83238a2cb660cf3e129e2a5d65fd63f1397ae7ddb5f2d87f52d6e85df9a8f918R172-R181) in `get-request-body.ts` and was mostly thinking that the conditional there would fail if the submit happened a "day" after the form was opened.

tegan-temporal: the scenario I was thinking for startTime being either missing or more specific than just a date is if you were using the ui to edit something that was created via the CLI

tegan-temporal: and I think I accounted for missing time just because of the way the protos have all fields as optional. You can't assume basically any value is really provided which is rather unfortunate.

laurakwhit: Yeah 😞 We can totally just call this a potential extreme edge case and leave as-is.

src/lib/components/schedule/utilities/get-request-body.test.ts
```diff
@@ -0,0 +1,226 @@
...
+
+  it('emits the cron string for cron specs', async () => {
+    const body = await getRequestBody(buildForm());
+
+    expect(body.schedule.spec.cronString).toEqual(['0 12 * * *']);
+  });
+
+  it('drops invalid cron specs', async () => {
+    const body = await getRequestBody(
+      buildForm({ specs: [{ kind: 'cron', cronString: '0 12 * * *' }] }),
```
laurakwhit: It looks like this is a valid cron string. Should we be testing a invalid one here?

src/lib/components/schedule/schedule-form/schedule-details-card.svelte
```diff
@@ -0,0 +1,248 @@
...
+              inputmode="numeric"
+              pattern="[0-9]+"
+              label={translate('schedules.occurrences-label')}
+              labelHidden
+              bind:value={
+                () => $form.endAfterOccurrences?.toString(),
+                (v) => ($form.endAfterOccurrences = Number(v))
+              }
+              placeholder={translate('schedules.occurrences-placeholder')}
+              min={1}
```
laurakwhit: Looks like the there isn't an error for "0 occurrences", which means an unlimited schedule. Do we want to guard against that? 

Image

src/lib/components/schedule/schedule-action-modals/trigger-schedule-modal.svelte
```diff
@@ -0,0 +1,80 @@
...
+    namespace: string;
+    scheduleOverlapPolicy: OverlapPolicy;
+  }
+
+  const overlapPolicyContent = getOverlapPolicyContent();
+  const identity = getIdentity();
+
+  let { scheduleId, namespace, scheduleOverlapPolicy }: Props = $props();
+
+  let selectedOverlapPolicy = writable<OverlapPolicy>('AllowAll');
```
laurakwhit: Should this reflect the `scheduleOverlapPolicy` instead of always defaulting to `AllowAll`?

laurakwhit (commented):
A few more small comments, but overall lgtm 🎉

src/lib/components/schedule/schedule-action-modals/trigger-schedule-modal.svelte
tegan-temporal: since this is a manual trigger, IMO it is expected that if I trigger it, it runs, which isn't really guaranteed a lot of the other policies. Anyway, I will move it back to the schedule's specified policy as the default as that was the original behavior.

laurakwhit (approved):

PR #3614, opened 2026-06-29
Fix SAA and WF consistency

laurakwhit:
## Description & motivation 💭 

The `/namespaces/{namespace}/activities` view 
* "Close Time" is now "End"
* "Start Time" is now "Start"
* "Activity Type" is now "Type"
* Hides the "Task Queue" column by default
* Added "Run ID" column by default that is filterable and contains a link
* Removed `ExecutionTime` as a filterable search attribute

The `/namespaces/{namespace}/activities{activityID}/{runID}/details` view
* "Scheduled Time" is now "Start"
* Removed "Last Started Time" from details at the top of page
* Uses the same "No workers polling" alert as the Workflow details page

| Before | After |
|-|-|
|Screenshot 2026-06-29 at 4 32 13 PM|Screenshot 2026-06-29 at 4 26 26 PM|
|Screenshot 2026-06-29 at 4 30 52 PM|Screenshot 2026-06-29 at 4 25 42 PM|

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
`DT-4202`

### DIFF
--- a/src/lib/components/standalone-activities/activities-summary-configurable-table/table-body-cell.svelte
+++ b/src/lib/components/standalone-activities/activities-summary-configurable-table/table-body-cell.svelte
@@ -22,7 +22,7 @@
   const { label } = $derived(column);
   const namespace = $derived(page.params.namespace);
 
-  const filterableLabels = ['Activity ID', 'Activity Type', 'Task Queue'];
+  const filterableLabels = ['Activity ID', 'Run ID', 'Type', 'Task Queue'];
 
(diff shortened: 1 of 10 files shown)


acav-temporalio:
LGTM Laura! thanks for taking the lead on this. I've also updated in Figma (kept the execution time only so we have record of where it needs to go when they add that back in).

acav-temporalio (approved):
LGTM - I've also updated Figma to match.

PR #3618, opened 2026-06-30
Add connection status and optional Latest Version column to deployments (DT-4174)

rossnelson:
## What

Implements [DT-4174](https://temporalio.atlassian.net/browse/DT-4174): surface compute-provider connection status on Worker Deployments, and add an optional Latest Version column to the deployments list.

CleanShot 2026-07-02 at 13 13 16

This PR now contains both halves of the ticket. The Latest Version column (second half) was developed in #3625 and merged into this branch.

**Connection status (list + details)**
- Types: `ProviderValidation` / `ComputeStatus` + `computeStatus?` on `VersionSummaryNew`, matching the new `compute_status.provider_validation` API field.
- `deriveConnectionStatus` derives `pending` / `connected` / `failed` from `error_message` + `last_check_time` (`last_check_time` checked first; empty and absent messages treated identically). Fully unit-tested.
- `formatConnectionCheckTime` formats freshness through i18n as hours-only ("36 hours ago"), with "less than an hour ago" for sub-hour / future / invalid times.
- `ConnectionBadge`: presentational dot + word chip with an informational tooltip. The API failure message is rendered directly (no client-side error-code mapping, per the backend decision). No click action.
- **List page**: the Current Version cell is a composite of build ID + compute + connection, with a "No Current Version" empty state.
- **Details page**: the Versions table gains a Connection column, shown for Current / Ramping / Draining versions and an em dash otherwise.

**Latest Version column (list)**
- An opt-in, hidden-by-default column exposed via the table's configurable-columns control (the same gear + drawer used by the Workflows / Schedules / Activities tables).
- Shows build ID + lifecycle badge + compute, with no connection indicator (connection stays on the Current Version cell). Shows "Same as Current" when the latest version equals the current one, and an em dash when there is no latest version.
- Store: a new `DEPLOYMENTS` table type with namespace-scoped persistence.

**Design polish**
- The build ID and compute badge sit together (removed the fixed min-width that left a large gap for short build IDs).
- Removed the build ID copy buttons from the version cells (not shown in the design); the deployment name keeps its copy button.

All new UI is behind the `serverScaledDeployments` capability; the on-demand validate-connection flow is unchanged.

## Why

The backend (temporalio/api#810, temporalio/temporal#10778, both merged) adds a periodic connectivity check for serverless worker deployments. This surfaces that signal in the UI so misconfiguration is visible before invocations fail. Per the DT-4174 decision log, the list defaults to the Current version, with Latest available as an optional column.

## Testing

- `pnpm check`: 0 errors
- `pnpm lint`: clean
- `pnpm test --run`: 2192 passed / 1 skipped (unit tests for the derivation + freshness utils and the deployments column store)

## Notes

- Freshness copy is hours-only; the "within 6h" cadence wording in the Pending tooltip should be confirmed against the backend's actual check interval.

[DT-4174]: https://temporalio.atlassian.net/browse/DT-4174?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/lib/utilities/connection-status.ts
```diff
@@ -0,0 +1,64 @@
...
+  return translate('deployments.connection-pending');
+};
+
+export const formatConnectionCheckTime = (time: ValidTime): string => {
+  if (!time) return translate('deployments.connection-checked-recently');
+  try {
+    const parsedDate = isTimestamp(time)
+      ? timestampToDate(time)
+      : new Date(time as string | number | Date);
+    const diff = Date.now() - parsedDate.getTime();
```
laurakwhit, line 35: NIT: Should we guard against `NaN` here since `getTime()` can return it silently for an invalid date?

src/lib/utilities/connection-status.ts
```diff
@@ -0,0 +1,64 @@
...
+    if (diff < 3_600_000) {
+      return translate('deployments.connection-checked-recently');
+    }
+    const hours = Math.round(diff / 3_600_000);
```
laurakwhit: Just want to verify we want to overstate vs. understate here.

```suggestion
    const hours = Math.floor(diff / 3_600_000);
```

src/lib/i18n/locales/en/deployments.ts
```diff
@@ -29,6 +30,7 @@ export const Strings = {
   version: 'Version',
   status: 'Status',
   'build-status': 'Build Status',
```
laurakwhit:
```suggestion
```

src/lib/stores/configurable-table-columns.ts
laurakwhit: Should probably make this default table-aware.

```suggestion
      const columns = state?.[namespace] ?? defaultColumns;
```

laurakwhit (approved):

PR #3619, opened 2026-06-30
COM-191 - prefill CloudFormation template params in Launch Stack link

rossnelson:
## Summary

The serverless worker deployment "Launch Stack" button (CloudFormation, under the "I don't have a role yet" accordion) previously linked to a bare CloudFormation console URL and prefilled nothing. This wires the form's field values into the launch URL using AWS's `param_<ParameterName>=<value>` convention so the CloudFormation quick-create form opens with the template and parameters prefilled.

## Changes

`compute-fields.svelte`: `launchStackHref` is now a `$derived.by()` that, when `cfnTemplateUrl` is set, builds a `#/stacks/quickcreate?templateURL=...` URL and appends:

- `param_AssumeRoleExternalId` from `roleExternalId`
- `param_LambdaFunctionARNs` from `lambdaArn`

`RoleName` is left to the template default (no form field exists for it). Each `param_*` entry is omitted when its source field is empty, and the template URL plus every param value are `encodeURIComponent`-encoded. When `cfnTemplateUrl` is falsy, the existing bare-console fallback is preserved.

## Decisions worth a look

- **RoleName**: left at the template default since there is no form field for it. Easy to expose a role-name input if we want one.
- **quickcreate vs create/review**: used `#/stacks/quickcreate` for the prefilled-but-editable UX. Can pin to `#/stacks/create/review` for strict parity with the S3 export / Kinesis audit-log forms if preferred.

## Notes for the integrator (out of scope here)

End-to-end this requires cloud-ui to pass a real, publicly-hosted `cfnTemplateUrl` and repack the tarball. The local `temporal-worker-role.yaml` is download-only and currently has no `Parameters` section; the parameterized hosted template is handled separately in cloud-ui.

## Test plan

- [x] `pnpm check` clean
- [x] `pnpm lint` clean
- [x] Verified generated href for sample inputs contains the encoded `templateURL`, `param_AssumeRoleExternalId`, and `param_LambdaFunctionARNs`, and omits empty params

### DIFF
--- a/src/lib/components/workers/serverless-worker-form/compute-fields.svelte
+++ b/src/lib/components/workers/serverless-worker-form/compute-fields.svelte
@@ -75,7 +75,7 @@
     if (lambdaArn) {
       params.push(`param_LambdaFunctionARNs=${encodeURIComponent(lambdaArn)}`);
     }
-    return `https://console.aws.amazon.com/cloudformation/home#/stacks/quickcreate?${params.join('&')}`;
+    return `https://console.aws.amazon.com/cloudformation/home#/stacks/create/review?${params.join('&')}`;
   });
 
   let showRoleHelp = $state(false);
--- a/src/lib/components/workers/serverless-worker-form/compute-fields.svelte
+++ b/src/lib/components/workers/serverless-worker-form/compute-fields.svelte
@@ -62,11 +62,21 @@
 
   const resolvedCfnTemplate = $derived(cfnTemplateProp ?? cfnTemplate);
 
-  const launchStackHref = $derived(
-    cfnTemplateUrl
-      ? `https://console.aws.amazon.com/cloudformation/home#/stacks/quickcreate?templateURL=${encodeURIComponent(cfnTemplateUrl)}`
-      : 'https://console.aws.amazon.com/cloudformation/',
-  );
+  const launchStackHref = $derived.by(() => {
+    if (!cfnTemplateUrl) {
+      return 'https://console.aws.amazon.com/cloudformation/';
+    }
+    const params = [`templateURL=${encodeURIComponent(cfnTemplateUrl)}`];
+    if (roleExternalId) {
+      params.push(
+        `param_AssumeRoleExternalId=${encodeURIComponent(roleExternalId)}`,
+      );
+    }
+    if (lambdaArn) {
+      params.push(`param_LambdaFunctionARNs=${encodeURIComponent(lambdaArn)}`);
+    }
+    return `https://console.aws.amazon.com/cloudformation/home#/stacks/quickcreate?${params.join('&')}`;
+  });
 
   let showRoleHelp = $state(false);
   let showScaling = $state(false);


src/lib/components/workers/serverless-worker-form/compute-fields.svelte
```diff
@@ -62,11 +62,21 @@
...
+      return 'https://console.aws.amazon.com/cloudformation/';
+    }
+    const params = [`templateURL=${encodeURIComponent(cfnTemplateUrl)}`];
+    if (roleExternalId) {
+      params.push(
+        `param_AssumeRoleExternalId=${encodeURIComponent(roleExternalId)}`,
+      );
+    }
+    if (lambdaArn) {
+      params.push(`param_LambdaFunctionARNs=${encodeURIComponent(lambdaArn)}`);
```
justinschoeff, line 76: Should we make this `param_LambdaFunctionARN` instead of the plural `param_LambdaFunctionARNs`?

src/lib/components/workers/serverless-worker-form/compute-fields.svelte
justinschoeff, line 95: Should we update this to use the same as URL and params now?

Similar to audit logs: https://github.com/temporalio/cloud-ui/blob/9fa9b01e6d23723482fd277e914d8e05dc22310b/src/routes/(app)/settings/audit-logs/edit/_kinesis-form.svelte#L165-L172

src/lib/components/workers/serverless-worker-form/compute-fields.svelte
justinschoeff: Nevermind. I see we are using plural here and docs. Was confused by https://github.com/temporalio/ui/blob/main/src/lib/components/workers/serverless-worker-form/temporal-worker-role.yaml, but that doesn't seem to be what actually gets downloaded. The docs are using this one, which aligns with this change. https://github.com/temporalio/cloud-ui/blob/9fa9b01e6d23723482fd277e914d8e05dc22310b/src/lib/workers/cfn-template-prod.yaml

src/lib/components/workers/serverless-worker-form/compute-fields.svelte
rossnelson: Switched the path to create/review to match. Left region off intentionally since this stack only creates global IAM resources and the ARN list can span regions, and kept encodeURIComponent on the values.

src/lib/components/workers/serverless-worker-form/compute-fields.svelte
```diff
@@ -62,11 +62,21 @@
...
   const resolvedCfnTemplate = $derived(cfnTemplateProp ?? cfnTemplate);
 
-  const launchStackHref = $derived(
-    cfnTemplateUrl
-      ? `https://console.aws.amazon.com/cloudformation/home#/stacks/quickcreate?templateURL=${encodeURIComponent(cfnTemplateUrl)}`
-      : 'https://console.aws.amazon.com/cloudformation/',
-  );
+  const launchStackHref = $derived.by(() => {
+    if (!cfnTemplateUrl) {
+      return 'https://console.aws.amazon.com/cloudformation/';
```
laurakwhit, line 67: NIT: We could define this as a variable to use here and on line 78 so we'd only ever have to update it in one place.

laurakwhit (approved):

src/lib/components/workers/serverless-worker-form/compute-fields.svelte
justinschoeff: Understood. This comment was more in reference to updating the `downloadCfnTemplate` function. This is using `resolvedCfnTemplate` instead of downloading the URL, which means this template may be different. Also, the audit logs code is using the cf utilities which will inject the form inputs into the downloaded template.

We should be able to remove `src/lib/components/workers/serverless-worker-form/temporal-worker-role.yaml`.

PR #3622, opened 2026-07-01
Add configurable UI extension slots

Alex-Tideman:
## Description & motivation 💭 

Adds configurable iframe-based UI extensions so operators can mount customer-owned UI into stable Temporal UI slots without forking the main UI bundle.

This PR:

- Adds `customUi` server configuration for enabling iframe extensions, including extension IDs, slot names, source URLs, allowed origins, route filters, sandbox options, sizing bounds, and permissions.
- Exposes the configured extensions through the settings API and maps them into client-side `settings.customUi`.
- Adds reusable `ExtensionSlot` and `IframeExtension` components that render configured extensions only when the slot, route pattern, iframe URL, and allowed origin all match.
- Adds extension host utilities for route matching, origin validation, sandbox token generation, bounded iframe sizing, scoped context permissions, and same-origin navigation requests.
- Adds a small versioned `postMessage` protocol:
  - Host sends `temporal-ui/context` and `temporal-ui/theme`.
  - Extensions can send `temporal-extension/ready`, `temporal-extension/resize`, and permitted `temporal-extension/navigate`.
- Mounts extension slots in the top navigation, a full-width row below the top navigation, and the workflow header after workflow details.
- Documents the extension configuration, available slots, route matching, URL/origin rules, sandboxing, permissions, messaging, and local development workflow in `src/lib/extensions/README.md`.

Examples 

Screenshot 2026-07-01 at 10 15 13 AM
Screenshot 2026-07-01 at 10 14 58 AM

### Design Considerations 🎨 

- Extension slots are intentionally narrow and stable so external UI can appear near existing Temporal context without changing the default UI when `customUi.enabled` is false.
- Iframes are sandboxed by default with only `allow-scripts`; additional capabilities are opt-in per extension.
- Workflow and user context are permission-gated so extensions only receive sensitive context when explicitly configured.
- Extension navigation requests are limited to same-origin paths and require `navigation:write`.

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

Local test commands run:

```sh
pnpm exec vitest run src/lib/extensions/iframe-extensions.test.ts src/lib/extensions/messages.test.ts src/lib/services/settings-service.test.ts
go test ./server/api/...
```

Results:

- Vitest: 3 files passed, 30 tests passed.
- Go: `github.com/temporalio/ui-server/v2/server/api` passed.

1. Configure `customUi.enabled: true` with one or more `iframeExtensions` in the UI server config.
2. Point an extension at a relative `src` with `allowedOrigin: self`, or at an HTTPS origin with a matching explicit `allowedOrigin`.
3. Use one of the mounted slots:
   - `app.top-nav.actions.before`
   - `app.top-nav.actions.after`
   - `app.top-nav.sub-nav`
   - `workflow.header.after-details`
4. Load a matching route and verify the extension iframe renders only when its slot and `routePatterns` match.
5. From the extension iframe, send `ready` and `resize` messages and verify the host responds with context/theme messages and clamps dimensions to configured bounds.
6. For a workflow header extension with `permissions: [context:workflow]`, verify workflow context is delivered.
7. With `permissions: [navigation:write]`, send a same-origin navigate message and verify the host navigates; verify off-origin URLs are ignored.

### Draft Checklist 

N/A

### Merge Checklist 

N/A

### Issue(s) closed 

N/A

### Any docs updates needed? 

Docs for the new extension framework are included in `src/lib/extensions/README.md`.


src/lib/extensions/iframe-extensions.test.ts
```diff
@@ -0,0 +1,254 @@
...
+  it('uses allow-scripts as the default sandbox', () => {
+    expect(buildIframeSandbox(extension().sandbox)).toBe('allow-scripts');
+  });
+
+  it('adds only configured sandbox permissions', () => {
+    expect(
+      buildIframeSandbox({
+        ...extension().sandbox,
+        allowForms: true,
+        allowPopupsToEscapeSandbox: true,
```
GiantRobots: `allowPopupsToEscapeSandbox` This seems sus I feel like we shouldn't allow sandbox escaping?

src/lib/extensions/iframe-extensions.test.ts
```diff
@@ -0,0 +1,254 @@
...
+    );
+    expect(
+      safeNavigationPath(
+        'https://other.example.com/namespaces/default',
+        currentOrigin,
+      ),
+    ).toBeNull();
+    expect(safeNavigationPath('javascript:alert(1)', currentOrigin)).toBeNull();
+  });
+});
```
GiantRobots, line 394: Can we have some tests around iframe allow headers on origin and remote and maybe an iFrame script to reach to the parent frame and a script from the parent frame trying to reach into the child frame?

src/lib/extensions/iframe-extensions.ts
```diff
@@ -0,0 +1,223 @@
...
+export const initialWidth = (
+  sizing: IframeExtensionSizing,
+): number | undefined => {
+  const defaultWidth = positiveOrUndefined(sizing.defaultWidth);
+  if (defaultWidth == null) return undefined;
+
+  const minWidth = positiveOrUndefined(sizing.minWidth) ?? MIN_WIDTH;
+  const maxWidth = positiveOrUndefined(sizing.maxWidth) ?? MAX_WIDTH;
+  return clamp(defaultWidth, minWidth, maxWidth);
+};
```
GiantRobots, line 411: It might be cool to have an @attach on the parent html container with a ResizeObserver debounced by RAF that sends the height/width of the container to the iframe so you could have the iframe reactive to the size of the parent html contianer

server/server/config/config.go
```diff
@@ -143,6 +145,41 @@ type (
 		DefaultErrorLink    string `yaml:"defaultErrorLink"`
 	}
 
+	CustomUI struct {
+		Enabled          bool              `yaml:"enabled"`
+		IframeExtensions []IframeExtension `yaml:"iframeExtensions"`
```
laurakwhit, line 151: I think static YAML is fine for now, but wondering if it might it be worth it to support getting extensions via a manifest URL? 

Something that wouldn't require a redeploy every time something needed to be changed e.g. 
```
  extensionManifests:
    - url: https://example.com/manifest.json  
      allowedOrigin: https://example.com       
      permissions: [context:workflow]
```

Alex-Tideman: Yes I think that's a great idea. Then that would also support Nexus endpoints - you could point the url to a Nexus endpoint and have it return a custom UI

rossedfort (approved):

PR #3569, opened 2026-06-17
feature(DT-3929): Timeline - Collapsible Idle Time

tegan-temporal:
## Description & motivation 💭 

Add ability to collapse / expand idle time

### Screenshots (if applicable) 📸 

https://github.com/user-attachments/assets/58352063-57fd-411e-a960-c651a1913151

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

PR #3627, opened 2026-07-02
feat(timeline): HTML virtualization + collapse idle time

tegan-temporal:
## Summary

Rebuilds the workflow **Timeline** as an HTML, scroll-driven virtualized graph (replacing the SVG implementation) and adds **Collapse idle time** to hide long idle gaps between events.

## Why

The SVG timeline janked on large histories. This version stays smooth on 40k+ event workflows, and collapsing idle spans lets you focus on where work actually happens.

## Highlights

- **HTML virtualization** — fixed row pool re-pointed across scroll (no per-row create/destroy), scroll-driven windowing via a closed-form visible-range calc.
- **Collapse idle time** toggle (DT-3929) — collapses idle timespans, deferred until history load completes so segments are stable.
- **Grouped-event buffer** decoupled from Svelte reactivity, so large histories don't drown the reactive graph.
- **Per-frame cost work** — `$state.raw` on the hot group/event buffers, `contain: layout` + `transform` row positioning, and category-guarded row derivations.
- **No full-height painted elements** — grid lines use a solid fill and the collapsed-idle zigzag is windowed to the visible band, so nothing rasterizes the tens-of-thousands-px canvas height (fixes a Safari invisible-grid bug and a Chromium GPU meltdown).

## Visual change

- Timeline grid/tick lines now render as **solid** faint lines instead of **dashed**. Same subtle grid, but a dashed border on the full-height canvas forced Chromium to rasterize thousands of dash segments (GPU thread pegged at ~177% of scroll time); a solid fill is cheap and renders in every browser. The collapsed-idle **zigzag stays dashed** — it's a tiled SVG `<pattern>`, now windowed to the viewport so it's cheap at any height.

## Perf (fast scroll, 4.44× CPU throttle)

| metric | before (SVG) | after |
| --- | --- | --- |
| long tasks >50ms | 63 | 1–3 |
| frame p90 | 52ms | 29ms |
| main-thread work / scroll event | 4.4ms | 3.2ms |
| GC | ~1s | ~65ms |
| GPU / scroll | — | ~4–9% |

Also verified ~60fps scroll in Safari on real hardware.

## Screenshots / video

https://github.com/user-attachments/assets/d1ef5747-647a-4069-9d3b-72c4e38fa780

## Testing

- `pnpm check`, `pnpm lint`, `pnpm test` all green
- Manually verified Timeline + Event History tabs, detail panel, and grid-line rendering in both Chromium and Safari

## Notes

- Targets `main` (rebased/merged up from the `SharedHistoryBuffer` base). Not intended to merge onto `SharedHistoryBuffer`.


tegan-temporal:
Supercedes https://github.com/temporalio/ui/pull/3569

GiantRobots (approved):
LETS GOOOOOOOOO

PR #3628, opened 2026-07-02
fix(cloud-nav): smooth label fade on side-nav collapse/expand

bilal-karim:
Follow-up to #3606.

### Problem
The Cloud side-nav item labels used a plain symmetric `opacity` transition. On collapse they stayed opaque while the panel narrowed over them (a visible "chop"), and expand/collapse felt abrupt (see Standalone Activities and Namespace Settings as an example).

### Change
Direction-aware label motion, paired with the nav's existing 150ms `transition-width`, scoped to the **Cloud nav only** (OSS unchanged). The panel width moves continuously in both directions (no stall); the label transitions on the roomy side of the width change:

- **Collapse** — the label *leads*: a fast 75ms `ease-out` fade so it clears before the shrinking panel reaches it.
- **Expand** — the label *lags*: waits 75ms for the panel to open, then fades in linearly over 150ms.

| Before | After |
|---|---|
| <video src="https://github.com/user-attachments/assets/97698827-eead-4421-8e06-741225f7c828">  |  <video src="https://github.com/user-attachments/assets/d3483f66-59b1-4bb4-8b3c-02757052a5db">  |

### Implementation
- Adds a `data-cloud` attribute on the nav container — mirrors the existing `data-nav` pattern the labels already read via `group-data-[nav=…]`.
- Timing is a few Tailwind utilities on the label in `navigation-item.svelte`, each scoped to Cloud + nav state; the OSS nav resolves to its original opacity-only transition.
- No keyframes, plugin, or design-system changes.

Diff: **2 files, +2/−1**.

### Testing
Verified against the real cloud-ui CSS bundle:
- Cloud collapse: 75ms ease-out; label ≈ 0.04 opacity by the time the panel edge reaches it (no chop).
- Cloud expand: panel opens 0–150ms, label fades in 75–225ms.
- OSS nav: 150ms opacity-only fade, default easing — unchanged.


src/lib/holocene/navigation/navigation-container.svelte
```diff
@@ -35,14 +35,13 @@
 
 <nav
   class={merge(
-    'group grid min-h-full grid-cols-[2rem] grid-rows-[fit-content(1.5rem)] gap-2 border-r border-subtle px-2 py-4 transition-width data-[nav=open]:w-auto data-[nav=open]:grid-cols-[100%] motion-reduce:transition-none',
+    'group grid min-h-full grid-cols-[2rem] grid-rows-[fit-content(1.5rem)] gap-2 border-r border-subtle px-2 py-4 transition-width data-[nav=closed]:w-[calc(3rem_+_1px)] data-[nav=open]:w-[13.125rem] data-[nav=open]:grid-cols-[100%] motion-reduce:transition-none',
```
laurakwhit: @bilal-karim any objection to making the transition apply for OSS as well? 

https://github.com/user-attachments/assets/391a1c3b-63c1-466d-850f-6151436424bd

bilal-karim: Let's do it!

laurakwhit (approved):

PR #3629, opened 2026-07-02
Improve Search Attribute deletion messaging

laurakwhit:
## Description & motivation 💭 

- Added an `isExisting` prop when adding a row so button to remove row is a close icon instead of a trash icon for new attributes. 
- In Cloud the trash button is disabled and shows a tooltip.
- The `getSupportedTypes` prop was removed and the content component now defines the type list internally.
- Added an optional description snippet prop with a fallback `<p>` that now includes a "Search Attribute docs" Link.
- `hideCancelButton` prop removed since it was not being used.

| Default | Cloud | Cloud w/ delete enabled|
|-|-|-|
|Screenshot 2026-07-02 at 3 52 40 PM|Screenshot 2026-07-06 at 2 57 32 PM|Screenshot 2026-07-02 at 3 50 41 PM|

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

For an existing search attribute
  - [ ] type select should be disabled
  - [ ] if the feature flag is enabled there should be a disabled delete button
  - [ ] if the feature flag **not** enabled there should be no delete button

For a new search attbibute
  - [ ] type select should be enabled
  - [ ] there should be a close button

### Issue(s) closed 
`DT-4190`

### DIFF
--- a/src/lib/components/search-attributes-form/search-attribute-row.svelte
+++ b/src/lib/components/search-attributes-form/search-attribute-row.svelte
@@ -46,6 +46,8 @@
     submitting || (disableTypeForExisting && isExisting),
   );
 
+  const showButton = $derived(!isExisting || isDeletable);
+
   const hasError = $derived(!!error);
 
   const handleNameInput: FormEventHandler<HTMLInputElement> = (e) => {
@@ -53,7 +55,7 @@
   };
 </script>
 
-<div class="grid grid-cols-[1fr,200px,auto] items-center gap-3">
+<div class="grid grid-cols-[1fr,200px,2rem] items-center gap-3">
   <Input
     id={`attribute-name-${index}`}
     value={name}
@@ -66,21 +68,23 @@
     oninput={handleNameInput}
   />
 
-  <Select
-    id={`attribute-type-${index}`}
-    value={type}
-    label={translate('search-attributes.type-label', {
-      index: index + 1,
-    })}
-    labelHidden
-    disabled={isTypeDisabled}
-    placeholder={translate('search-attributes.select-type-placeholder')}
-    onChange={onTypeChange}
-  >
-    {#each supportedTypes as type (`${type.label}:${type.value}`)}
-      <Option value={type.value}>{type.label}</Option>
-    {/each}
-  </Select>
+  <div class:col-span-2={!showButton}>
+    <Select
+      id={`attribute-type-${index}`}
+      value={type}
+      label={translate('search-attributes.type-label', {
+        index: index + 1,
+      })}
+      labelHidden
+      disabled={isTypeDisabled}
+      placeholder={translate('search-attributes.select-type-placeholder')}
+      onChange={onTypeChange}
+    >
+      {#each supportedTypes as type (`${type.label}:${type.value}`)}
+        <Option value={type.value}>{type.label}</Option>
+      {/each}
+    </Select>
+  </div>
 
   {#if !isExisting}
     <Button
--- a/src/lib/components/search-attributes-form/search-attribute-row.svelte
+++ b/src/lib/components/search-attributes-form/search-attribute-row.svelte
@@ -5,6 +5,7 @@
   import Input from '$lib/holocene/input/input.svelte';
   import Option from '$lib/holocene/select/option.svelte';
   import Select from '$lib/holocene/select/select.svelte';
+  import Tooltip from '$lib/holocene/tooltip.svelte';
   import { translate } from '$lib/i18n/translate';
 
   import type { SearchAttributeTypeOption } from './types';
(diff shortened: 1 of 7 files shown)


bilal-karim:
- Can we remove the `book-sparkles` icon from the docs links?
- In **Default** / **Cloud with delete enabled**, the inputs aren't aligned with the Type column heading.
- Is there a reason we're showing "search attributes cannot be deleted" in the card description rather than just the delete button tooltip? Displaying it in the description when no attributes exist may be unnecessary. Could we only show it when search attributes >= 1? Or surface it as an informational alert banner instead of card description text?

src/lib/components/search-attributes-form/search-attributes-form-content.svelte
```diff
@@ -140,9 +173,19 @@
...
-        </p>
+        {#if description}
+          {@render description()}
+        {:else}
+          <p class="text-sm text-secondary">
+            {translate('search-attributes.description')}
+            <Link
+              href="https://docs.temporal.io/search-attribute"
+              newTab
+              trailingIcon="book"
```
laurakwhit: > Can we remove the book-sparkles icon from the docs links?

@bilal-karim I added these based on designs from @acav-temporalio.

> In Default / Cloud with delete enabled, the inputs aren't aligned with the Type column heading.

 Fixed ⬇️ 

Image

> Is there a reason we're showing "search attributes cannot be deleted" in the card description rather than just the delete button tooltip? Displaying it in the description when no attributes exist may be unnecessary. Could we only show it when search attributes >= 1? Or surface it as an informational alert banner instead of card description text?

Will defer to @acav-temporalio.

acav-temporalio:
@bilal-karim 

> * Can we remove the `book-sparkles` icon from the docs links?
I thought this was the preferred icon for linking to Docs from within the platform? (as opposed to an external link with the --> arrow.

> * Is there a reason we're showing "search attributes cannot be deleted" in the card description rather than just the delete button tooltip? Displaying it in the description when no attributes exist may be unnecessary. Could we only show it when search attributes >= 1? Or surface it as an informational alert banner instead of card description text?

A few reasons why the content is not in the tooltip only. The tooltip requires user interaction (onHover/onFocus) whereas the content displayed is read without user interaction both visually and via screenreader. Because we have been getting feedback about this functionality and our goal is trying to avoid scenarios where people are carelessly adding search attributes that now require temporal support intervention to delete, this content treatment is added specifically to make two things clear: 

1) search attributes cannot be deleted if they are in use, period
2) they cannot be deleted self-serve - contacting support is the only way to delete them right now. 

The content is both preemptive (double check your work before adding a search attribute! covers the "no search attributes yet" scenario) and reactive (once these are added, you have to contact support to delete them). We want the content to exist regardless of whether there are custom search attributes added yet because we want to encourage intentional CSA additions to avoid getting support involved for deleting accidental typos, etc.

bilal-karim:
@acav-temporalio Ok, fine by me re: copy. Also, we're not really using the book sparkles icon anymore for learn more links. Good to remove?

acav-temporalio:
> @acav-temporalio Ok, fine by me re: copy. Also, we're not really using the book sparkles icon anymore for learn more links. Good to remove?

Yup if that's no longer standard, we can just use a regular link. @laurakwhit do you mind making that update (no external link icon, just link).

everything else LGTM.

acav-temporalio (approved):
approved - LGTM. Thanks Laura!

src/lib/components/search-attributes-form/search-attribute-row.svelte
```diff
@@ -63,30 +68,48 @@
...
+    <Tooltip
+      text={translate('search-attributes.cloud-delete-tooltip')}
+      hide={!isCloud}
+      topRight
+    >
+      <Button
+        variant="ghost"
+        size="xs"
+        on:click={onRemove}
+        disabled={submitting || isCloud}
```
andrewzamojc, line 108: The new show/hide logic makes sense to me.

src/lib/components/search-attributes-form/search-attributes-form.stories.svelte
```diff
@@ -173,7 +140,6 @@
     onSave: handleSave,
     onSuccess: handleSuccess,
     onCancel: handleCancel,
-    getSupportedTypes,
```
andrewzamojc, line 176: I see that getSupportedTypes is no longer passed in, and so removed from this storybook file. What about where the form is actually used? Like a page somewhere?

andrewzamojc: i see now, in cloud-ui 👍

andrewzamojc (approved):
LGTM 🚀

PR #3636, opened 2026-07-07
Use an editable combobox for the Cloud Run region field (DT-4232)

rossnelson:
## What

Replaces the free-text `gcpRegion` input on the Create Worker Deployment form (Cloud Run provider) with an editable Holocene `Combobox`, and makes the suggestion list configurable.

- Suggests known GCP Cloud Run regions from a new `GCP_REGIONS` list.
- Allows a **custom region** not in the list via `allowCustomValue`, so an unsupported-by-us region never blocks deployment creation.
- Keeps the field visible and editable (no auto-hide, no forced default). The existing `gcp-region-hint` is preserved below the field.
- Exposes an optional `gcpRegions?: string[]` prop, threaded page -> form -> `compute-fields`, so cloud-ui can pass its own region list. When omitted, the built-in `GCP_REGIONS` default applies.

## Combobox fix

While wiring this up, found a latent bug in the shared Holocene `Combobox`: `addCustomValue` unconditionally cleared the input (`displayValue = ''`) after selecting. For single-select this blanked the field even though the value was stored, so a newly added custom value looked unselected. That clear is only meant for multiselect (reset the input for the next chip), so it is now guarded with `if (multiselect)`. The only other `allowCustomValue` consumer (`nexus-form`) is multiselect, so this path was never exercised before.

## Why

The worker's region is always a GCP (Cloud Run) region, while the Namespace region may be AWS or Azure, so defaulting to the Namespace region is not reliable. Our `GetRegions` list only covers regions we support for hosting a namespace, not all GCP regions a Cloud Run worker pool can use. This offers helpful suggestions without limiting the user.

## Notes

- Skipped the optional Namespace-region pre-fill: the ticket flags the Namespace region as an unreliable GCP default and the plumbing was not warranted.
- No schema change needed; `gcpRegion` stays a required non-empty string, which allows custom values.

## Testing

- `pnpm check` passes (0 errors)
- `pnpm lint` passes (0 errors)

Jira: DT-4232


src/lib/components/workers/serverless-worker-form/compute-fields.svelte
```diff
@@ -49,6 +52,7 @@
     roleExternalId = $bindable(),
     gcpProject = $bindable(''),
     gcpRegion = $bindable(''),
+    gcpRegions = [...GCP_REGIONS],
```
tegan-temporal, line 55: how come this is a copy? does it get mutated somewhere?

rossnelson: It does, the user can a region that is not shown in the list. since its a static list, it made sense to also allow freeform input.

tegan-temporal: since this is a prop and it gets mutated, should it be bindable? it could be I still am react-brained, but if I passed in a gcpRegions prop, I wouldn't expect it to get mutated, especially if it wasn't bindable 🤔 OR the `gcpRegions` prop could always be copied instead of only the default value?

tegan-temporal: oh, I don't think gcpRegions gets mutated, it is just a copy because of GCP_REGIONS is `as const` which is readonly and fails a typescript check for string[]

src/lib/components/workers/serverless-worker-form/compute-fields.svelte
```diff
@@ -129,16 +133,27 @@
...
+        placeholder={translate('workers.gcp-region-placeholder')}
+        options={gcpRegions}
+        noResultsText={translate('common.no-results')}
+        valid={!errors.gcpRegion?.[0]}
+        error={errors.gcpRegion?.[0]}
+        allowCustomValue
+        showChevron
+        required
+      />
+      {#if !errors.gcpRegion?.[0]}
```
tegan-temporal, line 151: We should probably make hintText part of Combobox so it can associate the hint text with with the form field via aria-desbribedby. right now the hint text is not associated for accessibility tech users.

src/lib/components/workers/serverless-worker-form/compute-fields.svelte
rossnelson: Ahh, you are right. i thought it was attempting to modify it inside the combobox component but it spreads it again. So yes. just to appease the typing.

tegan-temporal (approved):

PR #3640, opened 2026-07-08
TS strict (4/4): enable strict mode in base tsconfig

tegan-temporal:
Final of a 4-PR split (was the combined branch). Stacked on `strict/3-behavior` (#3688).

Enables and enforces strict mode now that PRs 1–3 resolve every error:
- Flip `strict: true` in `tsconfig.json`
- Switch the Danger strict-error check from **warn to fail** (`dangerfile.ts`) so any strict error in a changed file blocks the PR, with 1-indexed line numbers (`scripts/count-strict-errors.ts`)

The enforcement flip lives here, not in the base PR, so the earlier PRs can merge incrementally without prematurely blocking unrelated PRs while the ~690 errors are still being resolved across the stack.

---
**Stack (merge bottom-up):**
1. #3686 `strict/1-base` — type annotations + dead-code cleanup (zero runtime)
2. #3687 `strict/2-mechanical` — strictNullChecks null-safety guards
3. #3688 `strict/3-behavior` — runtime-affecting model/form/service changes
4. #3640 `ts-strict-mode` — flip `strict: true` + enforce in CI

Split verified: stacking all four reproduces the original combined tree byte-for-byte.

### DIFF
--- a/dangerfile.ts
+++ b/dangerfile.ts
@@ -104,10 +104,15 @@ async function checkStrictModeErrors() {
     const relevantErrors: Record<string, StrictError[]> = {};
     let relevantErrorCount = 0;
 
+    // svelte-check and Danger both report repo-relative POSIX paths; compare
+    // by normalized equality (substring matching both missed regressions and
+    // false-blocked unrelated files that share a path suffix).
+    const normalizePath = (p: string) => p.replace(/^\.?\//, '');
+
     for (const [filename, errors] of Object.entries(result.errorsByFile)) {
+      const normalizedFilename = normalizePath(filename);
       const isChanged = Array.from(changedFiles).some(
-        (changedFile) =>
-          changedFile.includes(filename) || filename.includes(changedFile),
+        (changedFile) => normalizePath(changedFile) === normalizedFilename,
       );
 
       if (isChanged) {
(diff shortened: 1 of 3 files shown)


src/lib/components/codec-server-form/codec-server-form-content.svelte
```diff
@@ -133,7 +133,6 @@
 
       <!-- Info Alert -->
       <Alert intent="info" class="text-sm">
-        <Icon name="info" slot="icon" />
```
tegan-temporal: Alert does not have an icon slot... this wasn't being rendered to begin with. It just became noticeable due to the strictness change

andrewzamojc:
Hey @tegan-temporal . I know lots of these change are not runtime, but since they're mixed with runtime changes it ends up a being a big PR that needs review. Any chance you want to ship all the stuff that doesn't need real review (tests/types etc) first, which can be insta-approved. Then a second PR could be the runtime changes and get a proper review (~30 files).

tegan-temporal:
Superseded by #3688. The strict: true flip can't be a standalone follow-up: the standalone-activity form's superForm types only type-check under strictNullChecks, so the behavior changes and the flag must merge together. Folded the tsconfig/danger changes into #3688 and closing this.

PR #3652, opened 2026-07-09
refactor: add es-toolkit

tegan-temporal:
## Summary

Consolidates several small third-party utility packages and hand-rolled helpers onto `es-toolkit`. Note: es-toolkit was already present in the tree as a transitive dependency (via `@storybook/addon-svelte-csf`, at 1.43.0); this PR promotes it to a direct `dependency` (`^1.49.0`) — it's imported by bundled/published frontend code, matching the classification of the `just-*`/`kebab-case` packages it replaces — and dedupes the version.

- Dropped `lodash`, `just-debounce`, `just-throttle`, `kebab-case`; added `es-toolkit`.
- `lodash/isEmpty` → `es-toolkit/compat` `isEmpty` (oidc-server dev util; server-only)
- `just-debounce` → es-toolkit `debounce` 
- `just-throttle` → es-toolkit `throttle`
- `kebab-case` → es-toolkit `kebabCase` (cleaner output; only affects a fallback `id` in `search.svelte`, no test depends on the old leading-dash format)
- Local `omit` → es-toolkit `omit` (array-key signature; 5 call sites)
- Search-attribute decode `reduce`-with-spread → `mapValues` (also removes an O(n²) accumulator spread) in `workflow-execution.ts` and `standalone-activity-workers.svelte`
- Deleted dead `unique` util and the now-unused local `omit` util (+ their tests

## Bundle impact
Full production build compared before/after: net-neutral — raw JS +253 B, gzipped −92 B. Tree-shaking pulls only the used functions.

## Testing
- `pnpm check` → 0 errors
- `pnpm test -- --run` → all passing
- `pnpm lint` clean


src/lib/services/events-service.ts
```diff
@@ -1,4 +1,4 @@
-import throttle from 'just-throttle';
+import { throttle } from 'es-toolkit';
```
laurakwhit, line 1: `just-throttle` fired leading-only ([it forces](https://github.com/angus-c/just#just-throttle) trailing false when leading is true). Looks like es-toolkit's `throttle` defaults to leading and trailing. Do we want to update `throttleRefresh` to keep the leading-only behavior?

src/lib/components/search.svelte
```diff
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import kebabCase from 'kebab-case';
+  import { kebabCase } from 'es-toolkit';
```
laurakwhit, line 2: This is existing, but wondering why we have `kebabCase` on the default id here when it just removes the dash 🤔

laurakwhit (approved):

src/lib/components/search.svelte
tegan-temporal: I thought it was for if label or name had dashes already. not sure why it needed to be kebab anyways though

src/lib/services/events-service.ts
tegan-temporal: yeah I noticed that. I decided not to force trailing to false because IMO it's more intuitive to be have trailing on and I doubt having trailing false was intentional

tegan-temporal: I final refresh makes sense to me if we had invoked a refresh after the last one but before the throttle delay

PR #3654, opened 2026-07-10
a11y(2.4.3): keep toast live region announcing during a focus trap [DT-4252]

ardiewen:
Follow-up to #3598, from Laura's review. The inert-based trap inerts everything outside it up to `<body>`; the toast `aria-live` region lived inside the main app container, so toasts fired while a drawer/maximizable was open were **not announced** to screen readers (still shown visually — an AT-only gap).

**Fix:** hoist the live region to `<body>` from within `Toaster` (`data-inert-skip`), and have `inertBackground` skip live-region roots — matched on the element itself, so a container merely *holding* a live region (the app shell) is still inerted and containment is preserved.

Stacked on `wcag/2.4.3-focus-trap-inert` (#3598); retarget to `main` once it merges. cloud-ui picks this up on the next tarball repack — verify both its toaster mounts then (see DT-4252).

## Testing
- 3 new unit tests in `focus-trap.test.ts`: skips `data-inert-skip` / `aria-live` roots; still inerts a container with a *nested* live region.
- `pnpm check` clean, focus-trap suite green.
- [x] NVDA/VoiceOver: toast announced with a drawer/maximized surface open.

### DIFF
--- a/src/lib/utilities/focus-trap.test.ts
+++ b/src/lib/utilities/focus-trap.test.ts
@@ -141,7 +141,6 @@ describe('focusTrap focus management (inert-based)', () => {
   it('does not inert a background sibling marked data-inert-skip (hoisted live region)', () => {
     const liveRegion = document.createElement('div');
     liveRegion.setAttribute('data-inert-skip', '');
-    liveRegion.setAttribute('aria-live', 'polite');
     const background = document.createElement('div');
     background.appendChild(document.createElement('button'));
     const trapNode = document.createElement('div');
--- a/src/lib/utilities/focus-trap.test.ts
+++ b/src/lib/utilities/focus-trap.test.ts
@@ -138,6 +138,57 @@ describe('focusTrap focus management (inert-based)', () => {
     expect(document.activeElement).toBe(trigger);
   });
 
+  it('does not inert a background sibling marked data-inert-skip (hoisted live region)', () => {
+    const liveRegion = document.createElement('div');
+    liveRegion.setAttribute('data-inert-skip', '');
+    liveRegion.setAttribute('aria-live', 'polite');
+    const background = document.createElement('div');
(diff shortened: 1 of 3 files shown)


src/lib/utilities/focus-trap.test.ts
```diff
@@ -138,6 +138,57 @@ describe('focusTrap focus management (inert-based)', () => {
     expect(document.activeElement).toBe(trigger);
   });
 
+  it('does not inert a background sibling marked data-inert-skip (hoisted live region)', () => {
+    const liveRegion = document.createElement('div');
+    liveRegion.setAttribute('data-inert-skip', '');
+    liveRegion.setAttribute('aria-live', 'polite');
```
laurakwhit: Should probably just test the `data-inert-skip` here.

```suggestion
```

laurakwhit (approved):

ardiewen:
Downstream repack: **temporalio/cloud-ui#2929** (`release-ui`) — this change is the tip commit (`cc491a29`) of that pack.

**What needs to happen on the cloud-ui side (on #2929):**

1. **Verify announcements at both toaster mounts** — with a drawer/maximized surface open, confirm a toast is announced (NVDA/VoiceOver):
   - `src/routes/(app)/+layout.svelte`
   - `src/routes/(app)/get-started/+layout@.svelte`
2. **Check Playwright selectors** — the toast `LiveRegion` is now portaled to `<body>`, so integration tests that scope to a page container when querying `getByRole('status'|'alert'|'log')` or the `toast-live-region-*` testids may need updating (same pattern as #3560 → cloud-ui#2832). Expect a fix commit on the repack PR if those break.
3. **Out of scope:** native `<dialog>` modals still inert the document (incl. the hoisted live region) — a separate, pre-existing gap tracked in DT-4252, not addressed by this repack.

PR #3655, opened 2026-07-10
fix(schedules): default emptied second/minute/hour to 0 on submit

tegan-temporal:
Follow-up to #3653.

## Problem
That PR fixed the edit-load path, but the same "empty range → server matches nothing → no future runs" hazard exists on **create**: the zod schema only defaults second/minute/hour when they're `undefined`, so an emptied calendar time field (e.g. a cleared picker) slips through as `[]`.

## Fix
Guard the single submit choke point (`toStructuredCalendar`) so an empty second/minute/hour range defaults to `0`. Covers create and edit. Test added.

### DIFF
--- a/src/lib/components/schedule/utilities/get-request-body.ts
+++ b/src/lib/components/schedule/utilities/get-request-body.ts
@@ -154,9 +154,9 @@ function toStructuredCalendar(
   const { comment, ...ranges } = calendar;
   return {
     ...ranges,
-    second: ranges.second?.length ? ranges.second : DEFAULT_TIME_RANGE,
-    minute: ranges.minute?.length ? ranges.minute : DEFAULT_TIME_RANGE,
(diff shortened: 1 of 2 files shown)


src/lib/components/schedule/utilities/get-request-body.ts
```diff
@@ -143,11 +143,22 @@ async function getScheduleActionRequest(
 
 type SpecFormItem = FormScheduleSchema['specs'][number];
 
+const DEFAULT_TIME_RANGE = [{ start: 0, end: 0, step: 1 }];
```
laurakwhit: NIT: so it can't be mutated at any point. 
```suggestion
const defaultTimeRange = () => [{ start: 0, end: 0, step: 1 }];
```

laurakwhit (approved):

PR #3656, opened 2026-07-10
fix(nexus-operations): make start standalone nexus operation timeouts optional

rossedfort:
## Description & motivation 💭

The Start Standalone Nexus Operation form previously required that at least one of **Start to Close** or **Schedule to Close** timeout be set before the form could be submitted. This validation was too restrictive — all three timeout fields (Start to Close, Schedule to Close, Schedule to Start) should be optional, as the server can apply defaults.

Removed the Zod `superRefine` validation block that enforced the timeout requirement, the corresponding error display in the form, and the now-unused `form-timeout-required` i18n key.

N/A

### Design Considerations 🎨

None.

### How was this tested 👻

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

1. Navigate to the Start Standalone Nexus Operation form
2. Fill in the required fields (Endpoint, Service, Operation Name)
3. Leave all timeout fields empty
4. Submit the form — it should succeed without any timeout validation errors

### Any docs updates needed?

No docs updates needed.


src/lib/components/standalone-nexus-operations/start-standalone-nexus-operation-form/form.svelte
laurakwhit: Will there ever be a `timeoutError` here after the removal of the `superRefine`?

tests/integration/start-standalone-nexus-operation.desktop.spec.ts
```diff
@@ -64,51 +64,5 @@ test.describe('Start a Standalone Nexus Operation', () => {
...
-    await startNexusOperationPage.serviceInput.fill('my-service');
-    await startNexusOperationPage.operationNameInput.fill('my-operation');
-
-    await startNexusOperationPage.editPoliciesButton.click();
-    await startNexusOperationPage.startToCloseTimeoutInput.fill('0');
-
-    // Close the (now modal) policies drawer before submitting — values bind
-    // live to the form, so closing commits nothing but frees the main form
-    // from the drawer's focus trap so the submit button is interactive.
-    await startNexusOperationPage.updatePoliciesButton.click();
```
laurakwhit, line 87: Looks like we're no longer using `updatePoliciesButton`, should we remove the locator?

src/lib/components/standalone-nexus-operations/start-standalone-nexus-operation-form/form.svelte
```diff
@@ -81,59 +81,35 @@
...
-      idConflictPolicy: z.string().optional(),
-      summary: z.string().optional(),
-      details: z.string().optional(),
-      nexusHeader: z
-        .array(z.object({ key: z.string(), value: z.string() }))
-        .default([]),
-    })
-    .superRefine((data, context) => {
-      if (
-        !isPositiveDuration(data.startToCloseTimeout) &&
```
laurakwhit, line 118: If we're no longer using `isPositiveDuration` does that mean `0s` is a valid duration for these inputs?

src/lib/components/standalone-nexus-operations/start-standalone-nexus-operation-form/form.svelte
rossedfort: Nope, fixed.

src/lib/components/standalone-nexus-operations/start-standalone-nexus-operation-form/form.svelte
rossedfort: Based on the [protobuf doc comments](https://github.com/temporalio/api/blob/bd41196044da1791481b552616e8a8cec28b506f/temporal/api/workflowservice/v1/request_response.proto#L3355), yes, `0s` is treated the same as unset (which defaults to no timeout).

laurakwhit (approved):

PR #3668, opened 2026-07-13
Show IANA region names in schedule timezone picker (DT-4250)

tegan-temporal:
## What
Schedule form timezone picker now lists IANA region names (e.g. `America/New_York`) with their year offset range instead of descriptive names.

- Options built from `Intl.supportedValuesOf('timeZone')`; label + value are the region, so it's searchable by name (the combobox filters on label).
- New `getTimezoneOffsetRange` util shows the full-year offset range in UTC notation (e.g. `America/New_York (UTC-05:00/UTC-04:00)`) via `Intl` `longOffset` — DST-aware, preserves half-hour/45-min zones. Avoids the buggy `formatOffset`.

**Why this PR avoids `formatOffset`:** it only handles whole-hour offsets. The `:00` is hardcoded, so sub-hour zones render wrong (India `+05:30` → `+05:00`, Nepal `+05:45` → `+05:00`), and `Math.floor` pushes negative fractional offsets the wrong way (Marquesas `-09:30` → `-10:00`). `getTimezoneOffsetRange` uses `Intl` `longOffset`, which is accurate for all of these.

**Why a new function instead of fixing `formatOffset`:** `formatOffset`/`getUTCOffset` are shared by the nav timezone selector, workflow filters, and time-format store. Changing their output format could have far-reaching effects across those surfaces, which is out of scope for the schedules timezone selector. Fixing it repo-wide is worth a separate follow-up.

## Why
Customers see `America/New_York` on existing schedules but couldn't find it in the picker (grouped under "Eastern Standard Time" and not searchable by value).

## Test
Unit tests for `getTimezoneOffsetRange`; `pnpm check`/`lint`/`test` all pass.

CleanShot 2026-07-13 at 14 47 24@2x

### DIFF
--- a/src/lib/components/schedule/schedule-form/schedule-details-card.svelte
+++ b/src/lib/components/schedule/schedule-form/schedule-details-card.svelte
@@ -12,7 +12,7 @@
   import Tooltip from '$lib/holocene/tooltip.svelte';
   import { translate } from '$lib/i18n/translate';
   import type { DescribeFullSchedule } from '$lib/types/schedule';
-  import { getTimezoneOffsetRange } from '$lib/utilities/timezone';
+  import { ianaTimezoneComboboxOptions } from '$lib/utilities/timezone';
 
   import type { FormScheduleSchema } from '../schema/form';
   import { fromCalendarDateStr, toCalendarDateStr } from '../utilities/date';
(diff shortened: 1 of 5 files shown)


src/lib/components/schedule/schedule-form/schedule-details-card.svelte
```diff
@@ -29,11 +29,10 @@
 
   const timezoneComboboxOptions = $derived.by(() => {
```
laurakwhit: The zone options appear to be static. Can we move this out of `$derived` and put it in module scope? Maybe even in `src/utilities/timezone.ts`?

tegan-temporal: sure, I made it an exported constant called `ianaTimezoneComboboxOptions`

src/lib/components/schedule/schedule-form/schedule-details-card.svelte
```diff
@@ -198,7 +187,7 @@
         id="timezoneName"
```
laurakwhit, line 187: NIT: Anyway we can make these line up visually?

Image

laurakwhit (approved):

src/lib/components/schedule/schedule-form/schedule-details-card.svelte
tegan-temporal: I made the input gap 1.5 and it has them aligned:

Screenshot 2026-07-14 at 3 05 44 PM

PR #3674, opened 2026-07-14
[COM-193] Lock compute provider for new deployment versions

rossnelson:
## Summary

- Resolve the compute provider already used by an existing Worker Deployment before rendering the create-version form.
- Initialize new versions with that provider and hide every alternative so a deployment cannot move providers between versions.
- Preserve the existing configurable/default provider behavior for deployments without versions.
- Fail closed when deployment loading fails or the existing provider is missing, unknown, mixed, or excluded by configuration.
- Expose `canCreateServerlessDeployment` so consumers can hide the empty-state serverless create action when no provider is supported, while keeping the self-managed deployment link visible.

## Test plan

- [x] Focused provider-lock tests passed during implementation (25/25).
- [x] Pre-commit ESLint, Prettier, and Stylelint hooks passed for both commits.
- [x] `git diff --check` passed.
- [x] Static architect review approved both changes.
- [ ] CI checks pending. Per request, local test workers were not restarted after the final static-only adjustments.

### DIFF
--- a/src/lib/components/deployments/deployments-empty-state.svelte
+++ b/src/lib/components/deployments/deployments-empty-state.svelte
@@ -8,9 +8,14 @@
   interface Props {
     createHref: string;
     error?: string;
+    canCreateServerlessDeployment?: boolean;
   }
 
-  let { createHref, error = '' }: Props = $props();
+  let {
+    createHref,
+    error = '',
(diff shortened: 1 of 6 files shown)


andrewzamojc (approved):
LGTM 👍

PR #3685, opened 2026-07-17
feat: Add resizable columns to the workflows table.

tristal:
## Description & motivation 💭
The workflows table has no way to control column widths. Columns autosize to their content, and the only lever is the dense/comfortable toggle.

This adds drag-to-resize on the workflows table header:

- Drag the divider at the right edge of any column header to set its width.
- Widths persist per-namespace in the existing column config (`localStorage`), alongside label/order/pinned.
- Double-click a divider to return that column to autosizing. A reset button appears next to the density toggle once any column is resized, to clear all widths.
- Keyboard accessible: tab to a divider, arrow keys resize in 16px steps, Enter/Space resets. Minimum width 80px.

2-widened
3-shrunk

### Design Considerations 🎨 
- Dividers are invisible at rest so the table looks unchanged; hovering the header row reveals them all as subtle 1px lines (discoverability), and the divider under the cursor thickens to 2px in the accent color.

### How was this tested 👻

- [X] Manual testing
- [ ] E2E tests added
- [X] Unit tests added


bilal-karim:
This is a great addition. Thanks for creating this PR. I did add a few more updates:
- Added `role="separator"` with live `aria-valuenow`/`aria-valuemin`/`aria-valuetext` so the resize width is announced to Screen Readers.
- `Esc` on keyboard now cancels an in-progress drag and reverts to the pre-drag width.
- Focus ring switched to the app-standard `ring-2 ring-primary`, using `ring-inset` to avoid clipping in the sticky header.
- Changed the hover divider color from `bg-subtle` to `border-secondary` for better contrast.
- Divider is now a fixed 16px that grows to 24px on hover/focus/drag with an eased transition for better hover interaction.

PR #3686, opened 2026-07-17
TS strict (1/4): type annotations, dead-code cleanup, strict-error CI infra

tegan-temporal:
First of a 4-PR split of #3640 (enable TypeScript strict mode).

**Zero runtime impact, no CI/enforcement changes.** Contains only: noImplicitAny-driven type annotations, `import type` conversions, Storybook/test typing, and dead-code removal. 36 files.

Global `strict` stays off and the Danger strict-error check stays in warn mode until the final PR — so merging this early blocks nothing. All files verified strict-clean.

---
**Stack (merge bottom-up):**
1. #3686 `strict/1-base` — type annotations + dead-code cleanup (zero runtime)
2. #3687 `strict/2-mechanical` — strictNullChecks null-safety guards
3. #3688 `strict/3-behavior` — runtime-affecting model/form/service changes
4. #3640 `ts-strict-mode` — flip `strict: true` + enforce in CI

Split verified: stacking all four reproduces the original combined tree byte-for-byte.

### DIFF
--- a/src/lib/components/codec-server-form/codec-server-form-content.svelte
+++ b/src/lib/components/codec-server-form/codec-server-form-content.svelte
@@ -133,7 +133,6 @@
 
       <!-- Info Alert -->
       <Alert intent="info" class="text-sm">
-        <Icon name="info" slot="icon" />
         {translate('codec-server.info-message')}
       </Alert>
 
--- a/src/lib/models/event-groups/get-group-id.test.ts
+++ b/src/lib/models/event-groups/get-group-id.test.ts
@@ -6,7 +6,10 @@ import { getGroupId } from './get-group-id';
 
 import allEventTypesFixture from '$fixtures/all-event-types.json';
 
-const events = allEventTypesFixture as unknown as CommonHistoryEvent[];
+const events = allEventTypesFixture as unknown as Record<
+  string,
+  CommonHistoryEvent
+>;
 
 describe('getGroupId', () => {
   it('should get the correct ID for WorkflowExecutionCanceled events', () => {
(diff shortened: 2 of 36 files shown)


src/lib/components/codec-server-form/codec-server-form-content.svelte
```diff
@@ -133,7 +133,6 @@
 
       <!-- Info Alert -->
       <Alert intent="info" class="text-sm">
-        <Icon name="info" slot="icon" />
```
tegan-temporal, line 136: this slot does not exist in Alert and apparently never did??? the Icon is chosen from the Alert `icon` prop which defaults to `intent` prop

src/lib/models/event-groups/get-group-id.test.ts
```diff
@@ -6,7 +6,10 @@ import { getGroupId } from './get-group-id';
 
 import allEventTypesFixture from '$fixtures/all-event-types.json';
 
-const events = allEventTypesFixture as unknown as CommonHistoryEvent[];
+const events = allEventTypesFixture as unknown as Record<
```
tegan-temporal, line 9: This is the accurate shape, it's not an array

andrewzamojc (approved):
LGTM. All safe stuff 👍

src/lib/utilities/decode-payload-test-fixtures.ts
```diff
@@ -1,4 +1,6 @@
-export const workflowStartedEvent = {
+import type { PotentiallyDecodable } from './decode-payload';
```
tegan-temporal, line 1: Review this file with by hiding whitespace changes, the diff is weird due to the whitespace change

PR #3687, opened 2026-07-17
TS strict (2/4): strictNullChecks null-safety guards

tegan-temporal:
Second of a 4-PR split of #3640. Stacked on `strict/1-base`.

Mechanical null-safety introduced by strictNullChecks: nullish-coalescing defaults (`?? ''`, `?? 0`, `?? []`), optional chaining, `{#if}` render guards, and a few generics-driven refactors. Intended to be behavior-preserving — no server-request or form-default semantics change (those are in 3/4). 189 files, verified strict-clean against the base PR.

---
**Stack (merge bottom-up):**
1. `strict/1-base` — type annotations + CI infra (zero runtime)
2. `strict/2-mechanical` — strictNullChecks null-safety guards
3. `strict/3-behavior` — runtime-affecting model/form/service changes
4. `ts-strict-mode` (#3640) — flip `strict: true`

Original combined branch preserved in #3640, now reduced to the final flag flip. Split verified: stacking all four reproduces the original tree byte-for-byte.


andrewzamojc:
Because this PR doesn't target main, I don't think TS and such ran. How about merging the base (1/4) to main since it's safe and ready, then this one could target main and the TS checks will run again.

src/lib/components/schedule/schedule-form/schedule-spec-card.svelte
```diff
@@ -48,7 +49,7 @@
       return;
     }
 
-    $form.specs = [...$form.specs, { kind: 'cron', cronString: '' }];
+    $form.specs = [...$form.specs, getFormSpecInitialData('cron')];
```
tegan-temporal, line 52: this change was necessary because `{ kind: 'cron', cronString: '' }` is incomplete, getFormSpecInitialData gives us the full shape.

andrewzamojc (approved):
LGTM!

PR #3688, opened 2026-07-17
TS strict (3/3): runtime model/form/service changes + enable strict mode

tegan-temporal:
Final PR of the strict-mode split (base #3686 and mechanical #3687 already merged).

Contains the genuine runtime-affecting changes **and** the `strict: true` flip, because they're coupled: the standalone-activity form's superForm types only resolve correctly under `strictNullChecks`, so this code can't type-check on a `strict: false` main — the flag has to land together with it. (Supersedes the separate flag PR #3640, now closed.)

- WorkflowExecution optional fields (+ snapshot tests), schedule-form zod default seeding, standalone-activity retry fields as strings coerced at the boundary, duration-input defaults, requestFromAPI's `Promise<T | undefined>`
- Flip `strict: true` in `tsconfig.json`
- Switch the Danger strict-error check from warn to fail (1-indexed line numbers)

Verified: `svelte-check` reports 0 errors under strict on this branch.


andrewzamojc (approved):
Awesome. I was able to go through each runtime change and see why it's safe or that its covered by a test etc. Good stuff 👍

PR #3689, opened 2026-07-17
 Fix live timeline refresh for completed event groups

GiantRobots:
Refresh the workflow timeline on all grouped-event buffer changes so live completion events update existing rows, not just newly-created group heads. Merge live groups back into the history-backed pool when the bidirectional fetch later
 claims their head, clear stale pending activity/nexus metadata when live events resolve a group, and keep grouped-array
 caches fresh after follower mutations.
 Add coverage for history/live split ordering across activity, timer, child workflow, workflow task, update, and Nexus
 groups, including far-ahead live followers and live-only pending metadata.

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### DIFF
--- a/src/lib/layouts/workflow-history-layout.svelte
+++ b/src/lib/layouts/workflow-history-layout.svelte
@@ -27,11 +27,10 @@
     getWorkflowTaskFailedEvent as getBufferWftFailedEvent,
     getEventArray,
     getGroupArray,
-    onLatestGroup,
   } from '$lib/services/grouped-event-buffer';
   import { clearActives } from '$lib/stores/active-events';
   import { eventFilterSort, eventViewType } from '$lib/stores/event-view';
-  import { pauseLiveUpdates } from '$lib/stores/events';
+  import { bufferVersion, pauseLiveUpdates } from '$lib/stores/events';
   import { eventCategoryFilter, eventTypeFilter } from '$lib/stores/filters';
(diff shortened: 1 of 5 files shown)


tegan-temporal:
Small suggestion on the rAF throttle in `workflow-timeline-layout.svelte`: it could live entirely inside the `$bufferVersion` effect rather than as component-level state (`groupUpdatePending`, `groupUpdateFrame`, `scheduleBufferGroupRefresh`).

The effect's cleanup runs before each re-run, so a `$bufferVersion` bump cancels the previously-armed frame and schedules a fresh one — same "at most one `getGroupArray()` per frame" coalescing, without the `groupUpdatePending` guard. It also lets the `onMount` teardown go away, since the effect cancels its own frame on unmount.

```svelte
$effect(() => {
  const _bufferVersion = $bufferVersion;

  let frame: number | null = requestAnimationFrame(() => {
    frame = null;
    bufferGroups = getGroupArray({ excludeWorkflowTasks: true });
  });

  return () => {
    if (frame !== null) cancelAnimationFrame(frame);
  };
});
```

tegan-temporal:
I believe workflow-history-layout `workflowTaskFailedError` needs to be reactive on $bufferVersion` now as well, otherwise errors might not surface?

tegan-temporal (approved):

PR #3693, opened 2026-07-21
fix(events): reflect pending-activity commands without reload

tegan-temporal:
Pausing a pending activity from the event/history view didn't reflect until a full reload: `enrichGroup` mutated pending state in place, and since pause appends no history event, the stable group reference never triggered a re-derive.

**Fix:** `enrichGroup` returns a fresh group ref when pending metadata changes. `EventGroup` accessors move to a shared prototype so `createGroupFor` and the new `cloneEventGroup` produce one V8 shape (no megamorphic slowdown in the timeline).

**Tests:** Playwright regression (verified failing without the ref swap) + unit coverage for swap-on-change / reuse-when-unchanged.

Also adds `pnpm preview:workflows` for manual verification.

https://github.com/user-attachments/assets/dba17a5f-feb4-4646-9868-933b5ef640e8

https://github.com/user-attachments/assets/0b5b7404-26f7-4746-a9c5-056e70058f0d

### DIFF
--- a/src/lib/services/grouped-event-buffer.ts
+++ b/src/lib/services/grouped-event-buffer.ts
@@ -1,5 +1,6 @@
 import {
   addEventToGroup,
+  cloneEventGroup,
   createEventGroup,
   createWorkflowTaskGroup,
 } from '$lib/models/event-groups/create-event-group';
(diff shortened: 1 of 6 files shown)


src/lib/services/grouped-event-buffer.ts
```diff
@@ -344,33 +345,42 @@ function absorbLiveGroupIntoPool(poolIdx: number, liveGroup: EventGroup): void {
...
   if (isActivityTaskScheduledEvent(initial)) {
     const pa = byActivityId.get(
       initial.activityTaskScheduledEventAttributes?.activityId ?? '',
     );
-    if (pa && !hasTerminalActivityEvent(group)) {
-      group.pendingActivity = pa;
-    } else {
-      delete group.pendingActivity;
+    const next = pa && !hasTerminalActivityEvent(group) ? pa : undefined;
+    if (group.pendingActivity === next) {
```
laurakwhit, line 364: It looks like `toPendingActivities/toPendingNexusOperations` build fresh objects every fetch. Do we want some kind of `samePendingActivity()` check that compares the fields here instead?

src/lib/services/grouped-event-buffer.ts
laurakwhit, line 262: Looks like `clearResolvedPendingState` clears `pendingActivity/pendingNexusOperation` with an in-place delete. Do we want to do something similar here as you did for `enrichGroup` to prevent similar behavior? If so, maybe worth a follow-up PR.

laurakwhit (approved):

tegan-temporal:
latest commit fixes the legend appearing underneath event details 
Screenshot 2026-07-22 at 9 55 46 AM

PR #3700, opened 2026-07-23
[DT-4296] Add Terraform setup instructions to the GCP Cloud Run deployment form

rossnelson:
## Summary

- Add expandable Terraform setup guidance to the shared GCP Cloud Run worker deployment fields.
- Provide a copyable module example that reactively prefills the GCP project and clearly marks the required impersonator identity placeholder.
- Preserve the existing Lambda guidance while covering create, create-version, and edit flows through the shared component.

## Jira

https://temporalio.atlassian.net/browse/DT-4296

## Test plan

- [x] Focused Cloud Run component tests pass: rendering, module link, copy/reactivity, and Lambda isolation.
- [x] pnpm check passes with zero errors.
- [x] Targeted Prettier, ESLint, and Stylelint checks pass.
- [ ] Full Vitest suite not run locally because its unbounded worker pool exhausted local memory; focused tests passed with a single fork.

### DIFF
--- a/src/lib/components/workers/serverless-worker-form/compute-fields.svelte
+++ b/src/lib/components/workers/serverless-worker-form/compute-fields.svelte
@@ -10,9 +10,13 @@
   import { translate } from '$lib/i18n/translate';
 
   import { GCP_REGIONS } from './gcp-regions';
+  import cloudRunTerraformTemplate from './serverless-worker-cloud-run.tf?raw';
   import defaultTerraformTemplate from './serverless-worker-lambda.tf?raw';
   import { interpolateTerraformTemplate } from './shared';
   import cfnTemplate from './temporal-worker-role.yaml?raw';
+
+  const CLOUD_RUN_PROJECT_TOKEN = '"__TEMPORAL_GCP_PROJECT_ID__"';
+  const CLOUD_RUN_PROJECT_PLACEHOLDER = '<YOUR-GCP-PROJECT-ID>';
 
   interface Props {
     provider?: string;
@@ -88,6 +92,12 @@
       },
     ),
   );
+  const resolvedCloudRunTerraformTemplate = $derived(
+    cloudRunTerraformTemplate.replace(
+      CLOUD_RUN_PROJECT_TOKEN,
+      JSON.stringify(gcpProject || CLOUD_RUN_PROJECT_PLACEHOLDER),
+    ),
+  );
 
   const launchStackHref = $derived.by(() => {
     if (!cfnTemplateUrl) {
@@ -106,6 +116,7 @@
   });
 
   let showRoleHelp = $state(false);
+  let showCloudRunHelp = $state(false);
   let showScaling = $state(false);
   let activeRoleHelpTab = $state<'cloudformation' | 'terraform'>(
     'cloudformation',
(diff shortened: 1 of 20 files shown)


vitest.config.ts
```diff
@@ -7,6 +7,7 @@ import { configDefaults } from 'vitest/config';
 export default defineConfig({
   plugins: [svelte({ hot: false })],
   resolve: {
+    conditions: ['browser'],
```
laurakwhit: To clarify, was the intention here to replace [the default ](https://vite.dev/config/shared-options#resolve-conditions)`['module', 'browser', 'development|production']` here with just `['browser']`?

src/lib/components/workers/serverless-worker-form/compute-fields.svelte
```diff
@@ -74,10 +79,17 @@
     cfnTemplateUrl,
     cfnTemplate: cfnTemplateProp,
     terraformTemplate,
+    cloudRunTerraformTemplate,
     errors = {},
   }: Props = $props();
 
   const resolvedCfnTemplate = $derived(cfnTemplateProp ?? cfnTemplate);
+  const resolvedCloudRunTerraformTemplate = $derived(
+    (cloudRunTerraformTemplate ?? defaultCloudRunTerraformTemplate).replace(
```
laurakwhit: Do we want to replace only the first occurrence here or should we use a `replaceAll`?

src/lib/components/workers/serverless-worker-form/compute-fields.svelte
```diff
@@ -295,17 +308,53 @@
...
+              >{translate('workers.cloud-run-terraform-module-link')}</Link
+            >{translate('workers.cloud-run-terraform-description-after')}
+          </p>
+          {#if cloudRunTerraformTemplate === undefined}
+            <p class="text-sm text-warning">
+              {translate('workers.cloud-run-impersonator-warning')}
+            </p>
+          {/if}
+          <CodeBlock
+            content={resolvedCloudRunTerraformTemplate}
```
laurakwhit, line 355: Do we want to pass a label here similar to sibling Lambda block?

src/lib/components/workers/serverless-worker-form/compute-fields.svelte
```diff
@@ -295,17 +308,53 @@
...
+      >
+        <div class="-mt-8 flex flex-col gap-3 border-t border-subtle pt-3">
+          <p class="text-sm text-secondary">
+            {translate('workers.cloud-run-terraform-description-before')}<Link
+              href="https://github.com/temporalio/terraform-modules/tree/main/modules/serverless-workers/gcp/cloud-run"
+              newTab
+              >{translate('workers.cloud-run-terraform-module-link')}</Link
+            >{translate('workers.cloud-run-terraform-description-after')}
+          </p>
+          {#if cloudRunTerraformTemplate === undefined}
```
laurakwhit: Instead of checking for `undefined`, should this be checking that it doesn't include the default placeholder value?

vitest.config.ts
rossnelson: oops, that should not be in there. Ill remove it.

laurakwhit (approved):

PR #3707, opened 2026-07-23
Update holocene copyable/button to runes syntax

tegan-temporal:
Migrates `copyable/button.svelte` to Svelte 5 runes. Click now forwards through `{...rest}` (`Props extends HTMLButtonAttributes`) instead of the legacy `on:click` directive, so all 5 consumer usages (across `detail-list-value`, `workflow-run-layout`, `code-block`, and `copyable/index`) switch `on:click` → `onclick`. No cloud-ui consumers.


rossedfort (approved):
Just a reminder we'll want to track down if there's any usage of this component in cloud and update the `onclick` prop accordingly.

PR #3713, opened 2026-07-23
Update holocene range-input to runes syntax

tegan-temporal:
Migrates `input/range-input.svelte` to Svelte 5 runes: `$props()` (`value` is `$bindable()`), `$:` → `$derived` (the DOM-measurement offset is a reassignable `$derived` so the window-resize handler can force a recompute), `on:input`/`on:resize` → `oninput`/`onresize`, `$$restProps` → `...rest`.

Single ui consumer (`workflow-json-navigator`) uses `bind:value` — kept working via `$bindable()`, so no consumer change. No cloud-ui consumers.

### DIFF
--- a/src/lib/holocene/input/range-input.svelte
+++ b/src/lib/holocene/input/range-input.svelte
@@ -1,54 +1,34 @@
 <script lang="ts">
   import type { HTMLInputAttributes } from 'svelte/elements';
 
-  import { omit } from 'es-toolkit';
-
   import Label from '$lib/holocene/label.svelte';
 
-  interface $$Props extends HTMLInputAttributes {
-    value: number;
+  interface Props extends HTMLInputAttributes {
+    value?: number;
     id: string;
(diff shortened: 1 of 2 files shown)


src/lib/holocene/input/range-input.svelte
```diff
@@ -103,11 +99,11 @@
           type="range"
           class="h-0 w-full cursor-pointer appearance-none rounded border-y border-primary"
           bind:value
-          on:input={handleInput}
+          oninput={handleInput}
           {min}
           {max}
           {step}
-          {...omit($$restProps, ['class'])}
+          {...rest}
```
rossnelson, line 113: This spreads `...rest` after `oninput={handleInput}`. A consumer-provided `oninput` replaces internal validation. If not intentional, destructure and compose the callback, or spread props before the internal handler.

rossnelson (approved):

PR #3714, opened 2026-07-23
Update holocene accordion-light to runes syntax

tegan-temporal:
Migrates `accordion/accordion-light.svelte` to Svelte 5 runes: `$props()` (`open` is `$bindable()`), `on:click` on the native trigger button → `onclick`, named slots `title`/`description`/`action` → snippet props, and the scoped default slot `<slot {open}>` → `children` snippet (`Snippet<[boolean]>`). `title`/`children` are `Omit`ted from the extended `HTMLAttributes` since they'd collide with the DOM `title` string / base `children`.

Single consumer (`workflow-callbacks`) converted: `slot="title"` → `{#snippet title()}`, `let:open` default slot → `{#snippet children(open)}`. No cloud-ui consumers.

### DIFF
--- a/src/lib/components/workflow/workflow-callbacks.svelte
+++ b/src/lib/components/workflow/workflow-callbacks.svelte
@@ -15,17 +15,21 @@
 <WorkflowCallback callback={firstCallback} {link}>
   {#if callbacks.length > 1}
     <div class="py-2">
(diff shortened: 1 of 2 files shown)


rossnelson (approved):

PR #3718, opened 2026-07-24
Fix deployment connection validation status refresh

rossnelson:
## Summary

- validate the persisted deployment compute configuration by sending an empty change set from the manual Validate Connection action
- keep candidate compute-config validation unchanged for create and edit flows
- refresh deployment data once after a resolved validation without remounting the row or dismissing the result modal
- ignore stale refresh results when the route or active request changes

## Why

The validation API supports two distinct modes. Supplying scaling groups validates a candidate configuration as a dry run and does not update persisted validation status. Sending no scaling-group changes requests validation of the currently persisted configuration, allowing the backend to update and signal its validation status.

## Backend dependency

This UI flow depends on the worker-controller/backend accepting an empty ValidateSpec request and persisting the resulting validation status. Backend implementation is intentionally outside the scope of this UI PR.

## Test plan

- [x] Focused deployment service Vitest suite: 28 tests passed with one worker
- [x] Svelte check: 0 errors
- [x] ESLint and Prettier
- [x] Production build
- [x] Diff check


andrewzamojc (approved):
Seems aok. LGTM 👍

PR #3719, opened 2026-07-24
Update holocene toggle-switch to runes syntax

tegan-temporal:
Migrates `toggle-switch` to Svelte 5 runes and updates its consumers (`on:change` → `onchange`).

Consumer follow-up in cloud-ui: temporalio/cloud-ui#3007


rossnelson (approved):

PR #3724, opened 2026-07-24
Update holocene textarea to runes syntax

tegan-temporal:
Migrates `textarea` to Svelte 5 runes.

The `error` **slot** becomes an `errorSnippet` prop — Svelte 5 can't have both a prop and a snippet named `error` (the `error` string prop keeps that name), so the slot name couldn't survive the migration regardless.

Consumer follow-up in cloud-ui: temporalio/cloud-ui#3008


rossnelson (approved):

PR #3727, opened 2026-07-24
Add runes button component (strangler migration, part 1)

tegan-temporal:
First step of migrating `button` to Svelte 5 runes using a strangler-fig approach, since `button` can't be migrated atomically — Svelte 5 has no interop between the legacy `on:click` directive and the `onclick` prop, so a single-PR swap would force ~150 ui + ~190 cloud-ui files to change at once (and drag in icon-button/split-button).

**This PR:**
- Adds `button-runes.svelte` — a runes version with real `onclick`/`onkeydown` callback props (wrapped to preserve the legacy `stopPropagation` behavior), alongside the untouched legacy `button.svelte`. Styles are duplicated deliberately so the hot legacy component (87 consumers) carries zero risk.
- Migrates 4 initial consumers (`form-error`, `event-shortcut-keys`, `news-feed-widget`, `payload-code-block`) to prove the pattern: import swap + `on:click` → `onclick`.

**Plan:**
- Subsequent PRs move consumers over in small batches (ui + cloud-ui).
- Final PR: once legacy `button.svelte` has no consumers, delete it and rename `button-runes.svelte` → `button.svelte`.

No consumer behavior changes; `pnpm check` clean, all tests pass.

### DIFF
--- a/src/lib/components/event/event-shortcut-keys.svelte
+++ b/src/lib/components/event/event-shortcut-keys.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-  import Button from '$lib/holocene/button.svelte';
+  import Button from '$lib/holocene/button-runes.svelte';
   import DrawerContent from '$lib/holocene/drawer-content.svelte';
   import Drawer from '$lib/holocene/drawer.svelte';
   import Shortcut from '$lib/holocene/keyboard-shortcut/shortcut.svelte';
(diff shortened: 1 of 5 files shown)


rossnelson (approved):
Non-blocking: link buttons accept an onkeydown prop but don’t attach it to the <a>. Maybe we can forward it or exclude it from the link-button type.

PR #3728, opened 2026-07-24
Migrate app-level button consumers to runes button (part 2)

tegan-temporal:
Part 2 of the button strangler migration (stacked on #3727).

Migrates ~78 app-level consumers (`src/lib/components`, `src/lib/pages`, `src/routes`) from legacy `button.svelte` to `button-runes.svelte`: import swap + `on:click`/`on:keydown` → `onclick`/`onkeydown`. All mechanical, done via a scoped codemod that only touches `<Button>` tags (other components' `on:click` left intact).

Also exported the prop types (`ButtonWithoutHrefProps`/`ButtonWithHrefProps`/`ButtonProps`) from `button-runes.svelte` for wrapper consumers, and fixed `add-search-attributes` to use `ComponentProps<typeof Button>`.

Still on legacy button after this: holocene-internal renderers (modal, toast, banner, paginated, stories), the `icon-button`/`split-button` wrappers, and 2 type-only importers — follow-up batches.

`pnpm check` 0 errors, eslint clean, 2548 tests pass. Base is #3727; retarget to main once that merges.

### DIFF
--- a/src/lib/components/activity/activity-commands.svelte
+++ b/src/lib/components/activity/activity-commands.svelte
@@ -3,7 +3,7 @@
 
   import { page } from '$app/state';
 
-  import Button from '$lib/holocene/button.svelte';
+  import Button from '$lib/holocene/button-runes.svelte';
   import Tooltip from '$lib/holocene/tooltip.svelte';
   import { translate } from '$lib/i18n/translate';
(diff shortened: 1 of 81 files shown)


rossnelson (approved):

PR #3735, opened 2026-07-27
Update holocene number-input to runes syntax

tegan-temporal:
Migrates `input/number-input.svelte` to Svelte 5 runes. Independent of the button migration — branches off `main`.

- `export let` → `$props()` (`Props extends Omit<HTMLInputAttributes, 'value' | 'class'>`); `value = $bindable()`; `$$props.class` → `class` prop.
- `$:` validation + `let valid` state → a single `$derived`; `errorId` → `$derived`; the `name = id` default → a `$derived` fallback.
- Native event forwards (`on:input/change/focus/blur/keydown`) now flow through `{...rest}` (typed via `HTMLInputAttributes`).
- Removed a dead `valid={…}` prop from `search-attribute-input` — `valid` was never a real prop (computed internally from min/max); the adjacent `max={Number.MAX_SAFE_INTEGER}` already drives that validation.

No consumer event/prop changes needed (no ui or cloud-ui consumer forwarded events or used the dead prop). `pnpm check` 0 errors, eslint clean, 2548 tests pass.

Paired cloud-ui PR (removes an unused `error` prop from the retention form): temporalio/cloud-ui#3039

### DIFF
--- a/src/lib/components/workflow/search-attribute-input/index.svelte
+++ b/src/lib/components/workflow/search-attribute-input/index.svelte
@@ -107,7 +107,6 @@
       <NumberInput
         label={translate('common.value')}
         id="attribute-value-{id}"
-        valid={value < Number.MAX_SAFE_INTEGER}
         hintText="Number is too large"
         bind:value
         max={Number.MAX_SAFE_INTEGER}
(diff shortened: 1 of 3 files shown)


rossnelson (approved):
Requires a paired Cloud UI PR before the updated UI package is consumed: remove the unused `error` prop from the [retention form](https://github.com/temporalio/cloud-ui/blob/main/src/lib/components/namespace/settings/retention/form/content.svelte#L98). `hintText` already receives the same validation message.

PR #3737, opened 2026-07-27
Update holocene tabs and tab-button to runes syntax

tegan-temporal:
Migrates `tab/tabs.svelte` and `tab-buttons/tab-button.svelte` to Svelte 5 runes. Independent of the button migration — branches off `main`.

**tabs** (context provider): `context="module"` → `<script module>`, `export let`→`$props()`, default `<slot/>` → `children`, `$$restProps`→`...rest`. Keeps the `TABS` context export + register/select logic unchanged.

**tab-button**: kept the sound `AnchorProps | ButtonProps` discriminated union; `$app/stores` $page → `$app/state` page; `on:click` → `onclick`; `$$slots.default` → `children` check; `svelte:element` a/button preserved.

4 ui `TabButton` consumers updated (`on:click`→`onclick`) + both stories. **No cloud-ui changes**: all 20 cloud-ui `Tabs` consumers use default-slot content (auto-maps to `children`), and `tab-button` has no cloud-ui consumers — verified 0/0 against a pristine dist overlay.

`pnpm check` 0 errors, eslint clean, 2548 tests pass.

### DIFF
--- a/src/lib/components/event/event-summary.svelte
+++ b/src/lib/components/event/event-summary.svelte
@@ -72,21 +72,21 @@
         data-testid="feed"
         icon="feed"
         class="h-10"
-        on:click={onAllClick}>All</TabButton
+        onclick={onAllClick}>All</TabButton
       >
       <TabButton
         active={$eventViewType === 'compact'}
         data-testid="compact"
         icon="compact"
         class="h-10"
-        on:click={onCompactClick}>Compact</TabButton
+        onclick={onCompactClick}>Compact</TabButton
       >
       <TabButton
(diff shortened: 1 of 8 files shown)


rossnelson (approved):
Optional nit: would it be useful to preserve the native `onclick` type in each discriminated variant (`HTMLAnchorAttributes['onclick']` / `HTMLButtonAttributes['onclick']`)? That would retain element-specific `currentTarget` typing and `null` compatibility. If the broader `MouseEvent` type is intentional, no change is needed.

PR #3738, opened 2026-07-27
Update holocene pagination to runes syntax

tegan-temporal:
Migrates `pagination.svelte` to Svelte 5 runes. Independent of the button migration — branches off `main`. (`api-pagination` is handled separately — it's deprecated; see follow-up.)

- `generics="T"`, `export let` → `$props()`, `$:` → `$derived`/`$effect`, `$app/stores` $page → `$app/state` page, `svelte:window` `on:resize`/`on:keydown` → `onresize`/`onkeydown`.
- Hyphenated named slots → camelCase snippet props (`action-top-left` → `actionTopLeft`, `pagination-top` → `paginationTop`, etc.); scoped default slot → `children({ visibleItems, initialItem, activeRowIndex, setActiveRowIndex })`.
- 3 ui consumers updated (`namespaces/+page`, `child-workflows-table`, `live-child-workflows-table`) — incl. aliasing a consumer's `children` prop to avoid shadowing the new `children` snippet.

No cloud-ui consumers. `pnpm check` 0 errors, eslint clean, 2556 tests pass.

### DIFF
--- a/src/lib/components/workflow/child-workflows-table.svelte
+++ b/src/lib/components/workflow/child-workflows-table.svelte
@@ -47,57 +47,62 @@
 
 <Pagination
   items={formattedAll}
-  let:visibleItems
   aria-label={translate('workflows.child-workflows')}
   pageSizeSelectLabel={translate('common.per-page')}
   previousButtonLabel={translate('common.previous')}
   nextButtonLabel={translate('common.next')}
 >
-  <div slot="pagination-top"></div>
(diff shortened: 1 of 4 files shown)


rossnelson (approved):
Optional nit: would it make sense for `setActiveRowIndex` to accept `number | undefined`, matching the underlying store and preserving its clear/reset behavior? If snippet consumers are intentionally limited to numeric indexes, no change is needed.

PR #3750, opened 2026-07-29
Migrate paginated-table cluster to runes syntax

tegan-temporal:
Migrates the paginated-table cluster — `index.svelte`, `paginated.svelte`, `api-paginated.svelte` — to Svelte 5 runes. This was the last button-entangled holocene cluster (it renders Button/IconButton); unblocked now that the runes button has merged.

## Component changes
- `export let`/`$$Props`/`$$restProps`/`$$slots` → `$props()` + `...rest`; `$:` → `$derived`/`$effect`; `type Item = $$Generic` → `generics="Item"`; `context="module"` → `<script module>`.
- `paginated.svelte`: `$app/stores` (`$page`) → `$app/state` (`page`).
- `api-paginated.svelte`: `<svelte:window on:keydown>` → `onkeydown`; internal `error` state renamed → `fetchError` so the `error` snippet name is free.
- All slots → snippets, forwarded through `index` (`caption`/`headers`/`empty` pass straight through; the internal `actions-*` become snippets defined in the wrappers).

## Public API changes (consumers updated in this PR)
- Named slots → snippet props: `slot="caption"`→`caption`, `slot="headers"`→`headers` (scoped `{ visibleItems }`), `slot="empty"`→`empty`, `slot="actions-end-additional"`→`actionsEndAdditional` (scoped `{ visibleItems, page }`), `slot="error"`→`error`.
- **Scoped default slot → `rows` snippet** (not `children`). Svelte special-cases `children` to also accept the no-arg form (`Snippet<[]> & Snippet<[{ visibleItems }]>`), which a required-param snippet can't satisfy — so the scoped default content is now `{#snippet rows({ visibleItems })}`.
- `index.svelte` internal API: `slot="loading"`→`loadingContent` (avoids collision with the `loading` boolean prop), `slot="actions-start/center/end"`→`actionsStart`/`actionsCenter`/`actionsEnd` (used by cloud-ui's `api-pagination`).

## Verification
- 10 ui consumers updated. `pnpm check` 0 errors; 2570 tests pass; eslint clean.
- Two `state_referenced_locally` warnings in `api-paginated` (store created once from the initial `pageSizeOptions`) are behavior-equivalent to the original.

**Paired cloud-ui PR** (9 consumers, incl. the `api-pagination` wrapper that imports `index` directly): temporalio/cloud-ui#3038


src/lib/holocene/table/paginated-table/api-paginated.svelte
```diff
@@ -47,55 +48,72 @@
...
+  $effect(() => {
     if (pageSizeChange) {
       store.resetPageSize($store.pageSize);
       initalDataFetch();
     }
-  }
+  });
 
   async function initalDataFetch() {
     const fetchData = await onFetch();
```
laurakwhit, line 117: Should we clear the error here as well to make sure we aren't rendering a stale error?

```suggestion
 async function initalDataFetch() {
   clearError();
   const fetchData = await onFetch();
```

src/lib/holocene/table/paginated-table/api-paginated.svelte
```diff
@@ -47,55 +48,72 @@
     previousButtonLabel: string;
     nextButtonLabel: string;
     pageSizeOptions?: string[];
+    debounceDelay?: number;
+    caption?: Snippet;
+    headers?: Snippet<[{ visibleItems: T[] }]>;
+    rows?: Snippet<[{ visibleItems: T[] }]>;
+    loading?: Snippet;
```
laurakwhit: Should we use `loadingContent` here (like in `src/lib/holocene/table/paginated-table/index.svelte`) to avoid a collision?

src/lib/holocene/table/paginated-table/paginated.svelte
```diff
@@ -106,57 +128,56 @@
...
-      icon="arrow-right"
-    />
-  </nav>
-
-  <slot name="empty" slot="empty" />
+  {/snippet}
+
+  {#snippet actionsCenter()}
+    <div class="hidden items-center gap-2 md:flex">
+      {#each $store.pageShortcuts as page}
```
laurakwhit: NIT: to avoid conflict with page from `$app/state`.
```suggestion
      {#each $store.pageShortcuts as pageShortcut}
```

tegan-temporal:
@laurakwhit thanks! I implemented your suggestions

src/lib/holocene/table/paginated-table/api-paginated.svelte
```diff
@@ -189,19 +211,42 @@
...
+  const adjustedTotal = $derived(
+    !$store.hasNext && $store.indexEnd !== total ? $store.indexEnd : total,
+  );
 </script>
 
-<svelte:window on:keydown={handleKeydown} />
+<svelte:window onkeydown={handleKeydown} />
+
+{#snippet indexEmpty()}
+  {#if $store.loading}
```
laurakwhit: It looks like `empty` in `<PaginatedTable />` is only used if `loading` is `false`. I think we can just delete the condition and `loadingContent` here?

laurakwhit (approved):

PR #3753, opened 2026-07-29
Update holocene toast to runes syntax

tegan-temporal:
Migrates `toast.svelte` to Svelte 5 runes.

- `export let` → `$props()`; `$:` → `$derived`.
- `dismiss` event → `onDismiss` callback prop; default slot → `children`.
- Updated its only consumer, `toaster.svelte`, and the story.

No cloud-ui changes (it only imports `Toaster`, whose props are unchanged).

### DIFF
--- a/src/lib/holocene/toaster.svelte
+++ b/src/lib/holocene/toaster.svelte
@@ -25,8 +25,8 @@
 
   const announcements = toasterStore.announcements;
 
-  const dismissToast = (event: CustomEvent<{ id: string }>) => {
-    pop(event.detail.id);
+  const dismissToast = (id: string) => {
+    pop(id);
   };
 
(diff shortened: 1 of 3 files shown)


src/lib/holocene/toaster.svelte
```diff
@@ -59,7 +59,7 @@ inertBackground walk from re-inerting it as a body-level sibling. -->
       {closeButtonLabel}
       variant={variant ?? 'primary'}
       id={id ?? ''}
-      on:dismiss={dismissToast}
+      onDismiss={dismissToast}
```
laurakwhit: Looks like we could drop the `dismissToast` wrapper entirely.
```suggestion
      onDismiss={pop}
```

laurakwhit (approved):

PR #3756, opened 2026-07-29
Update holocene copyable to runes syntax

tegan-temporal:
Migrates `copyable/index.svelte` to Svelte 5 runes.

- `export let` → `$props()`; `$$props['container-class']`/`$$props.class` → declared `container-class`/`class` props (same names).
- Default slot → `children` (fallback `<span>` when no content); `$$slots.default` checks → `children`.
- `select-all` on the fallback span (always applied) now uses Svelte 5 array-class merging instead of a `class:select-all` directive.

No consumer changes in either repo — prop names and default-slot usage are unchanged.

### DIFF
--- a/src/lib/holocene/copyable/index.svelte
+++ b/src/lib/holocene/copyable/index.svelte
@@ -1,13 +1,31 @@
 <script lang="ts">
+  import type { Snippet } from 'svelte';
+
   import { copyToClipboard } from '$lib/utilities/copy-to-clipboard';
 
   import CopyButton from './button.svelte';
 
-  export let content: string;
-  export let visible = false;
-  export let clickAllToCopy = false;
-  export let copyIconTitle: string;
-  export let copySuccessIconTitle: string;
+  interface Props {
+    content: string;
+    visible?: boolean;
+    clickAllToCopy?: boolean;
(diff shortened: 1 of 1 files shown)


src/lib/holocene/copyable/index.svelte
```diff
@@ -1,13 +1,31 @@
...
-  export let clickAllToCopy = false;
-  export let copyIconTitle: string;
-  export let copySuccessIconTitle: string;
+  interface Props {
+    content: string;
+    visible?: boolean;
+    clickAllToCopy?: boolean;
+    copyIconTitle: string;
+    copySuccessIconTitle: string;
+    class?: string;
```
laurakwhit, line 14: Can we use type `ClassNameValue` from `tailwind-merge` here?

laurakwhit (approved):

src/lib/holocene/copyable/index.svelte
tegan-temporal: I think that needs to happen separately, there are type conflicts with how Svelte className type is larger than what tailwind-merge expects. The code now isn't a behavior change and introducing tailwind-merge would be as well so it introduces more risk in a big migration

PR #3757, opened 2026-07-29
Update holocene zoom-svg to runes syntax

tegan-temporal:
Migrates `zoom-svg.svelte` to Svelte 5 runes.

- `export let` → `$props()`; `width`/`height` are `$bindable` (bound to the container's `clientWidth`/`clientHeight`).
- `viewBox` is a reassignable `$derived({ x: 0, y: 0, width, height })` — mutated in place by pan/zoom, and resets when the container dimensions change (same behavior as the old `$:`).
- Native handlers `on:*` → `on*`; `$$restProps.class` → `class` prop.
- Slots → snippets: `slot="controls"` → `controls`; the scoped default slot → a `graph` snippet passing `{ width, height, zoomLevel }`.
- Split the a11y `svelte-ignore` into one comment per rule — Svelte 5 only honored the first rule when multiple were on a single line.

Updated the one consumer, `workflow-family-tree.svelte`, to the new snippets. No cloud-ui consumers.

### DIFF
--- a/src/lib/holocene/zoom-svg.svelte
+++ b/src/lib/holocene/zoom-svg.svelte
@@ -55,8 +55,11 @@
 
   function panBy(dx: number, dy: number) {
     if (!pannable) return;
-    viewBox.x += dx * viewBox.width;
-    viewBox.y += dy * viewBox.height;
+    viewBox = {
+      ...viewBox,
+      x: viewBox.x + dx * viewBox.width,
+      y: viewBox.y + dy * viewBox.height,
+    };
   }
 
   function zoomBy(factor: number) {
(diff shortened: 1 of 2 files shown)


src/lib/holocene/zoom-svg.svelte
laurakwhit: With `viewBox` being derived, should we be reassigning instead of mutating here?

```suggestion
    viewBox = {
      ...viewBox,
      x: viewBox.x + dx * viewBox.width,
      y: viewBox.y + dy * viewBox.height,
    };
```

tegan-temporal: oh dang, good catch

src/lib/holocene/zoom-svg.svelte
```diff
@@ -1,40 +1,65 @@
...
+    graph?: Snippet<[{ width: number; height: number; zoomLevel: number }]>;
+  }
+
+  let {
+    containerHeight = 600,
+    initialZoom = 1,
+    maxZoomIn = 0.25,
+    maxZoomOut = 2.5,
+    width = $bindable(600),
+    height = $bindable(400),
```
laurakwhit: Not a blocker, but are these actually being set anywhere? It seems like we should probably just drop these defaults/props and make them internal `$state` that rely on the DOM measurements ⬇️ 

```js
<div
  bind:clientWidth={width}
  bind:clientHeight={height}
  ...
>
```

laurakwhit (approved):

PR #3759, opened 2026-07-29
Migrate holocene modal to runes syntax

tegan-temporal:
Migrates `modal.svelte` to Svelte 5 runes, in place, with all 38 ui consumers updated in the same PR.

- `export let` → `$props()`; `createEventDispatcher` `cancelModal`/`confirmModal` → `onCancelModal`/`onConfirmModal` callback props.
- Slots → snippets: `slot="title"` → `titleSnippet` (renamed to avoid the `HTMLAttributes.title` collision), `slot="content"` → `content`, `slot="footer"` → `footer`.
- `open`/`error` are `$bindable`; `...rest` forwards remaining attributes to the `<dialog>`; `toggleModal` is now internal (had no external consumers).
- Story updated to the new API.

Two consumers needed judgment: `view-modal.svelte` drops the now-unused `event` arg on its confirm handler (the form handler already calls `preventDefault`), and `ramp-unversioned-modal.svelte`'s footer snippet renders an empty `<div>` when `onRemove` is falsy to keep the legacy spacer's right-alignment.

**Breaking change** to the published `modal` API — paired cloud-ui PR migrates its consumers: temporalio/cloud-ui#3046


src/lib/components/activity/activity-pause-confirmation-modal.svelte
```diff
@@ -47,30 +47,34 @@
...
-    </Link>
-    <Input
-      id="activity-pause-reason"
-      class="mt-4"
-      placeholder={translate('common.reason-placeholder')}
-      label={translate('common.reason-optional')}
-      bind:value={reason}
-    />
-  </div>
+  {#snippet titleSnippet()}
```
laurakwhit, line 53: Wonder if we really need a title snippet here or if we could just pass the title copy instead.

src/lib/components/schedule/schedule-action-modals/backfill-schedule-modal.svelte
```diff
@@ -88,7 +88,7 @@
   loading={$actionPending}
   error={$serverError}
```
laurakwhit, line 89: We should probably make this bindable at all call sites or ensure all the parents are clearing the error on cancel so a stale error message isn't kept in the store. 

```suggestion
  bind:error={$serverError}
```

laurakwhit (approved):
Some things for a potential follow-up PR but otherwise 🚀

PR #3762, opened 2026-07-29
Remove dead workflow-filters component

tegan-temporal:
Removes `workflow/workflow-filters.svelte` — dead code with no importers in ui or cloud-ui, and no story (only carried along by mechanical migration sweeps). The similarly-named `to-list-workflow-filters` utility is unrelated and untouched.

Note: but verified unused by both ui and cloud-ui.


laurakwhit (approved):
🗑️

PR #3764, opened 2026-07-29
Migrate batch-operations header to runes syntax

tegan-temporal:
Migrates `batch-operations/header.svelte` to Svelte 5 runes.

- Removes `createEventDispatcher`; `toggleAutoRefresh` event → `onToggleAutoRefresh` callback prop.
- Updates its one consumer (`pages/batch-operation.svelte`): `on:toggleAutoRefresh` → `onToggleAutoRefresh`, handler takes `{ checked }` instead of a `CustomEvent`.

No cloud-ui consumers.

### DIFF
--- a/src/lib/components/batch-operations/header.svelte
+++ b/src/lib/components/batch-operations/header.svelte
@@ -9,7 +9,7 @@
 
   interface Props {
     operation: BatchOperation;
-    onToggleAutoRefresh?: (detail: { checked: boolean }) => void;
+    onToggleAutoRefresh?: (checked: boolean) => void;
   }
 
   let { operation, onToggleAutoRefresh }: Props = $props();
@@ -20,7 +20,7 @@
     }
 
     const { checked } = event.currentTarget;
-    onToggleAutoRefresh?.({ checked });
+    onToggleAutoRefresh?.(checked);
     $autoRefresh = checked;
   };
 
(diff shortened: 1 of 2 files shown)


src/lib/components/batch-operations/header.svelte
```diff
@@ -11,21 +9,18 @@
 
   interface Props {
     operation: BatchOperation;
+    onToggleAutoRefresh?: (detail: { checked: boolean }) => void;
```
laurakwhit: Can we flatten this since a detail object is no longer required?

```suggestion
    onToggleAutoRefresh?: (checked: boolean) => void;
```

src/lib/components/batch-operations/header.svelte
```diff
@@ -11,21 +9,18 @@
...
 
   const handleToggleAutoRefresh = (event: Event) => {
     if (!(event.currentTarget instanceof HTMLInputElement)) {
       return;
     }
 
     const { checked } = event.currentTarget;
-    dispatch('toggleAutoRefresh', { checked });
+    onToggleAutoRefresh?.({ checked });
     $autoRefresh = checked;
```
laurakwhit: Since the parent hander is updating the `fetchKey` and the block that re-runs `fetchBatchOperation()` (which reads `$autoRefresh`) is keyed to `fetchKey` we should probably switch the order of these. 

```suggestion
    $autoRefresh = checked;
    onToggleAutoRefresh?.({ checked });
```

laurakwhit (approved):

PR #3765, opened 2026-07-29
Migrate workflow-family-node-tree to runes syntax

tegan-temporal:
Migrates `workflow-family-node-tree.svelte` (the recursive SVG relationship tree) to Svelte 5 runes.

- `export let` → `$props()`; `$:` reactive statements → `$derived` / plain helper closures (`getPosition`/`isExpanded`/`isCurrent`/`isActive` read reactive values at call time).
- Native `on:click`/`on:keydown` on the `<g>` nodes → `onclick`/`onkeydown`.
- `<svelte:self>` → a self-import (`WorkflowFamilyNodeTree`), the runes-idiomatic recursion.

Behavior preserved (the pre-existing unkeyed `{#each}` is left as-is). Single consumer (`workflow-family-tree.svelte`) unchanged — public prop API is the same. No cloud-ui consumers.

### DIFF
--- a/src/lib/components/workflow/relationships/workflow-family-node-tree.svelte
+++ b/src/lib/components/workflow/relationships/workflow-family-node-tree.svelte
@@ -17,26 +17,50 @@
 
   import { showFullTree } from '../workflow-relationships.svelte';
 
-  export let root: RootNode;
-  export let width: number;
-  export let height: number;
-  export let zoomLevel: number;
-  export let rootX = 0;
-  export let rootY = 0;
-  export let generation = 1;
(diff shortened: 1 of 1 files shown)


src/lib/components/workflow/relationships/workflow-family-node-tree.svelte
```diff
@@ -141,7 +168,7 @@
 {#each root?.children as child, index}
```
laurakwhit, line 168: NIT
```suggestion
{#each root?.children as child, index (index)}
```

laurakwhit (approved):

PR #3768, opened 2026-07-30
Migrate start-workflow and workflow-query pages to runes syntax

tegan-temporal:
Migrates the last two legacy pages to Svelte 5 runes — `pages/start-workflow.svelte` and `pages/workflow-query.svelte`. These slipped the earlier component sweep because they have no props (only `$:` reactive statements + `$page`).

- `let` locals → `$state`; `$:` → `$derived`; `$app/stores` `$page` → `$app/state` `page`.
- `workflow-query`: `queryType` (default-to-first, but `bind:`-bound and user-overridable) uses `$state` + an `$effect` for the default — it can't be a `$derived` because it's two-way bound.

No public prop API (both read `page.params`), so cloud-ui's `<StartWorkflow />` / `<WorkflowQuery />` usages are unaffected — verified, no paired PR.

### DIFF
--- a/src/lib/pages/start-workflow.svelte
+++ b/src/lib/pages/start-workflow.svelte
@@ -3,7 +3,7 @@
 
   import { onMount } from 'svelte';
 
-  import { page } from '$app/stores';
+  import { page } from '$app/state';
 
   import CodecServerErrorBanner from '$lib/components/codec-server-error-banner.svelte';
   import PayloadInputWithEncoding from '$lib/components/payload-input-with-encoding.svelte';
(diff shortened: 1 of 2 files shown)


src/lib/pages/start-workflow.svelte
```diff
@@ -232,16 +232,19 @@
...
-    !workflowCreateDisabled($page);
+      !!taskQueue &&
+      !!workflowType &&
+      !!inputValid &&
+      !workflowCreateDisabled(page),
+  );
 
-  $: checkTaskQueue(taskQueueParam ?? '');
+  $effect(() => {
+    checkTaskQueue(taskQueueParam ?? '');
```
laurakwhit, line 246: Since `checkTaskQueue` reads `namespace` (and it's derived) maybe we should throw an untrack on this? 

```suggestion
    const queue = taskQueueParam ?? '';
    untrack(() => checkTaskQueue(queue));
```

laurakwhit (approved):

src/lib/pages/start-workflow.svelte
tegan-temporal: tbh if the namespace actually changes we probably want to recheck it don't we?

laurakwhit: Hmm yeah I guess I was thinking it looks like the `$effect` here should just be tracking the `taskQueueParam` but it's also impacted by the `namespace` used in `checkTaskQueue()`.

PR #3772, opened 2026-07-30
Upgrade to Storybook 10 and migrate stories to defineMeta

tegan-temporal:
Upgrades Storybook 8 → 10 and migrates all stories to the addon-svelte-csf v5 `defineMeta` API.

- Bump Storybook 8.6 → 10.5; drop addons now folded into core
- Migrate all `.stories.svelte` from `satisfies Meta<T>` + `<Template>` to `defineMeta`, with a default `render` snippet and `asChild` for static stories
- Consolidate imports into the module script; instance scripts keep only reactive state
- gitignore `storybook-static`

https://temporalio.atlassian.net/browse/DT-4319


.storybook/preview.ts
```diff
@@ -32,9 +32,8 @@ const preview: Preview = {
     }),
   ],
   parameters: {
-    actions: { argTypesRegex: '^on[A-Z].*' },
     backgrounds: {
-      disable: true,
+      disabled: true,
```
laurakwhit: Should this still be "disable"?

```suggestion
      disable: true,
```

.storybook/preview.ts
```diff
@@ -32,9 +32,8 @@ const preview: Preview = {
     }),
   ],
   parameters: {
-    actions: { argTypesRegex: '^on[A-Z].*' },
```
laurakwhit, line 35: Does this mean no story logs actions to the "Actions" panel anymore?

laurakwhit (approved):

.storybook/preview.ts
tegan-temporal: it doesn't log automatically, no. storybook was warning about using it which why I removed it. Also, it wasn't logging things like onclick or on:click, it had to specifically be camel cased like that regex e.g., `onClick`

the recommended way is to use fn spies when defining the meta https://storybook.js.org/docs/essentials/actions#via-storybooktest-fn-spies

PR #3776, opened 2026-07-31
Track pointerDown and ignore subsequent click in clickoutside

laurakwhit:
## Description & motivation 💭 

When pressing inside the value input in the search attribute filter menu and then dragging past the menu's edge to select a long string, and releasing outside the menu, the click target became `<body>` (or another outer container). `node.contains(event.target)` was false, the handler ran, and the menu snap shut mid-selection.

`outside-click.ts` now also tracks `pointerdown` and ignores the subsequent click when the gesture started inside the node. This also fixes the same drag-select behavior in the two other clickoutside consumers (`date-picker.svelte` and `drawer.svelte`). 

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

Go to `/workflows` > select "Add Filter" > select "WorkflowId" and type in a value the exceeds the width of the input > attempt to select the entire string in the input such that the mouse is dragged over the edge of the menu 
   - [ ] Verify the menu does not close

### Issue(s) closed 
`DT-4152`

### DIFF
--- a/src/lib/holocene/outside-click.test.ts
+++ b/src/lib/holocene/outside-click.test.ts
@@ -1,0 +1,122 @@
+import { afterEach, describe, expect, it, vi } from 'vitest';
+
+import { clickoutside } from './outside-click';
+
+const dispatch = (target: Element, type: string, init: MouseEventInit = {}) => {
+  target.dispatchEvent(
+    new MouseEvent(type, {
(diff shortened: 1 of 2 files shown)


ardiewen (approved):
nice catch!

PR #3777, opened 2026-08-01
fix(auth): decouple refresh token cookie lifetime from access token expiry

### LINKED ISSUE #3210: OIDC Refresh doesn't work due to bad expiration date

thdxg:
## Description & motivation 💭

Fixes #3210.

`SetUser` in `server/server/auth/auth.go` derives the `refresh` cookie's `MaxAge` from `oauth2.Token.Expiry`, with a comment claiming that field "typically reflects the refresh token's lifetime". It doesn't: `oauth2.Token.Expiry` is populated from the token response's `expires_in`, which per [RFC 6749 §5.1](https://datatracker.ietf.org/doc/html/rfc6749#section-5.1) is the **access** token's lifetime. OAuth2/OIDC never communicates the refresh token's lifetime in the token response.

The consequence is fatal for the refresh flow: the frontend only refreshes reactively on a 401 (`src/lib/utilities/oss-provider.svelte.ts` → `ossPostResponse`), and a 401 can only occur once the access token has expired — which is the exact moment the browser deletes the `refresh` cookie. So the cookie is always gone before the first `/auth/refresh` attempt, and every refresh fails with `token_refresh_failed reason=missing_refresh_token`. Refresh tokens are effectively unusable with Google OIDC (`expires_in` ≈ 3599s) and any other IdP that returns a standard `expires_in`.

This PR decouples the cookie TTL from the access-token expiry:

- New config option `auth.refreshTokenLifetime` (`TEMPORAL_AUTH_REFRESH_TOKEN_LIFETIME` in the Docker config), so operators can match their IdP's refresh-token/SSO-session policy.
- Defaults to 7 days when unset; the existing 30-day cap is kept.
- `authenticateCb`/`refreshTokens` now take `config.Auth` instead of a bare `maxSessionDuration`, threading the new value the same way `MaxSessionDuration` already flows.

We verified the defect in a production deployment (v2.49.1, Google OIDC with `access_type=offline` + `prompt=consent` options): Google issued refresh tokens, the server logged a cookie `MaxAge` of 3598s, and 100% of refresh attempts failed with `missing_refresh_token`. Running a binary patched to a fixed 7-day cookie TTL (the same behavior as this PR's default) resolved it — sessions now survive access-token expiry via `/auth/refresh`.

N/A (server-side change).

### Design Considerations 🎨

N/A.

### How was this tested 👻

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

Unit tests cover the three lifetime paths (unset → 7-day default, configured value, > 30 days → capped) and assert that `oauth2.Token.Expiry` no longer influences the cookie. The underlying approach (fixed cookie TTL instead of `Expiry`-derived) has been running in our production deployment against Google OIDC.

1. Configure OIDC auth against Google (`providerUrl: https://accounts.google.com/`, provider `options: { access_type: offline, prompt: consent }`).
2. Log in and inspect the `refresh` cookie: before this change its Max-Age is ~3599s (the access token lifetime); after, it is `refreshTokenLifetime` (7 days by default).
3. Wait for the access token to expire (~1h) and perform any action: before this change the UI logs `token_refresh_failed reason=missing_refresh_token` and the user is logged out; after, `/auth/refresh` succeeds.
4. `cd server && go test ./server/auth/...`

### Issue(s) closed

- #3210

### Any docs updates needed?

`AUTHENTICATION.md` is updated in this PR (new config option, corrected Google note, troubleshooting entry).

### DIFF
--- a/AUTHENTICATION.md
+++ b/AUTHENTICATION.md
@@ -40,6 +40,7 @@ auth:
 | `enabled`            | boolean  | Enable or disable authentication                                                                                               |
 | `redirectToProvider` | boolean  | Skip the Temporal UI login page and redirect unauthenticated users directly to the configured OIDC provider                    |
 | `maxSessionDuration` | duration | Maximum session duration before forced re-login (e.g., `8h`, `24h`, `168h`). Set to `0` or omit for unlimited session duration |
+| `refreshTokenLifetime` | duration | Lifetime of the refresh token cookie (e.g., `24h`, `168h`). Should match your IdP's refresh token or SSO session policy. Defaults to 7 days; capped at 30 days |
 | `providers`          | array    | List of auth providers (currently only the first is used)                                                                      |
 
 #### Provider Settings
@@ -66,6 +67,8 @@ Temporal UI supports two mechanisms for session management:
 1. **Token Refresh**: When access tokens expire, the UI automatically refreshes them using the refresh token. Users stay logged in seamlessly.
 
 2. **Max Session Duration**: Forces users to re-authenticate after a specified time, regardless of token validity.
+
+The refresh token is stored in an HttpOnly cookie whose lifetime is controlled by `refreshTokenLifetime` (default 7 days). OAuth2 token responses only communicate the access token's lifetime, not the refresh token's, so set `refreshTokenLifetime` to match your IdP's refresh token or SSO session policy.
 
 ### Configuring Max Session Duration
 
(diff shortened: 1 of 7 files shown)


Albert-Coding:
@thdxg why close this PR?

thdxg:
@Albert-Coding I realized there's another PR open for the same issue

PR #386, opened 2022-03-25
/import/events route to import JSON file for event debugging

Alex-Tideman:
## What was changed
A new /import/events route was added to support importing JSON files for viewing event history

Screen Shot 2022-03-28 at 11 32 40 AM
Screen Shot 2022-03-28 at 11 33 18 AM

## Why?
This will allow the SDK team and others to import JSON files to view and troubleshoot events

## Checklist

2. How was this tested:
Ran test suite, retested all workflow pages with RainbowStatus workflows to confirm existing pages still all work, added new /import Cypress tests

3. Any docs updates needed?

PR #3778, opened 2026-08-03
VLN-1585: remediate claude-code-action-unhardened

picatz:
> 🏕️ This pull request was created by [camper](https://github.com/temporalio/camper), an automated security campaign tool.

## Finding

<table>
  <tr><td><strong>Rule</strong></td><td><code>claude-code-action-unhardened</code></td></tr>
  <tr><td><strong>Severity</strong></td><td>HIGH</td></tr>
  <tr><td><strong>Repository</strong></td><td><code>temporalio/ui</code></td></tr>
  <tr><td><strong>Ticket</strong></td><td><a href="https://temporalio.atlassian.net/browse/VLN-1585">VLN-1585</a></td></tr>
</table>

## Summary

- `.github/workflows/claude.yml`: Pinned `anthropics/claude-code-action` to the vetted `v1.0.183` commit; retained job-level `id-token: write` for action token minting and retained checkout credentials for the action pre-authentication fetch; disabled full output and the `WebFetch`/`WebSearch` tools; capped execution at 10 turns; restricted allowed tools to read, inline comments, `gh pr comment`, and `gh pr diff`; and removed every package-executing `pnpm` command from the Bash allowlist.

## Instructions

- **Approve** to merge this fix
- **Request changes** to trigger a new remediation attempt
- `/camper rebase` — rebase onto the base branch
- `/camper close` — close this PR without merging
- `/camper retry` — regenerate the fix from scratch against the current base

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -27,20 +27,21 @@ jobs:
       contents: read
       pull-requests: read
       issues: read
-      id-token: write
       actions: read # Required for Claude to read CI results on PRs
     steps:
       - name: Checkout repository
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
+          persist-credentials: false
           fetch-depth: 1
 
       - name: Run Claude Code
         id: claude
-        uses: anthropics/claude-code-action@2cc1ac1331eac7a6a96d716dd204dd2888d0fcd2 # v1.0.112
+        uses: anthropics/claude-code-action@be7b93b1907a4abad570368f3c74b6fe3807510b # v1.0.183
         with:
           anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
           track_progress: true
+          show_full_output: false
           # This is an optional setting that allows Claude to read CI results on PRs
           additional_permissions: |
             actions: read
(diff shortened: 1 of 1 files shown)


picatz:
This PR has had no activity for 7 days and may need attention.

**Actions you can take:**
- Review and approve if the changes look good
- Close if this fix is no longer needed
- Comment `/camper rebase` to rebase onto the latest base branch
- Comment `/camper retry` to regenerate the fix

picatz:
Pushed a correction: `b0a0e0d`.

Two of this PR's original changes were wrong. The same pair broke a docs bot in `temporalio/saas-external-observability`, corrected there in #385 and #386.

- **`id-token: write` is not unused.** With no `github_token` input, the action exchanges a GitHub OIDC token for its GitHub App token, so removing the permission fails the action at startup. Restored.
- **`persist-credentials: false` is incompatible with the action** on the checkout it operates on. In tag mode `setupBranch()` runs `git fetch origin --depth=N <branch>` before `configureGitAuth()` installs any credential of its own, so the fetch depends on what `actions/checkout` persisted. Open upstream: anthropics/claude-code-action#1236, #1559, #1510. Removed from `claude.yml` (it sets `track_progress: true`, which forces tag mode).

Where a workflow sets a `prompt:` it runs in agent mode, where the action configures credentials before its only fetch, so the setting is correct there and stays. It also stays on any checkout the action never fetches through, such as a plugin marketplace or an untrusted PR head under `path:`.

The description's bullets about the OIDC permission and persisted checkout credentials are superseded. The rest of the hardening stands: the SHA pin, tool scoping, `--max-turns`, and `show_full_output: false`.

rossnelson (changes requested):
Requesting changes on the existing P1 thread: the allowed `pnpm install`, lint, test, and build commands can execute PR-controlled package scripts and lifecycle hooks inside this secret-bearing `issue_comment` workflow. Disabling WebFetch/WebSearch does not remove shell network access or the resulting exfiltration path.

Please keep the workflow's Claude tools read/diff/comment-only and remove package-executing Bash entries. The remaining hardening changes look sound.

rossnelson (approved):

PR #3779, opened 2026-08-03
Fix worker deployment actions for self-hosted deployments

rossnelson:
## Summary

- Self-hosted deployments regain Set/Unset Current and ramping actions.
- Compute-only Create/Edit/Validate/Ramp-to-Unversioned actions remain gated by compute config.
- Regression tests cover both cases.

## Test plan

- `pnpm test --run src/lib/pages/deployment.svelte.test.ts` (4 passed)
- `pnpm check` (0 errors, 62 pre-existing warnings)
- Prettier and git diff check clean


GraceGardner (approved):
excellent!

PR #3786, opened 2026-08-04
fix(timeline): recolor group bar when a completion arrives live

tegan-temporal:
A running group's timeline bar/label stayed in its in-progress color until reload when the completion arrived via the live poll.

- `appendLiveEvent` extended the group's `eventList` in place (stable reference); the timeline row pool keys on group identity, so the slot was reused and the classification-derived color/label never re-rendered.
- Swap in a fresh `cloneEventGroup()` after a live follower extends a group (live-group and pooled-group paths) — the same fresh-reference pattern `enrichGroup` already uses for pending metadata (#3693).
- Adds an integration test that renders in-progress, delivers the completion via the live poll, and asserts the bar relabels to Completed without reload.


tegan-temporal:
Combined into #3787 (retargeted to `main`, now includes this fix plus the copy-on-write refactor).

PR #3787, opened 2026-08-04
fix(timeline): recolor a group when a later event extends it

tegan-temporal:
A group's timeline bar and label kept their in-progress color and text after the group completed, until reload.

The row pool reuses a row while its group reference is unchanged (`timeline-graph.svelte:427`), and the buffer extended already-rendered groups in place. The `eventCount` prop still changes, so the new dot gets drawn — but `lineColor` and `accessibleName` derive from `group` alone and never re-derive. When testing by hand, the dot appearing is not evidence the bug is gone.

Three paths now copy the group and publish the new reference:

- the live poll extending a group loaded by the fetch
- the live poll extending a group the poll itself created
- the fetch attaching a follower whose head arrived in an earlier page (histories over 1000 events)

`processEvent` re-syncs its local reference after the pending-follower flush, which can now swap `meta.group`. Unit tests cover each path, plus an integration test asserting the label flips without a reload.

The copy is shallow and shares `eventList`, so a superseded reference must be treated as dead; #3790 removes that sharp edge by deriving groups on read.

https://github.com/user-attachments/assets/c3c6495b-156b-46bc-a943-de6cade30f3c

### DIFF
--- a/src/lib/services/grouped-event-buffer.ts
+++ b/src/lib/services/grouped-event-buffer.ts
@@ -267,6 +267,26 @@ function clearResolvedPendingState(group: EventGroup): void {
   }
 }
 
+/**
+ * Returns a copy of `group` with `event` added. Callers must store the returned
+ * reference back over the old one: the timeline row pool reuses a row when its
+ * group reference is unchanged (timeline-graph.svelte), so extending a rendered
+ * group in place leaves the row showing its old classification — a completed
+ * activity keeps rendering as in-progress until reload.
+ *
+ * The copy is shallow and shares `eventList` and `links` with the original, so
+ * the pre-copy reference is left inconsistent and must be treated as dead once
+ * the caller has stored the new one.
+ */
+function withAddedEvent(group: EventGroup, event: WorkflowEvent): EventGroup {
+  const next = cloneEventGroup(group);
+  insertEventById(next.eventList, event);
+  next.timestamp = event.timestamp;
+  addEventToGroup(next, event);
+  clearResolvedPendingState(next);
+  return next;
+}
+
 function growArrays(newSize: number): void {
   if (newSize <= eventSlots.length) return;
   eventSlots.length = newSize;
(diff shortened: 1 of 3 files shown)


src/lib/services/grouped-event-buffer.ts
```diff
@@ -267,6 +267,26 @@ function clearResolvedPendingState(group: EventGroup): void {
...
+ * group reference is unchanged (timeline-graph.svelte), so extending a rendered
+ * group in place leaves the row showing its old classification — a completed
+ * activity keeps rendering as in-progress until reload.
+ *
+ * The copy is shallow and shares `eventList` and `links` with the original, so
+ * the pre-copy reference is left inconsistent and must be treated as dead once
+ * the caller has stored the new one.
+ */
+function withAddedEvent(group: EventGroup, event: WorkflowEvent): EventGroup {
+  const next = cloneEventGroup(group);
```
laurakwhit, line 282: We were previous just mutating (with `addEventToGroup`), now we're doing a clone. Could this cause any performance issues with large event histories?

src/lib/services/grouped-event-buffer.ts
```diff
@@ -547,6 +563,9 @@ export function processEvent(
       attachFollowerToPool(poolIdx, followerSlotIdx);
     }
     pendingFollowers.delete(slotIdx);
+    // attachFollowerToPool replaced meta.group — re-sync the local ref so the
+    // flushes below, and the group this call publishes, aren't stale.
+    group = meta.group ?? group;
```
laurakwhit, line 568: Would it be worth adding a test specifically for this case?

src/lib/services/grouped-event-buffer.ts
tegan-temporal: the mutation is why things were stale. by doing this shallow clone we ensure svelte reactivity is triggered due to the reference change. For what it's worth I have https://github.com/temporalio/ui/pull/3790 which is a much larger change that rethinks the whole event buffer entirely. I just wanted to get this fix in since it's a small immediate fix and the larger refactor is not rushed.

tegan-temporal: also for what it's worth the clone is pretty cheap, it's adding only ~16ms in a synthetic benchmark with 200k events:

```
┌────────────────┬──────────┬────────┬──────┬──────┐
│                │  median  │  mean  │ min  │ max  │
├────────────────┼──────────┼────────┼──────┼──────┤
│ main (mutate)  │ 1573ms   │ 1611ms │ 1550 │ 1813 │
├────────────────┼──────────┼────────┼──────┼──────┤
│ branch (clone) │ 1588.5ms │ 1618ms │ 1564 │ 1793 │
└────────────────┴──────────┴────────┴──────┴──────┘
```

src/lib/services/grouped-event-buffer.ts
tegan-temporal: added

laurakwhit (approved):

PR #3790, opened 2026-08-05
refactor(events): derive event groups on read in the grouped-event-buffer

tegan-temporal:
Rewrites the grouped-event-buffer so event groups are derived on read and memoized, instead of built at ingest and mutated in place, and gives the module a reactive read API so views stop hand-rolling their own subscribe / throttle / re-read effects.

## Correctness

Group identity becomes a pure function of content version: an unchanged group reads back as the same object, a changed one always as a new object. The timeline row pool keys on group identity, so extending a rendered group in place is a recurring source of stale rows — this makes the invariant structural instead of a convention every call site has to remember.

The buffer holds pending activity/nexus metadata and applies it when a group is built, so a terminal event and a stale pending entry resolve the same way in either arrival order, and a head event that arrives after the refresh listing it still picks its own up.

## Simplification

- **One ingestion path.** `ingestHistoryEvent` replaces `processEvent` + `appendLiveEvent`, which each kept their own group store, pending-follower map, and dedup set, plus a reconcile step between them.
- **Deletes unreachable API.** The Pixi track-layout machinery, the promise-per-row streaming API, and the group-registration listeners had no callers — 15 of 28 exports were reachable only from tests.
- Six manual cache-invalidation sites collapse to one revision counter, and `getEventArray` / `getWorkflowTaskFailedEvent` become single passes with no sort or dual-store merge.
- Module state drops from 22 variables to 11, excluding read caches. The buffer stays Svelte-free and unit-testable; the reactive wrapper is a separate `.svelte.ts`.

## Views read a value, and only materialize what they render

```svelte
const lazyGroups = $derived(eventBuffer.lazyGroupsWithoutWorkflowTasks);
```

That replaces ~20 lines of `$state.raw` + `void $bufferVersion` + rAF + re-read in each of the two workflow layouts, and producers no longer announce their own writes.

Nothing upstream of a rendered row needs a full group — the type filter reads `category`, the sorts read `isPending` and `initialEvent.id`, the status filter reads `classification`, and segment layout reads the head and last event times. `GroupRecord` carries those, so `getLazyGroups()` returns the records themselves and views call `materializeGroup` per rendered row: ~54 pooled rows on the timeline, one page of 100 in the compact view.

The feed view still takes full groups. Its gutter graph does cross-group layout needing `eventList`, `level`, and reverse event→group lookups (`history-graph`, `positioning`) — a separate piece of work.

## Performance

Measured in Chromium against a production build, running the same scripted interaction on both branches: load a mocked 50k-event history on the timeline, wait for the fetch, then 20 fixed scrolls. Three runs each, medians, run-to-run spread under 2%.

| | main | this branch |
|---|---|---|
| first row painted | 1103ms | **914ms** |
| fetch complete | 1580ms | **1392ms** |
| total blocking time | 324ms | **156ms** |
| longest task | 355ms | **186ms** |
| busy CPU | 998ms | **830ms** |

The blocking figures are the ones a user feels — on main that single 355ms task is a visible freeze during load.

Micro-benchmarks on a 200k-event / 80k-group history show where it comes from:

| | before | after |
|---|---|---|
| ingest | 1592ms | 738ms |
| `getEventArray` | 10.4ms | 1.5ms |
| open a view (cold) | 548ms | 2.9ms |

That last row is the point of deriving on read: opening a view used to build every group in the history, and now builds one screen. The 738ms that remains in ingest is 88% `toEvent` conversion, which this PR doesn't touch — deferring that is the next available win.

## API surface

Three changes worth a look, all in published modules:

- `EventSummaryTable`'s props are a union discriminated on `compact`, so `items` narrows without a cast and a caller can't pair the wrong list with the wrong mode. Callers whose mode is dynamic spread one object.
- `hoveredEventId` was a bindable prop on `EventSummaryTable` but is only ever set by its own rows, so it became local state. No consumer in this repo or cloud-ui referenced it.
- `enrichGroups` becomes `setPendingMetadata`. Pending state is now applied during the initial load rather than only after it completes.

cloud-ui imports three of the files this touches — the two view layouts and `workflow-run-layout` — and none of their prop contracts changed.

## Risk

A `LazyGroup` and its materialized `EventGroup` must agree field for field, or a view filters on one and renders the other. Rather than test two implementations against each other, `isPending` and `category` now share one function, so those cannot diverge; the rules themselves have direct tests. The remaining fields are compared at every prefix of every group type, over a field list whose `satisfies Record<keyof LazyGroup, boolean>` makes TypeScript reject a new `LazyGroup` field until it is classified.


src/lib/services/grouped-event-buffer.ts
```diff
@@ -257,868 +306,308 @@ function hasTerminalNexusEvent(group: EventGroup): boolean {
...
+  if (!(slot >= 0)) return false;
 
-  eventSlots[slotIdx] = raw;
-
-  if (slotIdx > latestEventSlotIdx) {
-    latestEventSlotIdx = slotIdx;
-    latestEventRef = raw;
-  }
+  grow(slot);
+  if (events[slot]) return false;
```
laurakwhit, line 410: Is there a potential issue r.e. total billable events if we return before calling `toWorkflowEvent` here?

src/lib/components/event/event-summary-table.svelte
```diff
@@ -117,42 +142,43 @@
       </TableHeaderRow>
     {/snippet}
     {#snippet rows({ visibleItems })}
-      {#each visibleItems as event, index (iterableKey(event))}
-        {#if isEventGroup(event)}
+      {#each visibleItems as item, index (iterableKey(item))}
+        {@const row = isLazyGroup(item) ? materializeGroup(item) : item}
```
laurakwhit: This never re-derives. Is that expected?

src/lib/services/grouped-event-buffer.test.ts
```diff
@@ -1,43 +1,33 @@
...
-  onLatestGroup,
-  processEvent,
+  type LazyGroup,
+  materializeGroup,
   reset,
-  resetLive,
-  setEstimatedGroupCount,
   setFailedEvent,
+  setPendingMetadata,
+  SHARED_WITH_EVENT_GROUP,
```
laurakwhit: Duplicate import 🗑️ 
```suggestion
```

src/lib/utilities/sort-timeline-groups.ts
```diff
@@ -19,34 +23,37 @@ import { orderGroupsByPending } from './order-groups-by-pending';
  * @param descMinId     - lowest event ID seen from the descending cursor;
  *                        0 means no descending page has arrived yet
  */
-export const sortGroupsDuringLoading = (
-  groups: EventGroups,
+export const sortGroupsDuringLoading = <T extends SortableGroup>(
```
laurakwhit, line 26: How does this differ from `orderGroupsByPending`?

src/lib/components/lines-and-dots/timeline-graph/timeline-graph.svelte
```diff
@@ -410,24 +415,32 @@
...
-      if (prev && prev.index === index && prev.group === group) {
-        slots[slot] = prev;
+      const slotIndex = index % poolSize;
+      const lazy = filteredLazyGroups[index];
+      const prev = prevSlots[slotIndex];
+      if (
+        prev &&
+        prev.index === index &&
+        prev.lazy === lazy &&
+        prev.version === lazy.version
```
laurakwhit, line 439: Should we make version be required if we're using it as a check here?

src/lib/services/grouped-event-buffer.ts
```diff
@@ -28,184 +34,229 @@ import {
...
+
+  get classification(): WorkflowEvent['classification'] {
+    return this.initialEvent.classification;
+  }
+
+  get finalClassification(): WorkflowEvent['classification'] {
+    return this.lastEvent.classification;
+  }
+
+  get isPending(): boolean {
```
laurakwhit, line 161: Potential issue flagged by Claude ⬇️ 
> `GroupRecord.pendingActivity`/`pendingNexusOperation` are only refreshed when the head event lands or when `setPendingMetadata` is re-called, so once a terminal event arrives the lazy `isPending` getter stays `true` while `materializeEventGroup` (which gates on `hasTerminalActivityEvent`/`hasTerminalNexusEvent`) reports `false` — breaking the stated LazyGroup/EventGroup invariant that views filter on one and render the other.

src/lib/services/grouped-event-buffer.ts
tegan-temporal: if an event is in events[slot] it means it was ingested already which accounts for the total billable events.

src/lib/components/event/event-summary-table.svelte
tegan-temporal: No that's a good catch. to fix it, I've updated it so the iteration key includes the version so it gets outright replaced when it changes.

src/lib/utilities/sort-timeline-groups.ts
tegan-temporal: `orderGroupsByPendin`g partitions on `isPending` alone. This one also requires the head to sit in the region the descending cursor has already covered, so groups in the ascending section that only look open because their companion hasn't loaded yet stay in event-ID order instead of jumping to the top. Once fetchComplete is true we call `orderGroupsByPending` directly.

src/lib/components/lines-and-dots/timeline-graph/timeline-graph.svelte
tegan-temporal: it doesn't need to exist if it's immutable (from graph-widget), `undefined === undefined` is a valid comparison

src/lib/services/grouped-event-buffer.ts
tegan-temporal: Fixed

src/lib/components/event/event-summary-table.svelte
```diff
@@ -76,10 +101,21 @@
...
+    }
+
+    if (isPendingActivity(event)) {
+      return `pending-activity-${event.id}`;
+    }
+
+    if (isLazyGroup(event)) {
+      // Version, because a lazy group keeps its identity as it gains events —
+      // key on id alone and the row below renders a stale materialized group.
+      return `group-${event.id}-${event.version}`;
```
laurakwhit: NIT: If the version changes (e.g. the group gains an event) does this mean it's going to remount and potentially lose the "expanded" state for the row?

src/lib/services/grouped-event-buffer.ts
laurakwhit: Got it. Might be worth adding in test case for this so we know to keep it before `toWorkflowEvent`.

laurakwhit (approved):

src/lib/services/grouped-event-buffer.ts
tegan-temporal: added

src/lib/components/event/event-summary-table.svelte
tegan-temporal: yes! I reverted that change and moved to use a `materializeRowReactive` function instead of changing the key.

PR #3795, opened 2026-08-05
Split worker deployment API response types

rossnelson:
## Summary

Separates the ListWorkerDeployments and DescribeWorkerDeployment response models so each matches its API shape. Updates deployment consumers and compute-provider helpers to use endpoint-specific fields, and replaces unsafe fixture casts with shape-checked fixtures.

## Test plan

- [x] Full Vitest suite: 184 files passed
- [x] Type checking: 0 errors
- [x] Prettier, ESLint, and Stylelint passed for touched files
- [x] Independent architecture and code reviews approved


andrewzamojc (approved):
LGTM

PR #3799, opened 2026-08-06
fix(schedules): resolve recent run statuses from visibility

tegan-temporal:
Recent run statuses on the schedule detail page can be wrong or significantly delayed -- paused statuses don't show as paused and status changes might not show up for minutes or more. While ultimately we are just rendering `info.recentActions.startWorkflowStatus` we can at least side-step the problem by relying on the visibility store for the specific workflows so our status reflects the actual status of the workflow. 

- Resolve the status from visibility instead, falling back to the recorded one whenever the query hasn't answered or a run isn't indexed yet
- Scope the query to the runs on screen and exclude `ContinuedAsNew`, bounding the result to one execution per row — visibility results can't be sorted, so a page must never be able to truncate them
- Add `fetchWorkflowsForQuery`, which rejects instead of raising a toast so a failed enrichment query degrades silently
- Unit tests for the run mapping, status merge, and query construction

https://github.com/user-attachments/assets/96ebea5f-2a36-4ef7-a1b9-ddf8da6b0805

### DIFF
--- a/src/lib/components/schedule/schedule-view/workflow-runs-recent.svelte
+++ b/src/lib/components/schedule/schedule-view/workflow-runs-recent.svelte
@@ -5,10 +5,15 @@
   import { timestamp } from '$lib/components/timestamp.svelte';
   import Link from '$lib/holocene/link.svelte';
   import { translate } from '$lib/i18n/translate';
-  import type { DescribeFullSchedule } from '$lib/types/schedule';
-  import { getEpochMilliseconds } from '$lib/utilities/format-time';
+  import {
+    fetchRecentScheduleRunStatuses,
+    toRecentScheduleRuns,
+  } from '$lib/services/schedule-service';
+  import type {
+    DescribeFullSchedule,
+    RecentScheduleRun,
+  } from '$lib/types/schedule';
   import { routeForWorkflow } from '$lib/utilities/route-for';
-  import { toWorkflowStatusReadable } from '$lib/utilities/screaming-enums';
 
   import WorkflowRunsEmpty from './workflow-runs-empty.svelte';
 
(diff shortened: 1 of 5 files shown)


src/lib/services/schedule-service.ts
```diff
@@ -94,6 +100,90 @@ export async function fetchSchedule(
...
+): Promise<RecentScheduleRun[]> => {
+  const workflowIds = [
+    ...new Set(runs.map((run) => run.workflowId).filter(Boolean)),
+  ];
+  if (!workflowIds.length) {
+    return runs;
+  }
+
+  const ids = workflowIds
+    .map((workflowId) => `"${workflowId.replace(/"/g, '\\"')}"`)
```
laurakwhit: NIT: Might want to account for id that could end in a backslash.

src/lib/services/schedule-service.ts
```diff
@@ -94,6 +100,90 @@ export async function fetchSchedule(
...
+      !current ||
+      getEpochMilliseconds(workflow.startTime) >=
+        getEpochMilliseconds(current.startTime);
+    if (isNewer) {
+      latest.set(workflow.id, workflow);
+    }
+  }
+
+  return runs.map((run) => {
+    const workflow = latest.get(run.workflowId);
```
laurakwhit: We should probably use the runId, as well as the workflowId, to find the correct workflow here.

src/lib/components/schedule/schedule-view/workflow-runs-recent.svelte
```diff
@@ -28,20 +33,19 @@
...
+          namespace,
+          scheduleId: schedule.schedule_id,
+          runs: recordedRuns,
+        })
+      : Promise.resolve(recordedRuns),
+  );
 </script>
 
-{#if !sortedRecentRuns.length}
+{#if !recordedRuns.length}
```
laurakwhit, line 46: NIT: Looks like we guard against empty runs in three places (here, the ternary in `runsPromise` and the early return in `fetchRecentScheduleRunStatuses`). Maybe we could simplify?

tegan-temporal: I'll remove 1 of the guards (the runsPromise ternary)

src/lib/services/schedule-service.ts
tegan-temporal: I'll improve the backslash escaping

src/lib/services/schedule-service.ts
tegan-temporal: Our query is `TemporalScheduledById="${scheduleId}" AND WorkflowId in (...) AND ExecutionStatus != "ContinuedAsNew"`, which filters out the older runs for the same workflows. I think this is what we want, if we used runId a `ContinuedAsNew` status would be stuck there instead of the status the user actually cares about (the latest run for that workflow).

It could introduce some confusion since the schedule response points at a runId that's different from the run we're showing the status for, but it still seems like the more useful thing to show. What do you think?

laurakwhit: I think I was more so thinking that we aren't overriding the runId like we are with status on line 144, which means we could be leaving the runId as the originally recorded first run and it could potentially not be the same as the runId in the workflow we get from latest. Does that make sense? Maybe we just override runId here?

```js
return { ...run, status: workflow.status, runId: workflow.runId }.
```

tegan-temporal: yeah that makes sense!

laurakwhit: Actually if we're using the `...run` we should probably make sure the runId matches whatever workflow we get from latest and not just override the runId 😅

laurakwhit (approved):

PR #3804, opened 2026-08-10
feat(DT-4359): Schedules content updates

tegan-temporal:
Content and label updates for the schedule detail page.

- Recent/Upcoming run tabs show counts, e.g. "5 Recent Runs" / "3 Upcoming Runs"
- "View All Runs" link becomes "View all N runs" (sentence case), using the
  schedule's total workflow count
- Workflow total is now labeled "Total Workflows (All Time)"
- Reworded the upcoming-runs empty state to point at the Schedule Spec
- Extracted `toUpcomingScheduleRuns` alongside `toRecentScheduleRuns` so the card
  and the upcoming panel share one definition of the list; the default tab now
  derives from the filtered recent-run count


laurakwhit (approved):

KATIETOLER (approved):
super cool, lgtm

PR #3806, opened 2026-08-10
fix(auth): derive auth cookie Secure attribute from the OIDC callback URL

ardiewen:
## Description & motivation 💭

The auth cookies (`user*`, `refresh`, `session_start`, OAuth `state`/`nonce`, and the
logout clears) took their `Secure` attribute from `c.Request().TLS` — ui-server's own
inbound connection, which isn't necessarily how the browser reaches the deployment.

Derive it from the scheme of the configured OIDC callback URL instead: that URL is the
browser-facing address the identity provider redirects back to, so it's the
authoritative record. `https` callback ⇒ cookies are `Secure`; `http` ⇒ they aren't.
No new configuration, and this brings the auth cookies in line with `_csrf`.

An unrecognized callback-URL scheme is now a startup error rather than a silent
`false`, so a malformed `callbackUrl` fails loudly.

### Alternatives considered

**Reusing `CORS.CookieInsecure`** (i.e. `secure := !serverCfg.CORS.CookieInsecure`) was
the first approach, and mirrors what the CSRF cookie already does in `server.go`. We
rejected it on three counts:

- **It overloads a knob that is documented and surfaced as CSRF-specific.** The field
  comment says "allows CSRF cookie to be sent…", and `config/docker.yaml` exposes it as
  `TEMPORAL_CSRF_COOKIE_INSECURE`. Operators who set that for CSRF reasons would have
  been changing auth cookie behavior without knowing it.
- **It regresses a configuration that is correct today.** A deployment terminating TLS
  directly *and* setting `cookieInsecure: true` currently gets `Secure` auth cookies
  from `Request().TLS`; keying off the flag alone would take it away.
- **It breaks plaintext deployments on upgrade.** The shipped default is
  `cookieInsecure: false` (`development.yaml`, `e2e.yaml`, `with-auth.yaml`, and the
  docker template), so any HTTP deployment that never touched the knob would start
  emitting `Secure` cookies and fail to log in.

**A dedicated `auth.cookieInsecure` option** was also considered. It has the cleanest
semantics but adds a config key, an env var, and helm/docker template surface — and it
doesn't avoid the upgrade break, since plaintext deployments would still have to opt
out. The callback URL already states the deployment's topology; asking operators to
restate it is redundant.

**`X-Forwarded-Proto`** is client-supplied and would need trusted-proxy configuration
to rely on.

## Testing 🧪

- [x] Unit tests added

`TestCookieSecureForCallbackURL` covers scheme handling (https/http/uppercase/missing/
unsupported/unparseable); `TestSetUserSecureFlag` asserts the cookies honor the flag.
Full `server` suite passes.

### DIFF
--- a/server/server/auth/auth.go
+++ b/server/server/auth/auth.go
@@ -58,7 +58,7 @@ func stripBearerPrefix(token string) string {
 	return strings.TrimPrefix(token, "Bearer ")
 }
 
-func SetUser(c echo.Context, user *User) error {
+func SetUser(c echo.Context, user *User, secure bool) error {
 	if user.OAuth2Token == nil {
 		return errors.New("no OAuth2Token")
 	}
@@ -89,7 +89,7 @@ func SetUser(c echo.Context, user *User) error {
 			Name:     "user" + strconv.Itoa(i),
 			Value:    p,
 			MaxAge:   int(time.Minute.Seconds()),
-			Secure:   c.Request().TLS != nil,
+			Secure:   secure,
 			HttpOnly: false,
 			Path:     "/",
 			SameSite: http.SameSiteStrictMode,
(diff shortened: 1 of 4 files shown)


rossnelson (approved):
Thanks!

GiantRobots (approved):

PR #3816, opened 2026-08-12
Manual on-call dependency review with constrained Claude remediation

rossnelson:
## Description & motivation 💭

This branch makes the weekly dependency-security review a manual, dry-run-first on-call workflow. The later commits correct the planner, which did not act on any Dependabot alert before them.

Two credential boundaries hold the design together. Claude holds no credential and can edit one file. Only the second job holds a write token, and that job runs no package code.

```mermaid
flowchart TD
  subgraph J1["Job 1. Read-only token."]
    A["Audit the open Dependabot alerts"] --> B["Plan the remediation"]
    B --> C{"Action type"}
    C -->|npm| D["Claude edits package.json.<br/>Read and Edit only.<br/>No shell, no network."]
    D --> E["Byte-match gate.<br/>Reject anything the plan<br/>did not authorize."]
    E --> F["pnpm install --lockfile-only"]
    C -->|go| G["go get, then go mod tidy.<br/>The toolchain applies the change.<br/>No model reads a Go file."]
    F --> H{"Does every remediated<br/>alert close?"}
    G --> H
    H -->|no| X["Fail the run"]
    H -->|yes| I["Attest the file hashes"]
  end

  I --> J

  subgraph J2["Job 2. GitHub App token."]
    J{"Publish gate"}
    J -->|pass| K["Update one draft pull request"]
    J -->|fail| Y["Publish nothing"]
  end
```

Job 1 is the only job that installs or runs package code, and its token cannot write. Job 2 holds the write token and runs no package code. It receives the candidate files with their hashes, checks each hash again, and stages only the four approved files.

## The defect

`ROOT_MANIFEST` held the value `/package.json`. GitHub never sends that path. Every alert therefore classified as unsupported.

The path values across all 321 alerts in this repository:

| value | count |
| -- | -- |
| `pnpm-lock.yaml` | 254 |
| `server/go.mod` | 33 |
| `package.json` | 26 |
| `package-lock.json` | 3 |
| `e2e/go.mod`, `e2e/go.sum` | 4 |
| `server/go.sum` | 1 |
| `/package.json` | 0 |

The test suite passed because its fixture held the same wrong value. The tests checked the code against itself.

## What the planner now does

- It accepts the manifest paths GitHub sends, for npm and for Go.
- It matches an override selector that carries a version range. The earlier code split the selector on `>`, which also occurs in `>=`. 12 of the 36 selectors in this repository did not match. That fault also disabled the guard that refuses a direct bump which an override would supersede.
- It raises a compound override range and keeps the upper bound. The earlier code refused these ranges. A raise without this care produced `>=7.1.0 <7`, which no version satisfies.
- It reads `pnpm-lock.yaml` and reports the resolved version of each package. A new `alreadySafe` status covers a package that the advisory no longer reaches. The planner refuses a direct bump when the lockfile holds a copy that the bump cannot reach.
- It remediates Go modules in `server/go.mod`. A Go major version belongs to the module path, so the planner refuses a change of major version. It also refuses a pseudo-version, because two pseudo-versions do not compare as releases.

## Proof that a remediation closes its alert

A new script reads `pnpm-lock.yaml` and `server/go.mod` after the install. It compares each resolved version against the advisory range of every alert the plan remediated. The run fails when a copy stays inside the range, and it fails when the file no longer holds the package. The script writes the resolved version into the report.

The publish gate then requires that proof. It requires Claude authorship only for the npm actions, so the Go path does not need a claim that no model made:

```text
publish when
  the report says verification passed
  and the plan holds at least one action
  and if the plan holds an npm action
        Claude authored it and the byte-match gate accepted it
```

## Live result

The five open alerts produce one authorized action:

```
safe    protobufjs      alerts [338,300,299]  resolved 7.5.8   raise override to ^7.6.5
manual  @grpc/grpc-js   alerts [295,294]      resolved 1.14.3  no override exists
```

The two `@grpc/grpc-js` alerts need a new override. That work is out of scope here. See UI-139.

### How was this tested 👻

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

77 tests pass through `pnpm test:github-automation`. Prettier and ESLint report no problem. Both workflow files parse.

A local rehearsal covers the full path for both ecosystems against the live alert data.

npm:

1. The plan authorized one action.
2. The byte-match gate accepted the manifest and set `changed=true`, `applied=true`, and `resolver=claude-code`.
3. `pnpm install --lockfile-only --ignore-scripts` resolved protobufjs to 7.6.5.
4. Verification passed and recorded the resolved version for each of the three alerts.
5. The publish gate returned true.

Go:

1. A stale `server/go.mod` held `v0.51.0` and `v1.79.3`.
2. `go get` and `go mod tidy` moved both modules to `v0.52.0` and `v1.82.1`. Those versions match the versions the team chose by hand in 506d4f72c. The net difference is zero.
3. Verification passed for all five alerts.

The publish gate was checked directly with jq across nine cases:

| case | result |
| -- | -- |
| npm only, Claude authored, verified | publish |
| npm only, not Claude authored | block |
| npm only, wrong resolver | block |
| go only, verified | publish |
| go only, not verified | block |
| mixed, Claude authored, verified | publish |
| mixed, not Claude authored | block |
| no action | block |
| no verification block | block |

Run the rehearsal. Open Actions, select **Weekly On-Call Review PR Test**, select **Run workflow** on this branch, and set the mode to `apply`.

The rehearsal sets the new `publish` input to false, so the job that holds the GitHub App token does not run and no pull request is touched. The publish job also refuses a pull request event, so the rehearsal has two independent guards.

### Merge Checklist

- [ ] Run the apply-mode rehearsal on this branch. Three items need a real run: the Claude action itself, `pnpm install --frozen-lockfile --ignore-scripts` followed by the tests and the build, and the artifact transfer of `server/go.mod` to the publish job.
- [ ] Confirm that the install without lifecycle scripts still builds. `onlyBuiltDependencies` names `esbuild`, `@swc/core`, and `@temporalio/core-bridge`, and the install skips all three.

### Issue(s) closed

UI-132, UI-133, UI-134, UI-136. UI-135 stays open for the digest work.

### Any docs updates needed?

`.github/WORKFLOWS.md` describes the release workflows only. It does not describe the weekly review. A separate change should add that section.

### DIFF
--- a/.github/scripts/apply-weekly-dependency-remediation.mjs
+++ b/.github/scripts/apply-weekly-dependency-remediation.mjs
@@ -10,7 +10,7 @@
 import { readFile, writeFile } from 'node:fs/promises';
 import path from 'node:path';
 
-const ROOT_MANIFEST = '/package.json';
+const REMEDIABLE_MANIFESTS = new Set(['package.json', 'pnpm-lock.yaml']);
 const DIRECT_DEPENDENCY_SECTIONS = ['dependencies', 'devDependencies'];
 const VERSION =
   /^(?<major>0|[1-9]\d*)\.(?<minor>0|[1-9]\d*)\.(?<patch>0|[1-9]\d*)(?:-[0-9A-Za-z.-]+)?(?:\+[0-9A-Za-z.-]+)?$/;
@@ -52,6 +52,37 @@ export function isVersionAtLeast(version, floor) {
   );
 }
 
+const RANGE_CONSTRAINT = /^(>=|<=|>|<|=)?\s*(.+)$/;
+
+function satisfiesConstraint(comparison, operator) {
+  if (operator === '<') return comparison < 0;
+  if (operator === '<=') return comparison <= 0;
+  if (operator === '>') return comparison > 0;
+  if (operator === '>=') return comparison >= 0;
+  return comparison === 0;
+}
+
+export function isVersionVulnerable(version, range) {
+  const target = parseSemver(version);
+  if (!target || typeof range !== 'string') return true;
+  const constraints = range
+    .split(',')
+    .map((part) => part.trim())
+    .filter(Boolean);
+  if (constraints.length === 0) return true;
+
+  for (const constraint of constraints) {
+    const match = RANGE_CONSTRAINT.exec(constraint);
+    if (!match) return true;
+    const bound = parseSemver(match[2]);
+    if (!bound) return true;
+    if (!satisfiesConstraint(compareSemver(target, bound), match[1] ?? '=')) {
+      return false;
+    }
+  }
+  return true;
+}
+
 function rangeLowerBound(range) {
   if (typeof range !== 'string') return null;
   const match = VERSION_IN_RANGE.exec(range.trim());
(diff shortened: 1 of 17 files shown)


.github/scripts/apply-weekly-dependency-remediation.mjs
```diff
@@ -0,0 +1,539 @@
...
+ * Plans the smallest safe package.json change for Dependabot alerts.
+ *
+ * This script deliberately does not run a package manager or inspect/edit a
+ * lockfile. The workflow owns `pnpm install --lockfile-only` after a reviewed
+ * plan has changed the manifest.
+ */
+import { readFile, writeFile } from 'node:fs/promises';
+import path from 'node:path';
+
+const ROOT_MANIFEST = '/package.json';
```
laurakwhit: Looks like GitHub returns a manifest_path without a leading slash, so probably need to remove it here. 

```suggestion
const ROOT_MANIFEST = 'package.json';
```

laurakwhit: Also, should we be checking `pnpm-lock.yaml` as well here?

.github/scripts/weekly-dependency-security-audit.mjs
```diff
@@ -0,0 +1,627 @@
...
+  }
+
+  return { kind: 'other', ticket: null, reason: null };
+};
+
+export const isRemediationPullRequest = (
+  pullRequest,
+  remediationBranch = DEFAULT_REMEDIATION_BRANCH,
+) =>
+  pullRequest.head?.ref === remediationBranch ||
```
laurakwhit: This reads `pullRequest.head?.ref` but it looks like it's called with normalized PR objects that only expose `headRef`. 

Might need something like 
```js
  const headRef = pullRequest.headRef ?? pullRequest.head?.ref;
```

.github/scripts/apply-weekly-dependency-remediation.mjs
```diff
@@ -0,0 +1,539 @@
...
+function findDirectDependency(packageJson, packageName) {
+  for (const section of DIRECT_DEPENDENCY_SECTIONS) {
+    const version = packageJson?.[section]?.[packageName];
+    if (typeof version === 'string') return { section, version };
+  }
+  return null;
+}
+
+function overrideTargetsPackage(selector, packageName) {
+  const target = selector.split('>').at(-1)?.trim();
```
laurakwhit: This takes everything after the last >. For a range-qualified one (e.g. `example@>=10`) would this break?

.github/scripts/apply-weekly-dependency-remediation.mjs
```diff
@@ -0,0 +1,539 @@
...
+  );
+}
+
+function rangeLowerBound(range) {
+  if (typeof range !== 'string') return null;
+  const match = VERSION_IN_RANGE.exec(range.trim());
+  if (!match?.groups) return null;
+  return {
+    operator: match.groups.operator ?? '',
+    version: `${match.groups.version}.${match.groups.minor}.${match.groups.patch}`,
```
laurakwhit: This captures the prerelease tag but rebuilds version as major.minor.patch only. Do we want to preserve the prerelease here?

laurakwhit (approved):
Found some things we might want to investigate 🕵️ but otherwise 🚀 

In `workflows/weekly-dependency-security-review.yml` the artifact name uses `github.run_id` and `upload-artifact@v4+` rejects a duplicate artifact name within a run, so re-running the failed job will likely die at the upload step.

Also might be worth looking into the `item.package ?? item.packageName ?? item.title ?? String(item)` chain. Looks like it might fall through to `String(item)` because `classification()` always sets `packageName`, using `null` when no package was identified.

.github/scripts/apply-weekly-dependency-remediation.mjs
rossnelson: I've got another iteration on this. Its going to add that as well as the go deps.

rossnelson:
Both findings were real. Thank you. Fixed in b068f5eff.

**Artifact name.** You are right. `upload-artifact` refuses a name that already exists in the same run, the upload step carries `if: always()`, and `github.run_id` does not change between attempts. So attempt one uploaded the artifact even when the job failed, and the re-run died at the upload step. The name now also holds `github.run_attempt`.

The same fault was present in `weekly-vln-review.yml` and `weekly-external-contributor-review.yml`. Both are fixed in the same commit.

Both re-run paths now behave. A re-run of only the failed publish job reads `artifact_name` from the successful audit job, so it still finds the artifact that attempt one uploaded. A re-run of the audit job computes a new name and uploads to it.

**The fallback chain.** Your reading is exact. I reproduced it:

```
• [object Object] — Alert did not identify an npm package.
```

`classification()` always sets `packageName`, and it sets null when the alert names no package. A classification also carries `alertIds`, not `number`, so the first branch does not apply and the chain reaches `String(item)`.

The label now falls back to the alert numbers, then to a fixed string:

| classification | label |
| -- | -- |
| no package name, one alert id | `#77` |
| no package name, several ids | `#77, #78` |
| no package name, no ids | `unidentified alert` |
| a named package | unchanged |

A test covers the case. 78 tests pass.

One item is still open, and it is not from your review. No apply-mode run has executed, so the Claude step, the install without lifecycle scripts, and the artifact transfer of `server/go.mod` are unproven. The merge checklist holds both items. You can run the rehearsal from Actions, **Weekly On-Call Review PR Test**, with the mode `apply`. It sets the new `publish` input to false, so no pull request is touched.

PR #3822, opened 2026-08-14
refactor(icons)!: replace holocene's icon registry with Io icon components

tegan-temporal:
## Description & motivation 💭

Replaces holocene's name-based icon registry with the Io design system's icon components, and deletes the registry.

`<Icon name="close" />` resolved a glyph at runtime from `holocene/icon/paths.ts`. Icons are now individual components:

```svelte
<IconClose />
<Button LeadingIcon={IconAdd}>Add</Button>
```

- 222 Io icons under `src/lib/io/icon/icons`, each a component around a shared `IconSvgWrapper`
- `holocene/icon/` deleted
- Icons render at `1.143em`, so they scale with the surrounding text instead of a flat 16px
- SDK logos, the Temporal mark and the login provider marks come from Io rather than static SVGs

Paired with temporalio/cloud-ui#3111, which migrates cloud-ui's call sites and can't merge until this ships in a release.

### Screenshots (if applicable) 📸

Some glyphs differ from the ones they replaced — Chromatic will show them. The most visible is `spinner`, now a filled arc rather than a stroked three-quarter arc, which appears in every loading `Button`.

### Design Considerations 🎨

Icon sheet in Figma: https://www.figma.com/design/4j6ZH2Fvoh74MPz19e3YyC/Io-Design-Library?node-id=1212-13530&m=dev

Every icon here is drawn by design. Two glyphs still have no Io equivalent and use non-Io artwork: a hamburger/bars for cloud-ui's mobile nav, and the solid directional triangles in `holocene/keyboard-shortcut/arrow-*` (Io now has `play-solid`, but only right-facing).

## Breaking changes for consumers

- `holocene/icon` is gone. Import from `$lib/io/icon` and pass components, not names.
- `Button`/`Link`/`IconButton` take `LeadingIcon`/`TrailingIcon`/`Icon` component props. `Link`'s deprecated `icon` prop is removed.
- `NavigationButton`'s `animate`/`active` props are removed; they were spread onto the `<svg>` as inert attributes.
- `IconButton`'s `data-track-intent` is now `{variant}-{label}` — a component reference can't be interpolated into a string.

### How was this tested 👻

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

`pnpm check`, `pnpm lint`, the unit suite and the integration suite all pass. Every migrated surface was checked in a browser. No new tests — this swaps one icon implementation for another and adds no logic.

### Steps for others to test: 🚶🏽‍♂️🚶🏽‍♀️

[...]

PR #3852, opened 2026-08-20
feat(deployments): show worker deployment connection status by default

rossnelson:
## Summary

- show the connection status by default: the Connection column on the
  deployment detail page, and the status inside the compute pill on the
  deployments list
- keep `showConnectionStatus`, so a consumer can still hide the status without
  a release of this package
- restore the Validate Connection guidance in the Pending tooltip
- report a check that does not finish as unknown, rather than as an invalid
  connection
- extend the unit and integration tests to cover the re-exposed path, which
  includes the detail row colspan

https://github.com/user-attachments/assets/15388342-e1a3-48ac-938e-ddc3ba8c14a8

## Why

PR #3718 hid the connection status behind `showConnectionStatus`, which
defaults to false. The UI was correct. Two backend defects made the status show
a permanent Pending:

- `validateValidateSpec` rejected an empty `ValidateSpec`, so
  `handleValidateSpec` never ran.
  Fix: temporalio/temporal-auto-scaled-workers#99, merged 2026-07-25.
- `version_workflow.go` never sent `syncSummary()`, so an updated
  `ComputeStatus` never reached the deployment workflow that
  `DescribeWorkerDeployment` reads.
  Fix: temporalio/temporal#11273, merged 2026-08-03.

Both fixes are merged.

## Where the flip belongs

The prop defaults to false, and no route in this repo passes it, so the status
never appears. Three options:

1. **Flip the default here.** One change, one repo, and every consumer gets the
   status. This PR takes this option.
2. **Pass true from each consumer.** This needs a change in every consumer plus
   a package bump, and it leaves a dead prop here. It does not solve the staged
   rollout problem either, because a consumer also deploys everywhere at once.
3. **Gate on a server capability.** This is the only mechanism that lets an
   environment with the fixes show the status while an environment without them
   hides it, because the server reports its own support. No such capability
   exists. `serverScaledDeployments` is true with or without these fixes, so it
   selects the broken set rather than the fixed set. A capability that did track
   the fixes needs a change to temporalio/api and to the server, which takes
   longer than the rollout it would guard.

Option 1 with the prop kept gives the same rollback lever at lower cost: a
consumer can pass `showConnectionStatus={false}` if an environment turns out to
be behind.

## Pending tooltip copy

Commit e6efe42a9 removed "or use Validate Connection to check now" because the
on-demand validation was a dry run. temporal-auto-scaled-workers#99 makes that
statement correct again. `handleValidateSpec` now writes the validation status
and signals the version workflow when the request carries no scaling group
changes, which is what the manual action sends. The row already refreshes after
a validation, so the badge changes in place.

The 6h claim stays. The default of
`workercontroller.periodic_validation_interval_s` is 21600 seconds.

## Client contract

`validateCurrentWorkerDeploymentVersionComputeConfig` keeps its `void` return.
`ValidateWorkerDeploymentVersionComputeConfigResponse` is an empty message, so
the earlier `{ valid, message }` shape never matched the server. The verdict
does not need that shape.

## A check that does not finish is not a failed connection

The modal treated every error as a verdict:

```js
const isValid = $derived(!result?.message);
```

A gateway timeout therefore read as "Connection is invalid", which is wrong. The
check never returned a result, so the connection state is unknown. The endpoint
runs a workflow update that calls a provider activity while the caller waits, so
a slow provider producing a timeout is an ordinary outcome, not an edge case.

The modal now has three states:

| Outcome | Trigger | Modal |
| --- | --- | --- |
| valid | 200 | Connection is valid |
| invalid | 400 only | Connection is invalid, plus the server message |
| indeterminate | 5xx, transport failure, 401, 403, 404, 429 | Could not complete the check, status unknown |

`InvalidArgument`, which maps to 400, is the only status the backend uses for a
verdict, so it is the only one treated as one. `resolveValidationOutcome` holds
this mapping so it is testable away from the component.

The row now refreshes for every outcome. A check that does not return to the
browser can still finish and persist a status.

## Issue(s) closed

FE-191

## Test plan

- [x] `pnpm lint`: 0 errors
- [x] `pnpm check`: 0 errors
- [x] `pnpm test -- --run`: 3041 passed, 2 skipped
- [x] `pnpm test:integration`: 317 passed, 1 skipped
- [ ] manual check against an environment carrying both backend fixes

### DIFF
--- a/src/lib/pages/deployment.svelte
+++ b/src/lib/pages/deployment.svelte
@@ -27,7 +27,7 @@
     showConnectionStatus?: boolean;
   }
 
-  let { showInstancesLink = true, showConnectionStatus = false }: Props =
+  let { showInstancesLink = true, showConnectionStatus = true }: Props =
     $props();
 
   const { namespace } = $derived(page.params);
--- a/src/lib/pages/deployment.svelte
+++ b/src/lib/pages/deployment.svelte
@@ -18,6 +18,7 @@
     setRampingUnversionedWorkers,
   } from '$lib/services/deployments-service';
   import type { DescribeWorkerDeploymentResponse } from '$lib/types/deployments';
+  import { deploymentShowsConnectionStatus } from '$lib/utilities/connection-status';
   import { deploymentHasComputeConfig } from '$lib/utilities/deployment-has-compute-config';
   import { decodeURIForSvelte } from '$lib/utilities/encode-uri';
   import { routeForWorkerDeployments } from '$lib/utilities/route-for';
@@ -131,6 +132,8 @@
   {/if}
   {@const info = deployment.workerDeploymentInfo}
   {@const hasComputeConfig = deploymentHasComputeConfig(info)}
+  {@const connectionColumnVisible =
+    showConnectionStatus && deploymentShowsConnectionStatus(info)}
   {@const unversionedRampingPercentage =
     !info.routingConfig?.rampingDeploymentVersion &&
     info.routingConfig?.rampingVersionPercentage != null
@@ -179,7 +182,7 @@
           <th>{translate('deployments.build-id')}</th>
           <th>{translate('deployments.lifecycle')}</th>
           <th>{translate('deployments.compute')}</th>
-          {#if showConnectionStatus}
+          {#if connectionColumnVisible}
             <th>{translate('deployments.connection')}</th>
           {/if}
           <th>{translate('deployments.deployed')}</th>
@@ -194,7 +197,7 @@
             {namespace}
             {deploymentName}
             conflictToken={deployment.conflictToken}
-            {showConnectionStatus}
+            showConnectionStatus={connectionColumnVisible}
             onChange={reload}
             onValidationComplete={reload}
           />
--- a/src/lib/components/deployments/deployment-client-test-runner-entry.ts
+++ b/src/lib/components/deployments/deployment-client-test-runner-entry.ts
@@ -50,7 +50,7 @@ export async function renderDeployment({
 }: {
   fetchDeployment: unknown;
   fetchDeploymentVersion: unknown;
-  showConnectionStatus: boolean;
+  showConnectionStatus?: boolean;
 }) {
   setPageState();
   resetDeploymentsServiceMock({ fetchDeployment, fetchDeploymentVersion });
@@ -59,7 +59,7 @@ export async function renderDeployment({
   mounted.push(
     mount(Deployment, {
       target,
-      props: { showConnectionStatus },
+      props: showConnectionStatus === undefined ? {} : { showConnectionStatus },
     }),
   );
   await Promise.resolve();
@@ -90,7 +90,7 @@ export function openMenu(controls: string) {
 
 export function renderRows({
   computeStatuses,
-  showConnectionStatus = false,
+  showConnectionStatus,
 }: {
   computeStatuses: Record<string, ComputeStatus>;
   showConnectionStatus?: boolean;
@@ -106,7 +106,9 @@ export function renderRows({
         props: {
           deployment: deployment(status.toLowerCase(), computeStatus),
           columns: [{ label: 'Current Version' }],
-          showConnectionStatus,
+          ...(showConnectionStatus === undefined
+            ? {}
+            : { showConnectionStatus }),
         },
       }),
     );
--- a/src/lib/components/deployments/deployment-table-row.svelte
+++ b/src/lib/components/deployments/deployment-table-row.svelte
@@ -35,7 +35,7 @@
   let {
     deployment,
     columns,
-    showConnectionStatus = false,
+    showConnectionStatus = true,
     onChange,
   }: Props = $props();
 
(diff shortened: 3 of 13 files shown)


rossnelson:
## Preview is blocked by an unrelated breaking change on main

The first preview build failed, and the cause is not in this PR.

Run: https://github.com/temporalio/cloud-ui/actions/runs/32403417274 (the pack
build succeeded)
Generated PR: temporalio/cloud-ui#3130, branch `test-ui-92f4af39`

On that PR, Vercel, `check-types`, and `unit-tests` all fail with errors of this
shape:

```
Object literal may only specify known properties,
and '"trailingIcon"' does not exist in type 'ButtonWithHrefProps'.
```

These come from #3822, `refactor(icons)!: replace holocene's icon registry with
Io icon components`, which merged earlier today as commit 403e5b975. cloud-ui
still uses the removed props: 27 uses of `trailingIcon` and 50 of `leadingIcon`.
This branch starts from 4 commits after that refactor, so it inherits the
breakage.

Evidence that this PR is not the cause: the last test PR before the refactor,
temporalio/cloud-ui#3128 from ui commit `99b026bf`, passed Vercel.
Commit 99b026bf is not a descendant of 403e5b975. The failing `check-pack` job
is also normal for a test PR, because it reports the expected `.ui-sha` change.

To see the feature, a second preview builds the same two commits on the commit
before the icon refactor. It is branch `preview/reexpose-connection-status-pre-icons`,
and it is a throwaway for the preview only. Its deployment change set is
identical to this PR: 4 changed lines in each of the four gated files and the
same string change.

cloud-ui needs an update for #3822 before a preview of this branch as it stands
can build.

rossnelson:
## Preview is live, built from a pre-refactor base

Preview URL: https://cloud-ui-git-test-ui-54483a7f.preview.thundergun.io
Generated PR: temporalio/cloud-ui#3131, branch `test-ui-54483a7f`
Run: https://github.com/temporalio/cloud-ui/actions/runs/32404421290

Vercel reports `DEPLOYED` and the URL answers 200.

This preview carries the same two commits as this PR, applied to `dd80eca05`,
which is the commit before the icon refactor. The change set is identical: 4
changed lines in each of the four gated files, plus the same string change. Its
branch and its cloud-ui PR are throwaway. Delete both after use.

What the preview shows is the UI. It does not confirm the feature end to end.
The cell it points at is not known to run 3.161. Against an older cell the badge
shows a permanent Pending, which is the backend gap this PR waits on, not a UI
defect.

Still to check by hand, with a signed-in session:

- [ ] Connection column on the detail page, status inside the compute pill on
      the list page
- [ ] Pending, Connected, and Failed states, each with the correct colour and
      tooltip
- [ ] Validate Connection changes the badge in place after the row reloads
- [ ] the restored Pending tooltip sentence reads correctly at width

rossnelson:
## Rebuilt preview, with the gateway timeout fix

Preview URL: https://cloud-ui-git-test-ui-2bff5f80.preview.thundergun.io
Generated PR: temporalio/cloud-ui#3133, branch `test-ui-2bff5f80`
Run: https://github.com/temporalio/cloud-ui/actions/runs/32430114522

Vercel reports `DEPLOYED` and the URL answers 200. This build supersedes
cloud-ui#3131, which predates commit `0aa28cf3`.

To check the corrected path, run Validate Connection against a version whose
provider is slow enough to time out at the gateway. The modal must read "Could
not complete the check" with a neutral icon, must not read "Connection is
invalid", and the badge must hold the status the server last reported.

src/lib/pages/deployment.svelte
```diff
@@ -27,7 +27,7 @@
     showConnectionStatus?: boolean;
   }
 
-  let { showInstancesLink = true, showConnectionStatus = false }: Props =
+  let { showInstancesLink = true, showConnectionStatus = true }: Props =
```
laurakwhit, line 30: Do we want to gate `showConnectionStatus` to `hasComputeConfig` on the `/namespaces/[namespace]/workers/deployments/[deployment]` page since it now defaults to `true` here? Or do we want to show the Connection column for for self-hosted/non-serverless deployments too?

laurakwhit (approved):

andrewzamojc (approved):
LGTM 👍

src/lib/pages/deployment.svelte
rossnelson: Good catch — gated it, but on the versions in the table rather than on `hasComputeConfig`.

A self-hosted deployment has no provider to report on, so the column held nothing but em dashes. `deploymentShowsConnectionStatus(info)` now returns true only when at least one version would render a badge, and the page passes that result down as `showConnectionStatus`, so the header, the cells, and the detail row colspan stay in agreement.

I did not reuse `hasComputeConfig` because it reads only the current and ramping versions. A draining compute-backed version sitting under a self-hosted current version would lose a status it actually has. The new helper mirrors the row's own rule — a compute provider plus current, ramping, or draining — and the row now calls the same function, so the cell rule cannot drift from the column rule.

The prop stays as the consumer-facing opt-out: `showConnectionStatus={false}` hides the column on every deployment.

0f8170c37

PR #3861, opened 2026-08-25
Make only keyboard focus pin Tooltip open

laurakwhit:
## Description & motivation 💭 
`handleFocusIn` previously set `isFocused = true` for any focus. In Chrome and Firefox, clicking a `<button>` focuses it, so a click left `isFocused` stuck `true` and the Tooltip stayed up until focus moved elsewhere. 

`:focus-visible` is `true` for tab and other keyboard focus but `false` for a plain mouse click on a button. The `isKeyboardFocus` guard forwards keyboard focus, but drops mouse focus and still meets requirements for WCAG 1.4.13 (Content on Hover or Focus). 

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

* Go to a button with a tooltip (e.g. Saved Views buttons in the collapsed view) > click the button
   - [ ] Verify tooltip goes away without having to click focus away from the button
* Tab to the button
   - [ ] Verify tooltip shoes up on button

### DIFF
--- a/src/lib/holocene/tooltip.svelte
+++ b/src/lib/holocene/tooltip.svelte
@@ -125,7 +125,7 @@
     }
   }
 
-  function isKeyboardFocus(target: EventTarget | null) {
+  function isFocusVisible(target: EventTarget | null) {
     if (!(target instanceof Element)) return true;
 
     try {
@@ -136,7 +136,7 @@
   }
 
   function handleFocusIn(event: FocusEvent) {
-    if (!isKeyboardFocus(event.target)) return;
+    if (!isFocusVisible(event.target)) return;
 
     isFocused = true;
   }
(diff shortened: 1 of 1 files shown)


Alex-Tideman (approved):

andrewzamojc (approved):
LGTM

src/lib/holocene/tooltip.svelte
```diff
@@ -125,7 +125,19 @@
     }
   }
 
-  function handleFocusIn() {
+  function isKeyboardFocus(target: EventTarget | null) {
+    if (!(target instanceof Element)) return true;
```
andrewzamojc, line 129: Strange to read, but maybe aok in practice. "If target isn't an element, isKeyboardFocus is true"

laurakwhit: Renamed to `isFocusVisible` to make this a little less confusing.

PR #3865, opened 2026-08-26
Render Markdown links in Event summary

Alex-Tideman:
## Description & motivation 💭 
Users can find it useful to add links to the Activity summary field to quickly link out to related urls. This PR adds support for Markdown rendering for the Activity summary field.

- Render event summaries with the existing Markdown preview renderer.
- Display Markdown links as clickable links in Timeline labels, History rows, and expanded event details.
- Add a compact inline preview mode that preserves the existing dense layouts.
- Keep plain-text summaries rendered as text.
- Add coverage for inline link rendering and the existing block-renderer behavior.

### Timeline Pool Updates
Markdown summaries are rendered in iframe-backed previews. The timeline’s previous modulo-based row-pool assignment changed whenever the pool size changed, particularly when opening the details panel which caused visible summary iframes to reload. This change preserves DOM-slot identity for groups that remain visible, reusing only vacated slots as the window moves. It keeps virtualization bounded while preventing iframe churn and visual flicker.

### UI Update
Moved the retry icon within the attempt count badge so it's more visible and connected to the count.

### Catalog Updates
Added 3 new workflows to the catalog that create 2,000 activities: with no summary, with text summary, and with link summary.

Screenshot 2026-08-26 at 10 28 09 AM
Screenshot 2026-08-26 at 10 28 19 AM

### How was this tested 👻 

- [x] Manual testing
- [x] E2E tests added
- [ ] Unit tests added


rossnelson (approved):

src/lib/components/event/event-summary-row.svelte
```diff
@@ -362,7 +363,19 @@
           class="flex max-w-xl items-center gap-2 first:pt-0 last:border-b-0 md:w-auto"
         >
           <p class="whitespace-nowrap text-right text-xs">Summary</p>
-          <PayloadSummary value={currentEvent.userMetadata.summary} />
```
rossedfort, line 365: Could we not just have the `<PayloadSummary />` component use `<Preview />` internally instead of what it's doing now? We could control the behavior via a prop, `renderAsMarkdown` or something similar.

rossedfort (approved):
Looks good, just one minor implementation question

src/lib/components/event/event-summary-row.svelte
Alex-Tideman: We could but there's a lot of props to pass in the Preview component, I'd rather have it more composable like this than pass a bunch of props to PayloadSummary

PR #3877, opened 2026-09-03
 Wrap long filter chips instead of overflowing the filter bar

laurakwhit:
## Description & motivation 💭 

A filter chip with a long value ran past the right edge of the filter bar instead of staying inside it. The raw-query toggle also got pushed onto its own line whenever a chip filled the row.

| Before | After |
|-|-|
|Screenshot 2026-09-02 at 9 53 16 PM|Screenshot 2026-09-02 at 9 38 14 PM|

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

1. On Workflows, add a `WorkflowId` filter whose value is ~300 characters.
2. Confirm the chip stays inside the filter bar and wraps, with `WorkflowId =` unbroken on the first line and the value continuing on the same line.
3. Confirm the raw-query toggle stays on the same row as "Clear all" instead of dropping to its own line.
4. Narrow the window to mobile width and repeat.
5. Select several execution statuses and confirm the status chip behaves the same way.

### Issue(s) closed 
`FE-556`

### DIFF
--- a/src/lib/components/search-attribute-filter/filter-bar.svelte
+++ b/src/lib/components/search-attribute-filter/filter-bar.svelte
@@ -6,6 +6,7 @@
 
   import Button from '$lib/holocene/button.svelte';
   import Tooltip from '$lib/holocene/tooltip.svelte';
+  import { translate } from '$lib/i18n/translate';
   import { IconCode } from '$lib/io/icon';
   import type { SearchAttributeFilter } from '$lib/models/search-attribute-filters';
   import type { SearchAttributeOption } from '$lib/stores/search-attributes';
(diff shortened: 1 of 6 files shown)


src/lib/components/search-attribute-filter/filter-bar.svelte
```diff
@@ -62,9 +63,11 @@
       {statusAttribute}
       {includeNullConditions}
     />
-    <div class="flex items-center gap-1">
+    <div class="flex shrink-0 items-center gap-1">
       <Tooltip
-        text={viewManualQuery ? 'Hide raw query' : 'View raw query'}
+        text={viewManualQuery
+          ? translate('common.hide-raw-query')
+          : translate('common.view-raw-query')}
```
andrewzamojc, line 70: nice

andrewzamojc (approved):

PR #3881, opened 2026-09-03
Honor modifier-clicks on links

laurakwhit:
## Description & motivation 💭 

Ctrl/cmd/shift-clicking a Saved View applied it in the current tab instead of opening it in a new one, because the views were rendered as `<button>`s. This makes them real links so the browser's default behavior works, while a plain click keeps doing exactly what it did before.

- **`Button`** — only `metaKey` and middle-click were honored, so ctrl and shift were swallowed on *every* href Button in the app. It now honors ctrl/shift/alt and non-primary buttons, and skips its own `goto` when a consumer already called `preventDefault`. `role="button"` is dropped from the anchor branch so links announce as links rather than buttons.
- **`MenuItem`** — the `href` branch never invoked `onclick`, wasn't reachable with arrow keys, and dropped the `selected` checkmark and `description`/`leading`/`trailing` snippets. It also fell back to a **full page load**: `Menu` stops click propagation on its `<ul>`, so SvelteKit's router (which listens on `documentElement`) never sees clicks inside a menu. Same-origin hrefs now route through `goto`. This also makes the schedule Actions → Edit item and the deployment version links navigate client-side instead of reloading the app.
- **Accessibility** — below `xl` the saved view label is visually hidden, which also removed it from the accessibility tree, leaving unnamed links. The label is now `max-xl:sr-only` so it's announced at every width.

New shared `isModifiedClick` utility applies the same rule SvelteKit's own router uses. `link.svelte` already had this logic inline while `Button` was the outlier. Now both use `isModifiedClick`. 

### Design Considerations 🎨 

- **Alt/option-click now downloads** instead of navigating — for Saved Views *and* for every `Link` in the app, since `Link` previously ignored `altKey`. That's standard browser behavior and what SvelteKit does, but it is a change.
- **`newTab` menu items now close the menu on click** (e.g. the user menu's community links). They used to stay open.

### How was this tested 👻 

- [ ] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

1. Go to a Namespace's Workflows page.
2. Ctrl/cmd-click a system view (Running, Today, …) → opens that view in a new tab; the current tab is unchanged, including its filter chips.
3. Shift-click one → new window. Middle-click → new tab.
4. Plain-click one → applies in place exactly as before (no scroll jump, filter chips update).
5. Open the Saved Views menu and repeat 2–4 on a custom view. On a modified click the menu should stay open so you can open several.
6. Keyboard: tab to a system view and press Enter → applies. Ctrl/cmd+Enter → new tab. In the menu, arrow keys move between views and Space activates one.
7. Regression check on the client-side routing change: schedule Actions → Edit, and the deployment version menu links, should navigate without a full page reload.
8. Narrow the window below 1280px and check with a screen reader that each system view still announces its name.

### Issue(s) closed 
`FE-564`

### DIFF
--- a/src/lib/components/saved-query-views/saved-views-menu.svelte
+++ b/src/lib/components/saved-query-views/saved-views-menu.svelte
@@ -22,7 +22,8 @@
     draftView?: SavedQuery;
     dirty?: boolean;
     maxQueries: number;
-    onSelect: (view: SavedQuery) => void;
+    onSelect: (view: SavedQuery, event?: MouseEvent) => void;
+    viewHref: (view: SavedQuery) => string;
   }
 
   let {
@@ -33,6 +34,7 @@
     dirty = false,
     maxQueries,
     onSelect,
+    viewHref,
   }: Props = $props();
 
   const menuId = $derived(`${id}-saved-views-menu`);
@@ -164,7 +166,8 @@
       <MenuItem
         data-view-item
         selected={view.id === activeView?.id}
-        onclick={() => onSelect(view)}
+        href={viewHref(view)}
+        onclick={(event) => onSelect(view, event)}
         data-testid={view.name.toLowerCase().replace(/\s+/g, '-')}
         data-track-name="user-query-menu-item"
       >
--- a/src/lib/components/saved-query-views/saved-views.svelte
+++ b/src/lib/components/saved-query-views/saved-views.svelte
@@ -19,6 +19,7 @@
   } from '$lib/stores/saved-queries';
   import type { SearchAttributes } from '$lib/types/workflows';
   import { copyToClipboard } from '$lib/utilities/copy-to-clipboard';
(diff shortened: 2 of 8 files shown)


src/lib/components/saved-query-views/saved-views-menu.svelte
```diff
@@ -164,7 +166,8 @@
       <MenuItem
         data-view-item
         selected={view.id === activeView?.id}
-        onclick={() => onSelect(view)}
+        href={viewHref(view)}
+        onclick={(event) => onSelect(view, event)}
```
andrewzamojc, line 170: I like this approach. It feels like progressive enhancement, where the link would keep working even if JS were disabled. Neat.

andrewzamojc (approved):
LGTM. I think you found a sane solution to a bit of a tricky UX.

tegan-temporal (approved):

PR #3882, opened 2026-09-03
Add a compact render mode and a title to the markdown preview

rossnelson:
`Preview` renders into an iframe served by `/render`. That page pads the body by `1rem` and lays out `main`/`p` as blocks — right for a document, wrong for a short string. A one-line summary arrives in a padded box, and because `getRenderedHeight` adds `verticalBodyPadding`, the padding is measured into the frame's height too.

## What this adds

Two additive props on `Preview`, both defaulting to today's behaviour:

- **`compact`** — drops the page padding and lets the block wrappers flow, so a short string renders as a line of text instead of a padded box. Wrapping stays on, so content of a sentence or two still fits its frame rather than being clipped.
- **`title`** — sets the frame's accessible name, which was hardcoded to `"output"`. A page with several of these frames announced them all identically.

It also sets `border-0` on the iframe, which otherwise picks up the UA default border in some browsers.

`compact` is plumbed the same way the existing options are: a query param on `/render`, turned into a body class, keyed in `markdown.reset.css`. The CSP, the sanitizer and the render pipeline are untouched.

## Credit, and why this is narrow

The approach is @tidemand's, from #3865 — that PR needs the same body-class plumbing for inline event summaries, and I'd have written something worse from scratch.

This is deliberately narrower than #3865. It has **no shrink-to-fit width measurement**, so that PR's `inline` mode (which measures rendered width and uses `white-space: nowrap` for one-line timeline labels) can layer on top rather than conflict. The two are complementary: `compact` is "no chrome, still wraps"; `inline` is "no chrome, one line, hug the text".

Split out because a consumer needs the renderer part now and #3865 carries catalog and timeline changes alongside it.

## Consumer

cloud-ui renders a Namespace description — up to 255 characters of inline markdown — in the Namespace header. It needs the tight layout and a meaningful frame title. cloud-ui keeps its own fork of `/render` and `markdown.reset`, so the equivalent change is going in there too; this keeps the two from drifting.

## Verification

- `pnpm check` — 0 errors
- `pnpm lint` — 0 errors
- New `render-server.test.ts` — 6 tests over the real `GET`: the CSP is still locked down (`default-src 'none'` + `sandbox`), block markdown still renders as blocks by default, compact mode is opt-in, it wraps, and images/scripts are still stripped in **both** modes
- Full unit suite: 3151 passed. The 13 failures in 5 files are pre-existing date/time tests (`format-date`, `format-time`, `time-format`, `datetime-filter-parse`, `to-duration`) and fail on `main` without this change


src/markdown.reset.css
```diff
@@ -13,11 +13,49 @@ body {
...
+  display: block;
+  margin: 0;
+}
+
+/* Vertical padding on an inline code span grows the line box. */
+body.compact code {
+  padding: 0 0.375rem;
+  font-size: 13px;
+  line-height: 20px;
+}
```
Alex-Tideman, line 57: I think you are gonna have to add this to ui.go SetRenderRoute as well

andrewzamojc (approved):
LGTM 👍 Cool feature!

PR #3884, opened 2026-09-03
feat(workers): add Amazon Bedrock AgentCore as a compute provider

rossnelson:
## Description & motivation 💭

Adds **Amazon Bedrock AgentCore** as a compute provider in the serverless Worker forms.

| Create | Edit |
|--------|--------|
| 05a-form-create | 05-form-edit | 

| deployment | worker-instances | deployment-versions |
|--------|--------|--------|
| 01-deployment-list | 02-worker-instances | 03-deployment |

| version-details | workflows | workflow-details |
|--------|--------|--------|
| 04-version-expanded | 06-workflows-by-version | 07-workflow-detail |

The provider itself is already merged upstream — [temporal-auto-scaled-workers#117](https://github.com/temporalio/temporal-auto-scaled-workers/pull/117) / [#118](https://github.com/temporalio/temporal-auto-scaled-workers/pull/118) added it, and [cli#1177](https://github.com/temporalio/cli/pull/1177) shipped the `--aws-agentcore-*` flags in v1.8.3. This is the UI catching up.

AgentCore is invoke-based in temporal-auto-scaled-workers, so it pairs with the `no-sync` scaler and takes the same assumed-role Access fields as Lambda. It differs from Lambda in exactly one payload key: `endpoint_arn` rather than `arn`.

### Who decides which providers are offered

Self-hosted offers **all** of them, selectable. There is no per-account entitlement to express, and a Service that cannot run a provider rejects the Version with a reason — so gating the picker would only hide a choice behind a badge nobody can act on.

Restriction belongs to whichever caller has grounds for it, passed in as `providers`. Temporal Cloud already does this: it derives the list from the Namespace's own cloud, so an AWS Namespace offers Lambda and AgentCore while a GCP Namespace offers Cloud Run. That path is untouched here — Cloud always passes the prop, so it never used these defaults.

⚠️ **This changes existing behaviour for Cloud Run.** It was previously gated on `serverScaledProviderCloudRun` and rendered "Coming Soon" on self-hosted. Since no server has ever set that capability, Cloud Run has been permanently unselectable there. It is now selectable, and a Service that cannot run it says so on create.

That also let the local `Capabilities` augmentation go: nothing gates on a provider capability any more, so the UI no longer depends on a field that does not exist, and does not need the one temporalio/api#867 replaced it with.

### Three judgment calls worth a look

**1. The field is labeled "Agent Runtime Endpoint ARN", not "Agent Runtime ARN".**

The provider parses the runtime id and endpoint name out of the ARN and requires the four-part resource `runtime/<id>/runtime-endpoint/<name>` — see `parseAgentCoreEndpointARN` in temporal-auto-scaled-workers. A bare Runtime ARN fails validation server-side, so the design mock's label would have sent people down a dead end. The schema validates that shape up front, and it is checked against an ARN AWS actually generated.

**2. The CloudFormation / Terraform role helper stays Lambda-only.**

Both templates grant `lambda:InvokeFunction`. Reusing them would hand out a role that cannot invoke an AgentCore runtime. AgentCore gets the IAM Role ARN and External ID fields with no helper accordion.

⚠️ **This leaves a gap:** there is no equivalent IAM setup path for AgentCore yet. Worth deciding whether that blocks enabling the provider, or lands separately.

**3. `buildComputeConfigFromForm` replaces the per-page provider ternaries.**

The three create/edit pages each carried their own `data.provider === 'cloud-run' ? gcp : lambda`, which a third provider would have turned into three three-way branches. Mapping a provider onto a `ComputeConfig` is now one function in `shared.ts`.

### Also fixed here: the Version edit form

A Version's provider cannot be changed, so the edit form now shows only the provider in use. Previously it fell through to the shared default list, which — combined with the old capability gating — rendered the provider a Version was **actively running on** as selected, disabled, and badged "Coming Soon" at the same time.

`lockComputeProvider` also failed closed when the configured provider was disabled or hidden, producing the same fallback on the create page. It now declines only when configuration does not know the provider at all.

Neither bug was AgentCore-specific: editing a Cloud Run Version on self-hosted hit both.

Storybook covers the picker with no server: **Workers/Compute Provider Picker** → "Self-hosted (every provider selectable)" and "Locked to the provider a Version uses".

### Design Considerations 🎨

- The label change in (1) departs from the mock. It is driven by what the backend accepts.
- The missing AgentCore IAM setup helper in (2) is an open question for Design/Docs.
- Cloud Run becoming selectable on self-hosted is a behaviour change, called out above.
- One unverified detail: the "Open AgentCore Console" button points at `https://console.aws.amazon.com/bedrock-agentcore`. **Someone with console access should confirm that path.**

### How was this tested 👻

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

- `svelte-check`: 0 errors across 885 files
- `pnpm lint`: 0 errors
- Full unit suite under `TZ=UTC` (what CI uses): **3261 passing**, 2 skipped, no unhandled errors

Driven end to end against a real Bedrock AgentCore runtime in `us-west-2`, not just unit-tested — and the Version was created **through this form**, not by the CLI:

```
The form created Version run-1789000863665
The Worker inside AgentCore dialed back through the tunnel and registered "agentcore-tq"
Greet completed on the AgentCore Worker and returned:
  "hello demo from AgentCore worker on localhost"
```

Nothing polled from the machine running it, so only a Worker AWS started could have run that workflow.

New coverage:
- `deployments-service.test.ts` — AgentCore config writes `endpoint_arn` (not `arn`), pairs with `no-sync`, round-trips through its decoder, and the Lambda/AgentCore decoders do not cross provider types.
- `shared.test.ts` — endpoint ARN required, a bare Runtime ARN rejected, role and external id required as with Lambda.
- `compute-fields.test.ts` — renders the endpoint ARN field not the Lambda one, reuses the Access fields, hides the Lambda IAM helper, and offers invoke-based scaling rather than replica controls.
- `lock-compute-provider.test.ts` — 23 tests, up from 15: locks to AgentCore when a Version uses it, and keeps a configured provider selectable where gating would previously have disabled or hidden it.

One thing worth calling out, since it cost a CI round: `version-compute-details.svelte` calls `decodeAgentCoreProviderDetails`, and `deployment.svelte.test.ts` swaps the real `deployments-service` for `deployments-service-client-test-double.ts`. The missing export there surfaced as an **unhandled rejection** rather than a failed assertion — every test still reported passing while the job exited 1. If you add an export a component consumes, the test double needs it too.

Storybook needs nothing. For the real thing, this branch carries a one-command scenario (temporalio/ui#3892, merged into this branch):

```bash
pnpm demo start agentcore-serverless-worker
```

agentcore

It provisions the AgentCore runtime, opens a tunnel so the Worker can dial the frontend back, **drives this form with Playwright** to create the Version, then waits for the Worker to register and runs a workflow on it. It needs AWS credentials that can reach `bedrock-agentcore`, ECR and IAM; it names the exact policies if any are missing, and prints teardown commands for everything it creates.

The form is the point. The CLI could already create an AgentCore Version before this PR existed, so a scenario that shelled out to `create-version` would prove nothing about the change under review. `createVersion: 'cli'` falls back for a machine with no browser, and `headed: true` shows the browser doing it.

Two things the run surfaced that are worth knowing while reviewing:

- **The Access fields are load-bearing.** `require_role_and_external_id: false` makes the IAM role *optional, not ignored* — the server assumes whatever role the config carries. The CLI can omit it with `--aws-agentcore-skip-role-and-external-id`; this form requires both fields and always sends them, so it needs a role that genuinely assumes. The scenario provisions one. That leaves a real gap, tracked separately: a self-hosted operator who turned the requirement off can create a Version by CLI and cannot by form.
- **Each run needs its own deployment**, because of the server-side issue noted below. Once a deployment's Versions all go inactive, the server stops reporting their `ComputeConfig`, `lockComputeProvider` cannot name a provider it cannot see, and this page renders an error where the form should be. That behaviour is unchanged from `main` and affects Lambda and Cloud Run identically — it is not introduced here — but it is easy to hit twice in a row.

### Draft Checklist

- [ ] Confirm the AgentCore console URL
- [ ] Decide whether an AgentCore IAM setup helper (CFN/Terraform) blocks enabling the provider
- [ ] Confirm Cloud Run becoming selectable on self-hosted is wanted in this PR, or should be split out

Companions:

- **temporalio/ui#3892** — the runnable demo scenario, merged into this branch
- The Temporal Cloud side (offering AgentCore in AWS Namespaces, and the account entitlement it is gated on) is tracked separately and does not block this PR.

### Any docs updates needed?

Serverless worker docs will need an AgentCore section, including that the Resource field takes the Runtime **Endpoint** ARN.

---

### Not in this PR

- **Saving a Version end to end** needs a server whose `auto-scaled-workers` dependency contains `312f95f` (2026-08-24), which registers the `aws-agentcore` provider. `temporalio/temporal` main still pins `v0.0.0-20260811170210`. The scenario on this branch bumps a fetched checkout itself, which is the only reason the end-to-end run above was possible.
- **temporalio/api#867** proposes reporting providers and their release stage per Namespace, which would let the hardcoded `defaultReleaseStage` table in `shared.ts` go away. Not a dependency of this PR.
- **The server drops `ComputeConfig` from inactive Version summaries**, so superseded Versions show no compute provider, and a deployment whose Versions have all gone inactive cannot be added to through this form. Server-side, unchanged from `main`, and affects Lambda and Cloud Run too. Tracked separately.


Alex-Tideman (approved):

src/lib/components/deployments/version-compute-details.svelte
```diff
@@ -12,10 +13,22 @@
...
+  );
   const gcpDetails = $derived(decodeGcpCloudRunProviderDetails(computeConfig));
   const scalerParams = $derived(decodeScalerDetails(computeConfig));
+  // Lambda and AgentCore share the assumed-role fields; only one decodes.
+  const iamRoleArn = $derived(
+    lambdaDetails.iamRoleArn ?? agentCoreDetails.iamRoleArn,
+  );
+  const roleExternalId = $derived(
+    lambdaDetails.roleExternalId ?? agentCoreDetails.roleExternalId,
+  );
```
justinschoeff, line 27: These are not being displayed in [version-details screenshot](https://private-user-images.githubusercontent.com/146089/648965022-598dc18c-1262-4757-b1e5-6d256d5eb1ee.png?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3ODkwNjMzMjMsIm5iZiI6MTc4OTA2MzAyMywicGF0aCI6Ii8xNDYwODkvNjQ4OTY1MDIyLTU5OGRjMThjLTEyNjItNDc1Ny1iMWU1LTZkMjU2ZDVlYjFlZS5wbmc_WC1BbXotQWxnb3JpdGhtPUFXUzQtSE1BQy1TSEEyNTYmWC1BbXotQ3JlZGVudGlhbD1BS0lBVkNPRFlMU0E1M1BRSzRaQSUyRjIwMjYwOTEwJTJGdXMtZWFzdC0xJTJGczMlMkZhd3M0X3JlcXVlc3QmWC1BbXotRGF0ZT0yMDI2MDkxMFQxNzU3MDNaJlgtQW16LUV4cGlyZXM9MzAwJlgtQW16LVNpZ25hdHVyZT00YzJiYTE3MmM0ZmEwN2NiYTFlYmQ5Yjg4ZmQyYmUxZjA1MWMwYTFlMWM2YzllOWEzMzNmOWI2M2YwYTNlMmM4JlgtQW16LVNpZ25lZEhlYWRlcnM9aG9zdCZyZXNwb25zZS1jb250ZW50LXR5cGU9aW1hZ2UlMkZwbmcifQ.O3Ex3jrhiZLxY1APm1yfTUM5vZ_ldH5zATDSf3GwS2Q). Was this added after or maybe something is missing?

rossnelson: @justinschoeff The ui will render it if its in the payload. We allow it to be empty since the cli allows you to bypass the role and external id. 

Maybe it's here? https://github.com/temporalio/api/blob/2af5c1dea2577eb478ba4e7a7991a16b8f74e1c5/temporal/api/compute/v1/config.proto#L65-L68

We need to add provider details?

rossnelson: Traced this. The proto line I pointed at is the wrong one: `ComputeConfigScalingGroupSummary` is the list-view type used by `version_summaries`, and this component never reads it. Its own comment says it is "optimized for list views", which is right — adding provider details there would bloat list responses without fixing this.

`version-row-details.svelte` calls `fetchDeploymentVersion` (`DescribeWorkerDeploymentVersion`) and passes `workerDeploymentVersionInfo.computeConfig`. The API already types that as a **full** `ComputeConfig` — [deployment/v1/message.proto#L165](https://github.com/temporalio/api/blob/2af5c1dea2577eb478ba4e7a7991a16b8f74e1c5/temporal/api/deployment/v1/message.proto#L165) — so no api change is needed for this path.

It is the server, in two places:

1. **It never sets the field.** [`client.go#L1709`](https://github.com/temporalio/temporal/blob/706e0b4372fb7339a3a50b5663857061f037867d/service/worker/workerdeployment/client.go#L1709) builds `WorkerDeploymentVersionInfo` and assigns `ComputeConfig` zero times, so it is always nil.
2. **There is nothing to set it from.** [`makeNewVersionState`](https://github.com/temporalio/temporal/blob/706e0b4372fb7339a3a50b5663857061f037867d/service/worker/workerdeployment/util.go#L447) takes a `*computepb.ComputeConfigSummary` and stores that into `VersionLocalState`, and [`wciSpecToComputeConfigSummary`](https://github.com/temporalio/temporal/blob/706e0b4372fb7339a3a50b5663857061f037867d/service/worker/workerdeployment/compute_util.go#L82) flattens the config to `{TaskQueueTypes, ProviderType}` on the way in. The internal `VersionState.compute_config` is typed as the summary at [`message.proto#L152`](https://github.com/temporalio/temporal/blob/706e0b4372fb7339a3a50b5663857061f037867d/proto/internal/temporal/server/api/deployment/v1/message.proto#L152).

So the details are discarded on create and never persisted.

justinschoeff (approved):

PR #3888, opened 2026-09-08
docs: add serverless-worker-aws skill

rossnelson:
## Description & motivation 💭

Adds a `serverless-worker-aws` skill covering the AWS side of Temporal serverless worker deployments — the piece you need standing up before the UI's Create Serverless Deployment flow does anything.

It documents:

- The IAM role, trust policy, and external ID the Worker Controller assumes (the external ID is required — `workercontroller.compute_providers.aws.require_role_and_external_id` defaults to `true`)
- Lambda function settings (`provided.al2023`, `bootstrap` handler, 75s timeout)
- Deploying the Go worker from [temporalio/temporal-serverless-worker](https://github.com/temporalio/temporal-serverless-worker) via its `scripts/deploy.sh`
- Which values must match between that repo's `.env` and the UI form — `DEPLOYMENT_NAME` and `BUILD_ID` appear in three places and silently misbehave when they disagree
- Staging (Temporal Cloud, base64 mTLS certs) as the default path, with local Temporal + ngrok as a secondary one that links out to the existing `local-temporal` skill
- A symptom → cause table for the failures that actually happen: architecture mismatch, stale ngrok `HOST_PORT`, credentials exported after `make start`, timeout under 60s

Content is sourced from `SETUP.md` in `temporal-serverless-worker` and corrected while moving: that doc still describes the old local-only ngrok workflow, a `lambda-worker/` directory that no longer exists, and a root-level `deploy.sh` that has since moved to `scripts/` and now handles env vars from `.env`.

### How was this tested 👻

- [x] Manual testing

Docs-only change — no runtime code touched. Verified every command, path, env var, and default in the skill against the current `temporal-serverless-worker` sources (`main.go`, `scripts/deploy.sh`, `.env.example`) rather than copying the stale `SETUP.md` text. Formatted with the repo's prettier config.

Ask Claude Code something like "how do I set up the Lambda for a serverless worker deployment?" in this repo and confirm the skill loads and the instructions are followable end to end.

### Any docs updates needed?

`SETUP.md` in `temporal-serverless-worker` is now largely superseded by this skill and is stale in a few places. Trimming it down to a pointer would be a good follow-up in that repo — not done here since it is a different repo.

### DIFF
--- a/.claude/skills/serverless-worker-aws/SKILL.md
+++ b/.claude/skills/serverless-worker-aws/SKILL.md
@@ -1,0 +1,232 @@
+---
+name: serverless-worker-aws
+description: Set up the AWS side of a Temporal serverless worker — IAM role, Lambda function, and deploy — so the Worker Controller can scale workers from zero and you can exercise serverless worker deployments in the UI. Use when asked how to create the Lambda or IAM role for serverless workers, what to put in the Create Serverless Deployment form, or why an invoked Lambda never connects.
+---
+
+# Serverless Worker AWS Setup
+
+Serverless worker deployments let the Worker Controller Instance (WCI) invoke an AWS Lambda when a task queue has backlog, so workers scale to zero when idle. Testing the UI's serverless deployment flows needs a real Lambda and IAM role behind them.
+
+The worker itself lives in [temporalio/temporal-serverless-worker](https://github.com/temporalio/temporal-serverless-worker) — a Go Lambda that connects back to Temporal, polls a task queue, runs for 60 seconds, then exits. This skill covers the AWS setup that repo depends on and how it connects to the UI.
+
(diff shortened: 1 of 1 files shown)


andrewzamojc (approved):
LGTM

rossnelson:
Heads up: this PR was blocked because the "Require tests to pass" ruleset needs `unit-tests`, `integration-tests` and `e2e-tests`, but both workflows that produce them had a `paths-ignore` for `**.md`. A docs-only PR never ran the required checks, so it could never go green. I dropped that filter from `lint-and-test.yml` and `playwright.yml` in e540e24 so the checks run on every PR. Happy to split that into its own PR if you would rather keep this one docs-only.

PR #3893, opened 2026-09-09
Keep portal menus anchored when the layout shifts

laurakwhit:
## Description & motivation 💭

Portal-based menus (e.g. the search attribute filter bar dropdowns) could become detached from their button if the layout changed while they were open. This PR tracks the anchor's rect in a `requestAnimationFrame` loop while the portal is open, repositioning only when it actually changes.

### How was this tested 👻

- [X] Manual testing
- [ ] E2E tests added
- [X] Unit tests added

1. Go to the workflows list and add a search attribute filter or two.
2. Open one of the filter chip dropdowns.
3. With the menu still open, change the layout around it. E.g. add another filter chip, remove a chip ahead of it, or narrow the window until the chip row wraps to a second line.
4. The menu should stay next to its button.

### Any docs updates needed?

### DIFF
--- a/src/lib/holocene/portal/portal.svelte
+++ b/src/lib/holocene/portal/portal.svelte
@@ -6,7 +6,11 @@
   import { twMerge as merge } from 'tailwind-merge';
 
   import { portal } from './portal-action';
-  import { calculatePosition, getElementRect } from './position-calculator';
+  import {
+    calculatePosition,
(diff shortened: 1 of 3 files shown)


andrewzamojc (approved):
LGTM 👍

tegan-temporal (approved):

PR #3900, opened 2026-09-10
Color the favicon by the status of the workflow being viewed

MoLow:
## Description & motivation 💭

A workflow run page is a page you leave open and come back to — while a backfill grinds, while a flaky run retries. Until you come back, the tab tells you nothing: every Temporal tab carries the same black mark, so finding the run that just failed means clicking through them.

This tints that mark. On a workflow run page the favicon's disc takes the colour of the status badge's scheme, and goes back to the static icon everywhere else. It follows the status live — terminate a running workflow with the tab open and the icon turns without a reload, because it is derived from the same `workflowRun` store the header renders from.

The glyph, its proportions and the disc are the ones already shipped in `favicon.ico` — only the disc's fill changes.

Top row at 64px, bottom row the actual 16px rendering (pixel-doubled):

workflow status favicons

### Design Considerations 🎨

Two choices worth a second opinion:

- **Colour source.** Each status maps to the same colour *scheme* its `BadgeStatus` uses, taken at step 11 of that scheme's scale — the step dark enough that the knocked-out white glyph still holds at 16px. `Canceled` is the one exception: its badge scheme is `neutral`, and neutral's step 11 is so close to black that the icon would be indistinguishable from the static one, so it uses `slate[9]`.
- **Statuses that share a colour.** `Running`/`Paused` and `Completed`/`ContinuedAsNew` share a colour, because their badges do. Happy to split them if you would rather the icon be more granular than the badge.

The mark is drawn as an inline SVG `data:` URI. The path is a copy of `vendor/logo.svg` — a `data:` URI cannot reference an external asset, so it has to be inlined somewhere.

Also happy to open an issue first and discuss the idea before this gets reviewed, if that is the preferred order.

### How was this tested 👻

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

Unit tests cover the status → icon mapping and the store's fallback to no icon once the workflow run is cleared.

Manually verified in Chromium against a local `temporal server start-dev`, reading `link[rel~="icon"]` off the live page:

```
on open              {"badge":"Running","disc":"#0d74ce"}
after terminate      {"badge":"Terminated","disc":"#ab6400"}
back on the list     {"badge":"Running","disc":"(static: favicon.ico)"}
```

`pnpm lint` and `pnpm check` are clean, and `pnpm test --run` passes.

1. `pnpm dev:temporal-cli`
2. `temporal workflow start --task-queue demo-q --type DemoWorkflow --workflow-id wf-1`
3. Open that workflow's run page — the tab icon is blue.
4. `temporal workflow terminate --workflow-id wf-1 --reason demo` and watch the icon turn amber without reloading.
5. Navigate back to the workflow list — the icon returns to the static one.

### Any docs updates needed?

No.

### DIFF
--- a/src/lib/components/page-title.svelte
+++ b/src/lib/components/page-title.svelte
@@ -1,4 +1,5 @@
 <script lang="ts">
+  import { statusFavicon } from '$lib/stores/favicon';
   import apple from '$lib/vendor/apple-touch-icon.png';
   import banner from '$lib/vendor/banner.png';
   import favicon from '$lib/vendor/favicon.ico';
@@ -19,7 +20,7 @@
 
 <svelte:head>
   <title>{title}</title>
-  <link rel="icon" href={favicon} />
+  <link rel="icon" href={$statusFavicon ?? favicon} />
   <link rel="manifest" href={manifest} />
   <link rel="apple-touch-icon" href={apple} />
 
--- a/src/lib/stores/favicon.test.ts
+++ b/src/lib/stores/favicon.test.ts
@@ -1,0 +1,33 @@
+import { get } from 'svelte/store';
+import { afterEach, describe, expect, it } from 'vitest';
+
+import { statusFavicon } from '$lib/stores/favicon';
+import { initialWorkflowRun, workflowRun } from '$lib/stores/workflow-run';
+import type { WorkflowExecution } from '$lib/types/workflows';
+import { getWorkflowStatusFavicon } from '$lib/utilities/get-workflow-status-favicon';
+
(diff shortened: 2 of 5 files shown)


Alex-Tideman:
Love this idea, appreciate the PR! We'll get it reviewed.

rossedfort (approved):

MoLow:
@Alex-Tideman @rossedfort this pr needs a workflow approval
should have a banner like this: 
image

MoLow:
@Alex-Tideman @rossedfort the two failures seem to be because this pr is from a fork. is there anything else I should do on my side or is this ready to land?

PR #3673, opened 2026-07-14
Support markdown tables in Workflow details

stpierre:
## Description & motivation 💭 

We want to be able to display rich tabular data in Workflow details. Currently, it's not possible to render a table; basic Markdown only supports tables via raw HTML `<table>`, but HTML tags are stripped out of WF details before rendering.

### Screenshots (if applicable) 📸 

image

### Design Considerations 🎨 

This modifies table presentation for many markdown renders, including:

* WF details
* User metadata
* News feed
* Nexus endpoint descriptions
* Markdown editor preview

### How was this tested 👻 

- [X] Manual testing
- [ ] E2E tests added
- [X] Unit tests added

### Steps for others to test: 🚶🏽‍♂️🚶🏽‍♀️ 

Super simple test:

1. Start a dev server.
2. http://localhost:3000/render?content=%7C%20Name%20%7C%20Status%20%7C%20Notes%20%7C%0A%7C%20%3A---%20%7C%20%3A----%3A%20%7C%20----%3A%20%7C%0A%7C%20foo%20%7C%20Running%20%7C%20see%20%5Bdocs%5D(https%3A%2F%2Ftemporal.io)%20%7C%0A%7C%20bar%20%7C%20Completed%20%7C%20done%20%7C&theme=light

This just calls the `render` endpoint directly so no style is applied, but it's sufficient to see a table rendered from the URL-encoded Markdown table.

For more detailed testing, emit a markdown table in the WF details of a real WF running on a local dev server.

PR #3463, opened 2026-05-27
fix(markdown): allow nested lists to render as block

rossnelson:
## Summary

- Scope the markdown reset's inline-display rule from `li *` to `li > p`, so nested `<ul>`/`<ol>` inside a list item no longer collapse to `display: inline`.
- Workflow "Current Details" now renders nested lists correctly instead of flattening every level onto a single line.

The original `li * { display: inline }` was meant to keep paragraph wrappers (mdast emits `<p>` inside `<li>` for loose lists) on the same line as the bullet. Narrowing the selector to direct `<p>` children preserves that behavior while letting nested lists fall back to their default block layout.

Reported in [DT-4047](https://temporalio.atlassian.net/browse/DT-4047). Fixes the nested-list portion; the separate `&nbsp;` literal-render report on the same ticket is still pending repro.

## Test plan

- [x] `pnpm lint` (stylelint clean on the changed file)
- [x] Verified via `/render` endpoint that the markdown pipeline emits nested `<ul>` inside `<li>` — with the fix, the browser's default `display: block` on `<ul>` now applies.
- [ ] Visual confirmation in deploy preview: workflow Current Details panel with a markdown nested list renders as a hierarchical list rather than a flat run-on line.

[DT-4047]: https://temporalio.atlassian.net/browse/DT-4047?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #3905, opened 2026-09-11
Generate the render stylesheet so both routes share one source

rossnelson:
Stacked on #3882 — review that one first.

## The problem

`/render` has two implementations. A packaged `ui-server` serves it from Go (`server/server/route/ui.go`); the SvelteKit route at `src/routes/(app)/render` only runs where there is a SvelteKit server. Both need the same stylesheet, and the Go one has been carrying a copy pasted in when the route was added in #2430.

**None of the differences were chosen.** Every change to `src/markdown.reset.css` since #2796 landed on one side only:

| Date | PR | Touched the Go copy? |
|---|---|---|
| 2025-08-07 | #2796 User metadata | yes |
| 2026-05-22 | #3430 `fix(a11y)`: rem font-size, unitless line-height | no |
| 2026-05-28 | #3463 `fix(markdown)`: nested lists render as block | no |
| 2026-07-14 | #3673 Support markdown tables | no |
| 2026-08-17 | #3785 Workflow Catalog | no |
| 2026-09-09 | #3849 Io design system | no |

So a packaged server is missing two shipped bug fixes (#3430, #3463), has no table styles, and **renders markdown in the 2024 Holocene palette** because #3849 never reached it.

## Why it could not just be hand-synced

Two hard constraints:

- The Io colours only exist as TypeScript (`ioThemeToCssVariables`), which Go cannot import.
- `server/` is a separate Go module, so `//go:embed` cannot reach `src/`.

## The change

One pure function composes the stylesheet, called from two places:

- `src/lib/utilities/markdown-stylesheet.ts` — `composeMarkdownStylesheet(reset)`, extracted from the route
- the SvelteKit route calls it per request
- `scripts/generate-markdown-css.ts` calls it to write `server/server/route/markdown.gen.css`, which Go embeds

`ui.go` loses 251 lines of duplicated CSS.

## Drift is now unmergeable

Editing the reset or an Io colour without running `pnpm generate:markdown-css` fails `scripts/generate-markdown-css.test.ts`. CI runs that (`lint-and-test.yml`) and the Go tests (`test.yml` → `make test`).

## Verification

- Dumped the CSS each route actually serves and diffed them: **byte-identical** apart from the generated-file banner comment
- `pnpm test`: 251 files, 3209 passed, 2 skipped
- `pnpm check`: 885 files, **0 errors**
- `go test ./server/route/`: pass, `gofmt` and `go vet` clean
- Confirmed each new test fails without its fix

## Deliberately left out

- The SvelteKit route still reads the reset with `fs.readFileSync` at request time. A build-time `?raw` import would be nicer, but vitest stubs `.css` imports by default and working around that means changing the shared vitest config — not worth it in this PR.
- Only the generated file is new; `src/markdown.reset.css` stays the hand-edited source.


tegan-temporal (approved):
Do you think we could use something like this for cloud-ui as well? the markdown styles are also duplicated over there from what I can tell

PR #3912, opened 2026-09-11
Fix portaled menus drifting while their trigger button is pressed

laurakwhit:
## Description & motivation 💭 

Menu's using `usePortal` shift a pixel or two as it opens and closes, then drifts back after the button is released. This was caused because `Portal` positions itself by measuring its anchor with `getBoundingClientRect()`, which **includes** transforms and `Button` has `active:scale-[0.98]` with `transition-all duration-200` that animates down to 98% while the mouse is held and animates back on release. `Portal` re-checks the anchor every frame in a `requestAnimationFrame` loop and repositions whenever `hasMoved()` reports a change.

The new `getLayoutRect()` in `position-calculator.ts` measures an element and then divides out its **own** transform, recovering the untransformed layout box in viewport coordinates. The press animation is untouched and buttons still scale on press, the menu just stops chasing them.

| Before | After |
|-|-|
|https://github.com/user-attachments/assets/f98bf4c4-416c-4fb0-8352-060e49f5d928|https://github.com/user-attachments/assets/36c291f9-d715-46b6-9146-85be75b29439|

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

### Issue(s) closed 
`FE-704`

### DIFF
--- a/src/lib/holocene/menu/menu.stories.svelte
+++ b/src/lib/holocene/menu/menu.stories.svelte
@@ -16,6 +16,7 @@
     variant?: ComponentProps<typeof MenuButton>['variant'];
     keepOpen?: ComponentProps<typeof Menu>['keepOpen'];
     position?: ComponentProps<typeof Menu>['position'];
+    usePortal?: ComponentProps<typeof Menu>['usePortal'];
     menuElement?: ComponentProps<typeof Menu>['menuElement'];
   };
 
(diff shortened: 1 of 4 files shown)


laurakwhit:
Closing in favor of https://github.com/temporalio/ui/pull/3913

PR #3921, opened 2026-09-14
Collapse common errors by default

laurakwhit:
## Description & motivation 💭 

Renders the workflow common errors as a single card with collapsable accordions instead of a flat list of alerts.
   - The header carries a per-severity count ("1 error, 2 warnings, 4 info") and the dismiss button.
   - Each error is an accordion with a severity-colored left border. The most severe one starts open while the rest collapse to a single line with a truncated description.
   - The list scrolls  at `max-h-64`, with a bottom fade that appear only while there is more content below. A run with many findings no longer pushes the rest of the page down.
   - Findings sort by severity (error → warning → info).

| Before | After |
|-|-|
|Screenshot 2026-09-14 at 2 51 12 PM|Screenshot 2026-09-14 at 2 43 43 PM|
|Screenshot 2026-09-14 at 2 50 57 PM|Screenshot 2026-09-14 at 2 43 59 PM|

### How was this tested 👻 

- [x] Manual testing
- [ ] E2E tests added
- [ ] Unit tests added

### Issue(s) closed 
`FE-420`


Alex-Tideman:
This looks so much better!

I wonder if Common Errors is a good title when there is errors/warnings/info. Maybe Helpful Hints?

bilal-karim:
@Alex-Tideman Agreed, so much better! Are there plans to make this customer-facing? 

How about "Diagnostics" for the title? It's more overarching because this panel includes a mix of info, warning, and error messages.

I also wonder if it should take up the full width of the page? We have the room, so why not? It would give the content more breathing room.

src/lib/components/common-errors/common-error-card.svelte
```diff
@@ -0,0 +1,132 @@
...
+            </Button>
+          </Tooltip>
+        {/if}
+      </div>
+    </div>
+    <div class="relative">
+      <div
+        bind:this={scrollEl}
+        id="{id}-errors"
+        class="max-h-64 overflow-y-auto"
```
laurakwhit, line 110: > How about "Diagnostics" for the title? It's more overarching because this panel includes a mix of info, warning, and error messages.
 > I also wonder if it should take up the full width of the page? We have the room, so why not? It would give the content more breathing room.

@bilal-karim here's what it would look like. Lmk if you'd like me to make those changes! 

Image

bilal-karim: This looks better. :shipit:

laurakwhit: @bilal-karim one last thing. How do you feel about this ⬇️ copy change as well? 

| Before | After |
|-|-|
|Screenshot 2026-09-17 at 1 28 55 PM|Screenshot 2026-09-17 at 1 33 24 PM|

bilal-karim: @laurakwhit Let's do it!

bilal-karim (approved):

PR #3924, opened 2026-09-15
Add Scheduled system view and last viewed Saved Views

Alex-Tideman:
## Description & motivation 💭

Saved query navigation had two gaps: scheduled workflows were not available as a built-in workflow view, and returning to a recently used custom view required reopening and searching the Saved Views menu. The menu trigger also changed its label and icon with the active view, which made the control move visually and blurred the distinction between the menu and the selected shortcut.

This PR:

- adds a **Scheduled** workflow system view using ``TemporalScheduledById is not null``;
- persists the last-viewed custom saved-view ID per view family and namespace;
- renders that view as a quick-access control immediately before the Saved Views menu;
- restores the last-viewed shortcut after reloads, updates it when another custom view is selected, and removes stale state when the view is deleted;
- immediately marks newly created, duplicated, and shared-link-created views as last viewed;
- keeps the menu trigger labeled **Saved Views** and removes its leading view icon so its purpose and width remain stable; and
- preserves modifier-click navigation without changing last-viewed state.

#### Stable Saved Views trigger

The dropdown keeps a consistent label and no longer shows a leading bookmark/view icon.

CleanShot 2026-09-15 at 2 57 16 PM@2x

#### Video

The recording shows filtering workflows, saving the query as a new view, and the new view appearing immediately as last viewed.

https://github.com/user-attachments/assets/8d610d8f-dabc-4fb1-bd2d-0eb20431d250

### Design Considerations 🎨

- The last-viewed shortcut appears before the Saved Views menu so the recent destination is encountered first in the left-to-right action flow.
- The bookmark icon remains on the specific saved-view shortcut, while the collection-level dropdown is text-only with its chevron.
- Only the saved-view ID is persisted. The rendered name, query, and icon continue to come from the canonical saved-query store, so edits remain in sync.
- Persistence is scoped by both saved-view family and namespace to avoid leaking a workflow, activity, worker, schedule, or Nexus selection into another surface.
- The shortcut has a descriptive `aria-label` containing the saved-view name and keeps existing button/link keyboard behavior.

## Testing 🧪

Integration coverage now verifies:

- initial empty last-viewed state;
- selection, persistence across reload, and query restoration;
- query and name updates;
- switching to a different custom view;
- deletion and stale-state cleanup;
- ordering before the Saved Views dropdown; and
- last-viewed behavior for normal creation, duplication, and shared-view creation.

Static validation completed with Prettier, ESLint, and `pnpm check` (0 errors; existing repository warnings unchanged). The full Saved Views Playwright suite completed for desktop and mobile: 23 scenarios passed on the first attempt and one timing-sensitive scenario passed on retry.

### How was this tested 👻

- [x] Manual testing
- [x] E2E tests added
- [ ] Unit tests added

1. Open the Workflows page and verify **Scheduled** filters on ``TemporalScheduledById is not null``.
2. Add and apply a workflow filter, then choose **Save as New** and name the view.
3. Verify the new view appears immediately before the **Saved Views** dropdown and is active.
4. Select **All**, reload the page, and verify the last-viewed shortcut remains available.
5. Select the shortcut and verify its saved query is restored.
6. Select another custom view and verify the shortcut updates; delete it and verify the shortcut is cleared.
7. Run `PW_MODE=integration pnpm exec playwright test tests/integration/saved-query-views.spec.ts`.

### Draft Checklist

N/A — ready for review.

### Merge Checklist

- [x] CI passes
- [ ] Reviewer approval

### Issue(s) closed

None.

### Any docs updates needed?

No documentation updates are needed; this changes Saved Views navigation behavior without changing a public API or configuration surface.


bilal-karim (approved):
@Alex-Tideman Shouldn't the Edit / Duplicate / Share be closer to the last saved view? It feels odd next to Saved View dropdown. Too far apart from the actual thing that's editable / duplicable / shareable.

Alex-Tideman:
@bilal-karim good call, I moved it to Dropdown / Last Saved View / Saved View actions

src/lib/components/saved-query-views/saved-views-menu.svelte
```diff
@@ -57,12 +57,8 @@
   const draftActive = $derived(
     Boolean(draftView) && activeView?.id === draftView?.id,
   );
-  const noneSelected = $derived(!activeUserView && !draftActive);
-  const label = $derived(
-    draftActive
-      ? (draftView?.name ?? '')
-      : (activeUserView?.name ?? translate('common.saved-views')),
-  );
+  const label = translate('common.saved-views');
```
laurakwhit: Are we sure we want to remove the derived label here? Seems a little harder to quickly interpret if the view is a new unsaved view or you are editing an existing saved view. 

Maybe for a new view the "Save as new" should go just to the right of the menu dropdown (before the last saved view) and when editing an existing saved view the "Unsaved" badge should go on the last saved view instead of in the menu button? 

| Before | Current | Proposed |
|-|-|-|
|Image|Image|Image|
|Image|Image|Image|


cc @bilal-karim

bilal-karim: @laurakwhit I like your solution. :shipit: 

Thanks for catching this.

Alex-Tideman: Updated to match this

PR #3925, opened 2026-09-15
FE-734 - Add resizable columns to the configurable tables

rossedfort:
## Description & motivation 💭

The configurable tables size every column to its content, and the only lever a user has over layout is the dense/comfortable toggle. One long Workflow ID, or a wide custom search attribute, squeezes every other column into uselessness, and there is no way to say "make this column wider and keep it that way."

This adds drag-to-resize to the header of all five configurable tables — workflows, schedules, activities, deployments, nexus operations.

Implementation notes:

- `width` is a new field on `ConfigurableTableHeader`, so it persists per namespace in the existing column-config `localStorage` keys alongside label and order, and is dropped automatically when a column is removed or renamed by `migrateColumnLabels`. The reducer keys off `TABLE_TYPE`, so one `CONFIGURABLE_COLUMN.RESIZE` case serves every table.
- Tables stay in `table-auto`. Rather than switching `Table` to `table-fixed`, a sized column pins `width`/`min-width`/`max-width` on its header and body cells and clamps its own content; unsized columns keep autosizing exactly as they do today.
- The handle is one Holocene primitive (`column-resize-handle.svelte`) that the five tables opt into. It is store-free, taking the minimum as a prop, and it finds the cell and table it lives in rather than having them threaded through. Dragging uses pointer capture, so there are no window-level move/up listeners to tear down.
- Passing `onResize` is what makes a column resizable. The archival and event-history tables share the workflows header cell and simply don't pass it, so they are unchanged. The workflows table also withholds it from the ephemeral "Versioning Behavior" column, which isn't in the persisted config.
- The keyboard decisions (step size, clamping, reset) live in a pure helper so they're unit-testable without a DOM. That helper also owns the propagation guard: both pagination components register their arrow-key shortcuts on `window`, so a handle that only called `preventDefault` would resize a column and page the table in the same keystroke. The handle now stops propagation for every key it consumes — `Space` included, since the paginated table maps it to bulk row selection.

Based on #3685, which built this for the workflows table; the store layer there is adopted as-is and the component layer is generalized. Worth deciding what happens to that PR.

The visual treatment is unchanged from #3685 — see the screenshots there. Still to add: fresh captures across the other four tables.

### Design Considerations 🎨

- Dividers are visible at rest as 1px lines in `border-primary`, the same token as the header rule they meet, so the tables read as resizable without having to find the cursor change at a column edge. The divider under the cursor, focus, or an active drag thickens to 2px in `border-brand`. The last column keeps its handle but not its resting line, which would otherwise float at the table's right edge with no column beside it.
- The 80px floor is the number in the ticket, and it holds. Worth a design look in comfortable density specifically: filterable cells reserve `pr-16` for the hover filter/copy buttons, so an 80px filterable column shows very little text before ellipsizing.
- Badge columns needed their own answer. `BadgeStatus` sized itself to its text and ignored its container, so an 80px Status column clipped the pill mid-word and the cell's `text-overflow` painted a stray ellipsis next to it. The badge now caps at the available width and truncates its label inside the pill, keeping the shape and colour that actually carry the status. Icon segments hold their intrinsic size, so the word gives way before a delayed or task-failure marker does.
- An 80px Status column leaves the badge 64px after cell padding, and only `paused`/`failed` (60px) fit; `running` needs 67px, `complete` 75px, `terminated` 90px, and `continued as new` 135px. So truncation at the floor is expected rather than avoidable — a Status column that never truncates would need a ~151px minimum. Flagging in case Design would rather give badge columns a higher floor than let them truncate.
- Keyboard users get the ARIA window-splitter pattern: focusable `role="separator"` with live `aria-valuenow`/`valuemin`/`valuemax`, arrow keys resizing in 16px steps, Enter/Space returning the column to autosizing, and Escape aborting a drag in progress.

### How was this tested 👻

- [x] Manual testing
- [ ] E2E tests added
- [x] Unit tests added

Unit tests cover the store reducer (targets one column only, clamps to the minimum, removes the width on reset, preserves sibling properties, leaves other namespaces and other tables alone) and the keyboard helper (step size, the 80px floor, reset keys, ignored keys, and that propagation is stopped for exactly the keys the handle consumes).

Manually verified per table against a local dev server: real mouse drag, keyboard stepping, the floor holding at exactly 80px, double-click and Enter returning a column to autosizing, reset-all clearing every width, persistence across a reload, namespace isolation, widths holding through the density toggle, and no handles in the workflows batch-actions colspan mode with widths intact on exit. Each table writes to its own storage key. No console errors.

1. Open any of `/workflows`, `/schedules`, `/activities`, `/workers/deployments`, `/nexus-operations`.
2. Every divider is visible at rest. Hover one — it thickens and picks up the brand colour. Drag it; the column follows the cursor and stops at 80px.
3. Double-click that divider to return the column to autosizing.
4. Tab to a divider: left/right arrows resize in steps, Enter or Space resets it, and Escape during a drag restores the pre-drag width.
5. Size a couple of columns, then use the reset button next to Configure Columns to clear them all — it only appears once something is sized.
6. Reload the page; widths come back. Switch namespace; widths don't follow.
7. On a table with more than one page of results, focus a divider and press the arrow keys — the column resizes and the page does not change.
8. Shrink the Status column to its 80px floor: the badge should stay a fully rounded pill with its label ellipsized inside it, with no stray `...` beside it, and any delayed/task-failure icon still visible.
9. Check the event-history and archival tables are visually unchanged.

### Draft Checklist

- [ ] Re-verify keyboard resizing on a multi-page table (the propagation fix landed after the manual pass above)
- [ ] Eyeball the at-rest dividers in both themes across all five tables — `border-primary` is a low-contrast alpha token, and always-on lines are a bigger visual shift than the old hover reveal
- [ ] Add screenshots for the four tables beyond workflows
- [ ] Verify the archival table on a server with archival enabled (the CLI dev server has it disabled, so only the shared code path was exercised via event history)
- [ ] Design sign-off on the 80px minimum in comfortable density
- [ ] Decide whether #3685 is closed in favour of this

### Merge Checklist

- [ ] Rename the Linear issue, which still says "workflows table"

### Issue(s) closed

[FE-734](https://linear.app/temporal/issue/FE-734/add-resizable-columns-to-the-workflows-table)

### Any docs updates needed?

No. The controls are discoverable in place and no configuration changes.


andrewzamojc (approved):
Nice interaction. I like the double click to reset. It all works well with the two-way scrolling in the table. Nice that it works well with keyboard! It could use an example in storybook so its easy to try out, not critical.

tegan-temporal:
Keyboard nav is triggering page change for me... liek I press right to expand a column but the data changes in addition (we load the next page of data)

https://github.com/user-attachments/assets/037d64c9-c639-425a-804e-468f903c1fab

bilal-karim:
Looks great. However, after reviewing this again, it’s not clear that columns can be resized unless you hit the edge of the column header.  

https://github.com/user-attachments/assets/270001e9-272f-43b1-8bd1-d6fea60d6b0a

We could either 1) make the divider visible anywhere within the header when hovered, or 2) show divider lines by default at 1px with `border-primary`. Then, these lines could be updated to 2px with `border-brand` on hover. I prefer option 2.  See an example below.

image


https://github.com/user-attachments/assets/d90c9b03-d26d-4433-87ae-2b4f08c690c0

Alex-Tideman:
CleanShot 2026-09-17 at 1 18 37 PM@2x

This is a silly tiny thing but it truncates (...) badges

PR #3930, opened 2026-09-16
(fix) Keep the Cloud nav toggle square when the nav is collapsed

bilal-karim:
## Description & motivation 💭

The collapse/expand toggle in the cloud nav rail had a different hover target in each state: 24×24 when the nav was open, but 16×24 when it was collapsed.

|  Before  |  After  |
|---|---|
| Screenshot 2026-09-16 at 1 00 16 PM | Screenshot 2026-09-16 at 12 59 55 PM |

The collapsed rail is `w-[calc(3rem_+_1px)]`, which leaves the nav a 32px content column (`grid-cols-[2rem]`). The toggle's `mx-[8px]` claimed 16px of that, leaving 16px for a `size-6` (24px) button — so it flex-shrank to 16px wide. The icon inside overflowed at its full size rather than being clipped, so the button still *looked* right; only the rounded `hover:bg-surface-overlay-primary` background gave it away. Hovering "Expand Navigation" lit a narrower, non-square box than hovering "Collapse Navigation".

The fix drops the horizontal margin while the nav is closed and adds `shrink-0`, so the 24px square centers in the 32px rail with 4px gutters. The open state is untouched.

### Design Considerations 🎨

`oss-nav-bar.svelte` has the same shape of problem, and worse: its toggle needs 56px in the same 32px column (`mx-2` + `group-data-[nav=closed]:p-2` + a 24px icon). I left it out of this PR to keep the change scoped to the reported cloud-nav issue — happy to fold it in if you'd rather fix both at once.

### How was this tested 👻

- [x] Manual testing

Measured the button's box in both states by reproducing the cascade (grid column → flex row → button margins/shrink) and reading `getBoundingClientRect()`:

| State | Grid column | Button box before | Button box after |
| --- | --- | --- | --- |
| Collapsed | 32px | **16 × 24** | 24 × 24 |
| Expanded | 239px | 24 × 24 | 24 × 24 |

No rail overflow in either state after the change.

Also packed this branch into cloud-ui locally (`svelte-kit sync` → `pnpm package` → `pnpm pack`) and ran it against the dev server to confirm the built `dist/` carries the change.

1. Open the cloud nav and hover the collapse button top-right of the rail — note the square hover background.
2. Collapse the nav, then hover the expand button in the rail.
3. The hover background should be the same size in both states. Before this change the collapsed one was 8px narrower and visibly non-square.

### Any docs updates needed?

No.


tegan-temporal:
I pushed a simplification that relies on a gap instead of funky margin shenanigans 
Screenshot 2026-09-16 at 10 55 48 AM
Screenshot 2026-09-16 at 10 56 06 AM

tegan-temporal (approved):

PR #3935, opened 2026-09-17
Set the markdown frame in the system font stack

rossnelson:
## The problem

The render frame's reset set the page in a bare `font-family: sans-serif`, which resolves to Helvetica on macOS. cloud-ui sets its pages in Tailwind's system stack, which is SF Pro on a Mac, so a namespace description rendered in the frame sat in a visibly different face from the heading above it. Reported in #vibrant-iguana with before/after screenshots.

## Why not load the page's font

The frame is sandboxed with `default-src 'none'` and no `font-src`, so it cannot fetch a webfont, and an `@font-face` declared in the parent does not reach a child document. Loading Inter into the frame for the open-source UI would need a `font-src 'self'` policy plus a stable font URL the static Go embed can reference, which is separate work.

## The change

Use the same system stacks Tailwind resolves to, for sans and mono, held in two custom properties so the mono stack is not repeated across `code` and `pre`. Emoji faces stay in the stack since a description can contain them.

- cloud-ui: exact match with the page, since its pages use the same stack
- open-source UI: closer than Helvetica, but its pages are in Inter, so not an exact match (see above)

`server/server/route/markdown.gen.css` regenerated. cloud-ui picks this up with the next release pack, now that temporalio/cloud-ui#3282 imports the stylesheet from the package.

## Verification

- `vitest run` on the generator and render route: 12 passed, including a new test that fails on the old `sans-serif` rule
- `go test ./server/route/`: pass; `gofmt`, `go vet` clean
- `eslint`, `prettier --check`: clean


andrewzamojc (approved):
LGTM

